Issue #218, opened 2015-03-14
Fused exterior twisted-clover fermions incompatible with tuning 


mikeaclark:
(bug report from Alex)

I'm still having problems with the fused kernels and tuning, I don't understand why. Disabling tuning yields the right result.

You can actually test it: if you use the same source (not point, fill the spinor with some random stuff), you can try to run invert_test with and without tuning on multiGPU with fused kernels, and you should see at the beginning:

"Prepared source = XXXXXXXXe+YY"

The number should be the same, regardless of the tuning, but this is not the case.  I found some problems with the labeling during tuning: the application of the clover term was not labeled properly, and the dslash kernels could be labeled wrong. After correcting this in dslash_quda.cu:

- Just add:

```
    if (twist == QUDA_TWIST_GAMMA5_DIRECT) {
      writeAuxString("direct");
    } else {
      writeAuxString("inverse");
    }
```
in the constructor, and change tuneKey to

```
    TuneKey tuneKey() const { return TuneKey(in->VolString(),
typeid(*this).name(), aux); }
```

And correcting dslash_twisted_clover.cu (removed tuneKey, so we use the standard one, and added fillAux with different endings depending on the kernel called) I tried again, but I'm getting the same wrong results.

mikeaclark:
Alex, is this problem still present in your latest development branch?  If so, I'll work on this tomorrow.

AlexVaq:
I think it is, although I indeed solved a problem with the tuning. Let me try to check it today.

AlexVaq:
Hi Mike, I can't easily test now on multiGPU. I have mostly access to my
own GPU in my linux box, but that's all, other GPUs are being used for
production. Single GPU works ok, but this is not telling us anything,
because the fused kernels are for multi GPU... so I don't know. I suppose
the problem still exists.

mikeaclark:
Alex regarding your comment on needing multiple GPUs to test this: not sure if you were aware, but you can test multiple GPU code on a single GPU using the `--partition N` flag, where N is a 4 bit integer dictating which communications to switch on, with bits 0,1,2,3 used for dimensions x,y,z,t.  For example, 
```
--partition 1: x dimension communication only
--partition 6: y and z dimension communication
--partition 15: all dimensions
```
So this flag can be used to emulate multi-GPU performance using loopback communication.  It should be suitable for doing the testing required for this bug.

AlexVaq:
[...]

PR #221, opened 2015-04-10
Feature/dynamic clover invert

AlexVaq:
I think this featuer is ready for merging. In the meantime, I'll work on dynamic clover for clover fermions as well.


AlexVaq:
Are we merging this into the develop branch? Is there anything blocking the merge?

mathiaswagner:
I guess we are. We are lacking just human time and experience with the process. Alexei will review your changes and merge it unless he has questions / comments.

This is the first real feature (I.e. Not only a small change) and thus there is still a lot of friction in the process.

On May 11, 2015, at 09:21, Alejandro Vaquero <notifications@github.com<mailto:notifications@github.com>> wrote:


Are we merging this into the develop branch? Is there anything blocking the merge?

alexstrel:
I'm reviewing this branch. Let you know today.

AlexVaq:
Thanks.

I have other question, but I don't know if it should go here.

I understand under the new development model that one should place new features under new branches and ask for pull request, place hotfixes in new branches... And what about the rest? I'm introducing some improvements in the contraction code which are not a hotfix, but they're not a new feature anyway. I'm placing them as a new feature, but I don't know if this is the right labeling...

mathiaswagner:
I guess right now everything that does not fix an urgent bug us considered a feature.

On May 11, 2015, at 10:42, Alejandro Vaquero <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks.

I have other question, but I don't know if it should go here.

I understand under the new development model that one should place new features under new branches and ask for pull request, place hotfixes in new branches... And what about the rest? I'm introducing some improvements in the contraction code which are not a hotfix, but they're not a new feature anyway. I'm placing them as a new feature, but I don't know if this is the right labeling...

alexstrel:
Before merging a few points to fix: 
1) since this feature is currently implemented solely for TM+Clover, it'd be desirable to trigger a compilation warning or better error if the preprocessor parameter DYNAMIC_CLOVER is enabled accidentally for pure clover. After glancing through the code, the dynamic clover stuff not always related to QUDA_TWISTED_CLOVER_DSLASH part of the code (see .e.g. lines 846-851 in interface_quda.cpp)
2) you should update make.inc.in and README

AlexVaq:
> Before merging a few points to fix: 
> 1) since this feature is currently implemented solely for TM+Clover, it'd be desirable to trigger a compilation warning or better error if the preprocessor parameter DYNAMIC_CLOVER is enabled accidentally for pure clover.
> 
Right. I didn't think about that. Does it have any effect, anyway? I'll check the interfaces...

> 2) you should update make.inc.in and README
> 
Thanks for pointing this out, I'll try to fix it tonight.

alexstrel:
seems that it has no effect, you are right (unless I overlooked some other place!) but I think it's safer to notify a potential user of a pure clover inverter if DYNAMIC_CLOVER is enabled.

AlexVaq:
Alexei, after checking the branch of the dynamic clover, I don't understand very well how to modify both the README and the make.inc.in files. The README, as I saw it, is some general statement about the library and doesn't go into deep details, so I'm not sure whether I should comment on the clover inversion there.

On the other hand, I checked the make.inc.in file in the repository, and to me it looks ok. It adds the option (literally copy-pasted from the repo):

#Dynamically invert the clover term for twisted-clover?
DYNAMIC_CLOVER = @DYNAMIC_CLOVER@	# Dynamic inversion saves memory but decreases the flops

and it also adds the corresponding flag for compiling:

ifeq ($(strip $(DYNAMIC_CLOVER)), yes)
  NVCCOPT += -DDYNAMIC_CLOVER
  COPT += -DDYNAMIC_CLOVER
endif

So I don't know what I should add there either. What did you have in mind?

AlexVaq:
Now, regarding the compilation error with the DYNAMIC_CLOVER flag, maybe it would be too drastic. I mean, if it really doesn't affect clover fermions (only twisted-clover), and this is clearly stated in the make.inc.in file, maybe it would be too much??? One could add a dependence, like, if you set DYNAMIC_CLOVER to yes, you compile TWISTED_CLOVER as well, but I don't like that because it's not a strict dependence, and maybe someone becomes crazy trying to save compiling time because he/she didn't notice that this flag was on.

I need to check in any case what happens if we don't compile with twisted-clover, but set DYNAMIC_CLOVER to yes and try to invert twisted-clover, if everything is all right (I mean, you get an error saying that you didn't compile the corresponding module), but this would be my only modification.

AlexVaq:
Does it make sense to g enerate a clover term, even if we didn't compile clover fermions? Maybe we should add some #ifdef statements, not only for DYNAMIC_CLOVER, but also for standard clover and twisted-clover in the interfaces, so no one tries to create clover term without compiling the corresponding modules.

alexstrel:
Alex, if DYNAMIC_CLOVER is a compilation option you might need to modify configure script as well, i.e. to allow dynamic clover inversions to be enabled via compiler option , something like --use-dynamic-clover-inversion. I don't think that a user should edit make.inc for this purpose. In README you may just notify the user about this option, e.g., something similar to lines 95-99 . This is not strictly needed if ./configure --help provides this information (which currently does not).
In make.inc.in line 39 : # Dynamic inversion saves memory but decreases the flops  => Should this increase the flops?

AlexVaq:
Ah, sure. I suppose it's my fault, since I never use the configure script. You're right, the configure script must be updated.

alexstrel:
Also, I suggest you put in DiracClover constructors some warning message like
#ifdef DYNAMIC_CLOVER
warningQuda("Dynamic clover inversion is currently not supported for pure Wilson-Clover dslash");
#endif 
It is better to let someone know that the option "--use-dynamic-clover-inverse" has currently no effect for this operator.

mathiaswagner:
I agree with Alexei. It's preferred to generate errors or warnings (either at compilation or at runtime) if someone wants to try something that is not supported. If possible without breaking anything, unsupported configurations should generate errors as warnings might not receive attention.

AlexVaq:
Ok, I properly modified the configure.ac file and dirac_clover.cpp. Now let me compile and check that the warnings indeed appear.

AlexVaq:
Everything should be ok now. By the way, can anyone of you compile gtest-all in the test folder and link it without -lpthread??? It seems to me that I need to set POSIX_THREADS=yes to be able to compile properly.

mathiaswagner:
I am not sure about the gtest-all question. I don't think it is meant to be compiled separately. It is used for example in dslash_test.

AlexVaq:
Exactly, for me compiling stops when trying to link gtest-all.o if POSIX_THREADS=no, which I think it shouldn't. I mean, if we really need to set POSIX_THREADS=yes for the library to work, then we should remove the option to set it to no. I was asking if you can confirm this (i.e., compilation of the tests is broken for POSIX_THREADS=no).

I think I'll look further into this and open a new thread with the bug.

mathiaswagner:
I have `POSIX_THREADS=no` in `make.inc` and can run `make` in `tests` without any issues.

mikeaclark:
The pthreads issue seems to be system dependent.  The google test API requires pthreads, and some linux installations automatically link in pthreads and others don't.  I think this problem can be easily solved by always explicitly linking with -lpthread (even when `POSIX_THREADS=no`, which is an option for the threaded dslash).

AlexVaq:
This looks like a solution. Actually, I think in some local branch I modified at some point the Makefile in “tests” to explicitly add -lpthreads.
Thanks Mike.

mathiaswagner:
Sounds good. So either create an issue as a reminder before this is lost in the discussion of the pull request or go for a pull request right away.

mikeaclark:
https://github.com/lattice/quda/pull/237

mathiaswagner:
Two things:

This is put **ON HOLD** until we know more about #218.

Please merge develop into this branch at some point to avoid merge conflicts.

mathiaswagner:
@AlexVaq I see now. Your previous  comment was about that this pull request is now ready for merge? The comment was on the particular commit and that confused me.

mathiaswagner:
@AlexVaq This is now ready for being merged into develop?

AlexVaq:
Sorry for the delayed response. I checked it now and yes, I think it's ready. So whenever you want we merge it.

Thanks.

AlexVaq:
Should I go on with the merge?

alexstrel:
I have no objections.

mathiaswagner:
I am at the airport now. Will check and probably merge later today.

mikeaclark:
Some questions, prior to merging:
* Lines 75-81 of interface_quda.cpp looks like a merge conflict
* The use of a tab space in the macros are inconsistent with the rest of QUDA, e.g. 
```
#if        DYNAMIC_CLOVER
``` 
instead of 
```
#if DYNAMIC_CLOVER
```
* Are different tune keys generated for the twisted clover kernels whether using dynamical inversion or not?

Looks great over all though!

AlexVaq:
1. Yes, I forgot to clean that mess up!
2. I know, I've been adapting my code, but it's a bit hard to change it all. All my new code doesn't use tabs... I tried to remove some, anyway.
3. No, and it's something I wanted to do and completely forgot. Thanks for pointing out. I committed a solution that should do the job (not very elegant, but it works).

mikeaclark:
Ok, I removed all the tab spacing (sat in a boring committee meeting....)

mikeaclark:
Merged.  Great work Alex!

mikeaclark:
I just noticed there was a compilation bug in this branch which breaks the compilation of QUDA when GPU_CONTRACT is not defined.  Will fix this, sadly this means that the develop branch will be broken for this commit.

AlexVaq:
Ok, I saw it. My fault, I always tested with GPU_CONTRACT=yes.

Issue #226, opened 2015-04-28
QUDA on OS X 10.10


lcosmai:
Installed CUDA 7.0 on OS X 10.10.3  (MacBook Pro, 1 Nvidia GeForce GT 750 M)

Downloaded QUDA 0.7.0 and configured with:
./configure --prefix=/Users/cosmai/quda --exec-prefix=/Users/cosmai/quda --enable-os=osx --enable-cpu-arch=x86_64 --enable-gpu-arch=sm_30  --disable-wilson-dirac --disable-twisted-mass-dirac --disable-ndeg-twisted-mass-dirac --disable-domain-wall-dirac --enable-hisq-fatlink --enable-gauge-force --enable-hisq-force --with-cuda=/usr/local/cuda CC=/opt/local/bin/gcc-mp-4.9 CXX=/opt/local/bin/gcc-mp-4.9/g++-mp-4.9

QUDA 0.7.0 compilation fails with the following error:
ptxas fatal   : Unresolved extern function '_ZNSt3__13powEdi'
make[1]: *** [unitarize_force_quda.o] Error 255

mathiaswagner:
I have not tried on Mac but is gcc supported by Cuda on Mac OS ? I guess clang is the default cc for OS X.

On Apr 28, 2015, at 09:00, lcosmai <notifications@github.com<mailto:notifications@github.com>> wrote:


Installed CUDA 7.0 on OS X 10.10.3 (MacBook Pro, 1 Nvidia GeForce GT 750 M)

Downloaded QUDA 0.7.0 and configured with:
./configure --prefix=/Users/cosmai/quda --exec-prefix=/Users/cosmai/quda --enable-os=osx --enable-cpu-arch=x86_64 --enable-gpu-arch=sm_30 --disable-wilson-dirac --disable-twisted-mass-dirac --disable-ndeg-twisted-mass-dirac --disable-domain-wall-dirac --enable-hisq-fatlink --enable-gauge-force --enable-hisq-force --with-cuda=/usr/local/cuda CC=/opt/local/bin/gcc-mp-4.9 CXX=/opt/local/bin/gcc-mp-4.9/g++-mp-4.9

QUDA 0.7.0 compilation fails with the following error:
ptxas fatal : Unresolved extern function '_ZNSt3__13powEdi'
make[1]: *** [unitarize_force_quda.o] Error 255

azrael417:
[...]

PR #228, opened 2015-05-04
Hotfix for lib/unitarize_force_quda.cu OSX with CUDA 7.0.

mikeaclark:
This should work around #226.

### DIFF
--- a/lib/unitarize_force_quda.cu
+++ b/lib/unitarize_force_quda.cu
@@ -11,6 +11,13 @@
 
 #ifdef GPU_HISQ_FORCE
 
+// work around for CUDA 7.0 bug on OSX
+#if defined(__APPLE__) && CUDA_VERSION >= 7000 && CUDA_VERSION < 7050
+typedef Real exponentT;
+#else
+typedef int exponentT;
+#endif
+
 namespace quda{
 namespace {
   #include <svd_quda.h>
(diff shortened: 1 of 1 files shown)


mathiaswagner:
Thanks for getting this one. I still assume it is a CUDA issue and I am not sure whether we ever want to merge that into develop.
It is a workaround and as such it should probably be removed at some point (when a fixed CUDA  is available).

So, my point is: Is the issue severe enough that we need the hotfix in develop  or can we just ask affected users (should not be a lot) to apply it as a patch ?

mikeaclark:
I have filed a bug internally, but given that it's unclear when the bug will be fixed, and also that CUDA 7.0 will always have this bug, we should roll apply the hotfix to the develop branch.  It would likely therefore be part of any subsequent QUDA releases until we deem CUDA 7.0 old enough that we don't need to support it.

mathiaswagner:
Ok, so I assume there won’t be any CUDA 7.0.1 or something like this to fix such a bug?
As Mac’s are not used for clusters / supercomputers but only as personal development machines I think it would be sage to assume that users update CUDA rather quickly.


Anyhow, not worth discussing about that for long so feel free to merge it. We might consider replacing

CUDA_VERSION >= 7000

with

CUDA_VERSION == 7000

and bet on the underlying bug to be fixed before we get a new release of CUDA.

And: Should we create a list of such workarounds so we can remove them? Otherwise we probably loose track of them.

mikeaclark:
I'd prefer to leave it with the >=.  The scope of the compiler fix is presently unclear, so I can't say when this will be fixed (we're relatively close to 7.5 release date).  Moreover, this hotfix will continue to work even if the compiler is fixed.  If we put in ==, and a future version of CUDA is released before the bug is fixed, then we will have to apply the patch again.

I do like the idea of having a wiki page devoted to work arounds though.  I'll start this, and make this hotfix the first entry.

mathiaswagner:
In addition to the wiki page what about creating a preprocessor Macro like
```
#workaround 228
```
which could be extended at compile time to issue a warning using `#warning` (we could even make this depend on something like `#ifdef warnworkaround` and include a link to the github issue for which we needed the workaround?

This way we could see all workarounds that are in the code and at compile time (we could place them in or outside of the if statements checking for cud versions, os ...)

mikeaclark:
I do like the idea of this also, but I think for this simple work around this is overkill.  There are *much* bigger workarounds in the code than this simple one.  (My opinion is biased by the fact I'm trying to debug the GPU_COMMS code at the moment which is a complete hodgepodge.)

mathiaswagner:
I don't disagree but we have to get started with doing things in a more organized way ay some point.
Let me see whether I am able to hack this together when I am having my espresso later and then we can still decide.

Hope I don't have to drink 10 espressi to get that finished. ;-)

mikeaclark:
Sure, if it takes a single espresso.  The GPU_COMMS code clean up should be measured in litres of espresso.

mathiaswagner:
I do have something that works (but not yet found the best header where to put it) but is suppressed by the ‘-w’ option to nvcc. Enabling this message again just for the host compiler does not work as the ‘-w’ gets appended to the host compiler options after the ‘-XCompiler’ stuff.

So, maybe we should add an option to the configure step that allows us to compile with or without warnings. We don’t need users to see the workaround warnings anyway so if developers decide to compile without ‘-w’ they should be able to see the options.

mikeaclark:
Ok, I've updated this now: the signature of the pow calls have been changed as discussed.  The only downside is that pow(Real,Real) is slower, than pow(Real,int), with the getUnitarizeForceField routine seeing a 35% performance regression on K40 (in both single and double precisions).  I'd say this is not a problem though since this routine is not on the performance critical path.

Mathias, can you review and merge?

nmrcardoso:
Using __nv_powi() still does not work for OSX?

mathiaswagner:
The fix looks good. I did not yet run the unit test.
The code path is indeed not critical but if there is any easy way to avoid the performance hit.
Any ideas?

Maybe something like ?
```
// work around for CUDA 7.0 bug on OSX
#if defined(__APPLE__) && CUDA_VERSION >= 7000
typedef Real exponentT;
#else
typedef int exponentT;
#endif
```
and the use  `static_cast<exponentT>(n)` in the `pow` calls ?

mikeaclark:
nv_pow is an internal function in the CUDA header, not intended for public consumption.  Using this in QUDA is dangerous, since it's possible that the CUDA header definition could change with a subsequent CUDA release breaking QUDA.  It also renders the code as CUDA specific rather than standard C++, which is a bad if we wanted in the future to have parts of QUDA compiled for the CPU.

mathiaswagner:
Guess your commit counts as an answer to my suggestion. Is it better to do the static_cast starting from ints or from doubles? I.e. should we also remove the `.0` in the argument of the cast?

mikeaclark:
It doesn't make a difference in performance at all.  It's cleaner as simple integers.

I did just notice that this commit breaks on OSX, as "Real" is undefined outside of the templated function.  Thus I propose to change the patch to
```
// work around for CUDA 7.0 bug on OSX
#if defined(__APPLE__) && CUDA_VERSION >= 7000 && CUDA_VERSION < 7050
typedef EXPONENT_TYPE Real
#else
typedef EXPONENT_TYPE int
#endif
```
This also only applies the redefinition to only 7.0 and not to 7.5 (the subsequent release which fixes this bug).

mathiaswagner:
Should have thought of this before ... go ahead. I am just rebooting my local test machine and will run the unit test on it (CUDA 7.0 linux). I cannot test on Mac OS but will compile on Mac OS with CUDA 7.0,

PR #235, opened 2015-05-06
enable warnings if we are not on a released version (a tagged commit …

mathiaswagner:
…currently indicates a released version)


mikeaclark:
Excellent feature!

Issue #230, opened 2015-05-06
Improved staggered dslash is broken


mikeaclark:
For certain volumes and grid partitioning, the asqtad/HISQ dslash fails to work correctly.  This is bug is present in the 0.7.0 release and the current develop branch.  To reproduce the bug, a multi-GPU build is required, though only a single GPU is needed.
```
./staggered_dslash_test --sdim 6 --tdim 6 --partition 3 --dslash_type asqtad --prec single
```
So far I have only tested this on a K40/K80.  Observations so far:
- For V=6^4, partitions 1, 3, 5, 6, 7, 11, 13, 14 fail
- For V=16^4, only partition 15 fails
- Setting CUDA_LAUNCH_BLOCKING=1 makes no difference, so this is likely not a stream race condition.
- The bug occurs with or without GPU_COMMS.
- The bug occurs with QUDA_FUSED_DSLASH or the default QUDA_DSLASH2
- The bug seems to only occur with single precision and not double or half precisions (at least for the volume / partitions tested).
- The bug does not occur for unimproved staggered fermions.

Once this bug is fixed we should issue a 0.7.1 release, as this is a critical bug that must be fixed.

mathiaswagner:
I am trying to reproduce this but had no luck so far. My build uses QMP and I get

```
Results: CPU = 1404591.946479, CUDA=1404591.938202, CPU-CUDA = 1404591.937778
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from dslash
[ RUN      ] dslash.verify
0 fails = 0
1 fails = 0
2 fails = 0
3 fails = 0
4 fails = 0
5 fails = 0
1.000000e-01 Failures: 0 / 3888  = 0.000000e+00
1.000000e-02 Failures: 0 / 3888  = 0.000000e+00
1.000000e-03 Failures: 0 / 3888  = 0.000000e+00
1.000000e-04 Failures: 0 / 3888  = 0.000000e+00
1.000000e-05 Failures: 0 / 3888  = 0.000000e+00
1.000000e-06 Failures: 1555 / 3888  = 3.999486e-01
1.000000e-07 Failures: 3604 / 3888  = 9.269547e-01
1.000000e-08 Failures: 3855 / 3888  = 9.915123e-01
1.000000e-09 Failures: 3886 / 3888  = 9.994856e-01
1.000000e-10 Failures: 3888 / 3888  = 1.000000e+00
[...]

PR #236, opened 2015-05-11
Break the degeneracy over number of faces in dslash packing when tuning.…

mikeaclark:
This fixes a bug present when using both staggered and improved staggered fermions (Issue #230).

### DIFF
--- a/lib/dslash_pack.cu
+++ b/lib/dslash_pack.cu
@@ -789,6 +789,16 @@ namespace quda {
       comm[4] = '\0'; strcat(aux,",comm=");
       strcat(aux,comm);
       if (getKernelPackT() || getTwistPack()) { strcat(aux,",kernelPackT"); }
+      switch (nFace) {
+      case 1:
+	strcat(aux,",nFace=1");
+	break;
+      case 3:
+	strcat(aux,",nFace=3");
+	break;
+      default:
+	errorQuda("Number of faces not supported");
+      }
     }
 
   public:


mathiaswagner:
Do we want a hotfix release for that or will just merge that into develop? At some point we will have features in develop that go beyond fixes and maybe don't want develop to become 0.7.1?

The issue looks like it is less likely to occur in a production environment.

mikeaclark:
Merge into develop I would say.  There are a number of fixes that have been or about to be merged into develop.  I'd say we should release 0.7.1 once these are in place.

When a significant feature goes into develop that merits the bump to 0.8.we should make a 0.7 bugfix branch for subsequent bug fixes.  Sound reasonable or does your development model have a clear path we should follow?

mathiaswagner:
Ok, I am fine with that for now. Right now we only have small corrections and basically now new features. I think the point at which we have to decide is when it comes to new features, i.e. like whether #221 is something that is considered to be included into 0.7.1.  

I think the decision what we want to do in a point release is up to us. 

As I never reproduced the bug I cannot run reasonable test but the changes look fine to me and I assume you ran your tests.

mikeaclark:
Yes, this passes tests.  

This raises an issue though: for some of the hotfixes, e.g., the next one I'm about to push which is with respect to GPU Direct RDMA, the person that pushes the change maybe the only one who understands the code.  For a feature pull, it certainly makes sense that we have another set of eyes look at it, but for simple hotfixes, is this necessary to enforce?  What about trivial changes, e.g., minor documentation fixes?

PR #237, opened 2015-05-12
Always link with pthreads library for building of google test API.

mikeaclark:
This fixes a bug on some systems where the pthreads library isn't automatically linked with the executable (POSIX threads are used by the Google Test API).


mathiaswagner:
Alex, does that work for you?

AlexVaq:
Yes, it solves the problem, and I don’t think it breaks anything. We should pull this asap.

Issue #231, opened 2015-05-06
GPU_COMMS and clover bicg


mathiaswagner:
Enabling GPU_COMMS results in large true residual in the bicgstab inverter with clover fermions. Alexei has reproduced this issue with the internal invert_test.

This is issue originates from #224.

alexstrel:
Also true for other fermion actions, e.g., Wilson fermions. This bug is related  to the "odd-odd" type of preconditioning.

mathiaswagner:
Might it be that we accidentally overwrite the solution?

The reliable updates seem to work?

On May 11, 2015, at 17:37, Alexei Strelchenko <notifications@github.com<mailto:notifications@github.com>> wrote:


Also true for other fermion actions, e.g., Wilson fermions. This bug is related to the "odd-odd" type of preconditioning.

alexstrel:
Revisited my HISQ eigCG, and it seems that this bug effected reported problem with the eigCG breakdown. With GPU_COMM=no, HISQ eigCG  works normally.

mikeaclark:
Just pointing out that I have a live branch (hotfix/gdr) at the moment for fixing some ongoing GPU_COMMS issues.  If you have a simple way to reproduce this, I can perhaps include the fix in this branch.

mathiaswagner:
Including mixed precision?

On May 12, 2015, at 18:41, Alexei Strelchenko <notifications@github.com<mailto:notifications@github.com>> wrote:


Revisited my HISQ eigCG, and it seems that this bug effected reported problem with the eigCG breakdown. With GPU_COMM=no, HISQ eigCG works normally.

mikeaclark:
Hopefully any outstanding bugs can be fixed in this branch.

alexstrel:
Forgot to mention, yes that was mixed precision. BTW, I had no problems with full precision eigCG, anyway. To reproduce the bug, at least you can run bicgstab inversion for wilson with odd-odd preconditioning, if this is successful then I can check eigCG as well.

mathiaswagner:
Does that mean that all the eigcg issues you mentioned (apart from features you still might want to add) in the call are due to GPU_COMMS and we don't need a separate issue. That would be great news.

alexstrel:
I think so, yes.

mikeaclark:
[...]

PR #238, opened 2015-05-13
Improve robustness of GPUDirect and fix silent errors

mikeaclark:
This significantly improved the error checking and robustness when using GPU_COMMS to avoid silent errors and removes a nasty hack when supporting staggered fermions.  The additional error checking also applies to the non-GPU_COMMS path and in either case is only performed when HOST_DEBUG=yes.

* Remove TIFR / naive staggered work around
* Check all GPU receive buffers when creating a message handler using cudaMemset
* Check all GPU send buffers by trying a cudaMemcpy
* Check all CPU receive buffers by trying a std::fill and catching any exception
* Check all CPU send buffer by trying a std:copy and catching any exception
* 2-d variants of all of the above for strided communicators
* Only allocate message handlers for the numbers of faces that are actually used 

With respect to the last point: that fact that message handlers for all numbers of faces were declared irrespective of the operator was a source of silent failures.  E.g., when doing Wilson fermions, the halo region / ghost zone is of depth one, but message handlers were previously created for depths (face) 1 through 3.  However, the size of the receive buffers allocated was only sized for depth 1.   This is a silent error if it goes unchecked, it didn’t causes any problems when running on GPU-aware MPI (e.g., creating a MPI handle using invalid memory), but I guess this is totally undefined behaviour, and could causes nasty things to happen on GDR.

### DIFF
--- a/include/comm_quda.h
+++ b/include/comm_quda.h
@@ -59,7 +59,9 @@ extern "C" {
   comm_declare_receive_relative_(__func__, __FILE__, __LINE__, buffer, dim, dir, nbytes)
 
   /**
-     Create a persistent strided message handler for a relative send
+     Create a persistent strided message handler for a relative send.
+     This should not be called directly, and instead the helper macro
+     (without the trailing underscore) should be called instead.
      @param buffer Buffer from which message will be sent
      @param dim Dimension in which message will be sent
      @param dir Direction in which messaged with be sent (0 - backwards, 1 forwards)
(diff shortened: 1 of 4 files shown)


mathiaswagner:
Changes look good to me but I did not yet compile (my local build machine is down again).
I am not in favor of the use of `#define` in general but here it looks to be a good choice - MILC does that a lot and it makes the code a pain to read.

The comments mention possible future improvements / optimizations. I have not yet checked but it might be good to have issues for them.

Automated testing would really be handy now ;-)

mikeaclark:
The use of the macros here follows that used in malloc_quda.h, and it is used to get access to the file, function, line info string for debugging.  I agree that in general macros should be avoided, but used sparingly they can be useful.

mathiaswagner:
I did get the intention and consider it to be a reasonable choice here.

mikeaclark:
On making new issues for the improvements in the comments: I believe these have already been implemented into Justin's peer-2-peer branch, but I have yet to verify that.  Once the immediate GPU_COMMS bugs have been verified fixed (issue #231) and this is merged in, one of my next tasks is to clean the peer 2 peer branch and get it merged in to (as a new feature).

mathiaswagner:
I don't think we can run any specific tests but as GPU_COMMS is anyway still experimental I will merge that in. The code looks good to me.

PR #242, opened 2015-05-15
Fix GPU_COMMS for full-field fermions

### LINKED ISSUE #231: GPU_COMMS and clover bicg

mikeaclark:
This pull request fixes a problem with GPU_COMMS with full-field fermions.  The ghost pointers need to be updated for the odd subset in a similar way that the v and norm pointers are redirected after the odd subsets have been updated.  This pull request should fix #231.

### DIFF
--- a/lib/cuda_color_spinor_field.cu
+++ b/lib/cuda_color_spinor_field.cu
@@ -217,10 +217,20 @@ namespace quda {
       even = new cudaColorSpinorField(*this, param);
       odd = new cudaColorSpinorField(*this, param);
 
-      // need this hackery for the moment (need to locate the odd pointer half way into the full field)
+      // need this hackery for the moment (need to locate the odd pointers half way into the full field)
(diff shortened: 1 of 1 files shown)


mikeaclark:
Alexei can you merge this once you've confirmed this fixes your bug?

alexstrel:
Sure, will check it.

alexstrel:
Very good, now BiCGstab works fine, thanks!

mathiaswagner:
Alexei, did you also run tests involving half and single precision?
The code looks safe but it might not hurt to fire up this test. Once that is done please merge the pull request.

alexstrel:
I also want to check eigCG, but you are right, other options must be tested.

alexstrel:
Half and single looks ok as well, so I think I can merge the pull request.

mathiaswagner:
So feel free do merge it.

Issue #243, opened 2015-05-15
Expose heavy_quark_check parameter in interface


mathiaswagner:
This is currently hardcoded to be 10 but for low iteration counts it might be desirable to use a lower value.

mathiaswagner:
I am working on this now and have exposed it as a parameter with default 10. 
However, in cases where we use only a heavy quark residual (L2 residual 0) as with some MILC applications we should probably set it to 1.
Right now as a first step I thought about doing that in the milc interface but we might better move that into QUDA and not a application specific interface?

mikeaclark:
Since you're working on the interface, I thought I'd make a **quick** attempt at formalizing interface conventions: https://github.com/lattice/quda/wiki/Updating-the-QUDA-Interface

mathiaswagner:
:-) :-)
Documentation. You gotta love it. 

I'll have a look before creating a pull request and try to comply with the guidelines.

Issue #232, opened 2015-05-06
Use fermilab residual more consistent in different solvers


mathiaswagner:
The CG solver has received some special treatment to allow restarts if the L2 residual maxed out in the given precision in addition to just doing reliable updates. This was necessary for some MILC runs with small fermilab residual. The bicgstab solver supports the fermilab residual but not the special treatment. We should check and fix this.

This issue originates from #224

PR #244, opened 2015-05-15
Feature/milc interface cloverinvert

mathiaswagner:
This allows qudaCloverInvert to be call with only a fermilab / heavy-quark residual to be specified. 
It goes back to issues reported by Carleton.

If removed the change to the frequency of `heavy_quark_check` from this branch and created #243 to track this.


mikeaclark:
Modulo the heavy-quark residual frequency, does this fix all outstanding issues with clover heavy-quarks with MILC?

mikeaclark:
(Never mind, I see that this still leaves #232 to be done.)

mathiaswagner:
All that have been reported. Carleton saw reasonable agreement with the QOPQDP results.

Yes, #232 remains to be done but until now this has not been seen in the wild. ;-)

PR #247, opened 2015-05-15
Added support for Maxwell to QUDA

mikeaclark:
This trivially adds support for setting the GPU architecture to sm_50 or sm_52 when configuring QUDA.


mathiaswagner:
configure / configure.ac start withs SM 1.1 while make.inc.in starts with 1.3 ? 
It is only a comment in the make.inc.in but I guess this is not intentional.

mikeaclark:
Fixed.  We no longer support < sm_13 (amd sm_13 is going the way of the dodo soon also).

mathiaswagner:
I also removed it from configure. If that is fine go ahead and merge.

PR #251, opened 2015-05-19
Merge in feature/peer_to_peer

maddyscientist:
This branch adds significant features to QUDA, and should drastically improve multi-GPU performance on high GPU density nodes and adds functionality for future development.
* Adds peer-to-peer communication support for direct GPU communication 
* Separates the ghost (halo) region into a separate buffer from the main spinor array
* Fuses the half-precision spinor and norm ghost arrays into a single buffer

These features involve a lot of additions and modifications to QUDA (e.g., all dslash files are modified to deal with the separate ghost arrays).  I'm creating this placeholder while I work through, clean this up and test it thoroughly.


mathiaswagner:
As this is a pretty large set of changes: shall we split of a 0.7.1 branch first?

I'm just wondering what is the best way to handle this with our development model.

mikeaclark:
This is definitely not for 0.7.1.  On the call this week we should discuss what we have to add to finish out 0.7.1, e.g., any remaining bugs that should be fixed.

mathiaswagner:
One idea to go one would be to create a quda-0.7.x branch and merge all things we want to go into a 0.7.x release into this branch. Everything for a future version would then still go into develop.

Obviously the quda-0.7.x branch should be merged into develop after merging features - except when the things added to 0.7.x are just workarounds where we do have a cleaner solution in develop that we don't want to or cannot merge into 0.7.x for whatever reason.

mikeaclark:
That is kind of what we've done in the past: after a major QUDA release has been done, we maintain a quda-0.x branch for bug fixes 0.x.y bug fixes etc.  You'll probably notice that we've got a bunch of these left over from the 0.2, 0.3, etc. releases.

mikeaclark:
I'm fairly confident now that I've fully verified the separated ghost array across different dslash versions.  This was not implemented properly for some of the twisted-mass fermions types and there were serious performance regressions coming from the fact that the arrays `DslashParam::ghostOffset` and `DslashParam::ghostNormOffset` were being indexed dynamically.

The advantage of fusing spinor and norm arrays means that only a single H2D cudaMemcpy is required instead of two separate calls.  This reduces host overhead, which means latency reduction.  In order to test this, I've done a comparison across QUDA versions of the half-precision Wilson dslash when running on a K40.  This nicely shows the performance improvements since 0.6.1 coming from 
* setting CUDA_DEVICE_MAX_CONNECTIONS=1
* rework autotuner to massively reduce tunecache look up (0.7.0)
* fused ghost array
* fused exterior kernel update (improves GPU occupancy, reduce API overhead)

dslash performance

As can be seen the improvement from 0.6.1->0.7.0 is largest (reworked auto tuner) but the addition of the fused ghost arrays and the fused exterior have continued this trend.  There's more improvement yet to come though (properly multi-threading the MPI/CUDA interactions) as well as the improvements from peer-to-peer on multi-GPU nodes.

Next up is to work on cleaning up and verifying the peer-to-peer communication in this branch.

mathiaswagner:
@mikeaclark Now that we do have separate release can we merge this into develop?

mikeaclark:
Well, peer-to-peer support isn't actually working yet (only matters if `--enable-p2p` is passed at configure time).  However, merging this pull request in could be advantageous as it gives performance benefit for half-precision.  I think I should do a full round of testing on the non-peer-to-peer code path in here before we consider merging this.

mathiaswagner:
@Jenkins test this please

mathiaswagner:
Is that still supposed to go in a 0.8 release?
I think it is more likely for 0.9.

maddyscientist:
0.9

mathiaswagner:
Is the build error here fixed in the MultiGrid PR?

maddyscientist:
Nope, but I'm working on that build issue right now :wink:

maddyscientist:
I think this pull is now ready to be merged.  @mathiaswagner you have the pleasure 😄 

To recap, this pull request make the following main changes:
* Adds peer-to-peer communication support for direct GPU communication.  MPI is used as a fall back for GPUs that cannot peer-to-peer.  This is enabled by default, but can be disabled at runtime using setting `QUDA_ENABLE_P2P=0`.
* Separates the ghost (halo) region into a separate buffer from the main spinor array.  This was required for peer-to-peer, but has the nice effect that the ghost zone can be allocated lazily, only when required and not for all ColorSpinorFields.  As a result, there can be a significant reduction in memory overheads.  It can also be resized which is useful the same field is used for different Dirac operators, e.g., improved and unimproved staggered.
* Fuses the half-precision spinor and norm ghost arrays into a single buffer.  This increases inter-GPU message size per transaction, reducing latency and thus improve strong scaling.
* The hacks in place for unimproved staggered fermions (`nFace=1`) are now removed (enables GPU Direct support for this fermion type).
* Some clean up: removal of a whole slew of compiler warnings and all unit test now use hyphenation and not underscores for setting command line parameters

As an example of the improvement possible with the peer-to-peer enablement, see the below plot from a 16-GPU node.

p2p performance

I've done quite a bit of testing with this, and things look good.  All changes in this branch should only have affected code that uses the `ColorSpinorField` ghost zones; this means dslash, force outer products and contraction code.  I've tested that MILC RHMC is working ok, all dslash types and a variety of solvers (including multigrid).

@alexstrel I had to make a couple of changes to the deflation algorithms due to the fact that `ColorSpinorfield::total_length` and  `ColorSpinorField::TotalLength()` have been removed, e.g., I just reverted these back to use `ColorSpinorField::Length()`.  It would be good if you could take a look to confirm all is working ok with this change done.

@AlexVaq I made some small changes to the contraction code, but I think it was fairly trivial.  It appears that your contraction code does its own halo exchange and doesn't utilize the one that dslash uses.

There's further work to be done here, as currently there are two different ghost field interfaces being used by ColorSpinorFields: that used by the Dirac operator on the CUDA fields, and that used by CPU fields and for coarse Dirac operators.  I think given how much peer-to-peer communication improves standard solvers on dense nodes, it's best to merge this sooner, and defer this ghost clean up for later.

AlexVaq:
> @AlexVaq <https://github.com/AlexVaq> I made some small changes to the contraction code, but I think it was fairly trivial. It appears that your contraction code does its own halo exchange and doesn't utilize the one that dslash uses.
> 
I assume you mean the covariant derivative (the contractions per se don’t need ghosts exchange). Yes, it does its own halo exchange because the spinors are not projected, so I need to transfer the four components of the spinor. To merge this in the dslash would have been very complicated, apart from the fact that I would have to overload a hell of a lot of functions just to make the covariant derivative work. I thnk it’s better this way.

I’m eager to test this, but most likely we’ll want to use it in combination with twisted-clover MG. Is someone reviewing that pull request?

maddyscientist:
@AlexVaq 
Yes, I meant the co-variant derivative.  For non-projected spinor exchange you can use the `void ColorSpinorField::exchangeGhost(QudaParity parity, int dagger)` function.  This is what the CPU fields and coarse-grid fields use; it's a generic exchange that doesn't do any projection and works regardless of the number of colors, or where the field is located.  You can then access the resulting ghost zone with the `void* const* ColorSpinorField::Ghost() const` accessor method.  This should do exactly what you want, and clean up the hacky exchange you have in there.  

Is the twisted-clover MG ready for review and merge now then?  I hadn't realized this was the case.

AlexVaq:
> For non-projected spinor exchange you can use the void ColorSpinorField::exchangeGhost(QudaParity parity, int dagger) function.
> 
Mmm… That calls for a clean up.
> Is the twisted-clover MG ready for review and merge now then? I hadn't realized this was the case.
> 
Yes, I tested the inverted against a CG from an old build and the results agreed perfectly well. We can improve performance at a later stage (for instance, I don’t really think that twisted-mass requires the AV fields. On the other hand, now we have twisted-clover nobody uses twisted-mass…).

maddyscientist:
One other comment on `exchangeGhost`: although the function asks for a parity argument, this exchange routine also works on full fields, not just single parity ones.  So it really is for all spins, color, location, checkboarding, etc.  Not sure what you need for co-variant derivatives, likely only parity fields, but I thought I would mention this for completeness.

AlexVaq:
Thank you. Definitely I'll update the covariant derivative.

alexstrel:
changes look ok.

CMakeLists.txt
```diff
@@ -413,39 +412,29 @@ mark_as_advanced(CUDA_NVCC_FLAGS_DEVICEDEBUG)
 # we might use:
 # set_property(TARGET tgt PROPERTY CXX_STANDARD 11) in the future but need to check support in older cmake version
 
-set(CMAKE_CXX_FLAGS_DEVEL  "${OpenMP_CXX_FLAGS} -std=c++11 -O3 -Wall" CACHE STRING
-    "Flags used by the CUDA compiler during regular development builds."
-    FORCE )
+set(CMAKE_CXX_FLAGS_DEVEL  "${OpenMP_CXX_FLAGS} -std=c++11 -Ofast -Wall" CACHE STRING
```
mathiaswagner: Particular reason why only DEVEL build uses -Ofast ?

maddyscientist: Fixed (reverted back to O3).  It was an accidental inclusion, best reverted since I don't know if Ofast is supported by all compilers (e.g., clang).

CMakeLists.txt
```diff
@@ -389,12 +389,11 @@ endif(QUDA_VERBOSE_BUILD)
 
 
 set(CUDA_NVCC_FLAGS_DEVEL ${QUDA_NVCC_FLAGS} -Xcompiler -Wno-unknown-pragmas -O3 -lineinfo CACHE STRING
-    "Flags used by the CUDA compiler during regular development builds."
-    FORCE )
```
mathiaswagner: removing `FORCE` here results in empty CUDA_NVCC_FLAGS_DEVEL for me and this breaks compilation. Not sure whether this is specific to my setup.

mathiaswagner:
Here we go and hope it doesn't break ...

PR #253, opened 2015-05-22
Gauge fixing, pure gauge and optimized gauge I/O routines

mikeaclark:
This pull request will add the following features to QUDA
* Gauge fixing
* Pure gauge generation (overrelaxation and heatbath algorithms)
* Added gauge fixing to interface_quda.cpp and milc_interface.cpp
* Optimized `gauge::FloatNOrder::load` and `gauge::FloatNOrder::save` routines (vectorized I/O)
* Modified unitarize_links_quda.cu in order to support link unitarization for 12/8 parameters
* Modified copy_gauge_extended.cu in order to support copy from extended to regular gauge

Gauge fixing files:
lib: gauge_fix_ovr_extra.cu, gauge_fix_fft.cu, gauge_fix_ovr_extra.h, gauge_fix_ovr.cu, gauge_fix_ovr_hit_devf.cuh, CUFFT_Plans.h
For pure gauge config generation:
pgauge_exchange.cu, pgauge_init.cu, pgauge_heatbath.cu, pgauge_plaquette.cu, random.cu

This pull request replaces #252.


mikeaclark:
Some benchmarks for copyGauge, taken at V=16^4 (old -> new)

* double -> single
   * 18 -> 18 (139 -> 176 GB/s)
   * 12 -> 12 (118 -> 175 GB/s)
   * 8 -> 8 (86 -> 150 GB/s)
* single -> half
   * 18 -> 18 (121 -> 247 GB/s)
   * 12 -> 12 (82 -> 213 GB/s)
   * 8 -> 8 (55 -> 113 GB/s)

While some of the half precision numbers look a bit too fast (L2 cache between runs perhaps?) the speedup is undeniable.  It looks like all kernels that use the new gauge::FloatNOrder accessor are much faster.  This really shows the strength of using generic accessor code, optimizing the accessor gives a speedup across the board.

mikeaclark:
I've added a new flag `--enable-gauge-alg` to enable these new algorithms.  This is very much needed as the gauge fixing code takes a long time to compile.

Nuno, having now had a cursory look at the gauge fixing and overrelaxation codes, I can see there is a lot of code and very little comments as to what is happening.  Also, the indentation is inconsistent with the rest of QUDA, which is mostly 2 space indents.  It would be nice to get this fixed before it it merged into develop (I know there are many other parts of QUDA that has similar issues as well).

nmrcardoso:
Thanks a lot Mike to share the bechmarks and for this configuration option.

I will try to address all this issues today, since tomorrow i'll be
traveling.

It also would be nice to have a formatter.
Some weeks ago a looked at several formatters,
https://github.com/lattice/quda/wiki/agenda-call-2015-05-07
and the most promising seems to be the uncrustify, we only need to setup a
configuration file, and then with a simple script run it for every code
file.

mathiaswagner:
Thanks for reminding me of the tools. I have created #254 to remind us to define some format guide lines that we can then also use for tools.

Nuno, will you create an issue to remind us of the missing MPI support if we want to use FFT gauge fixing w/ multi GPUs? 

It would be good to update the README file commenting on this restriction and (optional, if possible) catch this in the configuration process / at compilation time and not at runtime ( like we do for the asqtad force that also does not support Multi GPU).

nmrcardoso:
I created a remind issue for the gauge fixing with FFTs.
If you can add this to the configuration file would be great,
since I've never done a configure file.

Can you tell me a file code that follows the proper standard style
to see and apply the same style to the gauge fixing code?

mikeaclark:
@nmrcardoso I've created a quick guide on how to update the configure / makefile here: https://github.com/lattice/quda/wiki/Adding-new-QUDA-features

mikeaclark:
I've just push a change to this branch that enables the vectorization for the ghost I/O routines as well.  This brings a performance boost to these also, though not as significant as with bulk I/O routines: this is because the data volumes are much smaller here so they are more latency bound.  Nevertheless, this should give a nice boost to any extended gauge routines.

nmrcardoso:
Thanks a lot Mike, the guide is very helpful.
I added a warning msg when using --enable-gauge-alg and --enable-multi-gpu
"Gauge fixing with FFTs only supported for single-GPU. Use gauge fixing with overrelaxation in multi-GPU mode." 
to configure.ac and updated the configure.

mikeaclark:
Glad it helps.  Going forward, I'd like to instill a policy that whenever anyone asks a question on QUDA (like how to add a feature, make a change or a question on some parameter or whatever), that instead of someone writing an email response they spend 5 minutes more updating wiki pages and / or doxygen to answer the question.  This way we'll get much better documentation and the same questions will stop being asked :)

mathiaswagner:
👍
Good policy and while I doubt that questions will be asked again and again we can at least point them to documentation. 

I appreciate all the new wiki pages and now feel obligated to update the page for the git development model. 😉

mikeaclark:
I think I've noticed a problem here.  It would seem that `random.h` is included in `quda.h`, this causes external applications (e.g. MILC) to fail to build unless the CURAND includes are added to the application Makefile.

mathiaswagner:
Nuno, thanks for the doxygen comments. We already have some comments in develop by now. (Pull request #263 ). 
As github now displays that it cannot automatically merge in this request it is probably a good idea to resolve these conflicts. One possible solution is to merge develop into the gauge-fix branch and clean up all merge conflicts there. I have not checked and don't know which files might be affected ...

nmrcardoso:
Mathias, sorry about that.
I have changed the files directly in the github web page and since you have
updated the milc interface file, I used that file to add my comments in the
gauge fix feature branch.
Can you tell me how to do that, the github commands?

mathiaswagner:
Ok, so I will check where merge conflicts might show up (later) and see what needs to be done.

So if I do understand you correctly all doxygen comments from the develop branch are already in your modified milc interface file?

For now, you need to take no action. I'll get back to you.

nmrcardoso:
Thanks a lot Mathias.
Yes, the quda_milc_interface.h and the quda.h are the files from develop
branch
plus the gauge fixing routines.
Probably this was a bad idea, sorry.

mathiaswagner:
Just had a quick look and there are no serious conflicts. The merge is just not sure which changes to `quda_milc_interface.h` to use and so refuses to do that automatically. On my local checkout I just had to make sure to include the gauge fix stuff. If @mikeaclark  and @nmrcardoso agree I can merge develop into the gauge fix branch. 
Then we can merge back to develop whenever we're ready,

nmrcardoso:
I'm ok with that. Thanks Mathias

mikeaclark:
Sounds good to me.

mathiaswagner:
I merged develop in and github now can automatically merge again.
As we split of a release/v0.7.x branch:
Is there anything left that prevents us from merging?

Nuno, what tests do we have for the gauge fixing?

nmrcardoso:
I have only the file su3_testing.cpp in tests folder.
That test example includes gauge field generation, reunitarization and
the gauge fixing.

mathiaswagner:
@nmrcardoso @mikeaclark  Now that we do have separate release can we merge this into develop?

mikeaclark:
I need to push a fix to this due to an incompatible change I made with #260.

mikeaclark:
Ok, I've pushed my fix.

One thing I'd like to discuss is the test associated with these new features: the name isn't very descriptive (su3_testing).  It might be a good idea to rename it to something more descriptive.  The other thing is that this test isn't really so much of a unit test as a demonstration of how to use it.  It would be nice to have correctness tests for each of these components.  Also, @nmrcardoso, can you delete all the commented out debugging code and clean up the test a bit?

I think we can merge once this minor cleanup is done.

nmrcardoso:
Thanks, I cleaned up the code in su3_testing.cpp.
I was thinking in renaming it to su3_gaugefix.cpp in the begging, however
since it uses also gauge generation and the reunitarization, I don't have a
better name to it.
In which files can I see a correctness tests to better understand how this
is done in quda?

mikeaclark:
Look at the dslash_test or blas_test.  These return pass/fail using google test API to report correctness.

mikeaclark:
How about su3_pure_gauge_test?

mathiaswagner:
The dslash tests use some basic form of the google test framework which we want to use for the automated testing. I have a branch feature/gtest_dslash which contains some experiments for making better use of the framework.

mikeaclark:
@nmrcardoso, any progress on the unit tests as discussed in our call a week ago?  As we noted, Mathias or I can add the Google test framework to it, we just want some simple pass/fail metrics on these new algorithms.

nmrcardoso:
sorry for the delay. I updated the modified files directly on github page.
I added a new file pgauge_det_trace.cu to calculate the complex link mean
determinant and also the link mean trace.
Also I added to su3_testing the correctness check using the mean link
determinant.

I check the gauge fixing correctness by checking the mean plaquette value
before and after the gauge fixing.

To check the gauge generation evolution, I use the 6.2 for the beta and
then check if the value of the plaquette is in a range and also check the
link determinant.
The gauge generation  evolution checks can be a little tricky because of
the fluctuations for small lattice sizes and if someone decides to change
beta or decrease the number of steps.

Please let me know what do you think about the implementation.

mikeaclark:
This sounds great. We can instrument the test for pass / fail only at these parameters so that shouldn't be a problem.  Thanks :)

nmrcardoso:
I notice that if someone uses the partition this option cannot be used with
the gauge fixing and the gauge generation and gauge fixing.
This was the first time I use this option :S

The main reason is because I activate the exchange
if ( comm_size() > 1 )

Also the gauge fixing with FFTs uses the same check which it runs giving
wrong results if one uses the --partition...

Should I include in multi-gpu the check for partitioned dimensions instead?

mikeaclark:
Use the `int comm_dim_partitioned(int dim)` function instead.  This should fix this problem.

nmrcardoso:
@mikeaclark, thanks.
I made the changes, tested and now it fully supports the --partition option.

PR #257, opened 2015-05-27
Feature/expose heavy quark check

mathiaswagner:
This exposed the parameter `heavy_quark_check` used in solvers in the quda interface.
The default value is 10 as previously hardcoded in the solver implementations.

For MILC applications (i.e. that use the milc_interface) the value is set to 1 if a solver is called with only the fermilab / heavy-quark residual specified.

### DIFF
--- a/include/invert_quda.h
+++ b/include/invert_quda.h
@@ -53,6 +53,9 @@ namespace quda {
     i.e., how long do we want to keep trying to converge */
     int max_res_increase_total;
 
+    /**< After how many iterations shall the heavy quark residual be updated */
+    int heavy_quark_check;
+
     /**< Enable pipeline solver */
     int pipeline;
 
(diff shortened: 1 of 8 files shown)


mathiaswagner:
I updated also the fortan interface but did not compile or test it. 
Should we have some basic internal fortran based test to at least check compilation ?

mikeaclark:
At a minimum you should compile the fortran interface with `make fortran`.

I typically use BQCD for testing the Fortran interface.

mathiaswagner:
`make fortran` was fine.

BQCD testing is not something I have set up already so it might take some time. 

Anyway: Should we keep the interface completely fixed for the 0.7.x series or shall we allow for changes that do not require changes in the calling applications?

mikeaclark:
Keep the interface fixed I think. Let's get 0.7.1 with the current fixes out the door, use the quda-0.7 branch for further 0.7.x hot fixes and shift develop to 0.8.x before merging all the pending feature branches.

mathiaswagner:
I guess that can also be merged into develop now.

mikeaclark:
Done!

PR #258, opened 2015-05-27
Feature/gtest options

### LINKED ISSUE #250: Add gtest options to internal tests
The google test framework provides some options to control test execution and output. This might be useful when we go for automated testing.
As a first step we should add
`::testing::InitGoogleTest() ` to the `main` routines of the tests.

I have started the branch *feature/gtest_options* and added this to the *staggered_slash_test*.
The output of help is now quite long - if that is annoying we can add a flag to make.inc.in.

For the long run we might think about something like test_classes so that we can run different test from one executable and defining them as different google_tests. By moving this to a function or class we could also keep the current tests (strip the real test and just keep the infrastructure for running).

We could than have a dslash_gtest that tests the Dslash e.g. with the different precisions (each as a different test).

But maybe as a first step adding the gtest options is a good thing ?

Thoughts?

mathiaswagner:
This pull request is a minimal modification to make better use of the google test framework for automated testing.
It enables the command line options of the google test framework and returns the result of the tests defined using the framework.

This closes #250.

Future work is needed to go to better automated testing.

### DIFF
--- a/tests/dslash_test.cpp
+++ b/tests/dslash_test.cpp
@@ -880,7 +880,10 @@ TEST(dslash, verify) {
 
 int main(int argc, char **argv)
 {
+  // initalize google test, includes command line options
   ::testing::InitGoogleTest(&argc, argv);
+  // return code for google test
+  int test_rc = 0;
   for (int i =1;i < argc; i++){    
     if(process_command_line_option(argc, argv, &i) == 0){
       continue;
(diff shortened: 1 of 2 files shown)


mikeaclark:
Only one comment, why only comment out the previous the `::testing::InitGoogleTest()` and not delete it?  this looks like a permanent change and we need to stop stale comments from propagating.

mathiaswagner:
Just fixed that.

PR #260, opened 2015-05-28
Hotfix/tifr

mikeaclark:
These are fixes required to get the TIFR MILC clover rhmc benchmark working properly, as well as minor interface fixes.

* Renamed inconsistently named interface functions in `quda.h`
* Added new milc interface function `qudaResidentExtendedGaugeField` for creating an returning an extended gauge field from the persistent gauge field
* Disable reliable updates for pure double precision multi-shift solver in the milc interface (this matches the behaviour of the milc staggered interface)
* Add doxygen comments to some missing functions in `quda.h`
* Added support for backing up `gauge::FloatNOrder` objects for use in auto tuning
* Clover derivative computation now uses `gauge::FloatNOrder` structs and now supports reconstruction
* Added autotuning support to clover derivative and sigma trace kernels.

I think the risk in adding this functionality is minimal.  Although some of the quda interfaces have changed, these are only called by the milc interface (which is internal to QUDA) and by no other external application.  Thus I think this is fine to include in the 0.7.1 release.

One comment though: the additions made to `gauge_field_order.h` are incompatible with the changes that @nmrcardoso's gauge-fixing branch has made.  It's a trivial fix though, and I'll push this fix to the feature/gauge-fix branch once this pull request has been done.

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -635,31 +635,74 @@ extern "C" {
 
 
   /**
-   * Take a gauge field on the host, extend it and load it onto the device. 
+   * Take a gauge field on the host, load it onto the device and extend it.
    * Return a pointer to the extended gauge field.
+   *
+   * @param gauge The CPU gauge field (optional - if set to 0 then the gauge field zeroed)
+   * @param geometry The geometry of the matrix field to create (1 - scaler, 4 - vector, 6 - tensor)
+   * @param param The parameters of the external field and the field to be created
+   * @return Pointer to the gauge field (cast as a void*)
    */
-  void* createExtendedGaugeField(void* gauge, int geometry, QudaGaugeParam* param);
-
-  void* createGaugeField(void* gauge, int geometry, QudaGaugeParam* param);
-
-  void  saveGaugeField(void* outGauge, void* inGauge, QudaGaugeParam* param);
-
-  void  extendGaugeField(void* outGauge, void* inGauge);
-
+  void* createExtendedGaugeFieldQuda(void* gauge, int geometry, QudaGaugeParam* param);
 
   /**
-   *  Reinterpret gauge as a pointer to cudaGaugeField and call destructor.
+   * Allocate a gauge (matrix) field on the device and optionally download a host gauge field.
+   *
+   * @param gauge The host gauge field (optional - if set to 0 then the gauge field zeroed)
+   * @param geometry The geometry of the matrix field to create (1 - scaler, 4 - vector, 6 - tensor)
+   * @param param The parameters of the external field and the field to be created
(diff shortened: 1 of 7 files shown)


mathiaswagner:
I have not looked at the changes yet but about getting this in 0.7.1:
Apart from the slight risk (which is fine with me if my next if applies) I would only go for this if we do have a strong motivation. How urgent is this?

What tests has this passed? 

Anyway, I will have a look at the changes.

mathiaswagner:
First thing on my mind when changing the milc interface: While it completely misses doxygen comments right now this is not what we want for the future. Please add some comments so that at least the new stuff has them,

Should things like the turned off reliable updates for pure double precision solves also be mentioned in the doxygen comments then? I think doxygen comments are use documentation and this is something that might be of interest.

include/quda.h
```diff
@@ -635,31 +635,74 @@ extern "C" {
...
+   * @param mu mu direction
+   * @param nu nu direction
+   * @param dim array of local field dimensions
+   */
+  void computeCloverTraceQuda(void* out, void* dummy, int mu, int nu, int dim[4]);
 
-  void computeCloverTraceQuda(void* out, void* clover, int mu, int nu, int dim[4]);
-
+  /**
+   * Compute the derivative of the clover term
```
mathiaswagner: missing `@param`, might also use '@sa' ?

mikeaclark:
To answer the question on urgency, this is something that really should be in 0.7.1.  These changes are required for the TIFR clover rhmc code to work correctly, this is an important application for a new Cray XC30 installed in Mumbai.  I want them to be able to use a golden version of QUDA without having to rely on some development branch.

lib/milc_interface.cpp
```diff
@@ -878,14 +878,25 @@ void qudaInvert(int external_precision,
 
 #ifdef GPU_CLOVER_DIRAC
 
-void* qudaCreateExtendedGaugeField(void* gauge, int geometry, int precision)
+static inline void* createExtendedGaugeField(void* gauge, int geometry, int precision, int resident)
 {
   qudamilc_called<true>(__func__);
```
mathiaswagner: Just wondering whether we should us `qudamilc_called<>` only in exposed functions. 
However, we can keep it this way for now and I will create a separate cleanup issue and we can decide whether we also want to have the `qudamilc_called<false>` call before exiting for all exposed functions.

But as this is only verbose output it does not prevent the merge. It just made me aware of another inconsistency in the milc interface.

AlexVaq: Mathias, are we having the developers call? Did I miss something? Did I understand wrong the time???

PR #269, opened 2015-06-01
Fixes problem with tuning and fused kernels

AlexVaq:
The spinor was not saved during tuning of the fused kernels. Upon tuning, it was filled with garbage and never cleaned.

### DIFF
--- a/lib/dslash_quda.cuh
+++ b/lib/dslash_quda.cuh
@@ -381,7 +381,7 @@
 
     virtual void preTune()
     {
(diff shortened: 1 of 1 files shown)


mikeaclark:
Can you make the code more portable and readable, so it doesn't rely on the specific enumerations values, rather the , and use something like ```dslashParam.kernel_type != INTERIOR_KERNEL```

AlexVaq:
Actually that’s clever.

mathiaswagner:
This should preferably go into release/v0.7.x.  But develop is still identical with that.

PR #272, opened 2015-06-03
Small fix for APE smearing

AlexVaq:
A call to the APE smearing funciton would create a new gauge field called gaugeSmeared, copy gaugePrecise there, and perform the APE steps, saving the final result in gaugeSmeared and leaving gaugePrecise untouched. A subsequent call to the APE smearing function will start smearing again gaugeSmeared (which is already smeared), instead of reloading the gaugePrecise field. I think this is not what we want, although this can be the object of discussion.


AlexVaq:
Could we check this? I think it's ready and the modification is really small.

mathiaswagner:
@AlexVaq if this should go into the release branch for a 0.7.2 release please create a pull request to pull that into release/0.7.x and not develop.

PR #274, opened 2015-06-04
Fix dslash flops and update documents for 0.7.1 release

mikeaclark:
This pull request does the following
* Multi-GPU dslash kernels now correctly report their flops and bandwidth (to 2nd order edge effects)
* Fixes a critical bug where the 5-d domain-wall dslash was called twice when ever it was called
* Updates documentation for 0.7.1


mathiaswagner:
The changes do look good to me.

* Good catching the 5-d domain wall issue. I have not yet checked but assume that passes the internal tests?

* Flop and Bytes count: As this change is probably made now and then forgotten: Shall we have another issue to have these checked (at some point in the future)? Nothing that stops merging this pull request. More like: If we don't remind us to maybe do that at some point we probably never will.

mikeaclark:
The flops counts for all operators but the twisted mass variants match that reported by Dirac::flops, so I have good faith in the flops being correct.  Getting all of these to match is why I found the 5-d dwf double call and DiracStaggered::flops bugs.  Just the twisted-mass numbers need some minor tweaking, they are 90% correct, and will scale correctly, but need tweaking according to which preconditioning type is called.

PR #276, opened 2015-06-04
- Fixed uninitialized variable

AlexVaq:
As Nuno pointed out, this could be dangerous, although the code never gave me the wrong result.

### DIFF
--- a/lib/gauge_plaq.cu
+++ b/lib/gauge_plaq.cu
@@ -149,11 +149,11 @@ namespace quda {
 
       void apply(const cudaStream_t &stream){
         if(location == QUDA_CUDA_FIELD_LOCATION){
+          ((double *) arg.plaq_h)[0]    = 0.;
           TuneParam tp = tuneLaunch(*this, getTuning(), getVerbosity());
 
 	  LAUNCH_KERNEL(computePlaq, tp, stream, arg, Float, Gauge);
 
-//	  cudaMemcpy(arg.plaq_h, arg.plaq, sizeof(double), cudaMemcpyDeviceToHost); 
 	  cudaDeviceSynchronize();
 
 	  #ifdef MULTI_GPU
--- a/lib/gauge_plaq.cu
+++ b/lib/gauge_plaq.cu
@@ -79,46 +79,49 @@ namespace quda {
   template<int blockSize, typename Float, typename Gauge>
     __global__ void computePlaq(GaugePlaqArg<Gauge> arg){
       int idx = threadIdx.x + blockIdx.x*blockDim.x;
-      if(idx >= arg.threads) return;
-      typedef typename ComplexTypeId<Float>::Type Cmplx;
-      int parity = 0;
-      if(idx >= arg.threads/2) {
-        parity = 1;
-        idx -= arg.threads/2;
-      }
 
-      int X[4]; 
-      for(int dr=0; dr<4; ++dr) X[dr] = arg.X[dr];
-
-      int x[4];
-      getCoords3(x, idx, X, parity);
-#ifdef MULTI_GPU
-      for(int dr=0; dr<4; ++dr) {
-           x[dr] += arg.border[dr];
-           X[dr] += 2*arg.border[dr];
-      }
-#endif
       double plaq = 0.;
 
-      int dx[4] = {0, 0, 0, 0};
-      for (int mu = 0; mu < 3; mu++) {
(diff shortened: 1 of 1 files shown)


nmrcardoso:
Another change I would do is inside the kernel, see code below
because shared memory is not initialized with zero values, then in the if(idx
>= arg.threads) return; these threads that return are
not updating shared memory...

template<int blockSize, typename Float, typename Gauge> __global__ void
computePlaq(GaugePlaqArg<Gauge> arg){ int idx = threadIdx.x + blockIdx.x*
blockDim.x;
double plaq = 0.; if(idx >= arg.threads) { typedef typename
ComplexTypeId<Float>::Type Cmplx; int parity = 0; if(idx >= arg.threads/2) {
parity = 1; idx -= arg.threads/2; } int X[4]; for(int dr=0; dr<4; ++dr)
X[dr] = arg.X[dr]; int x[4]; getCoords3(x, idx, X, parity);#ifdef MULTI_GPU
for(int dr=0; dr<4; ++dr) { x[dr] += arg.border[dr]; X[dr] += 2
*arg.border[dr]; }#endif //double plaq = 0.; int dx[4] = {0, 0, 0, 0}; for (
int mu = 0; mu < 3; mu++) { for (int nu = (mu+1); nu < 4; nu++) {
Matrix<Cmplx,3> U1, U2, U3, U4, tmpM; arg.dataOr.load((Float*)(U1.data),
linkIndex3(x,dx,X), mu, parity); dx[mu]++; arg.dataOr.load
((Float*)(U2.data),linkIndex3(x,dx,X), nu, 1-parity); dx[mu]--; dx[nu]++;
arg.dataOr.load((Float*)(U3.data),linkIndex3(x,dx,X), mu, 1-parity);
dx[nu]--; arg.dataOr.load((Float*)(U4.data),linkIndex3(x,dx,X), nu,
parity); tmpM
= U1 * U2; tmpM = tmpM * conj(U3); tmpM = tmpM * conj(U4); plaq += getTrace
(tmpM).x; } }} typedef cub::BlockReduce<double, blockSize> BlockReduce;
__shared__ typename BlockReduce::TempStorage temp_storage; double aggregate
= BlockReduce(temp_storage).Sum(plaq); if (threadIdx.x == 0) atomicAdd((
double *) arg.plaq, aggregate); }

AlexVaq:
Ufff, that was a bit messy... Can someone else read it?

mikeaclark:
@nmrcardoso, can you reformat you comment above?

mathiaswagner:
I added the markdown formatting already. @nmrcardoso if you go to the github site you can edit and add some line breaks / indents there. Use the pen symbol to edit your original comment.

nmrcardoso:
Sorry about that, in that code also forgot to change the inside if,
basically the changes are:
1) change line 82:      if(idx >= arg.threads) return;
to
double plaq = 0.;
 if(idx < arg.threads) {

2) delete line 101, double plaq = 0.;

3) add } before line 125
typedef cub::BlockReduce<double, blockSize> BlockReduce;

mikeaclark:
The code above is still a complete mess.  @mathiaswagner, I don't see any changes, did you not save your changes?

mathiaswagner:
Strange. It shows it nicely formatted in the preview pane but when I save it its screwed up again. 
Sorry. Now idea what is going wrong there.

AlexVaq:
I see your point now, so the threads above "arg.threads" should call "BlockReduce" with a zero value of the "plaq" accumulator.

I'm not so familiar with cub as to say whether this is strictly necessary or not, and if it does, it will affect other kernels that we will need to fix (this is not the first one with cub reduction). Someone out there can cast some light? It's a simple modification anyway...

AlexVaq:
I did some tests today, I didn't appreciate any differences. Is cub also summing in threads that don't call to BlockReduce????

nmrcardoso:
I made a simple example to show the differences, using double and double2.
run the program for example with ./test size
put for example 11111 in the size
and you can notice the differences.

AlexVaq:
It worked for 11111, but it didn't for 7745237, so I suppose you're right: even though the BlockReduce is not called, there are extra sums. I'll change the code accordingly and check which other kernels might be affected.

AlexVaq:
So if you don't have any more issues, maybe this is ready for merging.

mikeaclark:
This looks good to me.  I've fallen afoul of this issue as well with respect to threads exiting prior to calling cub reduction operations, so it happens to the best of us!

PR #280, opened 2015-06-06
When using threaded dslash, threaded MPI must be used.

mikeaclark:
This fixes a bug that caused irreproducible behaviour with MVAPICH when the threaded dslash is enabled.


mathiaswagner:
Tests only so no potential for side effects.

PR #290, opened 2015-06-22
Small fix for APE smearing

AlexVaq:
A call to the APE smearing funciton would create a new gauge field called gaugeSmeared, copy gaugePrecise there, and perform the APE steps, saving the final result in gaugeSmeared and leaving gaugePrecise untouched. A subsequent call to the APE smearing function will start smearing again the fields in gaugeSmeared (which are already smeared), instead of reloading the gaugePrecise field. I think this is not what we want, although this can be the object of discussion.

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -5267,13 +5267,14 @@ void performAPEnStep(unsigned int nSteps, double alpha)
   if (gaugeSmeared == NULL) {
 //    gaugeSmeared = new cudaGaugeField(gParamEx);
     gaugeSmeared = new cudaGaugeField(gParam);
-    #ifdef MULTI_GPU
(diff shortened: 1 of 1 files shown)


AlexVaq:
New version for release

mathiaswagner:
Is everything fine with the commit? 
I am staring at the diff (on github, 'files changebut to me it looks like the commit only changes indentation.

AlexVaq:
No, I moved the code out of the brackets, so there is a difference.

I noticed the problem when simulating heatbath and computing the topological charge per configuration. The topological charge would be wrong if I don't take this out of the parenthesis, because instead of smearing a new heatbath generating configuration (stored in gaugePrecise), I'd be smearing gaugeSmeared all the time.

mathiaswagner:
Sorry, I simply did not see this.
Looks good to merge and I think nobody objects to having the code acting this way.

PR #291, opened 2015-06-23
Feature/staggered reconstruct

mikeaclark:
This is mainly a feature-add:
* Adds support for 8/12 reconstruction for naive staggered fermions.
* Fixes a bug for both naive and improved staggered fermions, whereby 8/12 reconstruction was broken for periodic boundary conditions or anti-periodic boundary conditions with temporal splitting over multiple GPUs.

While 9/13 reconstruction for naive staggered is not (yet) supported, this would be an easy addition if needed.  Similarly, for improved staggered, the fat-link field is only supported using no reconstruction but adding support for non-trivial reconstruction would be straightforward.  This would only be desirable when using stout smearing (or similar) for the fat-link.

### DIFF
--- a/lib/dslash_constants.h
+++ b/lib/dslash_constants.h
@@ -437,6 +437,13 @@ void initDslashConstants(TimeProfile &profile)
   float tProjScale_fh = (float)tProjScale_h;
   cudaMemcpyToSymbol(tProjScale_f, &tProjScale_fh, sizeof(float));
 
+  // set these for naive staggered
+  float coeff_fh = 1.0;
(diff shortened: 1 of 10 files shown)


mathiaswagner:
This request goes into develop. Shall we cherry pick b7a9101 for a hotfix into release/0.7.x ?

mikeaclark:
I don't think it's necessary:
* the bug only affects 8/12 reconstruct which doesn't apply to HISQ, only Asqtad
* reconstruction choice isn't even exposed to MILC

PR #295, opened 2015-06-24
Separated computation of Fmunu from the clover field computation

### LINKED ISSUE #277: Split the clover code into Fmunu tensor computation and clover computation
It can be advantageous to have independent code for the Fmunu tensor, without the Clover parameter dependence, so other potential applications (like the topological charge) don't have to duplicate the Fmunu calculation.

mikeaclark:
* Computation of the field strength tensor is now done in the field_strength_tensor.cu file, and the `computeFmunu` function is used to achieve this (closes #277)
* Use `gaugefield::FloatNOrder` accessors for accessing Fmunu elements.

@AlexVaq can you test this is working as it should be?  Also, in `createCloverQuda` (interface_quda.cpp,), why do you compute the clover term twice when doing twisted clover fermions?  If you need a second clover field, why not just make a copy of the first one?  Perhaps I am missing something.

### DIFF
--- a/include/gauge_tools.h
+++ b/include/gauge_tools.h
@@ -1,4 +1,29 @@
 namespace quda {
-  double	plaquette	(const GaugeField& data, QudaFieldLocation location);
-  void		APEStep		(GaugeField &dataDs, const GaugeField& dataOr, double alpha, QudaFieldLocation location);
+  /**
+     Compute the plaquette
+
+     @param gauge The gauge field upon which to compute the plaquette
+     @param location The location of where to do the computation
+     @return The plaquette
+   */
+  double plaquette (const GaugeField& gauge, QudaFieldLocation location);
+
+  /**
+     Apply APE smearing
+
+     @param dataDs Destination gauge field
+     @param dataOr Source gauge field
+     @param alpha Smearing parameter
+     @param location Location of the computation
+   */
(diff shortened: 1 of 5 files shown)


AlexVaq:
> @AlexVaq can you test this is working as it should be?
> 
Sure!

> Also, in createCloverQuda (interface_quda.cpp,), why do you compute the clover term twice when doing twisted clover fermions? If you need a second clover field, why not just make a copy of the first one? Perhaps I am missing something.
> 
In the early days, copying never worked for me and would give me wrong results. I don't know why because for mixed precision you have to do it anyway, and everything looks all right.

I'll test this too (substituting computing by copying). Maybe it's something related to the clover field parameters...

mikeaclark:
Ok, it would be good to know if there's a bug in the clover copy somehow.  Anyway, given that the field-strength tensor construction has been separated, at least that won't be called twice now.

mikeaclark:
I'm pretty sure this is ready for merging now.  It passes clover construction tests in both single- and multi-GPU.

AlexVaq:
It works for me perfectly well for twisted clover. I'm going to test the topological charge, which is a completely different application, to see if there are other issues.

AlexVaq:
I need to work a bit more with the topological charge, but I think you can safely merge. So far the Fmunu is only used for the clover term computation, and in that case this works perfect.

PR #296, opened 2015-06-25
use maxregcount only for the Tesla architecture

mathiaswagner:


mikeaclark:
What's with the failing continuous-integration checks?  I guess this is a work in progress?

mathiaswagner:
I had hoped to include some checks inspired by reading https://github.com/blog/1935-see-results-from-all-pull-request-status-checks 

I tried to check whether one can at least run some linter on the code (building does not work as travis-ci has no support for cuda) but apparently I messed that up. Sorry for the confusion. I hope it is turned off now. 

work in progress was what I hoped it would be but looks more like fail in progress (or fail completed for now)

PR #300, opened 2015-06-26
add nvtx markers to milc interface

### LINKED ISSUE #262: Use qudamilc_called only in exposed functions of the QUDA interface
`qudamilc_called` provided only debug output when entering and exiting quda functions, so we know whether control is within QUDA or MILC.
In the somewhat messed up MILC interface this is just another thing that needs to be cleaned up.

mathiaswagner:
This adds the option to have nvtx markers in the MILC interface.
This also required the consequent bracketing of all MILC interface functions with `qudamilc_called`.
Closes #262.


mikeaclark:
Looks good :+1:

PR #307, opened 2015-07-06
Added separate temporal-spatial plaquette

AlexVaq:
This should add the missing feature

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -799,8 +799,18 @@ extern "C" {
                       const void* const u_link,
                       const QudaGaugeParam* param);
 
+  /**
+   * Computes the spatial and temporal plaquette of the loaded gauge configuration.
+   */
+  void plaqCuda(double&, double&);
+
+  /**
+   * Performs APE smearing on gaugePrecise and stores it in gaugeSmeared
+   * @param nSteps Number of steps to apply.
+   * @param alpha  Alpha coefficient for APE smearing.
+   */
   void performAPEnStep(unsigned int nSteps, double alpha);
-  double plaqCuda();
+
   /**
   * Open/Close MAGMA library
   *
--- a/include/quda.h
+++ b/include/quda.h
@@ -800,9 +800,10 @@ extern "C" {
                       const QudaGaugeParam* param);
 
   /**
-   * Computes the spatial and temporal plaquette of the loaded gauge configuration.
+   * Computes the total, spatial and temporal plaquette averages of the loaded gauge configuration.
+   * @param Array for storing the averages (total, spatial, temporal)
    */
-  void plaqCuda(double&, double&);
+  void plaqQuda(double plaq[3]);
 
   /**
    * Performs APE smearing on gaugePrecise and stores it in gaugeSmeared
--- a/include/gauge_field.h
+++ b/include/gauge_field.h
@@ -153,12 +153,6 @@ namespace quda {
     /** Whether the staggered phase factor has been applied */
     bool staggeredPhaseApplied;
 
-    /**
-       This function returns true if the field is stored in an
-       internal field order for the given precision.
-    */ 
-    bool isNative() const;
-
   public:
     GaugeField(const GaugeFieldParam &param);
     virtual ~GaugeField();
@@ -178,7 +172,6 @@ namespace quda {
     const int* R() const { return r; }
     QudaGhostExchange GhostExchange() const { return ghostExchange; }
     QudaStaggeredPhase StaggeredPhase() const { return staggeredPhaseType; }
-
     /**
        Apply the staggered phase factors to the gauge field.
      */
@@ -193,6 +186,12 @@ namespace quda {
     int Nface() const { return nFace; }
   
     void checkField(const GaugeField &);
+
+    /**
+       This function returns true if the field is stored in an
+       internal field order for the given precision.
+    */ 
+    bool isNative() const;
 
     size_t Bytes() const { return bytes; }
     size_t PhaseBytes() const { return phase_bytes; }
--- a/include/gauge_tools.h
+++ b/include/gauge_tools.h
@@ -1,4 +1,4 @@
 namespace quda {
-  double	plaquette	(const GaugeField& data, QudaFieldLocation location);
-  void		APEStep		(GaugeField &dataDs, const GaugeField& dataOr, double alpha, QudaFieldLocation location);
+  void plaquette (const GaugeField& data, QudaFieldLocation location, double &plqS, double &plqT);
+  void APEStep   (GaugeField &dataDs, const GaugeField& dataOr, double alpha, QudaFieldLocation location);
 }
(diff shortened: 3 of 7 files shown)


AlexVaq:
I didn't check multiGPU. Could you do this for me? It's really late now here, and tomorrow I have to wake up at an absurd hour...

mikeaclark:
Sure, no problem.  I might not get to it today anyway, as I've 101 other things I'm working on...

mikeaclark:
Just had a quick look: this looks great, thanks for working on this so quickly :)  This will likely be an easy fix for me tonight, I'll try to get it done.  Note that there's no need to have the MULTI_GPU macro blocks in there as comm_dim and comm_reduce should just work in single GPU mode too; I'll clean this up a bit.

mathiaswagner:
Hate to be the bad guy: TESTS??? 😉

mikeaclark:
A plaquette test really requires a non-trivial gauge field: this either means loading a previously defined gauge field, computing a plaquette on some random gauge with known answer or doing some kind of gauge evolution.  We can't include a gauge field in QUDA, and the last option requires the merge of the gauge-fix branch.

We could include the plaquette computation in the regular tests directly in CPU code, but this is a lot of work.  This pull is an epsilon change, so the lack of testing isn't changing anything, just exposing a previously present problem.

Make a new issue for doing the reference plaquette computation on the host?

mathiaswagner:
Ok, so for now we should rely on some external tests (the code that will use the temporal and spatial plaquette  ?). 

For the long run we should create an issue for the test code. 

Also thinking about creating a QUDA reference data repository but that can be discussed at Fermilab. 

Sorry to be picky about that but plaquette is something we should not screw up.

AlexVaq:
Well, I meant I tested it with some non-trivial configurations and I didn't wqnt to recompile at 1:30am with h multiGPU. So what I meant is "just recompile with multiGPU and check that the results are correct".

I might do this today, if I find a spot and I don't fall sleep aroumd.

mikeaclark:
Ok, something I'm confused about here.  Shouldn't the convention be such that the spatial and temporal plaquettes each alway vary in the range [0,1].  As such, shouldn't the normalization be 
```c++
  arg.plaq_h[0].x /= 3.*(arg.threads*nNodes);
  arg.plaq_h[0].y /= 3.*(arg.threads*nNodes);
```
and then when computing the total plaquette, we should have
```c++
  plaq = 0.5 * (plaq_spatial + plaq_temporal);
```
Also, the flop and byte numbers look wrong, but I can fix these easily.

AlexVaq:
Yes, I didn't touch these things, and I wanted to warn you (it was really late). Since I never used this plaquette I didn't know the conventions, but I was feeling that it was wrong.

And the flops and the bytes... Yep, they were wrong all the time. So let me check today multiGPU with real confs in the hackathon, and maybe you can help me with these other things.

nmrcardoso:
that's correct, but since the getTrace dont't have the 1/3,   it should be
arg.plaq_h[0].x /= 9.*(arg.threads*nNodes);
arg.plaq_h[0].y /= 9.*(arg.threads*nNodes);

mikeaclark:
@nmrcardoso: you're correct.  Not sure how I divided 18 by 2 and got 3 :)

mikeaclark:
Ok, I've just pushed some changes that should address all of the above:
* Interface was C++: it needs to be valid C for linking to C/Fortran apps (also change to plaqQuda, instead of plaqCuda)
* Tuning flops / bytes should be correct
* Cleaned up unnecessary `#ifdef MULTI_GPU`
* Halved the number of kernel compiled using traits to only compile the precision / reconstruct / order types needed (we should probably apply this to other parts of the library to help compile time / footprint)
* Changed the `plaquette` function signature such that it returns a `double3` with (all, spatial, temporal) plaquette values returned 
* Updated the tests/su3_test to compute the plaquette after downloading the gauge field (this also provides a simple sanity check)
* Added Fortran interface to allow calling this from Fortran apps

AlexVaq:
Great! I’m compiling multiGPU now, will let you know if there’s something broken. I’ll test on a single GPU with the partition option, though. I think this is all right.

AlexVaq:
Placate works flawlessly with the partition option in multiGPU. Are we ready for merging?

AlexVaq:
Sorry, I meant plaquette. The autocorrector is evil...

mathiaswagner:
Fine with me. So before we merge:
1. @mikeaclark Do you want to test the code with the customer code before:
2. Create an issue for the tests: @AlexVaq maybe you can do that and add a short comment what you already did for testing?
3. I will add an issue to move the `atomicAdd` for double somewhere else.

Anything else?

mathiaswagner:
2. and 3. are done in #308 and #309.

mikeaclark:
Ok, I've made one more push.  This adds profiling info to `plaqQuda`, and I've also modified the gauge force computation, such that the extended gauge field is made resident if `QudaGaugeParam::make_resident_gauge=1` is set.

I've checked this code against the TIFR code, and it agrees to 12 decimal places, so I think we're good.  So this is ready for merging.

AlexVaq:
I'm working on singleGPU, CPU test for this.

include/quda.h
```diff
@@ -799,8 +799,19 @@ extern "C" {
                       const void* const u_link,
                       const QudaGaugeParam* param);
 
+  /**
+   * Computes the total, spatial and temporal plaquette averages of the loaded gauge configuration.
+   * @param Array for storing the averages (total, spatial, temporal)
+   */
+  void plaqQuda(double plaq[3]);
```
mathiaswagner: As we want to merge that into the 0.7.x release and this changes the interface: Shall we just for the sake of avoid to break anything also keep the `double plaqCuda()` signature (which should be `plaqQuda`).

This would only be relevant if this is used and would break a code.

mathiaswagner:
Thanks, @AlexVaq. Are you doing this in the same branch?

include/quda.h
AlexVaq: Well, that could be easily done. Actually, if you want, we can call the two functions the same (plaqCuda I assume) and overload them, since they have different arguments.

AlexVaq:
No, I assume this test is not urgent, and I’m touching only test_util.cpp (and its header), so I suppose this can go in any branch. Also I don’t expect to finish multiGPU anytime soon, but it can be a good addition anyway.

mikeaclark:
I would argue we should *not* support the `plaqCuda` function call:
* Only Alex would have been using it
* It uses C++ references as arguments, in an `extern C` block.  That just sounds yucky to me.

AlexVaq:
I won’t oppose removing it either, so do as you wish.

mathiaswagner:
Agreed. We can live without it. 
I just wanted to pause before rushing into interface changes. If only Alex used it I have no objections and will merge this.

mathiaswagner:
I merged that in but will keep the branch alive for now. 
@AlexVaq Let me know if you want to recycle that branch for the test you are working on or whether you want to use a separate branch.

AlexVaq:
I might very well use this one, but I can also create a new one, no problems.

My code works (singleGPU only), I didn’t push it because 1. it’s a bit dirty yet and 2. I couldn’t work with these neighborFullLattice functions and other stuff in test_util.cpp to pass from coordinates to indices and viceversa. I really don’t know what’s wrong, I just know that, until I wrote my own code, I was getting the wrong result.

I’ll have to look into that.

mathiaswagner:
I leave that up to you. Thanks again for putting that together!

PR #317, opened 2015-07-16
Hotfix/dslash cleanup

### LINKED ISSUE #315: Tuned bandwidth numbers for some 5-d operators looks wrong
Flops look correct, but the bandwidth numbers for the 5-d local kernels look way off.  A quick look at the code suggests that these aren't special cased, and the default `DslashCuda::bytes()` is being used, but I'll need to check for more than 30 seconds to work this out exactly.

```
Tuned block=(128,3,1), shared=24577 giving 71.00 Gflop/s, 1116.09 GB/s for N4quda16MDWFDslashPCCudaI7double2S1_EE with type=single-GPU,reconstruct=18,Dslash4pre
Tuned block=(32,4,1), shared=0 giving 170.86 Gflop/s, 347.94 GB/s for N4quda16MDWFDslashPCCudaI7double2S1_EE with type=single-GPU,reconstruct=18,Dslash4
Tuned block=(32,4,1), shared=16385 giving 155.89 Gflop/s, 181.87 GB/s for N4quda16MDWFDslashPCCudaI7double2S1_EE with type=single-GPU,reconstruct=18,Dslash5inv
Tuned block=(32,2,1), shared=24577 giving 64.99 Gflop/s, 895.91 GB/s for N4quda16MDWFDslashPCCudaI7double2S1_EE with type=single-GPU,reconstruct=18,Xpay,Dslash5
Executing 10 kernel loops...
```

mikeaclark:
This pull request:
* Only compile fused dslash kernels when `MULTI_GPU` is set
* Fixes the memory bandwidth numbers for the 4-d domain-wall kernels (closes #315)

### DIFF
--- a/lib/dslash_domain_wall_4d.cu
+++ b/lib/dslash_domain_wall_4d.cu
@@ -215,6 +215,28 @@ namespace quda {
       }
       return flops;
     }
+
+    long long bytes() const {
+      bool isHalf = in->Precision() == sizeof(short) ? true : false;
+      int spinor_bytes = 2 * in->Ncolor() * in->Nspin() * in->Precision() + (isHalf ? sizeof(float) : 0);
+      long long Ls = in->X(4);
+      long long bytes;
+
+      switch(DS_type){
+      case 0:
+	bytes = DslashCuda::bytes();
+	break;
+      case 1:
+	bytes = (x ? 5ll : 4ll ) * spinor_bytes * in->VolumeCB();
+	break;
+      case 2:
+	bytes = (x ? Ls + 2 : Ls + 1) * spinor_bytes * in->VolumeCB();
+	break;
+      default:
+	errorQuda("invalid Dslash type");
+      }
+      return bytes;
+    }
   };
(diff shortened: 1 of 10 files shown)


lib/dslash_domain_wall_4d.cu
```diff
@@ -215,6 +215,28 @@ namespace quda {
...
       return flops;
     }
+
+    long long bytes() const {
+      bool isHalf = in->Precision() == sizeof(short) ? true : false;
+      int spinor_bytes = 2 * in->Ncolor() * in->Nspin() * in->Precision() + (isHalf ? sizeof(float) : 0);
+      long long Ls = in->X(4);
+      long long bytes;
+
+      switch(DS_type){
```
mathiaswagner: While this is not introduced here and does not stop me from merging this pull request: is an enum a better choice for DS_type? In the long haul we might go for strongly typed C++11 enum but that is some time to go until we can drop support for CUDA versions without C++11 support.

PR #319, opened 2015-07-17
Hotfix/compile time

mikeaclark:
This pull significantly reduces the compilation time of algorithms that use `gauge::FloatNOrder` and `clover::FloatNOrder`.  This is done through utilizing C++ traits and the structs `gauge_mapper` and `clover_mapper` to only instantiate the correct field order templates for a given precision / reconstruct type.

There's more that could be done here, but I've left out some files that I know have been changed in the develop branch so we can easily merge this back into develop.

### DIFF
--- a/include/gauge_field_order.h
+++ b/include/gauge_field_order.h
@@ -945,4 +945,29 @@ namespace quda {
   };
 
 
-}
+  
+  // Use traits to reduce the template explosion
+  template<typename ,QudaReconstructType,int N=18> struct gauge_mapper { };
+
+  // double precision
+  template<int N> struct gauge_mapper<double,QUDA_RECONSTRUCT_NO,N> { typedef FloatNOrder<double, N, 2, N> type; };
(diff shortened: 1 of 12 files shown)


include/gauge_field_order.h
```diff
@@ -945,4 +945,29 @@ namespace quda {
   };
 
 
-}
+  
+  // Use traits to reduce the template explosion
+  template<typename ,QudaReconstructType,int N=18> struct gauge_mapper { };
```
mathiaswagner: There might be reason for the comma before `QudaReconstructType` but that seem strange to me. I have not yet asked the compiler for comments.

mikeaclark: I just left the typename unnamed.  It compiles fine :)

mathiaswagner: Seems like you never stop learning C++ 😉. Makes sense though but can't remember having seen that before.

PR #322, opened 2015-07-23
fix compilation error in gauge_plaq 

mathiaswagner:
about undefined plq if GPU_GAUGE_TOOLS is not defined


nmrcardoso:
there is a typo in the new version double2 -> double3, line 270.

mikeaclark:
That's not a typo.  The plaquette function on line 246 returns a double2.

PR #336, opened 2015-07-31
Hotfix/blas magma issue

### LINKED ISSUE #335: blas_magma.cpp does not compile
On the develop branch after the recent merge of the mixed-precision eigCG support (62868e52cd701ba86b9c27684a0ddbc77fcede16).  There are two issues that I can see:
* This is a C++ code yet it uses C99 complex numbers.  These are not recognised by strict C++ compilers (e.g., clang/LLVM) and these fail to compile.
* There are many parts of the code that are not protected from compilation if MAGMA_LIB is not defined.

There are dozens of errors, here's a sample :)
```
blas_magma.cpp:523:46: error: use of undeclared identifier 'magmaDoubleComplex'
    cudaHostRegister(pQR, ldqr*nrefls*sizeof(magmaDoubleComplex), cudaHostRegisterMapped);
                                             ^
blas_magma.cpp:1047:17: error: use of undeclared identifier 'creal'
  lwork = (int)(creal(qwork));
```

@alexstrel Please fix this as soon as possible inserting the appropriate macros to prevent compilation and replace the C99 complex numbers with C++ complex numbers (use quda::complex or std::complex).

alexstrel:
This should fix #335


mikeaclark:
This closes #355.

PR #346, opened 2015-08-05
Hotfix/dev timerprofile

### LINKED ISSUE #344: running timer
Running the gauge_alg_test in develop branch i'm getting the error:
ERROR: Cannot start an already running timer (rank 0, host hybrid, ../include/quda_internal.h:127 in Start())
       last kernel called was (name=N4quda9GaugePlaqIfNS_11FloatNOrderIfLi18ELi2ELi18EEEEE,volume=28x28x28x28,aux=threads=614656,prec=4)
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0

mikeaclark:
This pull request fixes some issues with the TimeProfile class that came up in the gauge-fixing routines after the global timer was added.
* Disable global timer in algorithms, only enable this in interface timing to prevent double starting the global timer
* Add trace information for TimeProfile Start / Stop routines.  Now `TimeProfile::TPSTART` and `TimeProfile::TPSTOP` must be called instead of calling Start / Stop directly
*Fixed a bug in quda_internal.h that resulted in double starting a timer

This closes #344.

### DIFF
--- a/include/quda_internal.h
+++ b/include/quda_internal.h
@@ -123,14 +123,20 @@ namespace quda {
 
   Timer() : time(0.0), last(0.0), running(false), count(0) { ; } 
 
-    void Start() {
-      if (running) errorQuda("Cannot start an already running timer");
+    void Start(const char *func, const char *file, int line) {
+      if (running) {
+	printfQuda("ERROR: Cannot start an already running timer (%s:%d in %s())\n", file, line, func);
+	errorQuda("Aborting");
+      }
       gettimeofday(&start, NULL);
       running = true;
     }
 
-    void Stop() {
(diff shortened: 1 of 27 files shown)


mikeaclark:
@nmrcardoso This all looks good to me.  You have verified this against MILC, to check that it still works ok haven't you?

nmrcardoso:
yes, no errors.

mathiaswagner:
Thanks for getting this fixed. I will have to look at why this failed. 

I had hoped to implement the global timer in a way that starting it while already running only increases a counter to make sure it is only stopped when that counter was decreased to 0. 

But I had never tested with the gauge fixing merged in.

PR #348, opened 2015-08-10
moved typedef for Complex in blas_magma into class scope 

mathiaswagner:
to avoid compilation issues because of additional typedef

```
../include/blas_magma.h:111: error: reference to ‘Complex’ is ambiguous
../include/blas_magma.h:14: error: candidates are: typedef struct std::complex<double> Complex
../include/eig_variables.h:13: error:                 typedef struct std::complex<double> quda::Complex
```

Might be that only the oooooooooold gcc on the local GPU development machine in Bloomington is picky about that but it might be safer to have that in the class scope anyway.

### DIFF
--- a/include/blas_magma.h
+++ b/include/blas_magma.h
@@ -11,9 +11,12 @@
 //MAGMA library interface 
 //required for (incremental) EigCG solver
 
-   typedef std::complex<double> Complex; //same notations
+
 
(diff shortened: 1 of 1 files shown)


mikeaclark:
@alexstrel Can you review this is ok?

alexstrel:
Just tested, it's ok.

PR #351, opened 2015-08-12
Feature/gauge reduce

### LINKED ISSUE #349: Kernels that use the RNG class are not reproducible when autotuning
I noticed when running the gauge_alg_test that the answer changes for the first test depending on whether autotuning is occurring.  This is because the state of the RNG generator is not being restored after the tuning has completed, giving a different answer if the tuning is disabled or if it has already previously been performed.  This affects the heatbath algorithm and the creation of an initial hot state.

We should restore the initial RNG state to ensure reproducibility regardless of whether autotuning is occurring or not.

@nmrcardoso can you take a look at this?  Since you wrote the RNG class, I think you are likely the best person to do this.

mikeaclark:
* Introduce `reduce` and `reduce2d` helper functions for easy reductions.
* Replace `atomicAdd` calls with calls to `reduce` in plaquette and gauge fixing algorithms.  For the plaquette, this results in a 2x performance boost. 
* Reproducible answers when tuning algorithms that use the RNG class (closes #349)
* Performance fixes for single-precision reconstruct-8 `gauge::FloatNOrder` accessor.  This doubles the performance of the equivalent plaquette computation.
* Small improvement in `unitarizationLinksQuda()` performance through some precalculation of divisors.
* Renamed random.h to random_quda.h to prevent include name collisions with external applications.
* Patched the TimeProfile class to prevent double-start Timer errors.
* Fixed mismatch of TimeProfile labels with the appropriate enum type


mikeaclark:
@mathiaswagner can you take a look when you get the chance?  I've tested fairly well, and it seems robust.

mathiaswagner:
Will do.

mathiaswagner:
I feel like everything is falling apart here:

The Cray
```
/opt/nvidia/cudatoolkit6.5/6.5.14-1.0502.9613.6.1//bin/nvcc  -O3 -D__COMPUTE_CAPABILITY__=350 -ftz=true -prec-div=false -prec-sqrt=false  -w -DMULTI_GPU -DGPU_WILSON_DIRAC -DGPU_DOMAIN_WALL_DIRAC -DGPU_STAGGERED_DIRAC -DGPU_CLOVER_DIRAC -DGPU_WILSON_DIRAC -DGPU_GAUGE_TOOLS -DGPU_TWISTED_MASS_DIRAC -DGPU_WILSON_DIRAC -DGPU_GAUGE_TOOLS -DGPU_GAUGE_ALG -DGPU_GAUGE_TOOLS -DGPU_UNITARIZE -DGPU_DIRECT -DBUILD_QDP_INTERFACE -DBUILD_MILC_INTERFACE -DNUMA_AFFINITY -arch=sm_35	   -I/opt/nvidia/cudatoolkit6.5/6.5.14-1.0502.9613.6.1//include -DMPI_COMMS  -I/include -I../include -Idslash_core -I.  gauge_fix_ovr.cu -c -o gauge_fix_ovr.o
terminate called after throwing an instance of 'std::bad_alloc'
  what():  St9bad_alloc
```
@nmrcardoso Sounds familiar ? But this is not Tegra K1 :cry:
I asked the admins to fix this as I believe it is an artificial memory limit.


and the local development machine has the known `const int Ns` issues with the old gcc version (as discussed at FNAL). I need to patch that  and then hopefully get it tested soon ...

tests/unitarize_link_test.cpp
```diff
@@ -36,7 +36,7 @@ static double unitarize_eps  = 1e-6;
 static bool reunit_allow_svd = true;
 static bool reunit_svd_only  = false;
 static double svd_rel_error  = 1e-4;
-static double svd_abs_error  = 1e-5;
+static double svd_abs_error  = 1e-4;
```
mathiaswagner: Is this required to make the test pass? Did it pass with the stricter error before the optimization?

mikeaclark: This represent the cut off when SVD is used, and so this is a stricter condition (higher number means SVD is used more). This change was needed for the test to pass at large volumes.

mathiaswagner: Ok, I probably was confused by the error name while this is rather a threshold / tolerance. Anyway, fine with that.

mikeaclark: It is poorly named.

mikeaclark:
Regarding the crashing build with 6.5.  Do you think this would go away if this file was split up more to reduce memory, reduce the number of kernels instantiated or if a serial build is used?

mathiaswagner:
After the parallel build failed I tried serial. I believe it is a system issue (they previously had a 2GB VRAM limit). 
Splitting probably reduces the memory footprint and might help. 
I have not tried on blue waters or Titan but am optimistic that this will not happen there. 
Not sure whether it is best to just wait for the admins to increase the VRAM limit.

mathiaswagner:
Just ran into
```
gauge_fix_fft.cu(282): error: argument list for class template "quda::ReduceArg" is missing
          detected during:
            instantiation of "void quda::gaugefixingFFT<Elems,Float,Gauge,gauge_dir>(Gauge, quda::cudaGaugeField &, unsigned int, unsigned int, Float, unsigned int, double, unsigned int) [with Elems=9U, Float=float, Gauge=quda::FloatNOrder<float, 18, 2, 18>, gauge_dir=4]" 
(1337): here
            instantiation of "void quda::gaugefixingFFT<Elems,Float,Gauge>(Gauge, quda::cudaGaugeField &, unsigned int, unsigned int, unsigned int, Float, unsigned int, double, unsigned int) [with Elems=9U, Float=float, Gauge=quda::FloatNOrder<float, 18, 2, 18>]" 
(1359): here
            instantiation of "void quda::gaugefixingFFT(quda::cudaGaugeField &, unsigned int, unsigned int, unsigned int, Float, unsigned int, double, unsigned int) [with Float=float]" 
(1404): here
```

while trying to compile.

mikeaclark:
That's weird that I haven't seen that issue.  GCC version issue perhaps?  I'm using 4.8.  Taking a look.

mathiaswagner:
Maybe gcc version again ... time to switch to a development machine with reasonably recent software. Will try on the Cray again.

mikeaclark:
Looking at it, I'm actually surprised that it compiles!  Will push a fix shortly.

mikeaclark:
Ok, this issue should be fixed now.

PR #352, opened 2015-08-14
Trigger reliable updates only from shift 0 in multi shift cg

mathiaswagner:
This is sufficient with the current implementation. Still keep the loop and arrays to be able to trigger on higher shifts again in the future without having to add back in that code.

The multi shift solve needs less reliable updates and is slightly faster (time including the refinement phase)

### DIFF
--- a/lib/inv_multi_cg_quda.cpp
+++ b/lib/inv_multi_cg_quda.cpp
@@ -227,11 +227,11 @@ namespace quda {
       //fixme: The loop below is unnecessary but I don't want to delete it as we still might find a better reliable update
       int reliable_shift = -1; // this is the shift that sets the reliable_shift
       for (int j=0; j>=0; j--) {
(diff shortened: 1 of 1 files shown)


mikeaclark:
Any thoughts on how we can make the refinement do double-half, but double-single in the solve?  Is this an easy patch?

mikeaclark:
Style comment: you've used 6 space indent here not the usual 2 space indents as is generally used (and advocated).

mathiaswagner:
sorry, that is messed up because the files is partially indented using tabs and I have set my tab width to two ....
I'll remove the tabs.

mathiaswagner:
No idea yet what is the easiest way to use a different sloppy precision ...

mikeaclark:
I guess we can either:
* use the cuda_prec_precondition precision for the refinement 
* make a new parameter in `QudaInvertParam`

I realize I should introduce this discussion elsewhere.  :)

PR #353, opened 2015-08-17
Fixed timer nomenclature

AlexVaq:
Compilation was broken due to the inclusion of different timer functions. We should fuse this part of the code with dslash_policy.cuh at some point...


AlexVaq:
Never mind, I'm tracking a much more serious bug, but I don't even know how to name it at this moment.

PR #356, opened 2015-08-17
Hotfix/magma fermi

alexstrel:


lib/Makefile
```diff
@@ -259,7 +259,7 @@ blas_magma.o: blas_magma.cpp $(HDRS)
 	$(CXX) $(MAGMA_INCLUDE) $(CXXFLAGS) $(MAGMA_FLAGS) $< -c -o $@
 
 blas_magma_gpu.o: blas_magma.cu $(HDRS)
-	$(NVCC) $(NVCCFLAGS) $(MAXREG) $< -c -o $@
+	$(NVCC) -DMAGMA_LIB $(NVCCFLAGS) $(MAXREG) $< -c -o $@
```
mathiaswagner: What is the point of always defining `MAGMA_LIB` here?
If it is always defined anyway, can't you just do that in the source file?

Or should that be a configure option?

alexstrel: corrected

mathiaswagner:
While looking at this: Is there a good reason why the naming is las_magma.cpp and blas_magma.cu and they generate a blas_magma.o and blas_magma_gpu.o ?

alexstrel:
it's right point, there is no reason now to have two files , that was an artifact of the former development process. I merged all into a single file blas_magma.cu.

mikeaclark:
This looks fine to merge now.  Thanks Alexei for the clean up.

PR #357, opened 2015-08-19
Fixed rare multiGPU bug

AlexVaq:
Added stream synchronization in the apply function to ensure that the output spinor is updated before operating with it.


AlexVaq:
This should be merged soon. We found the bug running into production issues for certain multiGPU node configurations.

I'm not aware of anyone using this code apart from my team, but just in case it should be updated.

AlexVaq:
With this commit (I hope) all the bugs that were affecting our latest production runs (we never had issues before) are fixed, so I guess that means the covariant derivative is fool-proof now.

Issue #323, opened 2015-07-23
Build quda using cmake


mathiaswagner:
I am working on a cmake build of quda.

This should come with:
* support for build directories (e.g. different directories for different SM)
* support for different build types (release, debug, ...)
* automatic dependency generation (faster recompilation)

This issue is just to document and track progress, missing features from the current autoconf/make build and discuss issues.

Everything is still in a private branch *mw/cmake*.

Building works but does not support any options from the configure script yet.

**Current configure options** (serves as a todo list)

- <del>enable-cpu-arch=arch  Set CPU arch: x86 or x86_64 (default: x86_64)</del> *Only 64 bit in the future*
- [x] enable-gpu-arch=arch  Set GPU arch: sm_13, sm_20, sm_21, sm_30, sm_35,
                          sm_50, sm_52 (default: sm_35)
- [x] enable-os=os          Set operating system: linux, osx (default: linux)
- [x] enable-host-debug     Enable debugging of host code
- [x] enable-device-debug   Enable debugging for device code
- [x] enable-verbose-build  Display kernel register usage
- [x] enable-qdp-interface  Build the QDP interface (default: enabled)
- [x] enable-milc-interface Build the MILC interface (default: enabled)
- [x] enable-cps-interface  Build the CPS interface (default: disabled)
- [x] enable-qdpjit-interface
                          Build the QDPJIT interface (default: disabled)
- [x] enable-bqcd-interface Build the BQCD interface (default: disabled)
- [x] enable-tifr-interface Build the TIFR interface (default: disabled)
- [x] enable-wilson-dirac   Build Wilson Dirac operator (default: enabled)
- [x] enable-clover-dirac   Build Wilson-clover Dirac operator (default:
                          enabled)
- [x] enable-domain-wall-dirac
                          Build domain wall Dirac operator (default: enabled)
- [x] enable-staggered-dirac
                       Build staggered Dirac operator (default: enabled)
- [x] enable-twisted-mass-dirac
                          Build twisted mass Dirac operator (default: enabled)
- [x] enable-ndeg-twisted-mass-dirac
                          Build non-degenerate twisted mass Dirac operator
                          (default: disabled)
- [x] enable-twisted-clover-dirac
                          Build twisted mass clover Dirac operator (default:
                          disabled)
- [x] enable-staggered-fatlink
                          Build asqtad link-fattening routines (default:
                          disabled)
- [x] enable-hisq-fatlink   Build HISQ link-fattening routines (default:
                          disabled)
[...]

PR #364, opened 2015-09-14
Feature/cmake

mathiaswagner:
I believe the cmake build is no ready for broader testing. 
There are still some open issues in #323 but most things should work by now.

It does not interfere with the current build system if using with out of source builds, e.g. like

```
mkdir <build-dir>; cd <build-dir>
cmake <quda-src>
```

Further options can be edited using e.g. `ccmake`, a graphical cmake version or editing `CMakeCache.txt`.


mikeaclark:
Ok, I've started some testing.  Some comments:
* I guess the ordering of the CMakeCache.txt file is alphabetical.  This means that the the interface options are not contiguous, and neither are the Dirac operator options.  This makes setting options a bit of a pain.  Can this ordering be made custom?
* When editing the `CMAKE_BUILD_TYPE` there is a typo 
"CMAKE_BUILD_TYPE: Choose the type of build, options are: Devel **Relaese** Debug Release HostDebug DeviceDebug"

mikeaclark:
A more serious issue: the plaquette seems to fail to build using cmake, but compiles fine with make.  The error it gives complains about the amount of shared memory use by the kernel:
```c++
ptxas error   : Entry function '_ZN4quda11computePlaqILi960EfNS_11FloatNOrderIfLi18ELi4ELi8EEEEEvNS_12GaugePlaqArgIT1_EE' uses too much shared data (0xc320 bytes, 0xc000 max)
`
It's only complaining about the kernels with 960, 992 and 1024 threads per block.

mathiaswagner:
* The ordering in the CMakeCache.txt unfortunately is a pain. The only option are better names for the options (i.e. prefix everything with QUDA). The GUI (I guess that will not really be used) has an option to generate at least one level of folding (using '_' as separator). See screenshot.
screen shot 2015-09-16 at 19 52 20

* I should fix that.

* CMake supports a boolean type (the on / off), strings, paths but unfortunately (I have been looking for that) no dropdown lists. We can however add a check within in cmake whether it is a valid build type.

mikeaclark:
So it sounds like to group together the build options we need to rename them.  E.g., to get the Dirac operators together we should use `BUILD_STAGGERED_DIRAC` -> `BUILD_DIRAC_STAGGERED`.

mathiaswagner:
We should probably rename all options for QUDA to be prefixed with QUDA.
Not sure whether we should rely on the grouping even more and use also QUDABUILD (which will show up as another group (cf screenshot above)).

mathiaswagner:
I have added a check for a vlid build type.

mikeaclark:
It's question time!
* Am correct in thinking that the build option names using cmake are completely independent of the names using configure/make?  So there's nothing to stop us from renaming these for cmake without touching the configure/make build system is there?
* What's the process of adding subdirectories using cmake?  Just thinking about source code clean up and moving to a more hierarchical structure.
* With cmake can we still build in the source quda directory if we want to (we just can't do this while also expecting to be able to use confugure/make.  Is that correct?

mathiaswagner:
Answer time:
* Yes. They are mostly the same now for historical reasons. As long as they trigger the correct defines. 
* We can do in-source builds. I currently have an option to forbid that as this would overwrite the current Makefiles.

mathiaswagner:
Subdirectories for sources should be easy. We could just specify the sources in the lib/CMakeLists.txt with path or we can use glob (I'd rather not).

mathiaswagner:
Oh and for headers we just have to tell cmake where to look. All dependencies should come for free.

mathiaswagner:
@mikeaclark I just managed to trigger the error for gaugeplag. I started with a fresh build directory and ran cmake. The NVCC_FLAGS were missing.
The problem could be fixed by running 'cmake .' again in the build directory. I need to figure out what is going on here.

bjoo:
HI Mathias, Kate,
 I have been trying to play with CMake, and wanted to try better integration with QMP. 
 I saw how Mathias did this in the CMakeLists file. I went for another approach, which 
is that I added a QMPConfig.cmake file to qmp (look in feature/cmake-config branch 
at github.com:usqcd-software/qmp.git )

This sets a bunch of CMake variables:

QMP_C_COMPILER   — the C compiler used to compile QMP
QMP_C_COMPILE_FLAGS — the compiler flags used to compile QMP       (turned into a list for ease of appending)
QMP_C_LINK_FLAGS         — linker flags used to link QMP test programs  (turned into a list for ease of appending)
QMP_C_LIBRARIES            — the various -l options to link wiht QMP          ( turned into a list for ease of appending)
QMP_C_INCLUDE_PATH    — the location of the include/ directory   
QMP_C_FOUND

In my test setup I used this the following way:

i) when I invoke cmake, I need to specify:   

> -DQMP_DIR="/home/bjoo/package-8-7-15/avx/install/qmp”


 to tell automake where to look, tho this can also be done by cc-make or the GUI or by appending to the CMAKE_PREFIX_PATH directly
  
ii) In my CMakeLists.txt I explicitly add QMP_DIR onto the prefix path if the user has defined QMP_DIR

>    if( QMP_DIR )
>      set( CMAKE_PREFIX_PATH "${CMAKE_PREFIX_PATH}" "${QMP_DIR}")
>    endif( QMP_DIR )
> 

iii) I can then do regular CMake 'find_package(QMP)’ and take regular configuration actions.


> # Find package QMP
> find_package( QMP REQUIRED )
> 
> # Check for QMP
> if ( ${QMP_C_FOUND} )
>   include_directories(${QMP_C_INCLUDE_PATH})
>   LIST(APPEND EXT_LIBS ${QMP_C_LINK_FLAGS} )
>   LIST(APPEND EXT_LIBS ${QMP_C_LIBRARIES} )
> endif( ${QMP_C_FOUND} )

So in the above case I add the include path to the global include_directories
and append the link options to a variable called EXT_LIBS

iv) when I then need to link an application I can do e.g. for some gtest program:

> add_executable(foo_test foo_test.cpp
> "${CMAKE_SOURCE_DIR}/external/gtest/gtest-all.cc")
> )
> add_compile_options(${EXT_CXXFLAGS})
> target_link_libraries(foo_test ${EXT_LIBS})

NB: Here I hava also an ${EXT_CXXFLAGS} which I didn’t set in step iii) but I may have it modified
by other find-packages (e.g. find_package ( Threads ) )

In any case I don’t know if this simpler than what is being done currently, but I am thinking it is heading more
towards the ‘CMake’ way (ie that packages like QMP should at least describe themselves etc). 

If you like this and it works for you, I’ll send a pull request to merge the addition of this QMPConfig.cmake file
into the master branch of QMP. 

Best, 
 B

> On Sep 18, 2015, at 12:43 AM, Mathias Wagner <notifications@github.com> wrote:
> 
> @mikeaclark I just managed to trigger the error for gaugeplag. I started with a fresh build directory and ran cmake. The NVCC_FLAGS were missing.
> The problem could be fixed by running 'cmake .' again in the build directory. I need to figure out what is going on here.
> 
> —
> 

Jefferson Lab
Newport News, VA 23606, USA
email: bjoo@jlab.org       

mathiaswagner:
I only had a quick look but if we can get QMP to provide the necessary configuration I always prefer that. 
Only issue is then we have to force people to use a recent enough QMP. 

I'm busy with some other thing this week but appreciate any feedback on cmake. 

Thanks, Balint!

mathiaswagner:
I know the cmake still has some rough edges that need polishing but for the automated builds it might be useful to get the cmake support into develop and keep the cmake branch alive for further improvements.
As I will need to setup different builds anyway this will also test the different builds using cmake so the automated builds first (pre-release ;-) ) milestone will be to test all the cmake options.

Does that sound useful?

mikeaclark:
As per @mathiaswagner's suggestion, merging this now, despite it not yet being perfect.  The present build system will continue to work while we migrate.

mathiaswagner:
So, this adds a small additional duty to reviewers and developers: If you modify the current build system please add these changes to the Cmake based build system or let me know so I can help with that.

For our current pull requests (from @alexstrel , @AlexVaq , @mikeaclark ) that are affected it might be the best idea to merge develop into the feature branch and then add the changes to the CMakeLists. Again, I am happy to help you out there.

mikeaclark:
Sounds like a good idea @mathiaswagner.  I guess this is similar to having to maintain the Fortran interface as well as the C interface.

mathiaswagner:
I keep working in the 'feature/cmake' branch. The issue with nvcc flags not being set correctly should now be fixed.

PR #365, opened 2015-09-16
Feature/deflation

alexstrel:
This will add GMRESDR solver (including mixed precision version).


mikeaclark:
@alexstrel Does the deflation_test include support for testing GMRES-DR?

alexstrel:
ups, no. will fix it.

alexstrel:
Ok, now should work

mikeaclark:
Thx, can you give some command-line instructions for how to test the different deflation algorithms (eigCG and GMRES-DR)?

mikeaclark:
I'm going through this, some question:
* Why is atlas a requirement, as opposed to generic blas?  
* On line 185 of make.inc.in, why is lib not added to LAPACK_HOME (but it is added to MAGMA_HOME / ATLAS_HOME) etc.?  This means that LAPACK_HOME has to be defined to include lib whereas the others don't

mikeaclark:
I've noticed there's huge scope for optimization of EigCG (and presumably GMRES-DR also).  EigCG spends most of time time doing BLAS1 linear algebra (e.g., 50% of the entire deflation_test is spent doing capxy and cDotProduct).  Most of the kernels have potential for kernel fusion, which will vastly reduce this overhead.  A small step in this direction to illustrate this is shown in commit 00f8a0804872732f53eb44935c9c69bd00ea6890.

alexstrel:
@mikeaclark : thanks a lot for comments and questions. GMRESDR now should work. (the problem with segfaults was not related to the algorithm). Concerning low accuracy of the eigenvectors: in how many iterations the eigCG finished? you can see how many eigenvector restarts was performed as well (Rayleigh-Ritz blocks). Any chance it was done at all? Otherwise it just shows that eigenvectors did not converge (not enough iterations?). could you try lower tolerance to increase iterations?  BTW, I set MAGMA version 1.7.0 as a default version for blas_magma.cu. Memory leak for the EigCG is fixed. Will answer other questions later on.

mikeaclark:
There is a simple solution to the problem with respect to whether the user is using cblas vs atlas vs openblas etc.  When MAGMA is installed it includes a configuration file for pkg-config:
```
more /usr/local/magma/lib/pkgconfig/magma.pc 
prefix=/usr/local/magma
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: magma
Description: Matrix Algebra on GPU and Multicore Architectures
Version: 1.7.0
Cflags: -I${includedir} -DADD_ -fopenmp -I/usr/local/cuda/include
Libs: -L${libdir} -lmagma -fopenmp -L/usr/local/cuda/lib64 -L/lib -lblas -llapack -lcublas -lcudart
Libs.private:
Requires: 
Requires.private:
```
This gives the information (on the `Libs` line) as to which BLAS and LAPACK libraries have been used to install MAGMA.  So the solution is:
1. Remove all mention of LAPACK_HOME / ATLAS_HOME from the configure / make process.
2. Replace the following lines from make.in.inc
```
  LIB += -fopenmp -L$(MAGMA_HOME)/lib -lmagma
  LIB += -L$(LAPACK_HOME) -llapack -lf77blas -lgfortran
  LIB += -L$(ATLAS_HOME)/lib -lcblas -latlas
  LIB += -L$(CUDA_INSTALL_PATH)/lib64 -lcublas
```
with
```
  LIB += $(shell pkg-config --libs magma)
```
This solution adds pkg-config dependency to QUDA, but this is a minor thing since it's included in all Linux installs that I know of, and it works on WIndows and Mac also.  The only very minor downside that I see at the moment is that the environment variable `PKG_CONFIG_PATH` has to be set
```
 export PKG_CONFIG_PATH=/usr/local/magma/lib/pkgconfig
```

@mathiaswagner I presume pkg-config support is easy with cmake?

Unless there are **strong** objections I will go ahead and make this change as this is needed to clean up the build system for the deflation branch.

mikeaclark:
@alexstrel the other thing that needs to be done is to replace the LAPACK calls with MAGMA ones to make it consistent between the different deflation algorithms.  Can you make this change?

alexstrel:
yes, will take care of it. I put this into highest priority. Thanks for the solution for proper MAGMA configuration in QUDA.

mathiaswagner:
Can one of the admins verify this patch?

mikeaclark:
@mathiaswagner Sorry, I don't follow what you want me to verify.

mikeaclark:
Scrub that I see that this was an automated message :smile:

mikeaclark:
All my questions above have been resolved now, with the exception of the choice of making MAGMA  dynamically link as opposed to statically linked into QUDA.  What was the motivation for this?

With respect to high eigenvector residuals that were being observed, this was purely because the mass of the operator was too high, so the solver converged very quickly, giving very imprecise solutions.  Running the solver with a lighter mass corrects this problem.

I've now removed the `ATLAS` and `LAPACK` dependencies from the configure / make build system, and so now it require that pkg-config be installed with `PKG_CONFIG_PATH` be pointed at `$MAGMA_DIR/lib/pkgconfig`.

I'm still working through the deflation test to understand how this works a bit better, but I think we should be able to get this merged next week.

mathiaswagner:
Just as a reminder: Once the build with Magma is cleaned up we need to add MAGMA support to CMake. Due to the ATLAS / LAPACK dependencies which seemed a bit odd to me I have pushed this behind. Remind me to pick up on here if I forget to add that to cmake.

mathiaswagner:
@Jenkins test this please

alexstrel:
@mikeaclark : I have no particular preference regarding static vs dynamic linking. For the previous versions of magma (i.e., prior v 1.5.X) only static linking were used.

alexstrel:
sorry closed by mistake

alexstrel:
@mathiaswagner : do we need to write a custom CMake module? Googling for "FindMAGMA.cmake" gave a couple of examples. Alternatively,  we can require to build deflation stuff with magma library simultaneously. This build process will be pretty long, though..

mathiaswagner:
Probably we can get along without FindMagma. I need to look into that again.

alexstrel:
MAGMA-based GMReS-DR solver is broken and requires immediate fix (no problem with LAPACK-based version)

alexstrel:
with commit  4641170d41f92293d18232e82ea16e826e3fcce7, GMResDR algorithm is now working as expected .

maddyscientist:
@mathiaswagner any progress on cmake for this?

mathiaswagner:
I did not look into that yet. Is the autoconf/make build now as desired? I didn't want to start porting from an old version. As soon as this is done I can get started.

maddyscientist:
Yes, autoconf / make changes are done.

mathiaswagner:
Great. I'll get started with this so we can merge this in. 
Are there any other open issues with this pr?

mathiaswagner:
Somewhat confused about the line
`MAGMA_INCLUDE = -I$(CUDA_INSTALL_PATH)/include -I../include -I${MAGMA_HOME}/include`

Do we really need the first two here? Anyway, I can add these to the CMakeLists

alexstrel:
@mathiaswagner there is no need for the first two flags please remove

mathiaswagner:
Ok,  I will. Thanks.

mathiaswagner:
I currently don't have MAGMA installed anywhere. As `make.inc` is calling `pkgconfig` can someone of you send me the output of that call ?

mathiaswagner:
Between the talks I made some progress. The only question remaining is whether we are fine with relying  on `pkg-config` like we do for autoconf/make (that is extremely simple as CMake supports using `pig-config` ) or (slightly more involved but also doable if needed) should have a fallback if `pig-config` is not available / configured for MAGMA.

alexstrel:
@mathiaswagner : here the example of the output : -fopenmp -L/phihome/astrel/install/magma-1.7.0-openblas-cuda-7.5/lib -L/usr/local/cuda-7.5/lib64 -L/phihome/astrel/install/openblas/lib -lmagma -lopenblas -lcublas -lcudart

maddyscientist:
I think having a pkg-config dependency is fine (or pig-config whatever that is :wink: ).  This is the portable way of doing things.

bjoo:
Hi All, 
 If this is only to link with MAGMA, you can also try to use a CMake finder for MAGMA. 
I found one on GitHub: 

https://gist.github.com/andrewseidl/60238a1fa7de304acbe0

In principle all you’ll have to do is
have MAGMA_ROOT point to the root of the MAGMA install
and MAGMA_USE_STATIC_LIBS set to true/false as needed

Then just:

find_package( MAGMA )   with additional [REQUIRED] and [QUIET] options.

and you should get back  (from the GitHub page)

#   MAGMA_FOUND               ... true if magma is found on the system
#   MAGMA_LIBRARY_DIRS        ... full path to magma library
#   MAGMA_INCLUDE_DIRS        ... magma include directory
#   MAGMA_LIBRARIES           ... magma libraries

If licensing terms allow, you can always create CMakeModules/ directory in the source,
put the FindMAGMA.cmake file there and ship it with QUDA. To make CMake find it,
that CMakeModules directory needs to be added to the CMAKE_MODULE_PATH e.g. as:

set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/CMakeModules/")

(with thanks to stack overflow http://stackoverflow.com/questions/10765885/how-to-install-your-custom-cmake-find-module for this last bit)

Hope this is helpful, 
 Best 
 B

> On Oct 13, 2015, at 12:22 PM, maddyscientist <notifications@github.com> wrote:
> 
> I think having a pkg-config dependency is fine (or pig-config whatever that is  ). This is the portable way of doing things.
> 
> —
> 

Jefferson Lab
Newport News, VA 23606, USA
email: bjoo@jlab.org       

mathiaswagner:
@bjoo Thanks. The potential issue with that is that it is not included in cmake so we would have to include that code in quda and licensing is not obvious for that.
So as we rely on pkg-config already for automaker it is just a simpler (not necessarily cleaner) solution to use that. I actually prefer your approach (also found that and @alexstrel pointed that out as well - see comments above).
Finding other libraries is not something that is something that will not really improve with cmake (also cmake supports better ways to do it).

mathiaswagner:
@maddyscientist Shall we then go *all-in* with pig-config (I am refusing to undo the autocorrect now) and also use that for include? Specifying MAGMA_HOME for include but then find the libraries only using pig-config seems inconsistent.

maddyscientist:
Going all in sounds good by me.  Can you make this change, as I was planning to do a full review of this pull (and write a page on the wiki based on some notes that @alexstrel sent me showing how use this.)  Sound like a plan?

mathiaswagner:
Sounds good. I did my best to screw up the old build systems so that you need to use cmake by default :wink:. But let me know if it is broken.

maddyscientist:
Ok, I'm about to merge this @alexstrel.  One question: you've used `printf` rather than `printfQuda` in many places.  Is this intentional, e.g., MAGMA could fail on one processor but not on others?  If this isn't intentional, I'd like to replace these all with `printfQuda`.

maddyscientist:
@alexstrel: I've been working through cleaning up my wiki article on the deflated solver (based on your email) and better understanding the parameters.  I've found a couple of things that need to be cleaned up before we merge, taken from inv_eigcg_quda.cpp:
```c++
const int max_restart_num = 3;//must be external for end-user tuning                                                                                                                                                                                                  
const double inc_tol = 1e-2;//must be external for the end-user tuning                                                                                                                                                                                                
```
The above need to be exposed to the interface, adding new parameters if necessary.

```c++
//In many cases, there is no need to use full precision accumulator, since low-mode deflation stabilizes                                                                                                                                                              
//the mixed precision solver. Moreover, full precision accumulation results in worse performance of the deflated solver, up to 15% in my experiments                                                                                                                   
//However, this parameter should be exposed to the enduser for the performance tuning, just in some rare cases when low-mode deflation will be insufficient                                                                                                           
//for stable double-half mixed precision CG.                                                                                                                                                                                                                          
initCGparam.use_sloppy_partial_accumulator = 0;
```
I'm confused by the comment above.  You set the solver to use high-precision accumulation, but your comment says it's better to use sloppy accumulation.  Either way, why not just use the interface setting for this (which defaults to 0 / high-precision accumulation).

```c++
initCGparam.delta = 1e-2; // might be a bit better than the default value 1e-1 (think about this)
```
Why is this hard coded and simply not left to the interface value?

maddyscientist:
Also, these parameters should be exposed in the interface as well
```c++
const bool use_reduced_vector_set = true;
const int eigcg_min_restarts  = 3;
```
And lastly: what is the significance of the `last_rhs` parameter?  Does this do memory freeing?  Why not an explicit memory free?  This seems slightly opaque, and it may not always be possible to tell when calling the QUDA solver that this is the last one you're going to do due the structure of the host application.

alexstrel:
@maddyscientist : I agree some of these parameters can be included into the interface (that is why you saw the comments). I need to check about initCGparam.use_sloppy_partial_accumulator and initCGparam.delta. As to max_restart_num, this is up to a user, in principal, but normally we don't allow more then 3 restarts (projection overhead will eliminate performance gain). Concerning printf's , I think these appear only in blas_magma.cu. My original plan was to make it "independent" of all other QUDA stuff, that is, not to include any QUDA specific  header into blas_magma.h (why that was I don't remember now).

alexstrel:
@maddyscientist  , yes last_rhs tells that the execution is done, and all resources must cleaned. Not sure what do you mean under explicitly clean. For a system with many RHS I suppose a user knows how many RHS to solve.

maddyscientist:
Thanks Alexei for addressing most of my list. 

What I meant with respect to memory management is that when running a bunch of measurements, the host application may not know when it is doing the last solve.  It is a scoping issue.  Take for example when the user wants to do some meson and baryon measurements.  These may be in completely different modules and will be done independently of each other.  There is no way to tell when the requested solve is the last one.

What we need instead is something like a `destroyDeflationQuda` interface that frees all the resources allocated by the deflated solver.  See for example how one done the multigrid solver interface (which is of course evolving).  Make sense?

If you like we can defer this interface change later, but it must be done, for example to make t consistent with the multigrid interface.

alexstrel:
@maddyscientist : I agree, it's indeed better to allow for  a (user controlled) separate call for cleaning resources. Not a serious change in the code either.

maddyscientist:
So shall we defer this change for now, and work on the interface in feature/multigrid?  Or is this something you can easily do now?

alexstrel:
It's not difficult, doing now.

maddyscientist:
Excellent :+1:

maddyscientist:
@alexstrel Your last commit seemed to break a lot of things
```
inv_eigcg_quda.cpp: In member function ‘int quda::IncEigCG::EigCG(quda::cudaColorSpinorField&, quda::cudaColorSpinorField&)’:
inv_eigcg_quda.cpp:608:104: error: no matching function for call to ‘quda::EigCGArgs<double, double2>::CheckEigenvalues(quda::cudaColorSpinorField*&, const quda::DiracMatrix*&, int&)’
          if(getVerbosity() >= QUDA_DEBUG_VERBOSE) eigcg_args->CheckEigenvalues(Vm, matDefl, eigvRestart);          
                                                                                                        ^
inv_eigcg_quda.cpp:608:104: note: candidate is:
inv_eigcg_quda.cpp:311:8: note: void quda::EigCGArgs<Float, CudaComplex>::CheckEigenvalues(const quda::cudaColorSpinorField*, const quda::DiracMatrix&, int) [with Float = double; CudaComplex = double2]
   void EigCGArgs<Float, CudaComplex>::CheckEigenvalues(const cudaColorSpinorField *Vm, const DiracMatrix &matDefl, const int restart_num)
        ^
inv_eigcg_quda.cpp:311:8: note:   no known conversion for argument 2 from ‘const quda::DiracMatrix*’ to ‘const quda::DiracMatrix&’
inv_eigcg_quda.cpp: In member function ‘void quda::IncEigCG::ExpandDeflationSpace(quda::DeflationParam*, int)’:
inv_eigcg_quda.cpp:800:46: error: request for member ‘isStaggered’ in ‘((quda::IncEigCG*)this)->quda::IncEigCG::matDefl’, which is of pointer type ‘const quda::DiracMatrix*’ (maybe you meant to use ‘->’ ?)
      cudaColorSpinorField *tmp2_p = !matDefl.isStaggered() ? new cudaColorSpinorField(*W, csParam) : &tmp;
                                              ^
inv_eigcg_quda.cpp:805:65: error: expression cannot be used as a function
        matDefl(*W, dpar->cudaRitzVectors->Eigenvec(j), tmp, tmp2);//precision must match!
                                                                 ^
inv_eigcg_quda.cpp: In member function ‘void quda::IncEigCG::ReportEigenvalueAccuracy(quda::DeflationParam*, int)’:
inv_eigcg_quda.cpp:862:46: error: request for member ‘isStaggered’ in ‘((quda::IncEigCG*)this)->quda::IncEigCG::matDefl’, which is of pointer type ‘const quda::DiracMatrix*’ (maybe you meant to use ‘->’ ?)
      cudaColorSpinorField *tmp2_p = !matDefl.isStaggered() ? new cudaColorSpinorField(*W, csParam) : &tmp;
                                              ^
inv_eigcg_quda.cpp:869:36: error: expression cannot be used as a function
          matDefl(*W2, *W, tmp, tmp2);
                                    ^
inv_eigcg_quda.cpp: In member function ‘void quda::IncEigCG::LoadEigenvectors(quda::DeflationParam*, int, double)’:
inv_eigcg_quda.cpp:946:46: error: request for member ‘isStaggered’ in ‘((quda::IncEigCG*)this)->quda::IncEigCG::matDefl’, which is of pointer type ‘const quda::DiracMatrix*’ (maybe you meant to use ‘->’ ?)
      cudaColorSpinorField *tmp2_p = !matDefl.isStaggered() ? new cudaColorSpinorField(*W, csParam) : &tmp;
                                              ^
inv_eigcg_quda.cpp:978:40: error: expression cannot be used as a function
              matDefl(*W2, *W, tmp, tmp2);
                                        ^
inv_eigcg_quda.cpp: In member function ‘virtual void quda::IncEigCG::operator()(quda::cudaColorSpinorField*, quda::cudaColorSpinorField*)’:
inv_eigcg_quda.cpp:1327:89: error: no matching function for call to ‘quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile*&)’
            if(cg_updates) initCG = new CG(*matSloppy, *matCGSloppy, initCGparam, profile);
                                                                                         ^
inv_eigcg_quda.cpp:1327:89: note: candidates are:
In file included from inv_eigcg_quda.cpp:9:0:
../include/invert_quda.h:313:5: note: quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile&)
     CG(DiracMatrix &mat, DiracMatrix &matSloppy, SolverParam &param, TimeProfile &profile);
     ^
../include/invert_quda.h:313:5: note:   no known conversion for argument 4 from ‘quda::TimeProfile*’ to ‘quda::TimeProfile&’
../include/invert_quda.h:306:9: note: quda::CG::CG(const quda::CG&)
   class CG : public Solver {
         ^
../include/invert_quda.h:306:9: note:   candidate expects 1 argument, 4 provided
inv_eigcg_quda.cpp:1373:109: error: no matching function for call to ‘quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile*&)’
                    if(!initCG && (r2 > stop)) initCG = new CG(*matSloppy, *matCGSloppy, initCGparam, profile);
                                                                                                             ^
inv_eigcg_quda.cpp:1373:109: note: candidates are:
In file included from inv_eigcg_quda.cpp:9:0:
../include/invert_quda.h:313:5: note: quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile&)
     CG(DiracMatrix &mat, DiracMatrix &matSloppy, SolverParam &param, TimeProfile &profile);
     ^
../include/invert_quda.h:313:5: note:   no known conversion for argument 4 from ‘quda::TimeProfile*’ to ‘quda::TimeProfile&’
../include/invert_quda.h:306:9: note: quda::CG::CG(const quda::CG&)
   class CG : public Solver {
         ^
../include/invert_quda.h:306:9: note:   candidate expects 1 argument, 4 provided
inv_eigcg_quda.cpp:1445:67: error: no matching function for call to ‘quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile*&)’
           initCG = new CG(*mat, *matCGSloppy, initCGparam, profile);
                                                                   ^
inv_eigcg_quda.cpp:1445:67: note: candidates are:
In file included from inv_eigcg_quda.cpp:9:0:
../include/invert_quda.h:313:5: note: quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile&)
     CG(DiracMatrix &mat, DiracMatrix &matSloppy, SolverParam &param, TimeProfile &profile);
     ^
../include/invert_quda.h:313:5: note:   no known conversion for argument 4 from ‘quda::TimeProfile*’ to ‘quda::TimeProfile&’
../include/invert_quda.h:306:9: note: quda::CG::CG(const quda::CG&)
   class CG : public Solver {
         ^
../include/invert_quda.h:306:9: note:   candidate expects 1 argument, 4 provided
inv_eigcg_quda.cpp:1478:65: error: no matching function for call to ‘quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile*&)’
         initCG = new CG(*mat, *matCGSloppy, initCGparam, profile);
                                                                 ^
inv_eigcg_quda.cpp:1478:65: note: candidates are:
In file included from inv_eigcg_quda.cpp:9:0:
../include/invert_quda.h:313:5: note: quda::CG::CG(quda::DiracMatrix&, quda::DiracMatrix&, quda::SolverParam&, quda::TimeProfile&)
     CG(DiracMatrix &mat, DiracMatrix &matSloppy, SolverParam &param, TimeProfile &profile);
     ^
../include/invert_quda.h:313:5: note:   no known conversion for argument 4 from ‘quda::TimeProfile*’ to ‘quda::TimeProfile&’
../include/invert_quda.h:306:9: note: quda::CG::CG(const quda::CG&)
   class CG : public Solver {
         ^
../include/invert_quda.h:306:9: note:   candidate expects 1 argument, 4 provided
make[1]: *** [inv_eigcg_quda.o] Error 1
make[1]: *** Waiting for unfinished jobs....
inv_gmresdr_quda.cpp: In member function ‘void quda::GMResDR::RunDeflatedCycles(quda::cudaColorSpinorField*, quda::cudaColorSpinorField*, quda::GMResDRDeflationParam*, double)’:
inv_gmresdr_quda.cpp:986:19: error: invalid conversion from ‘const quda::DiracMatrix*’ to ‘quda::DiracMatrix*’ [-fpermissive]
        sloppy_mat = matDefl;
                   ^
inv_gmresdr_quda.cpp:1003:44: error: request for member ‘flops’ in ‘((quda::GMResDR*)this)->quda::GMResDR::mat’, which is of pointer type ‘quda::DiracMatrix*’ (maybe you meant to use ‘->’ ?)
    double gflops = (quda::blas_flops + mat.flops())*1e-9;
                                            ^
inv_gmresdr_quda.cpp:1018:8: error: request for member ‘flops’ in ‘((quda::GMResDR*)this)->quda::GMResDR::mat’, which is of pointer type ‘quda::DiracMatrix*’ (maybe you meant to use ‘->’ ?)
    mat.flops();
        ^
inv_gmresdr_quda.cpp:724:9: warning: unused variable ‘heavy_quark_check’ [-Wunused-variable]
     int heavy_quark_check = 10; // how often to check the heavy quark residual
         ^
inv_gmresdr_quda.cpp: In member function ‘void quda::GMResDR::RunProjectedCycles(quda::cudaColorSpinorField*, quda::cudaColorSpinorField*, quda::GMResDRDeflationParam*, bool)’:
inv_gmresdr_quda.cpp:1090:20: error: invalid conversion from ‘const quda::DiracMatrix*’ to ‘quda::DiracMatrix*’ [-fpermissive]
         sloppy_mat = matDefl;
                    ^
inv_gmresdr_quda.cpp:1294:44: error: request for member ‘flops’ in ‘((quda::GMResDR*)this)->quda::GMResDR::mat’, which is of pointer type ‘quda::DiracMatrix*’ (maybe you meant to use ‘->’ ?)
    double gflops = (quda::blas_flops + mat.flops())*1e-9;
                                            ^
inv_gmresdr_quda.cpp:1309:8: error: request for member ‘flops’ in ‘((quda::GMResDR*)this)->quda::GMResDR::mat’, which is of pointer type ‘quda::DiracMatrix*’ (maybe you meant to use ‘->’ ?)
    mat.flops();
        ^
inv_gmresdr_quda.cpp:1160:10: warning: unused variable ‘heavy_quark_check’ [-Wunused-variable]
      int heavy_quark_check = 10; // how often to check the heavy quark residual
          ^
```

alexstrel:
now fixed. Still need to test the branch

maddyscientist:
@alexstrel and @mathiaswagner.  Thoughts on whether we should use use a consistent naming convention for freeing up QUDA resources.  We presently have two styles e.g.,
1. `freeGaugeQuda`, `freeCloverQuda`, etc.
2. `destroyMultigridQuda`, `destroyDeflationQuda`, etc.

Since 2.) hasn't been propagated into any applications yet, we can still standardize on `freeMultgridQuda` in favour of `destroyMultigridQuda`.  Or we could decide that we don't care that much.

alexstrel:
yep, in fact, in destroyDeflationQuda we just free memory; it might be better to reserve "destroy" for something more  then just memory operations.

maddyscientist:
@alexstrel I've merged this pull request after making some small changes and cleanup.  I've also cleaned up my wiki description of the deflation algorithms, though a couple of questions are outstanding:
* If `QudaInvertParam::eigcg_min_restarts` corresponds to the maximum number of refinement cycles why is called `min` and not `max`?
* For `QudaInvertParam::use_reduced_vector_set` how are the criteria set?  E.g., the number of eigenvectors to use or maximum residual norm admissible?

alexstrel:
@maddyscientist  : To the first question, yes u are right, my mistake!  
to the second question, if one decides to use a reduced vector set , then one needs to specify eigenval_tol parameter . I was not certain about as to whether to expose another parameter , a  number of eigenvectors to use. Currently it is set to full deflation dimension (see line 1507 in inv_eigcg_quda.cpp)
Strictly speaking, one can adjust the  number of eigenvectors by choosing an appropriate tolerance , so in some sense, they are related. that is why, I decided not to expose the number of vecs to the interface, but I don't exclude it totally. Let's keep it as an "undocumented feature", what do you think?

maddyscientist:
@alexstrel can you make a hotfix to rename eigcg_min_restarts to eigcg_max_restarts then?

mathiaswagner:
@maddyscientist @alexstrel This pull request breaks the MILC interface. It looks like a screwed up merge conflict from https://github.com/lattice/quda/commit/b982fb8423e415cbc065f0a0406401a7ec0d65a0.
If we build with 'GAUGE_ALG' the interface for the gauge fixing routines is included twice.

I guess this is something that Jenkins should catch (if it passes human review) at some point but it didn't since we just cannot cover the full space of parameters we have (we can increase that coverage for pull requests).

maddyscientist:
@alexstrel can you fix this as well in your hotfix?

alexstrel:
yes, sure.

mathiaswagner:
@alexstrel I just pushed the release/0.8.x so that we can get started with the preparation for the 0.8.0 release. When you create the pull request for your hot fix please use 'release/0.8.x' instead of develop. (Right now release/0.8.x is identical to develop but that may change ...)

alexstrel:
ok, thanks. I made changes in hotfix/deflation. After checking it will issue a pull request for release/0.8.x

maddyscientist:
@alexstrel One thing I just noticed: you've changed the interface to the deflated solvers to take a pointer rather than a reference for the matrix and profile instances.  What was the motivation for doing this?  This makes the interface inconsistent with the other solvers.

alexstrel:
that was because of destroyQudaDeflation(), and that was a quick hack. Just see constructor at Ln 441 in inv_eigcg_quda.cpp. Basically what I needed was to be able to pass null pointers. (the only reason for ln 441 was to clean/operate resources without using solvers. I understand this is very ugly approach but it works for now. see also ln 3255 and so on in interface_quda.cpp. Later I will replace all this stuff by something similar you did for multigrid solver)

maddyscientist:
understood

PR #369, opened 2015-10-01
Added topological charge computation

AlexVaq:
This adds a simple function that returns the topological charge of the loaded gauge field. If the field was smeared, the function uses the smeared field for the calculation. If there are no smeared fields present, the function uses the available gauge fields.

If smeared fields are present, the current implementation overwrites the extended resident gauge fields with an extended version of the smeared fields. This can easily be changed if people think it's not useful.

Just closed the previous branch request because the name of the new branch didn't comply with the new workflow for QUDA developing.

### DIFF
--- a/include/cub_helper.cuh
+++ b/include/cub_helper.cuh
@@ -50,6 +50,10 @@ namespace quda {
     a.y = 0.0;
   }
 
+  __device__ inline void zero(double &a) {
+    a = 0.0;
+  }
+
   __device__ unsigned int count = 0;
   __shared__ bool isLastBlockDone;
 
--- a/include/gauge_tools.h
+++ b/include/gauge_tools.h
@@ -79,4 +79,5 @@ namespace quda {
 		    const GaugeField& gauge, 
 		    QudaFieldLocation location);
 
+  double computeQCharge(GaugeField& Fmunu, QudaFieldLocation location);
 }
--- a/include/gauge_tools.h
+++ b/include/gauge_tools.h
@@ -79,5 +79,11 @@ namespace quda {
 		    const GaugeField& gauge, 
 		    QudaFieldLocation location);
 
+  /**
+     Compute the topological charge
+     @param Fmunu The Fmunu tensor, usually calculated from a smeared configuration
+     @param location The location of where to do the computation, currently supports only the GPU
+   */
+
   double computeQCharge(GaugeField& Fmunu, QudaFieldLocation location);
 }
--- a/include/quda.h
+++ b/include/quda.h
@@ -831,6 +831,11 @@ extern "C" {
   void performAPEnStep(unsigned int nSteps, double alpha);
 
   /**
+   * Calculates the topological charge from gaugeSmeared, if it exist, or from gaugePrecise if no smeared fields are present.
+   */
+  double qChargeCuda();
+
+  /**
    * @brief Gauge fixing with overrelaxation with support for single and multi GPU.
    * @param[in,out] gauge, gauge field to be fixed
    * @param[in] gauge_dir, 3 for Coulomb gauge fixing, other for Landau gauge fixing
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -37,7 +37,7 @@ QUDA_OBJS = gauge_phase.o timer.o malloc.o solver.o			\
 	pgauge_exchange.o pgauge_init.o pgauge_heatbath.o random.o	\
 	gauge_fix_ovr_extra.o gauge_fix_fft.o gauge_fix_ovr.o		\
 	pgauge_det_trace.o clover_outer_product.o			\
-	clover_sigma_outer_product.o
+	clover_sigma_outer_product.o qcharge_quda.o
 
 # header files, found in include/
 QUDA_HDRS = blas_quda.h clover_field.h color_spinor_field.h convert.h	\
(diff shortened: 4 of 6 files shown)


AlexVaq:
The implementation has been succesfully tested against CPU code, and I checked the multiGPU implementation in a single GPU using the -partition command line option.

Since it's a small add-on, and it doesn't interact with anything else, I think the merge should come soon.

mikeaclark:
@AlexVaq this looks like a fairly uncontroversial commit :+1:  A few small suggestions though
* Doxygen markup for the `computeQCharge` prototype in gauge_tools.h
* I think there's a bug: don't you need to do a `cudaDeviceSynchronize()` after the kernel before doing the `comm_allreduce`?  At the moment you have the synchronization outside of the apply so it happens after.
* Can you replace the use of the double-precision atomic if the `reduce` helper functor that I create in include/cub_helper.cuh?  As well as making it faster, it also will cleanup your code quite a bit.  See gauge_plaq.cu for an example of how to use this: the argument struct need to inheret from ReduceArg, and you can simply pass `tmpQ1` variable to this.
* Can you add conditional compilation and checks, such that this routine is only compiled if `GPU_GAUGE_TOOLS` is compiled?  The Fmunu computation already requires this. 
* Just a comment.  Once my feature/improved-staggered-rhmc pull is merged, I'll start removing all support for pre-Fermi architecture.  So there's no need to keep the guard in there.

That's all.  We should be able to get this merged with the above changes.  Thanks :)

mikeaclark:
ACtually one more thing: I just noticed you're using the legacy `loadLinkVariableFromArray` loading with explicit stride, etc.  Can you replace this with the `gauge::FloatNOrder` accessor?  I now refuse to merge any code that doesn't use this :stuck_out_tongue_closed_eyes:

AlexVaq:
I see that there is a lot of new stuff...

I also foubd a bug in the interface: the gaugeSmeared field is always an extended field for multiGpu, so no need to extend it twice.

I'll try to work a bit on this tomorrow.

mathiaswagner:
Can one of the admins verify this patch?

mathiaswagner:
ok to test

AlexVaq:
For single GPU (and multi GPU in a single GPU with the partition option) I already test the correctness of the results. But I'd appreciate if someone could check that multiGPU with several GPUs gives the same result as single GPU.

AlexVaq:
Opsss, the automatic build fails because of staggered_dslash_test.cpp (which has nothing to do with this pull request). Is Jenkins suppose to be 100% working, or it's been very recently introduced? I expect this problem to affect all other branches/pull requests based on develop.

mikeaclark:
The build failure is coming from staggered being broken on Fermi.  This is fixed in the feature/improved-staggered-rhmc branch so will go away when that's merged in.

mathiaswagner:
Jenkins is work in progress. But here it is working. It detects that something is broken. The issue is obviously develop but Jenkins cannot know that. Even if - with a build failing for any reason if you exclude that reason you (Jenkins) cannot know if the build might still fail.

Ideally in the future we won't merge anything into develop that breaks it. Then we are pretty sure a failed build for a pull request is failing because of things in that pull request.

mathiaswagner:
@Jenkins retest this please

mathiaswagner:
@Jenkins test this please

mathiaswagner:
While I am trying to improve Jenkins I noted a real failure for this pull request now:
```
interface_quda.cpp: In function ‘double qChargeCuda()’:
interface_quda.cpp:5549:7: error: ‘temp’ was not declared in this scope
       temp = gaugePrecise;
```

AlexVaq:
Weird, that line shouldn't be there. Let me check if my local repo coincides with the remote one (git push should make sure that).

AlexVaq:
Now I'll try to recompile with multiGPU and run with the partition option enabled, to see if I still get the right result.

mathiaswagner:
I would not fully exclude the option of an issue with Jenkins.

AlexVaq:
No, no, Jenkins was right this time. I forgot to remove those lines that affect only multiGPU, and I compiled singleGPU to check that everything was still right, after a few modifications. I hope everything is all right by now.

I actually find that Jenkins can perform a pretty good job detecting those broken branches.

mathiaswagner:
Looks like Jenkins just triggered the rebuild and we just have to get the SM 2.0 build fixed also in develop as this creates a lot of broken builds currently.

AlexVaq:
Yes, I see for Kepler everything looks right, but for Fermi there are the issues with the plaquette and with the staggered.

AlexVaq:
It gives the right result with multiGPU and the partition option.

mathiaswagner:
Argh. I tried to push to that branch but somehow it looks like that my git did not differentiate between 'feature/QCharge' and 'Feature/Qcharge'. It managed to push that into a new branch ...
It looks like nothing is screwed up on github and I only need to check how to clean that up locally (and maybe delete the duplicate branch on github).

AlexVaq:
Yeah well, my fault. I don't really know why I used upper case and I really feel that I shouldn't... Also, I think there isn't an easy way to rename a branch in github...

mikeaclark:
Ok, this all looks good now, so I am going to merge this.  Thanks @AlexVaq for a nice feature add.

PR #374, opened 2015-10-05
Feature/q charge

mathiaswagner:
This is just a dummy pull request to clean up my mess with `Feature/QCharge` and `feature/Qcharge`.
I hope this will still correctly show up in #369.


mikeaclark:
Hmm, I think this is still messed up.  I just merged in the main #369 but I just noticed that it doesn't have your CMakeList.txt addition, which I thought you said was fixed.  Also, even though I've merged, there is not option to delete the branch.  Any idea what's wrong?

PR #375, opened 2015-10-05
Use GIT commit id to ensure tunecache.tsv is valid

mikeaclark:
`GITVERSION` is now recorded in the tunecache.tsv file and is used to check that the file is valid.  This prevents stale tunecache files from being used when switching between development branches with the same `MAJOR` and `MINOR` version.


mathiaswagner:
Does that work if GITVERSION is not defined during the build process?

In interface QUDA we use
```C++

extern char* gitversion;

[...]

  if (getVerbosity() >= QUDA_SUMMARIZE) {
#ifdef GITVERSION
    printfQuda("QUDA %s (git %s)\n",quda_version.c_str(),gitversion);
#else
    printfQuda("QUDA %s\n",quda_version.c_str());
#endif
  }
```

I think my motivation here was to make sure if someone just downloads a zip file from github or the release version it does not come with the git files.
Also `version.cpp` contains `gitversion`as a string and `interface_quda.cpp` uses that thus we don't have to recompile the interface but just link with the updated version.o in the end.

No over engineering here but I hate having two solution for the same issue in the code and affecting the build system.

mikeaclark:
Feel free to adjust to suit 😀

mathiaswagner:
Doesn't reviewing pull requests mean that I can just ask others to do the work?
I will do that and add also '-debug' to the version string. This will probably not cause NANs but maybe bad tuning parameters.

mathiaswagner:
I hope the gitversion for the tunecache (including the  -debug) should now work. 
Only point I am not sure about: You might want to be able to turn on DEBUG and use a preexisting tunecache. That currently won't work unless you modify the tunecache. (Which I guess is fine because if you want to do that you are developer and are aware what you want to do).

It should however prevent people like me from running with a debug tune cache.

Should be ready for review and merge now.

PR #378, opened 2015-10-06
Hotfix/qmp gpuid

mikeaclark:
This pull fixes an issue when using the QMP backend on multiple nodes with multiple GPUs: the GPU ID assignment would result in multiple MPI processes sharing the same GPU if the process mapping is done round-robin between the nodes.

This fixes the issue by using the same mechanism adopted by the MPI backend, where we make a local array holding all the hostnames of all processes.  Locally, we iterate`i` from 0 to `comm_rank()`, and every time my hostname matches an entry in the array up to rank we increment the variable `gpuid`.  This ensures that each process will have a unique `gpuid`.

In QMP this is a bit of a hack since there is no all-gather algorithm, so we mimic this using reductions.  Furthermore, since there is no integer array reductions in QMP, and we wish to avoid having to do a reduction for every element of the hostname array, we use the `QMP_sum_float_array` to achieve this.

### DIFF
--- a/lib/comm_qmp.cpp
+++ b/lib/comm_qmp.cpp
@@ -37,7 +37,7 @@ void comm_init(int ndim, const int *dims, QudaCommsMap rank_from_coords, void *m
 
   // determine which GPU this rank will use
   char *hostname = comm_hostname();
-  char *hostname_recv_buf = (char *)safe_malloc(128*size);
+  char *hostname_recv_buf = (char *)safe_malloc(128*comm_size());
 
   // Abuse reductions to emulate all-gather.  We need to copy the
   // local hostname to all other nodes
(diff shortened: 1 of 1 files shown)


lib/comm_qmp.cpp
```diff
@@ -35,7 +35,33 @@ void comm_init(int ndim, const int *dims, QudaCommsMap rank_from_coords, void *m
...
+
+  // Abuse reductions to emulate all-gather.  We need to copy the
+  // local hostname to all other nodes
+  for (int i=0; i<comm_size(); i++) {
+    int data[128];
+    for (int j=0; j<128; j++) {
+      data[j] = (i == comm_rank()) ? hostname[j] : 0;
+    }
+    // check nasty int to float hack is valid
+    if (sizeof(float) != sizeof(int)) errorQuda("Broken");
```
mathiaswagner: Ok, I am not familiar with QMP internals so that hack is probably necessary. But the error message should be slightly more verbose than broken. Also, not for 0.7.2: Can we catch that at compile time? Assert?

mathiaswagner: I was thinking about static_assert and apart from various hacks it looks like it is C++11 (or later) only.

mikeaclark: Yes, I'm looking into pre C++11 options at the moment.  Note that C++11 is now required for the feature/multigrid branch but we certainly don't want this on release/0.7.x

mikeaclark:
Ok, compile-time check is now done.

PR #381, opened 2015-10-14
Hotfix/blas_tuning_speedup

maddyscientist:
When tuning the grid size of a kernel, set the maximum size grid size based on the streaming multiprocessor (SM) count instead of the arbitrary previous value of 256.  Here we use 2xSM count since it allows for two CTAs to be simultaneously resident in an SM, which may be desirable where we wish to allocate all resources of an SM, e.g. using all the shared memory on a GM200 requires 2 thread blocks per SM.

A comparison against the previous value (256) shows no performance degradation on GM200 (48 blocks), however the tuning time is cut by a factor of 5x.

### DIFF
--- a/include/tune_quda.h
+++ b/include/tune_quda.h
@@ -57,7 +57,7 @@ namespace quda {
     virtual bool advanceGridDim(TuneParam &param) const
     {
       if (tuneGridDim()) {
-	const unsigned int max_blocks = 256; // FIXME: set a reasonable value for blas currently
+	const unsigned int max_blocks = 2*deviceProp.multiProcessorCount;
 	const int step = 1;
 	param.grid.x += step;
 	if (param.grid.x > max_blocks) {


mathiaswagner:
Looks good to me. Do you have data (old tune caches?) that somehow show that we don't need to go beyond 2*number of SM? Just to be sure because a 5x improvement in tuning is nice but tuning is a one time thing (even more than compiling also I realize that this is shifting as we requiring retuning after each commit now). 

I am fine with an educated guess but even if we might lose real world performance for that I am not sure whether the 5x improvement is worth it.

But I am not insisting on that just want to make sure we are aware of this. 
Code change is obviously fine with me then.

maddyscientist:
I actually tested on GM200 with just `max_blocks=#SMs` and it showed zero performance regressions.  The factor of two was some additional safety.

maddyscientist:
Also I note I tested the full suite of blas tests at different volumes, using `max_blocks = 256, 1x#SMs, 2x#SMs, 4x#SMs`, and see essentially zero performance difference.

mathiaswagner:
I got tha. I was just wondering whether it's worth it and what may happen on older architectures and future ones. But on the other hand hard-coded 256 also was not great.

maddyscientist:
I think in general, if you have multiple work per thread, you should **never** need to launch more thread blocks than can be simultaneously resident.  In fact, this approach should in general be more efficient since there is less creation / destruction overhead.  This is explained in Duane's paper here https://sites.google.com/site/duanemerrill/policytuning_inpar12.pdf?attredirects=0

mathiaswagner:
Is tuning a general area we want to improve on? Just thinking about tuning on the fly (i.e. while we have not tested all launch configurations just tune the Kernel on the fly, i.e. try a new configuration when we execute the Kernel anyway). Probably not for reductions though. Is that of interest in terms on the tuning- in addition to the things we already have issues for?

maddyscientist:
I think I don't favour the gradual tuning approach:
* Many of the kernel launches in tuning run fail.  In all likelihood this shouldn't invalidate the data, e.g., in the case of a kernel that overwrites its input, but I suspect this behaviour is undefined.  If we backup and restore the input for every small step through tuning space the overhead would be huge (fields are backed up on the CPU not GPU).
* For large-scale task independent runs, the tunecache can be done in a small run up front and then used for multiple independent runs.  This would break if the tuning takes many iterations to complete.
* In fact the tuning may not complete on a typical run, so we'd have to instrument it to backup and restore the tuning state when starting a new job.

mathiaswagner:
I agree with all three of your points. Just wanted to bring it up to anyway to see whether it might get us somewhere.

PR #384, opened 2015-10-27
fixed calculation of griddim for multi GPU long link code

mathiaswagner:
For some local volumes that caused some of the longlinks to be zero.


maddyscientist:
I think this is ready to merge now.  I've fixed the llfat_test to work with the `--recon 12` option, which was broken previously.

PR #386, opened 2015-10-29
fix llfat_compilation

mathiaswagner:
typo that was not caught by Jenkins (and neither by the one that merged the pull request)


maddyscientist:
I don't understand why llfat_test would compile for me with the missing header include.  Could this somehow have been a cmake thing that doesn't show up with gmake?

mathiaswagner:
Strange, indeed. No idea. Jenkins passed as it dose not build fat link currently.

PR #396, opened 2015-11-17
fixed  bugs in DiracStaggered::M and  DiracImprovedStaggered::M metho…

alexstrel:
…ds: dslashXpay kernels were applied on zero field tmp1 in the operations a*tmp1+{hopping terms}


maddyscientist:
Why not just delete the tmp crest action and destruction instead of commenting them out?

alexstrel:
those were copied from the dirty local staggered-multigrid branch, now cleaned.

PR #397, opened 2015-11-23
Fixed a race-condition bug in multi-reductions

maddyscientist:
This resulted in incorrect evaluation when >1 vectors.  Also fixes a bug in autotuning which would result in parameter aliasing.

(This bug fix does not fix s-step CG.)


mathiaswagner:
I did not fix s-step CG but I saw irreproducible results in subsequent runs. S-step still crashes but at least is reproducible right now. That should make debugging a lot easier.

PR #406, opened 2015-12-10
Feature/no tesla arch

maddyscientist:
This pull request will remove all support for GPUs less than sm_20.  That is to say all GPUs prior to Fermi, which are the so-called Tesla generation.


mathiaswagner:
Just as a reminder to myself: Update CMakeLists.txt for python code generation.

mathiaswagner:
There is also still a `STAGGERED_TESLA_HACK`? Looks like a workaround but I guess we can remove that as well?

mathiaswagner:
CMake should be fine (at least a quick `make gen` on my Mac worked).

mathiaswagner:
There are still some references to Telsa left in the code:
```
blas_core.h:51:  // these can't be curried into the Spinors because of Tesla argument length restriction
blas_mixed_core.h:53:  // these can't be curried into the Spinors because of Tesla argument length restriction
dslash_clover.cu:31:#undef GPU_STAGGERED_DIRAC // do not delete - hack for Tesla architecture
dslash_constants.h:11:#ifndef STAGGERED_TESLA_HACK
dslash_constants.h:21:#ifndef STAGGERED_TESLA_HACK
dslash_improved_staggered.cu:43:#define STAGGERED_TESLA_HACK
reduce_core.h:231:  // these can't be curried into the Spinors because of Tesla argument length restriction
reduce_core_cub.h:415:  // these can't be curried into the Spinors because of Tesla argument length restriction
reduce_mixed_core.h:241:  // these can't be curried into the Spinors because of Tesla argument length restriction
```

Not sure but I assume if we don't remove them now they will clutter the code forever.

AlexVaq:
I agree we should clean everything up carefully. Fortunately it doesn't look like there are too many left.

maddyscientist:
I've removed the STAGGERED_TESLA_HACK macro guards.  For the blas and reductions, I suggest not touching this code since it's due for a major rewrite to clean it up and use a unified approach for multi-reductions.

mathiaswagner:
Looks good to me now.

PR #407, opened 2015-12-12
Feature/multigrid

maddyscientist:
Here it is, what everyone has been waiting for.  Here's a small subset of the features that this pull request adds

* Implements adaptive aggregation multigrid algorithm into QUDA
   * Null-space generation algorithm using BiCGstab
   * Block orthogonalization of the null-space vectors
   * Coarse grid gauge field construction
   * Coarse operator implementation (Wilson and Clover presently supported)
   * Restriction and prolongation operators for moving between grids
* Introduced new fine-grained accessor routines for generic indexing into fields (CPU and GPU supported)
* New generic halo region exchange for ColorSpinorField of arbitrary color and spin
* Moves QIO functionality directly into the library proper
* Support for load/storing to ColorSpinorFields (using QIO)
* Unified BLAS interface that operates on either CPU or CUDA fields
* All solver algorithms now operator on generic ColorSpinorField and not cudaColorSpinorField
* Improves functionality for dealing with full color-spinor fields
* Dirac operators now all act on ColorSpinorFields and not cudaColorSpinorFields (only GPU execution is supported on all but the coarse operators)
* C++11 is now used in QUDA, raising the minimum toolkit to 6.5

While this pull request will merge the critical framework for multigrid into the develop branch, there is still much left to be done before multigrid is in a useable state (from a performance point of view anyway).  The motivation here is to get the many framework changes into the mainstream to avoid the continual battles in merging between develop and feature/multigrid.


mathiaswagner:
This is some pretty disruptive pull request. As we just branched off 0.8 for all release work and it is probably impossible to review this pull request at all (I guess if we want to have code review for monsters like this we need a different workflow - see below) I will try to check some basic coding styles and then call it a year.

If we however do want a basic review this can only be done in an extensive walk through of the code.

For the future we should consider creating a special branch like develop/multigrid and merge repeatedly into this from feature branches to break down the review into digestible bites.

maddyscientist:
Having a branch solely for the purpose of code review sounds like a good idea.

Hopefully this will be the first and last such disruptive change though of this order.

mathiaswagner:
I am inlined to just merge this in. I am fine with having a somewhat messy state of the develop branch and just see where this takes us, even if we will have a lot of things to fix. But that's also what develop is for.

So, my requests so far:
* Do we have sufficient doxygen comments in this?
* Any disruptive changes for applications that don't use multigrid?

At least the Jenkins build (without MG) compile.

bjoo:
For code reviewing have you considered Geritt Code Review?

Apparently it is for this kind of thing — you wouldn’t need new branches, but could
still apply a code-review step before accepting new submissions. 

Apparently it is simple to integrate in front of GitHub (eg. http://gerrithub.io/)

I just thought I’d mention it as an alternative to code-review branches.
 
Best, 
 B
> On Dec 14, 2015, at 10:03 AM, Mathias Wagner <notifications@github.com> wrote:
> 
> I am inlined to just merge this in. I am fine with having a somewhat messy state of the develop branch and just see where this takes us, even if we will have a lot of things to fix. But that's also what develop is for.
> 
> So, my requests so far:
> 
> 	• Do we have sufficient doxygen comments in this?
> 	• Any disruptive changes for applications that don't use multigrid?
> At least the Jenkins build (without MG) compile.
> 
> —
> 

Jefferson Lab
Newport News, VA 23606, USA
email: bjoo@jlab.org       

maddyscientist:
I have been adding doxygen to the headers fairly consistently (certainly to new headers anyway). I do another pass through though if you like prior to merging.

maddyscientist:
In theory there should be no disruptions to non MG codes.  The solver interface (dis)functions the same as before.

mathiaswagner:
I am fine with that. I am just thinking about ways to best handle that now. As I said, we can probably just go on.
@bjoo  Just looking at gerrit now to see whether this can help here. But I am in favor of not adding another _tool_ to our workflow but maybe the features will just blow me away and I can push in this direction.

maddyscientist:
I've added some more doxygen comments to fill in a couple of gaps.  I think we're all good on that front now.

mathiaswagner:
Well, here we go ... multigrid is live.

maddyscientist:
Boom!  :boom:

PR #410, opened 2015-12-15
Warn about deprecation of configure / autoconf

mathiaswagner:
This introduces a warning message at the end of the `configure` run.
```
configure: WARNING:
    Note that of QUDA 0.8.0 configure (autoconf) is deprecated.
    Please consider switching to the new cmake based build system.
    See https://github.com/lattice/quda/wiki/Building-QUDA-with-cmake

    This warning will be turned into an error soon (use --enable-force-ac to turn it off).
    At some point configure (autoconf) will no longer be maintained and removed from QUDA.
```
This message should be later turned into a hard error.

A new flag `--enable-force-ac` is introduced to turn off the warning / error but prints a request to file a bug so we can make sure cmake can be used in the future.

```
configure: WARNING:
  If you require --enable-force-ac please file a bug on https://github.com/lattice/quda
  This will allow us to make sure you can build quda with cmake (required in the future).
```

### DIFF
--- a/configure
+++ b/configure
@@ -743,6 +743,7 @@ enable_magma
 enable_blas_tex
 enable_fermi_double_tex
 enable_numa_affinity
+enable_force_ac
 '
       ac_precious_vars='build_alias
 host_alias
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/lattice/quda/pull/410" target="_blank">ReviewNinja

PR #411, opened 2015-12-15
Some minor cleanup of solver interface.

### LINKED ISSUE #278: Add GPU-GPU interfaces
Aof now, QUDA is mostly oriented towards a CPU-GPU interface, where the data is read from CPU, used to compute something in the GPU and then written back in CPU. There are, though, some applications that require data generated in the GPU to stay in the GPU for further calculations. An example are the bilinear contractions and inversion of sources, which should be coupled together to avoid intermediate and useless CPU transfers. That's why it is necessary to develop some kind of GPU-GPU interfaces that hardly communicate to host.

maddyscientist:
Some minor cleanup
* Removed hacked `invertMDQuda` interface which is not used anywhere and was only there as an experiment
* Added trivial support to keep `invertQuda` solution resident instead of copying it back (closes #278)

### DIFF
--- a/include/quda_fortran.h
+++ b/include/quda_fortran.h
@@ -157,8 +157,6 @@ extern "C" {
    */
   void invert_quda_(void *h_x, void *h_b, QudaInvertParam *param);
 
-  void invert_md_quda_(void *hp_x, void *hp_b, QudaInvertParam *param);
-
   /**
    * Solve for multiple shifts (e.g., masses).
    * @param _hp_x    Array of solution spinor fields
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/lattice/quda/pull/411" target="_blank">ReviewNinja

mathiaswagner:
Not sure whether we ever need this, but do we need to expose a way to delete a resident solution?

maddyscientist:
Well `endQuda()` will delete any leftover resident solution, and the consumers of it (which is only `computeCloverForceQuda()` also delete it once consumed.  I think that's all we need for now.

PR #414, opened 2015-12-15
Small improvements to cmake build

mathiaswagner:
For builds with MPI or QMP catch missing MULTI_GPU option already in cmake (not only at build time).


review-ninja:
<a href="https://app.review.ninja/lattice/quda/pull/414" target="_blank">ReviewNinja

mathiaswagner:
We can also automatically set MULTI_GPU to true and taking it even further: Do we need to expose the MULTI_GPU option at all.
If we use MULTI_GPU we require either MPI or QMP and if we use either MPI or QMP we require MULTI_GPU.
So, we need the internal define but can get rid of the exposed option.

maddyscientist:
We could do.  Beyond debugging, I'm not sure what utility the single-GPU build has anymore anyway.

mathiaswagner:
ok, I'll do that. Only if neither MPI or QMP is requested we will use the single GPU build. Reduced the exposed options.

maddyscientist:
While you're playing with cmake, can you change the build options for the dirac operators so that the Dirac operators are grouped together: e.g., BUILD_WILSON_DIRAC -> BUILD_DIRAC_WILSON.  It will make parsing the cmake list much easier.

AlexVaq:
You can skip all the ghost exchanging in single-GPU builds (a bitheavy withgauge utilities, youneed to create all the extended fields...), and I think it makes the development easier, so I would keep the option. But I understand why you might want to remove it completely.

maddyscientist:
Yes, but a properly written multi-GPU code running on one GPU should have the same performance as a single-GPU code.  The only exception here is probably the dslash, which is extremely sensitive to branch latency overhead.  Anyway, I wasn't suggesting to kill it **yet**.

AlexVaq:
I agree we should eventually get rid of it.

PR #436, opened 2016-02-02
made c++11 a default flag and fix ; in CXX_FLAGS

mathiaswagner:


maddyscientist:
Confirmed this fixes the issue with cmake I was seeing.  I also pushed a micro fix for QIO building with cmake that broke when I moved qio_util.cpp into the QUDA library as opposed to being in the tests directory.

Closing.

PR #438, opened 2016-02-09
Feature/cub exact sum

maddyscientist:
This pull adds several features, most of which are experimental for now.  The intent is to merge these in, and continue these at a later date.  None of the experimental features are broken, rather just not performance tuned.
* Adds double-double number type, `doubledouble` to QUDA, for emulated quad-precision operations
* All reductions now use the cub functors, instead of the home-grown QUDA code.  This slightly improves performance of reductions at small volumes, with similar performance at large volumes.
* Update cub version to the latest 1.5.1
* Reductions can now be done in quad precision, for improved stability (**Experimental**, disabled by default)
* Improved performance and robustness of multi-reductions for s-step solvers (**Experimental**)
* Added support for pipeline GCR using multi-reductions (**Experimental**)


mathiaswagner:
cub killed the diff. But all I can see looks good so as we just released 0.8 ... merge in.

PR #440, opened 2016-02-24
check whether resident gauge is compatible with invert params from MILC

### LINKED ISSUE #439: MILC precision mismatch in rhmc (MILC 7.7.13 / quda 0.8.0)
QUDA internal tracking of bug report by @lcosmai https://github.com/milc-qcd/milc_qcd/issues/5

MILC 7.7.13 and probably also 7.8.0 fails in RHMC with QUDA

To reproduce build MILC for double precision and enable all QUDA acceleration for HISQ:

run
```
~/milc_qcd/ks_imp_rhmc/test$ ../su3_rhmc_hisq su3_rhmc_hisq.2.sample-in
```

error message:
```
GACTION: 5.960903e+00
ERROR: Solve precision 4 doesn't match gauge precision 8 (rank 0, host nvsocal2, /home/mathias/qudagit/lib/interface_quda.cpp:1904 in checkGauge())
       last kernel called was (name=N4quda22HeavyQuarkResidualNormI7double37double2S2_EE,volume=3x6x6x6,aux=vol=648,stride=756,precision=8)
```

mathiaswagner:
fixes #439

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -1900,9 +1900,7 @@ void MatDagMatQuda(void *h_out, void *h_in, QudaInvertParam *inv_param)
 
 namespace quda{
 bool canReuseResidentGauge(QudaInvertParam *param){
-
-  if (gaugePrecise == NULL) errorQuda("Precise gauge field doesn't exist");
-  return param->cuda_prec == gaugePrecise->Precision();
+  return (gaugePrecise != NULL) and param->cuda_prec == gaugePrecise->Precision();
 }
 }
 
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -1898,6 +1898,14 @@ void MatDagMatQuda(void *h_out, void *h_in, QudaInvertParam *inv_param)
   popVerbosity();
 }
 
+namespace quda{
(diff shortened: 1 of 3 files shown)


lib/interface_quda.cpp
```diff
@@ -1898,6 +1898,14 @@ void MatDagMatQuda(void *h_out, void *h_in, QudaInvertParam *inv_param)
   popVerbosity();
 }
 
+namespace quda{
+bool canReuseResidentGauge(QudaInvertParam *param){
+
+  if (gaugePrecise == NULL) errorQuda("Precise gauge field doesn't exist");
```
mathiaswagner: @maddyscientist This line is supposed to check whether a gauge field is allocated and halt if not.

maddyscientist:
Yes, my question is whether this is the correct thing to do? If there is no gauge field allocated shouldn't we just return false?

mathiaswagner:
Yes, that is probably ok. We will then call InvalidateGauge and this will call FreeGauge and If I looked at the right place FreeGauge will not fail if you try to fail a field that is not allocated.

maddyscientist:
This looks good now. Merging.

PR #441, opened 2016-03-04
Covdev wrongly assigns the neighbour

AlexVaq:
If splitted in more than 2 gpus per direction, the covariant derivative assigns a wrong neighbour and multiGPU gives the wrong result. This was fixed for 0.8.x, but a fix should be applied to the release/0.7.x for those wishing to stick with this old version.

### DIFF
--- a/lib/covDev.cu
+++ b/lib/covDev.cu
@@ -394,8 +394,8 @@ namespace quda
           MsgHandle *mh_send;
           MsgHandle *mh_from;
 
-          mh_send = comm_declare_send_relative	(send, dir, rel,      ghostBytes);
-          mh_from = comm_declare_receive_relative	(recv, dir, rel*(-1), ghostBytes);
+          mh_send = comm_declare_send_relative(send, dir, rel*(-1), ghostBytes);
+          mh_from = comm_declare_receive_relative(recv, dir, rel, ghostBytes);
           comm_start (mh_send);
           comm_start (mh_from);
           comm_wait (mh_send);


maddyscientist:
Curious: is there any reason, at this point, why someone would favour using 0.7.x over 0.8.x?

AlexVaq:
I suppose compatibility, or lazyness. Actually there is one case in my group, the code around QUDA is so big and complex that updating to 0.8.x would break everything.

But I agree, we should move on, and that's why I didn't bother to fix the issue before.

maddyscientist:
As far as I know there was no real interface changes with QUDA 0.8, so what breaks?

AlexVaq:
The code was really low level at some points. My custom interfaces for disconnected, for instance, were broken, but since it was a small piece of code, I updated it. This code is a supermonster that calculates everything, and requires many modifications to run with 0.8.x.

I lobby in my group to make people add their features to the QUDA repository (after they clean them), so we never run into this issue, but so far I was not very successful...

maddyscientist:
Ok understood.  Will getting multigrid convince them ;)

AlexVaq:
Sure! Actually I suggested them seriously contributing to that for the twisted-clover... At the moment, we're switching to openQCD, but I'd love to have multigrid for our regularization in QUDA.

maddyscientist:
@mathiaswagner do you have an opinion on whether we should issue a new 0.7 release with minor bug fixes?  Are there any others beyond this one?

mathiaswagner:
I certainly don't mind merging this pull request. On the other hand I'd really prefer people to move on to newer versions of quda so unless people stop using quda for this fix not being in an official release I'd rather not do a release.
@AlexVaq: Do we need a released version or are people fine with just pulling release/0.7.x (with the fixed included) from github? The way I understood it the group has some custom QUDA things anyway that don't make it back so they might just be fine.

AlexVaq:
I don’t think a new release is necessary. Actually people might not want to update to the new release for the same reasons they don’t update to 0.8.x (you know, if it works, don’t touch it… You can’t imagine how strong the inertia is). I’d just fix the released 0.7.x. All people I know that use the covariant derivative have been warned anyway.

maddyscientist:
Ok, I've merged this in.  If there is reason to make a new 0.7.x release we can do so.

PR #442, opened 2016-03-15
Feature/mobius optimize

maddyscientist:
This pull is a significant clean up of the 4-d preconditioned chiral operators:
* Added support for symmetric preconditioning
* Even/odd convention is now consistent with the rest of QUDA (this will break any external applications using this, and necessitate an interface update, e.g., `QUDA_EVEN_EVEN_ASYMMETRIC` -> `QUDA_ODD_ODD_ASYMMETRIC`
* Optimized performance of M5 inversion
* Replaced previous `DiracMobiusDomainWallPC` class with `DiracMobius` and `DiracMobiusPC` classes, for consistency with QUDA, and to add support for solving full system for Mobius fermions
* Clean up of host unit test code to reduce boilerplate
* Fixed bugs in multi-process host unit test code for 4-d preconditioned chiral operators
* Fixed bugs in reporting of memory bandwidth for mixed-precision BLAS kernels


mathiaswagner:
We should make sure to add a warning that this might break existing codes in the NEWS file but apart from that (and asking the users to test) this looks good.

PR #443, opened 2016-03-16
Compile cpp files first with cmake

mathiaswagner:
This is a bit of a hack. It seems to work though. Please give it a try.

One downside might be compilation time as all cpp files are build first before starting work on cu files.
However, errors in cpp files should show up faster.
Overall compilation time might still be the same as I guess without the hack all cu files had to compile before cmake started building cpp files.


maddyscientist:
This looks a nice simply work around.

PR #447, opened 2016-03-18
Multigrid optimization (and more)

maddyscientist:
This pull request provides a significant update for adaptive multigrid, as well as multiple performance and bug fixes
* Massively optimized adaptive multigrid algorithm
   * improved strong scaling
   * improved performance of the coarse dslash operator through auto-tuned fine grained parallelization
   * support for even-odd preconditioned multigrid at any or all levels
   * added partial support for half-precision on the top-level smoother
   * added support for QOPQDP-style recursive multigrid, where each level is wrapped by a GCR solver solved to some target precision
   * extended the interface to allow more variables to be set at runtime and not compile time
   * memory optimizations to reduce maximum GPU memory used when running multigrid on even-odd reduced system
   * by default double-precision multigrid routines are not built in order to reduce compile time
   * added new test `multigrid_benchmark` used for benchmarking performance of the coarse dslash operator
* Augmented the auto profiler to include cudaMemcpys (just replace `cudaMemcpy` with `qudaMemcpy` to enable this)
* Enabled auto-tuning of DslashPolicy, to automatically determine whether to use separate kernels for the halo regions or a single fused kernel
* Extended the auto tuner to allow for autotuning over arbitrary parameters: this is utilized when tuning for degree of parallelization in the the coarse dslash, and when tuning for dslash policy
* Fixed bug in non-degenerate twisted-mass fused halo kernel
* Small optimizations of fused halo kernels
* Improved debugging: when `HOST_DEBUG` is enabled, all allocated fields will be filled with `Nan` to better debug initialization errors
* Multiple minor bugs fixes and compiler warning removals

(this pull request should not be completed until after #442 has been merged, since this branch already been merged into feature/multigrid (it was required for doing some testing in feature/multigrid)


mathiaswagner:
Not sure whether this is an old error as it also appears in the P2P pull request but it looks like some shuffle instructions are included in the Fermi build. I'll continue looking at this later.

maddyscientist:
Ah yes I forgot about that.  I'll add a workaround for Fermi (slow but correct).

maddyscientist:
Separately, I had been wondering how important it is to maintain Fermi support for QUDA 0.9.  At least for multigrid, the memory requirements are going to make it unlikely I think that many users will be running this on Fermis.  @bjoo do you think it important that we get multigrid running on Fermi for the older Jlab clusters.

maddyscientist:
Ok, just pushed changes that fixes Fermi compilation.  There was another issue with the restrictor that needed to be fixed as well (Femri can't support as large aggregate sizes since it required more shared memory than Kepler when doing reductions due to the lack of a `__shfl` instruction.

mathiaswagner:
I guess it's fine to keep the multi grid branch alive?

maddyscientist:
Yes, still lots and lots to do there :wink:

Issue #448, opened 2016-03-28
CMAKE throwing "lib/gitversion.dummy" error


cpviolator:
Hi guys,

I'm trying to build a simple 0.8.0 master copy build on Centos. I go through the motions of making a build directory, running `cmake ../quda`, I set the compute mode for 2.0 turn off some Dirac operators builds with `ccmake .` in the build directory, compile then generate, then run `make -j 16`. The error that is thrown is not something I've been able to fix, it goes:

`[howarth@compphys bld_quda]$ make -j 16`
`[  3%] Built target quda_test`
`make[2]: *** No rule to make target "lib/gitversion.dummy", needed by` `"lib/CMakeFiles/quda.dir/version.cpp.o".  Stop.`
`make[1]: *** [lib/CMakeFiles/quda.dir/all] Error 2`
`make: *** [all] Error 2`
`[howarth@compphys bld_quda]$`

The same error gets thrown on Stampede. The now deprecated `./configure` and `make` works fine, but I'd rather get cmake working. Am I missing something spectacularly obvious?

maddyscientist:
What version of cmake, nvcc, and gcc do you have?

This is for @mathiaswagner, he's the cmake expert :wink:

maddyscientist:
One thing though: can you check if the problem also occurs in the develop branch?  There have been a number of minor changes to the cmake build in develop, so it would be good to rule this out as something that hasn't been fixed already.

cpviolator:
GCC: 4.4.7
NVCC: CUDA 6.0
CMAKE: 2.8.12.2
GIT: 1.7.1

mathiaswagner:
Yes, cmake version as well as CMakeLists.txt are helpful. Do you build in source or in a separate build directory?

maddyscientist:
I can see from what @cpviolator has written above, that he's using a separate build directory.

cpviolator:
I build in a separate directory. Here is my `CMakeLists.txt` from my local server. I'll post the TACC Stampede info too.

[CMakeLists.txt](https://github.com/lattice/quda/files/191422/CMakeLists.txt)

mathiaswagner:
Right. I see it now.

mathiaswagner:
Not my morning. I meant CMakeCache.txt. sorry.

cpviolator:
No worries, here's the `CMakeCache.txt` from my local. TACC Stampede data will follow.
[...]

PR #449, opened 2016-03-29
make cmake build depend on git HEAD state

mathiaswagner:
To fix unnecessary always recompile of version.cpp and issues with the gitversion.dummy hack


mathiaswagner:
@cpviolator This is an even improved version for #448. Do you mind giving that a try?

cpviolator:
Yep, that builds fine too, Centos 6.5.

mathiaswagner:
thanks!

Issue #450, opened 2016-04-13
Multigrid fails for unpreconditioned clover


AlexVaq:
There seems to be a bug with unpreconditioned clover and the multigrid solver, but to see it one needs to use a smooth configuration and calculate the clover term (these are not default settings of the test). What I see is that the null vectors are poorly computed, and the resulting solver has a really bad performance.

The issue doesn't exists in the preconditioned clover, or in the twisted-clover (preconditioned and unpreconditioned) at mu = 0.0, which is very surprising because twisted-clover is suppose to do the same operations than clover for the unpreconditioned case.

I observed this problem in the feature/multigrid and feature/tm-multigrid branches.

maddyscientist:
Is there a way for me to reproduce this problem using the internal QUDA tests?

AlexVaq:
I'm not sure: in a unit configuration the clover term is irrelevant. I didn't try a totally random configuration, maybe if you just create the clover term there you'll see it.

I'll try tomorrow.

maddyscientist:
@AlexVaq any update on this?

AlexVaq:
Not yet, I'll download the current develop and try again. It's curious that the BiCG solver directly fails...

If the bug is still there, I'll also try to find an easy way to reproduce it.

AlexVaq:
It seems to me that the bug affects clover fermions, and not specifically multigrid. The part that fails is the BiCG solver, and it fails right from the start.

In order to see the bug, with minimal changes in multigrid_invert_test.cpp, one can try this:

1. Change clover_coeff from 1 to 1e-20. This ensures our clover fermions will be like wilson. I mean, we should get the same result with wilson than with clover, but we will enforce QUDA to use clover fermions.

2. Change the unit configuration for a random one. This is achieved by commenting line 432 and uncommenting line 430.

3. Run the following two tests: 

./multigrid_invert_test --dslash-type wilson --sdim 12 --tdim 12 --prec single --prec-sloppy single --precon-type mr --solve-type direct

./multigrid_invert_test --dslash-type clover --sdim 12 --tdim 12 --prec single --prec-sloppy single --precon-type mr --solve-type direct

and compare. Everythin looks normal, but with wilson the solver converges in 10-15 iterations up to 1e-7, and with clover the solver stops after reaching the iteration limit without converging.

invert_test seems to work properly in all cases, although QUDA detects a memory leak when the solver is BiCG, I think I should look into that.

AlexVaq:
[...]

Issue #426, opened 2016-01-08
Add easy method to change mass/kappa to MG preconditioner 


maddyscientist:
We currently need to rerun the entire setup if we want to change the mass/kappa parameter for the MG preconditioner.  Since the mass parameter is just a diagonal constant, it should be easy make this a trivial change.

cpviolator:
Just pushed to hotfix/delta_mu to address this issue. Three new parameters are added to `include/quda.h`, namely `delta_muMG`, `delta_massMG`, and `delta_kappaMG`. These are multiplicative factors which change the respective values of `mu`, `mass` and `kappa` when going from the finest grid to coarser grids.

Passing values to `delta_XXXMG` from the command line is done with "`--delta-XXXMG <value>`" and the defaults are set to `1.0`.

maddyscientist:
Closing this issue, since more thought by me realized that this isn't actually that easy to solve.  Since we use even/odd preconditioning, the inverse clover matrix is included in the coarse link matrix definition.  Hence the mass is not included as a simple diagonal, and can't be rescaled after the fact.

Regardless, since the coase-link matrix computation has been heavily optimized, this is less of an issue that it once was, especially on Volta.

PR #451, opened 2016-04-13
Feature/tm multigrid

AlexVaq:
Adds twisted-mass and twisted-clover support for the multigrid solver.


maddyscientist:
I'll review this shortly.

Have you actually done testing of twisted-clover MG to see if you can get speedup yet?

Am I correct in thinking that I should be able to test this with real lattices with twisted-mass fermions and it should just work?

AlexVaq:
> Have you actually done testing of twisted-clover MG to see if you can get speedup yet?
> 
No, I didn't test performance. I was really discouraged because in my internal tests I couldn't even see a speed up with wilson. I understood that MG is sensitive to the input parameters (the first time I was using gcr as preconditioner, instead of mr, and the performance was really slow). Also maybe my configuration was not problematic enough.
> Am I correct in thinking that I should be able to test this with real lattices with twisted-mass fermions and it should just work?
> 
Yes, you should. I did many non-trivial tests (in my single GPU, but I enabled the partition option) and compared spinors obtained with different inverters, component by component. It should be 100% right, unless I forgot something. I see that MG was a very big update, so I might have overlooked/misunderstood some option or parameter.

maddyscientist:
Yes the parameters are a bit of an explosion.  I'll test twisted-mass and see how I get on.  With Wilson and clover, with the correct parameters, I don't ever see a slow down now for MG versus BiCGstab.  I should update the wiki with suggested parameters to use since more and more people are going to want to use this.

maddyscientist:
Doing some quick testing of twisted mass multigrid, and things are looking very good for zero tuning (I just used parameters that I use for Wilson testing)

First iteration count, which clearly shows that critical slowing down with mu is defeated 👍 

| mu | CG | BiCGstab | MG |
| --- | --- | --- | --- |
| 0.1 | 90 | 68 | 17 |
| 0.05 | 165 | 186 | 18 |
| 0.01 | 722 | - | 19 |
| 0.005 | 1333 | - | 20 |
| 0.001 | 3776 | - | 20 |

Second, time to solution, which shows we have speedup 👍 👍 

| mu | CG | BiCGstab | MG |
| --- | --- | --- | --- |
| 0.1 | 0.31 | 0.61 | 0.85 |
| 0.05 | 0.52 | 1.56 | 1.15 |
| 0.01 | 1.98 | - | 1.08 |
| 0.005 | 3.54 | - | 1.35 |
| 0.001 | 9.94 | - | 1.18 |

I did these tests using a regular Wilson 24^3x64 lattice, running at around kappa critical, using 3 GPUs (split along z axis).

Parameters for running multigrid_invert_test were:
```
--prec double
--prec-sloppy single
--prec-precondition half
--precon-type mr (smoother type)
--mg-levels 3
--mg-block-size 0 4 4 4 4 (first aggregation size)
--mg-block-size 1 3 3 1 8 (second aggregation size)
--mg-nu-pre 5 (pre smoothing steps)
--mg-nu-post 5 (post smoothing steps)
--tolhq 0.2 (tolerance of the GCR solver that wraps each coarse level, poorly named)
--recon 12
--recon-sloppy 8
--recon-precondition 8
```

Most of the code changes look, so it's almost good to merge.  A couple of things though:
* Support for dynamical clover inversion is FUGLY.  Can't this be rewritten using more modern QUDA C++ style and no macros?
* Why when do the `DiracTwistedClover::createCoarseOp` and `DiracTwistedCloverPC::createCoarseOp` methods pass `QUDA_CLOVER_DIRAC` and `QUDA_CLOVERPC_DIRAC` instead of `QUDA_TWISTED_CLOVER` and `QUDA_TWISTED_CLOVERPC_DIRAC`.  I understand that passing in the `a` parameters shows that it has a twist, but I think it would make the underlying coarse operator construction more clear if the twisted-clover enums were passed in

Great work though @AlexVaq!

AlexVaq:
By the way, @maddyscientist, I noticed that unpreconditioned clover doesn't
work properly. For some reason, it won't converge in my test configuration
(or converge really slowly), whereas twisted-clover with mu = 0 will give
me the right result. I tried to figure out in the code what was wrong, but
I couldn't find the culprit. Also, I was afraid this was a regression of my
commits, but I found the same behavior in the vanilla feature/multigrid
branch. Do you know anything about this?

This is issue #450.

AlexVaq:
I addressed your second comment (use QUDA_TWISTED_WHATEVER_DIRAC when calling CoarseOp), now I think everything is cleaner. I removed the twisted boolean because if we now exactly which dslash we're dealing with we don't really need it.

AlexVaq:
I'm thinking what to do with dynamic clover. It's really ugly as it is now, but the performance is good. If I try to write it as in clover_invert.cu, I'll be forced to stop reusing a lot of variables, and I'll roll-back some loops (that should be unrolled by the compiler anyway). If you remember, the performance impact in twisted-clover was huge (like a factor of 2).

Any advice?

kostrzewa:
@maddyscientist @AlexVaq Since this is in the early stages I would like to make a comment. I have extended the tmLQCD interface to be able to provide a direct cross-check of this for realistic configurations wit

kostrzewa:
As a further comment it would be nice to discuss at some point, likely after testing and speed-up have been confirmed on a realistic lattice, what kind of modifications and parameters would/might be necessary for running tm+clover @ m_phys efficiently. Presently I would expect that the null vector generation using BiCGstab could be a major bottleneck because the solver stalls, but maybe I'm missing something:

```
MG level 1 (GPU): BiCGstab: 497 iterations, <r,r> = 2.490378e-01, |r|/|b| = 4.590827e-05
MG level 1 (GPU): BiCGstab: 498 iterations, <r,r> = 2.532646e-01, |r|/|b| = 4.629622e-05
MG level 1 (GPU): BiCGstab: 499 iterations, <r,r> = 2.902742e-01, |r|/|b| = 4.956356e-05
MG level 1 (GPU): BiCGstab: 500 iterations, <r,r> = 2.513716e-01, |r|/|b| = 4.612288e-05
MG level 1 (GPU): WARNING: Exceeded maximum iterations 500
MG level 1 (GPU): BiCGstab: Reliable updates = 5
MG level 1 (GPU): BiCGstab: Convergence at 500 iterations, L2 relative residual: iterated = 4.612288e-05, true = 4.664247e-05
MG level 1 (GPU): Solution = 1065.88
MG level 1 (GPU): smoother has operator N4quda6DiracME
MG level 1 (GPU): Creating a MR solver
MG level 1 (GPU): Creating a MR solver
```

This is on the lightest (non-clover) twisted mass ensemble on an L=48, T=96 lattice. kappa=kappa_c, amu=0.0015

AlexVaq:
Oh, yes. @maddyscientist is aware of the problems of BiCG for twisted-mass/clover, she told me she has some ideas...

Thanks for the tmLQCD interface. That definitely will make things easier.

maddyscientist:
@kostrzewa To fix this basis issue is simple, the problem is due to the currently poor interface design that multigrid is using.  You need to set `QudaMultigridParam::invert_param->gamma_basis = QUDA_DEGRAND_ROSSI_GAMMA_BASIS`.  This only affects the basis the internal multigrid set uses, and doesn't involve any interaction with the application.  The basis that the final solver is determined by the instance of `QudaInvertParam` that is passed to `invertQuda`.  Make sense?

The interface for multigrid is going to get a total rewrite.  It's current poor design is the product of trying to get something working quickly, while at the same time not knowing exactly what would be needed in the interface.

maddyscientist:
@AlexVaq I definitely think we should go with the clean code for the multigrid setup.  For one thing, this hasn't actually been offloaded to the GPU yet so this performance issue is a theoretical one.  I'd really like this code to be clean, and manageable.

Moreover, I still definitely think the compiler should be able to handle this as there is nothing fundamentally hard there for the compiler to manage.  This was something I never got around to looking at; what would be great would be two kernels that we could compare directly against and benchmark, one written in high level C++ and the other in hand-unrolled macro hell.  Given we have a self-contained high-clover inversion routine (clover_invert.cu), perhaps you could make an alternate one that is hand optimized and then I can work on improving the high-level one (and / or file bugs at NVIDIA against the compiler 😉 ) until the performance matches.  Sound reasonable?

maddyscientist:
@kostrzewa Regarding BiCGstab stalling.  In the adaptive setup, convergence is only requested to 5e-6, and I've found when running on larger lattices with regular Wilson and clover that at small quark mass, 500 iterations often isn't enough to reach this requested tolerance (which was chosen arbitrrily), and it might only reach ~5e-5.  But regardless, the MG algorithm converges very well.

For twisted-mass fermions in general, BiCGstab is a poor solver due.  In the experiments I did yesterday though I found that while BiCGstab might not solve Ax=b, e.g., in the table above it fails to converge for small mu, it will converge for Ax=0 (with random initial x).  Similar behaviour was also observed with Wilson and clover, namely that the BiCGstab on the homogenous system converges much quicker.

So the convergence you report of 5e-5 looks absolutely fine.  I suspect when the solver is run through to completion (once the gamma basis issue is fixed), it will work out of the box with no changes.  (Don't quote me on that though :wink:

Once you got the basic working, you can play with this setting in lib/multigrid.cpp.  These are hard coded for now, but will exposed when the interface is rewritten.  For example, you could up the max iterations to 1000, to ensure that 5e-6 tolerance is obtained.
```c++
    // set null-space generation options - need to expose these                                                                                                                                                                                                                 
    solverParam.maxiter = 500;
    solverParam.tol = 5e-6;
    solverParam.use_init_guess = QUDA_USE_INIT_GUESS_YES;
    solverParam.inv_type = QUDA_BICGSTAB_INVERTER;
```
If it wasn't for this unexpected BiCGstab convergence on the homogenous system, then we would have had to use a different solver for this.  CG would be something to consider, though then we would in principle get components rich the singular vectors of the operator, and not necessary eigenvectors.  This may not matter though since when local aggregation (the geometric blocks) is used, eigenvectors and singular vectors seem to be fairly colinear.

AlexVaq:
Ok, I'll keep the dirty one for production, if we ever use dynamic inversion with MG, and upload a nice and clean code based on clover_invert.cu.

maddyscientist:
@AlexVaq How about move the dirty one into a separate .cuh file, and include / enable it with a macro setting or something like that?  That in itself would enable easy testing between the two variants, and we can then delete the dirty one once the clean is as fast.

kostrzewa:
@maddyscientist 
> To fix this basis issue is simple, the problem is due to the currently poor interface design that multigrid is using. You need to set QudaMultigridParam::invert_param->gamma_basis = QUDA_DEGRAND_ROSSI_GAMMA_BASIS. This only affects the basis the internal multigrid set uses, and doesn't involve any interaction with the application. The basis that the final solver is determined by the instance of QudaInvertParam that is passed to invertQuda. Make sense?

Thanks. In that case I need to use two sets of ```QudaInvertParam``` (currently my ```QudaMultigridParam::invert_param``` points to the ```QudaInvertParam``` that I pass to ```invertQuda```. This was for testing purposes only since I would obviously like to adjust kappa, mu and other parameters independently for the MG setup and the solve. 

> Regarding BiCGstab stalling. In the adaptive setup, convergence is only requested to 5e-6, and I've found when running on larger lattices with regular Wilson and clover that at small quark mass, 500 iterations often isn't enough to reach this requested tolerance (which was chosen arbitrarily), and it might only reach ~5e-5. But regardless, the MG algorithm converges very well.

Okay, thanks, this would be in line with experience from other MG solvers. An option would certainly be to set kappa [<]< kappa_c, mu=0.0 for the null vector generation (and possibly other parts of the MG). The low-mode eigenvectors, I guess, should be somewhat independent of the Wilson quark mass. Did you study what happens to the solve times when BiCGstab is terminated very early; after 100 iterations, say?

maddyscientist:
I haven't focused at all on the setup process yet, so the study of optimizing this has yet to happen.  The main initial application this MG was developed for requires ~100,000 solves per gauge field, so set up time is irrelevant.

Once this is cleaned up I do plan on drastically optimizing this with the goal of deploying it for HMC.

Also, note for now that I think setup and solve need to have the kappa value. It's on the to do list to fix this though ([#426](https://github.com/lattice/quda/issues/426)).

AlexVaq:
@maddyscientist I think you'll like this version much more. You can review the lib/dyninv_clover_mg.h file and the code that replaces what's in there in lib/coarse_op.cuh.

maddyscientist:
Ok, I've merged in the latest develop branch into this branch, and also I've also added the fix when using `QUDA_CHIRAL_GAMMA_BASIS` in copy_color_spinor.cuh.

maddyscientist:
Pushed some more micro patches:
* Fixed broken cmake compilation
* Removed compiler warning from coarse_op.cuh
* Removed the `nSpin==4` and `nColor==3` from the dynamical clover inversion kernels, since this will break when compiled for the GPU.  Added some sanity equivalent sanity checks though to the start of `calculateY` that are equivalent.

This looks good for merging now.  I will merge this once the compilation checks complete.

The coarse operator construction code is beginning to be a compile-time sink.  And this code isn't even instantiated as GPU code yet.  A headache for the future, especially once domain-wall and staggered operator support is added.

PR #452, opened 2016-04-13
in loadClover, set solve_pc = true also for QUDA_NORMERR_PC_SOLVE

kostrzewa:
loadClover, set solve_pc = true also for QUDA_NORMERR_PC_SOLVE

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -728,7 +728,8 @@ void loadCloverQuda(void *h_clover, void *h_clovinv, QudaInvertParam *inv_param)
 
   // determines whether operator is preconditioned when calling invertQuda()
   bool pc_solve = (inv_param->solve_type == QUDA_DIRECT_PC_SOLVE ||
-      inv_param->solve_type == QUDA_NORMOP_PC_SOLVE);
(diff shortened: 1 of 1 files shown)


mathiaswagner:

mathiaswagner:

kostrzewa:
I'm not sure whether this doesn't cause some regression down the line or why it was omitted in the first place. In tmLQCD, using QUDA_NORMERR_PC_SOLVE is convenient for us because it's what we do in our CPU routines. As a result, the residuals are easy to compare. For the NORMOP_PC_SOLVE we have to add fudge factors to get our desired residuals.

maddyscientist:
Thanks for the bug fix.  This was just an oversight on my part when I added support for `QUDA_NORMERR_PC_SOLVE`.

Is there any need for this to be in the next 0.8.x release?  If this goes into develop then it won't show up in a release until 0.9.0 which is still some time away.  If you would like this to be in the next official release (and in the stable branch) then I'd suggest closing this an resubmitting it for the release/0.8.x branch.  Otherwise I'll just merge this into develop.

kostrzewa:
Okay, thanks. I'll resubmit for the 0.8.x branch then. I guess we prefer to use "the stable branch" for production, also because of the issue with the gamma basis mentioned in #453

PR #457, opened 2016-04-23
Feature/field container

maddyscientist:
This pull introduces an arbitrary container `CompositeColorSpinorFieldDescriptor` that allows us to form a large `ColorSpinorField`, which is split into multiple constituents.  The full field can be addressed directly or the composite parts can be addressed individually.  The addition of this feature allows the deflation algorithms to run directly on the full system, and not just the preconditioned single parity operator.


maddyscientist:
@alexstrel I merged in the develop branch, as expected the conflicts were with respect the removal total_length and the change in ghost zone allocation.  This seemed fairly trivial to merge so I just went ahead and did this (mostly just deletion, since the change in ghost zones simplifies a lot of logic)..All seems fine with respect to regular solvers, but could you run a deflation test to confirm all is ok there?

include/invert_quda.h
```diff
@@ -720,6 +720,7 @@ namespace quda {
   };
 
   struct DeflationParam;//Forward declaration
+  typedef cudaColorSpinorField cudaColorSpinorFieldSet;
```
mathiaswagner: It seems this typedef is just used to make it more obvious that this spinor contains multiple components. Or is there more to it?

mathiaswagner:
Maybe I am missing something here but I am getting a bit confused here: We have
```typedef std::vector<ColorSpinorField*> CompositeColorSpinorField;```
which is obviously to be used as a vector of `ColorSpinorField`.

The `CompositeColorSpinorFieldDescriptor` name suggests that an object of that type provides a description for a `CompositeColorSpinorField`.

But then finally the `CompositeColorSpinorField` is just a member of a regular `ColorSpinorField` and in invert.h we have a typedef to a ColorSpinorFieldSet. 

It might be that I don't get the design idea behind that. Just one obvious thing that gives me a headache: I can put a ColorSpinorField which is a CompositeField in the Components of the ColorSpinorField and go on with that (like a Matryoshka doll).

File: include/invert_quda.h (hunk 1 above)
Comment (alexstrel) [inline include/invert_quda.h]: Correct.

**Comment by alexstrel** [comment]
to @mathiaswagner :  
> invert.h we have a typedef to a ColorSpinorFieldSet. 
this is an "internal" typedef specific for deflation solvers, one should not consider it in any other context. This indicates that a pointer to the ColorSpinorField object will in fact point to a set of eigenvectors and not to a single spinor field. It was introduced to avoid confusions, but if you think it makes it more obscure, we can revert to the original version (i.e. pointers to spinor fields )
> I can put a ColorSpinorField which is a CompositeField in the Components of the ColorSpinorField and go on
A  CompositeField is an extension of the QUDA_REFERENCE_FIELD_CREATE -based fields. One cannot create a composite field that is a component of another composite field , see , e.g., line 58 in color_spinor_field.h. But thanks for pointing out this moment, I will double check it.

**Comment by maddyscientist** [comment]
The composite field construction is a generalization of the even/odd reference members.  There is definitely scope for cleaning it up, but it serves it purpose for now.  I'd suggest this take place in the multi-src development though.

**Comment by alexstrel** [comment]
in QudaMemCopy class, cudaMemcpyKind member does not allow for cudaMemcpyDefault. Was that designed in purpose? inv_eigcg_quda.cpp and bals_magma.cu uses that one heavily. (I would prefer to modify QudaMemCopy class, if there are no objections)

**Comment by maddyscientist** [comment]
That was just an oversite by me.  Please update it to support `cudaMemcpyDefault`.

**Comment by alexstrel** [comment]
if there are no further questions/comments, this branch can be merged.

**Comment by mathiaswagner** [comment]
Yes, good to merge.

On 25.04.2016, at 19:40, Alexei Strelchenko <notifications@github.com<mailto:notifications@github.com>> wrote:


if there are no further questions/comments, this branch can be merged.

-
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub<https://github.com/lattice/quda/pull/457#issuecomment-214454169>

File: lib/cuda_color_spinor_field.cu [hunk 2]
```diff
@@ -1863,23 +1875,23 @@ namespace quda {
     return out;
   }
 
-//! for deflated solvers:
-  cudaColorSpinorField& cudaColorSpinorField::Eigenvec(const int idx) const {
+//! for composite fields:
+  cudaColorSpinorField& cudaColorSpinorField::Component(const int idx) const {
```
mathiaswagner: Any particular reason why `Component` is only around for `cudaColorSpinorField` and not `ColorSpinorField` ? Apart from the obvious only implement what is currently needed?

alexstrel: Correct, I did not consider more generic case since current implementation was sufficient for deflation methods.

PR #458, opened 2016-04-26
Feature/complex cleanup

maddyscientist:
This pull request is a significant clean up of the complex-number usage: we now use the complex-number class in complex_quda.h for supporting complex-valued matrices in quda_matrix.h.  This is a wide-ranging change, albeit a straightforward one, affecting all code that uses the matrix class (gauge fixing, force terms, plaquette, link smearing, etc.).

This pull request also fixes a bunch of compiler warnings.

### DIFF
--- a/include/complex_quda.h
+++ b/include/complex_quda.h
@@ -418,6 +418,14 @@ public:
       return *this;
     }
 
+  __host__ __device__
+    inline complex<ValueType>& operator*=(const ValueType z)
+    {
+      this->x *= z;
+      this->y *= z;
+      return *this;
+    }
+
   __host__ __device__ inline ValueType real() const volatile;
   __host__ __device__ inline ValueType imag() const volatile;
   __host__ __device__ inline ValueType real() const;
@@ -511,6 +519,14 @@ public:
       return *this;
     }
 
+  __host__ __device__
+    inline complex<float>& operator*=(const float z)
+    {
+      this->x *= z;
+      this->y *= z;
+      return *this;
+    }
+
   // Let the compiler synthesize the copy and assignment operators.
   __host__ __device__ inline complex<float>(const volatile complex<float> & z)
   {
@@ -591,6 +607,14 @@ public:
     }
 
   __host__ __device__
+    inline complex<double>& operator+=(const complex<float> z)
+    {
+      real(real()+z.real());
+      imag(imag()+z.imag());
+      return *this;
+    }
+
+  __host__ __device__
     inline complex<double>& operator-=(const complex<double> z)
     {
       real(real()-z.real());
@@ -609,6 +633,14 @@ public:
     inline complex<double>& operator/=(const complex<double> z)
     {
       *this = *this / z;
+      return *this;
+    }
+
+  __host__ __device__
+    inline complex<double>& operator*=(const double z)
+    {
+      this->x *= z;
+      this->y *= z;
       return *this;
     }
 
--- a/lib/clover_deriv_quda.cu
+++ b/lib/clover_deriv_quda.cu
@@ -13,14 +13,14 @@ namespace quda {
   
 #ifdef GPU_CLOVER_DIRAC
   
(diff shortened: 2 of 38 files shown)


include/complex_quda.h
```diff
@@ -511,6 +519,14 @@ struct complex <float> : public cuFloatComplex
       return *this;
     }
 
+  __host__ __device__
```
mathiaswagner: As this is a cleanup: Is there a reason not to template this?

lib/clover_deriv_quda.cu
```diff
@@ -69,42 +70,35 @@ namespace quda {
...
     const int& nu = arg.nu;
 
-    Matrix<Cmplx,3> thisForce;
-    Matrix<Cmplx,3> otherForce;
+    M thisForce, otherForce;
 
     // U[mu](x) U[nu](x+mu) U[*mu](x+nu) U[*nu](x) Oprod(x)
     {
       int d[4] = {0, 0, 0, 0};
+      M U1, U2, U3, U4, Oprod1, Oprod2;
```
mathiaswagner: At a first glance I did not realize that `M` is the type here. Might be better to be a little more verbose in the typedef?

include/complex_quda.h
maddyscientist: Not sure I follow your comment here?  Template what exactly here?

lib/clover_deriv_quda.cu
maddyscientist: How about?
```c++
typedef Matrix<complex<Float>,3> Link;
```

mathiaswagner:
What about using C++1's `using` for Complex 3x3 matrices, e.g.
```
template<class T>
using CMatrix3 = Matrix<complex<T>,3>; 
``` 
?

include/complex_quda.h
mathiaswagner: There are a couple of functions like
```
inline complex<float>& operator-=(const complex<float> z)
```

which could be

```
template<class T>
inline complex<T>& operator-=(const complex<T> z)
```

Same for things like operations between `complex<T>` and `T`, e.g.
```
inline complex<float>& operator*=(const float z)
```

lib/clover_deriv_quda.cu
mathiaswagner: I don't particular like `Matrix` and `matrix`. I like `Link`. This is not a general context and `Link` makes it obvious what this is.

mathiaswagner:
The code builds and I really like the amount of cleanup. So apart from adding some more **optional**  C++11 / template convenience I am perfectly happy.

include/complex_quda.h
maddyscientist: I don't think these can be further templated.  Note that there is a generic `complex<ValueType>` base class that implements operations like `-=` and `*=` etc., but there are also the specializations `complex<float>` and `complex<double>` that build upon cuComplex.h.  These latter specializations need to be fully reimplemented, hence the functions you refer to are written using explicit `float` and `double` types.  Agree?

mathiaswagner: Got it.

On 26.04.2016, at 17:54, maddyscientist <notifications@github.com<mailto:notifications@github.com>> wrote:


In include/complex_quda.h<https://github.com/lattice/quda/pull/458#discussion_r61110595>:

> @@ -511,6 +519,14 @@ struct complex <float> : public cuFloatComplex
>        return *this;
>      }
>
> +  __host__ __device__


I don't think these can be further templated. Note that there is a generic complex<ValueType> base class that implements operations like -= and *= etc., but there are also the specializations complex<float> and complex<double> that build upon cuComplex.h. These latter specializations need to be fully reimplemented, hence the functions you refer to are written using explicit float and double types. Agree?

-
You are receiving this because you were assigned.
Reply to this email directly or view it on GitHub<https://github.com/lattice/quda/pull/458/files/9b831966c56553288fbe635a0924020203b7f445#r61110595>

PR #459, opened 2016-04-26
Feature/unitarization cleanup

maddyscientist:
This pull request is primary a clean up, and does the following:
* Fixes virtually all compiler warnings when compiling QUDA with cmake
* Force unitarization computation rewritten to use gauge-field accessors
* Force and link unitarization parameters are now set when running, rather than frozen at first instance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,7 +163,7 @@ endif()
 if(${QUDA_MPI})
   add_definitions(-DMPI_COMMS)
   set(COMM_OBJS comm_mpi.cpp)
-  include_directories(${MPI_CXX_INCLUDE_PATH})
+  include_directories(SYSTEM ${MPI_CXX_INCLUDE_PATH})
 endif()
 
 if (${QUDA_QMP})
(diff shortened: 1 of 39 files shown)


mathiaswagner:
I am looking into clang compile. I might have to do some more cmake. The c++11 warning might come from the use of `CMAKE_C_COMPILER` as host compiler even if `CUDA_HOST_CPP` is set. Overriding this with `CMAKE_CXX_COMPILER` seems to get rid of some warnings. Not sure whether this breaks something else in cmake. 
I am also in favor of using `CMAKE_CXX_STANDARD` instead of manually specifying `--std=c++11`.
I will check and probably add that to the branch later.

mathiaswagner:
I have updated `CMakeLists.txt` and it builds fine for me. It removes the C++11 related warning for me.
The build with clang still triggers some more warning. Some might be due that I only quickly checked that build for single GPU.

mathiaswagner:
@maddyscientist: I am good to merge but please give the update cmake a quick try.

mathiaswagner:
I turned off some more warning from system headers

PR #468, opened 2016-05-10
Feature/milc rhmc cleanup

maddyscientist:
This pull request fixes some issues with MILC RHMC that creeped in, mainly due to lacking proper test coverage
* Fixed compilation bug when Wilson-type fermions are disabled, but staggered are enabled
* Fixed bugs HISQ force unitarization, and also autotuning now reports bandwidth for this kernel
* HISQ unitarization force test, CPU reference always done in double precision

### DIFF
--- a/lib/blas_core.h
+++ b/lib/blas_core.h
@@ -275,40 +275,39 @@ template <template <typename Float, typename FloatN> class Functor,
     errorQuda("blas has not been built for Nspin=%d fields", x.Nspin());
 #endif
   } else if (x.Precision() == QUDA_SINGLE_PRECISION) {
+    if (x.Nspin() == 4) {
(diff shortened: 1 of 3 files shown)


mathiaswagner:
Just a first quick (I did not yet look at the code in detail): MILC's strategy is to always do unitarization in double. I saw same changes related to precision here and assume they only clean up the code but wanted to make sure we also always do unitarization in double.

maddyscientist:
Yes we do always do it in double here too.  The changes here related to
1.) Make sure it works correctly when fields are in single precision (do computation in double)
2.) For CPU reference, ensure everything is done in double (storage and compute)

mathiaswagner:
Thanks. This was what I assumed but was not sure.

PR #476, opened 2016-05-31
When calling comm_abort() in (HOST)DEBUG builds raise SIGINT

mathiaswagner:
This allows to attach gdb to the process to e.g. get a backtrace where the error was caused, e.g. like

```
ERROR: wrong here (rank 0, host nvsocal2, /home/mathias/quda/tests/invert_test.cpp:126 in main())
Program received signal SIGINT, Interrupt.
0x00007ffff7bcc20b in raise (sig=2) at ../nptl/sysdeps/unix/sysv/linux/pt-raise.c:37
37      ../nptl/sysdeps/unix/sysv/linux/pt-raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff7bcc20b in raise (sig=2) at ../nptl/sysdeps/unix/sysv/linux/pt-raise.c:37
#1  0x000000000194dd6d in comm_abort (status=1) at /home/mathias/quda/lib/comm_mpi.cpp:339
#2  0x00000000018bf0f1 in main (argc=1, argv=0x7fffffffe558) at /home/mathias/quda/tests/invert_test.cpp:126

```

I did not do a lot of testing whether this somehow interferes with MPI but I believe it should be find and is anyway only enabled for (HOST)DEBUG builds.

### DIFF
--- a/include/util_quda.h
+++ b/include/util_quda.h
@@ -27,6 +27,7 @@ void popVerbosity();
 
 char *getPrintBuffer();
 
+
 // Note that __func__ is part of C++11 and has long been supported by GCC.
 
 #define zeroThread (threadIdx.x + blockDim.x*blockIdx.x==0)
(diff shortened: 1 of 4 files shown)


maddyscientist:
Excellent feature 😄

PR #481, opened 2016-06-11
Fixes a bug in asymmetric preconditioning

AlexVaq:
Very serious, the inversion will give wrong results, although since the inverter usually explodes and people were using symmetric preconditioning, it seems that nobody was affected.


AlexVaq:
Curiously, release branches DO NOT have this bug. It was introduced at some point surreptitiously, most likely because of a tiny, stupid, easy-to-overlook mistake.

PR #483, opened 2016-06-28
Fixed bug in clover inversion

maddyscientist:
Introduced with commit 38fda505cc1aec25c9c351ad8a093a5c2fcf8f0b.

@AlexVaq Can you merge and close this please?

### DIFF
--- a/lib/clover_invert.cu
+++ b/lib/clover_invert.cu
@@ -47,11 +47,11 @@ namespace quda {
       // FIXME use native ordering in the Cholseky 
       // factor of two is inherent to QUDA clover storage
       constexpr Float two = static_cast<Float>(2.0);
-      for (int i=0; i<6; i++) diag[i] = two*A[ch*36+i];
+      for (int i=0; i<6; i++) diag[i] = two*A[i];
 
       const int idtab[15]={0,1,3,6,10,2,4,7,11,5,8,12,9,13,14};
 #pragma unroll
-      for (int i=0; i<15; i++) tri[idtab[i]] = complex<Float>(two*A[ch*36+6+2*i], two*A[ch*36+6+2*i+1]);
+      for (int i=0; i<15; i++) tri[idtab[i]] = complex<Float>(two*A[6+2*i], two*A[6+2*i+1]);
 
       //Compute (T^2 + mu2) first, then invert (not optimized!):
       if (twist) {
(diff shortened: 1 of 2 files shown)


AlexVaq:
> @AlexVaq <https://github.com/AlexVaq> Can you merge and close this please?
> 


Ok, let me have a look.

AlexVaq:
Ok, I see. With the new accessor you were reading out of bounds. Most likely we would get garbage for the clover.

Looks correct now. I'll run a fast check and merge unless something goes really wrong.

maddyscientist:
Yep.  I've checked with a clover RHMC run, and it seems to be fine now.

AlexVaq:
It fails dslash_test 0, 1 and 3 with DYNAMIC_CLOVER. I’m trying to find out whether this is because the problem has not been fixed for the dynamic inversion or because the host routines failed unexpectedly.

I’m running without DYNAMIC_CLOVER, I’ll review the code for the inversion on-the-fly and let you know.

AlexVaq:
Removing DYNAMIC_CLOVER fixes it. Definitely there is a problem with the inversion on-the-fly. Looking into that now.

AlexVaq:
I'll have to check multigrid dynamic clover, so this might take long...

AlexVaq:
I believe the flaw lies in the test routine, that wrongly copies back the inverse clover term for the dynamic inversion. This would be good news, because the gpu implementation would be correct, but I want to ensure everything is right and the only way to do that is to fix this copy-back.

AlexVaq:
It seems the problem was in dslash_test.cpp: the inverse clover was not copied back to the CPU in the case of dynamic clover inversion. I added a fix, but before merging I'd like to do further testing.

AlexVaq:
Ok, in the end I think that single line does the job. @maddyscientist do you want to review this or should I go on with the merging? I passed dslash_test, invert_test and multigrid_invert_test, so I guess it's safe.

AlexVaq:
Merged #483.

maddyscientist:
@AlexVaq Thanks for fixing that last dslash_test bug.  I presume that regular clover continued to work ok?

When I have time I will plan on adding dynamic-clover support to regular clover as well to fully merge the clover twisted-clover paths.  At the same time I'll probably also make dynamic-clover a run-time policy rather than compile time, as it will make it easier to test.

AlexVaq:
Yes, clover is working as before. I didn't try clover multigrid, but I don't see how this should be affected.

A merging of paths would be very useful, it would simplify maintenance. And yes, the dynamic inversion should be selected at runtime.

PR #486, opened 2016-07-25
Enable MPS functionality

maddyscientist:
This pull adds MPS functionality to QUDA: the ability to over subscribe the GPUs if the number of processes requested per node, exceeds the number of GPUs per node.  This is enabled through the environment variable `QUDA_ENABLE_MPS=1` else it will error.


mathiaswagner:
Looks good. Nothing wrong with the code but wondering whether we should warn about an imbalanced oversubscription on a system with multiple nodes / GPU.

Issue #391, opened 2015-11-04
Single-precision restriction operators have sub-optimal memory access patterns


maddyscientist:
The memory access patterns of the restriction operators are suboptimal owing to the fact that thread blocks have a one-to-one mapping with aggregates.  Since the aggregates typically are of size `4x4x4x4`, the resulting checkerboarded x-dimension of width 2 is not wide enough to cover a cache line in single precision: `width x sizeof(complex<float>) = 16 < 32 bytes`.

Note this problem does not occur with the prolongator, since in general there is no one-to-one mapping between aggregate and CUDA thread block.  Here the CUDA thread block is free to extend between thread blocks and we have perfectly coalesced memory access patterns.

While this problem occurs for both the load of the input vector and the V rotation matrices, since the bulk of the memory operations correspond to the V rotation matrices, we need only address this problem for V to fix >90% of this.  We can address this problem in one of two ways:
1. Use a block ordering for the V fields.  This will ensure that the V loads will be contiguous across the entire block, instead of strided as is at present.
2. We can use float4 ordering for single precision V fields.  This doubles the size of each memory transaction, and as a result, consumes the entire cache line as is presently the case with double precision.

Option 2.) is much easier to implement, so I will likely go with that.  Having a native block ordered accessor function is clearly desirable though since it will simplify the block orthogonalization routines, and allow us to write them using accessors, which is needed for offloading these to the GPU.

alexstrel:
Option 2 looks more complicated to extend for staggered, while option 1 is more generic

maddyscientist:
I'm only suggesting this for the packed V field, where each space-time point has Ns*Nc*Nvec colors.  Even when Ns=1, as for staggered, Nvec will be >=2, so it shouldn't be an issue for staggered either.  The only constraint would be that Nvec must be even.

alexstrel:
ok, agree. even number Nvec for staggered is not a restriction (one more eigen/null vector makes no significant difference)

maddyscientist:
Closed with #488

PR #488, opened 2016-08-17
Significantly improved restrictor performance

maddyscientist:
This pull request addresses #391, and essentially fixes this problem.  To do so, I ended up taking an approach that is not described in #391, hence I will described it here.  The problem is that the block width, in terms of bytes, can be less than a cache line width (32 bytes).  This means that we end up with strided memory access patterns, and memory bandwidth is wasted bringing in unused partial cache lines.  To fix this we could have augmented each thread block to work on multiple MG blocks, such that each thread block requests an entire cache line line, but this complicates using cub reductions.  Instead, a simpler approach is to ensure that the thread block for the adjacent MG block is scheduled near by, e.g., on the same SM, so that it can share the cache lines between these blocks.

The CUDA scheduler, by default, will distribute thread blocks in round-robin fashion around the GPU, meaning that adjacent thread blocks, e.g., `blockIdx.x` and `blockIdx.x + 1` are run on different SMs.  To bring them closer together we need to swizzle how `blockIdx.x` maps to a given coarse grid point.

The default mapping is
```c++
   int x_coarse = blockIdx.x;
``` 
What we desire instead is to do a transposition of this mapping
```c++
  int swizzle = Nsm; // swizzle factor here is the number of SMs
  int i = blockIdx.x % swizzle;
  int j = blockIdx.x / swizzle;

  int x_coarse = i * (gridDim.x / swizzle) + j;
```
where the width of the "matrix" we are transposing is the number of SMs.  Trivially, we see that `swizzle=1` is the original mapping.

The only complication here is if the total number of blocks we are scheduling (`gridDim.x`), is not a multiple of the number of SMs then we will end up with nonsense.  In this case we have to deal with the remainder explicitly, for this we just use the original mapping.

Since this transposition can be done for an arbitrary swizzling factor, I've left the swizzle factor as an autotuning parameter, to ensure that the thread block schedule to maximize cache hit rate can be found regardless of GPU.  The autotuning will search in the space [1,2*Nsm] for the optimum point.

An example of before and after performance for this change (and some minor thread coarseness tweaks that give minor benefit):
* M6000, FP32, V=16^3x64, Nvec=24, B = (4,4,4,4)
* Before: 101 GB/s
* After: 238 GB/s
* For comparison the prolongator achieves 278 GB/s

The reason that performance isn't quite at the level of the prolongator is due to the fact that the input vector field is being read by multiple thread blocks since coarse color is split between threads blocks.  The fix for this is to keep the split between threads (else threads become too coarse), but to enable different coarse colors within the thread block (z dimension) and not just between blocks.  This is not enabled at the moment due to difficulty with row reductions with cub.  This can be revisited if needed in the future.

### DIFF
--- a/lib/restrictor.cu
+++ b/lib/restrictor.cu
@@ -5,6 +5,9 @@
 #include <typeinfo>
 #include <multigrid_helper.cuh>
 
+// enabling CTA swizzling improves spatial locality of MG blocks reducing cache line wastage
+#define SWIZZLE
+
 namespace quda {
 
 #ifdef GPU_MULTIGRID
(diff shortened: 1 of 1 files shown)


AlexVaq:
Nice piece of work, and quite non-trivial!

mathiaswagner:
Looks good. Curious: Does the auto tuner actually pick the number of SMs for swizzle ?

maddyscientist:
Sometimes it does, sometimes it doesn't, it depends on the volume (fine or coarse grid) and the block size.  It seems to never pick `swizzle=1` which is what was done before.  I think it varies because sometimes a non-trivial swizzle will lead to L2 cache locality which can be better than L1/TEX cache locality depending on the size of the thread block occupying the SM.

PR #493, opened 2016-09-01
Hotfix/numanvml

### LINKED ISSUE #223: numa_affinity broken ?
I noticed getting the message:

```Failed to determine NUMA affinity for device 0 (possibly not applicable)```

quite often. When looking in the code it looks for something like

```"/proc/driver/nvidia/gpus/%d/information", my_gpu```

with my_gpu being the device id (i.e. an integer).

I checked on some machines and see
```
[mwagner@gdev1 ~]$ ls /proc/driver/nvidia/gpus/
0000:02:00.0  0000:03:00.0  0000:83:00.0  0000:84:00.0
...
[mwagner@cream ~]$  ls /proc/driver/nvidia/gpus/
0000:02:00.0  0000:03:00.0  0000:83:00.0  0000:84:00.0
```

It looks like the directory names have changed from device-ids to Bus-Ids with the newer drivers. I don't know when that happened but we should update the numa_affinity.cpp to reflect this.

mathiaswagner:
This pull requests enables the use of NVML to set numa affinity again and removes the old and broken numa affinity code.
Note that this is only implement for the cmake build.

For more complex topologies a better solution is still required. Closes #223.

We should probably just give up autoconf anyway.


maddyscientist:
Have you checked that this works as it should on multi-GPU systems?

mathiaswagner:
Not fully. The nvml call runs fine and as far as I saw it sets the mask to the one reported by nvidia-smi but I need to run some more tests on different systems.

maddyscientist:
Also, any reason not to kill the old code?  Since it doesn't work on current drivers any reason to keeping it in there?

mathiaswagner:
No, I'll need to remove that.
I checked on a multi gpu system and it also there correct. It does not set the affinity to a single core but to a range. I think if we need finer affinity settings that is something that should be controlled by mpi / queuing ...
This here is only to provide some basic affinity setting. We certainly need more topology awareness for fat nodes.

maddyscientist:
Ok, I'll wait until you've remove the old code before merging this.  Perhaps we need a comment, or warning of some kind to note that the multi-GPU aspect of this only works from a certain driver revision, e.g., older drivers alias GPUs on the same NUMA domain to the same core.

maddyscientist:
Regarding configure support for this: I believe with CUDA 8.0, nvml.h is included with the toolkit so there would be no issue finding this file.

maddyscientist:
On second thoughts, I agree, no need to attempt to maintain NUMA NVML functionality with autoconf.  It's a waste of development time.  This is good as it, merging.

Thanks Mathias for getting this done.

PR #495, opened 2016-09-05
Hotfix/multigrid

maddyscientist:
This is mainly  cleanup and bug fix pull, and fills in some gaps
* Include coarse Dslash in the asynchronous profile by adding a dummy policy tuner around the ghost exchange / kernel application
* Set `GaugeField::nFace` parameter to prevent potential silent errors when copying GPU gauge field to CPU in coarse-link creation
* Do not allocate a ghost zone for the coarse clover fields (saves memory)
* Adds ability to backup / restore cpuGaugeField instances (necessary to add tunecache support for CPU routines)
* Include most time consuming CPU routines in the tunecache and profile for getting a full trace of MG setup (coarse-link creation, block orthogonalization, etc.)
* Added the ability to write out different profiles, and flush the profile, `flushProfile()`, to provide multiple profiles from one run, allowing for directed profiling.  Profiles have an ordinal in their filename indicating the order in which they have been created.
* When running MG, we save and flush the profile after the setup is complete to get a separate profile for the setup and solver.

### DIFF
--- a/include/gauge_field.h
+++ b/include/gauge_field.h
@@ -262,6 +262,16 @@ namespace quda {
      * @param[in] src Source from which we are copying
      */
     virtual void copy(const GaugeField &src) = 0;
+
+    /**
+       @brief Backs up the cpuGaugeField
+    */
+    virtual void backup() const = 0;
+
+    /**
+       @brief Restores the cpuGaugeField
+    */
+    virtual void restore() = 0;
   };
 
   class cudaGaugeField : public GaugeField {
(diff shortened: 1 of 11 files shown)


mathiaswagner:
The count feature for the profile name is to write several profiles within one application run? Just curious what's the use case for this?

maddyscientist:
Ah yes, I forgot to mention that.  Will augment description above.

mathiaswagner:
Got it.

PR #496, opened 2016-09-07
add CUDA_INCLUDE_DIRS to search path for nvml.h

mathiaswagner:
Improves FindNVML automatic detection by adding CUDA_INCLUDE_DIRS to the search path for the nvml.h header


maddyscientist:
Thanks for the quick fix 😄

PR #498, opened 2016-09-07
Create new "pinned" memory allocations on the device

maddyscientist:
Due to the fact that QDJIT/PTX uses a pool allocator redirecting all calls to `cudaMalloc`, yet the CUDA IPC peer-to-peer interface requires base pointers to be exchanged, means that peer-to-peer is broken when used with Chroma and QDPJIT/PTX.  E.g., in general we will never know the base pointer.  This patch creates new allocator and free functions, `device_pinned_malloc` and `device_pinned_free` that uses the CUDA driver API memory functions that are not redirected by QDPJIT to allow QUDA to manage directly memory allocations on the device.  This is used when creating the ghost field for a given quark field.  The overhead of this should be tiny, since this is only deployed for halo regions not entire fields, and in turn halos are only allocated for fields that have a dslash applied to them.

### DIFF
--- a/include/malloc_quda.h
+++ b/include/malloc_quda.h
@@ -13,19 +13,23 @@ namespace quda {
    * macros below instead.
    */
   void *device_malloc_(const char *func, const char *file, int line, size_t size);
+  void *device_pinned_malloc_(const char *func, const char *file, int line, size_t size);
   void *safe_malloc_(const char *func, const char *file, int line, size_t size);
   void *pinned_malloc_(const char *func, const char *file, int line, size_t size);
   void *mapped_malloc_(const char *func, const char *file, int line, size_t size);
   void device_free_(const char *func, const char *file, int line, void *ptr);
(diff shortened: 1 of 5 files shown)


maddyscientist:
@bjoo Can you confirm this fixes the QDJIT + p2p issue?

maddyscientist:
One consequence of this patch is that it will require the CUDA library to be linked into all executables that use QUDA (`-lcuda`), and not just the runtime library (`-lcudart`), as it presently the case.  Not a big deal, but it will likely break a few apps that will need (trivially) fixed.  This won't be a problem for Chroma + QDPJIT since it already use the driver API.

maddyscientist:
Also, adding a heads up to @fwinter about this issue.

bjoo:
Hi Kate! Thank you very much. I managed to run with the K80s in P2P mode. Please feel free to merge these into develop.

PR #499, opened 2016-09-08
Add multiblas functionality to QUDA

mathiaswagner:
This is an intermediate pull request on the way to get block solver into quda.
The block solver functionality is basically there but still needs cleanup and optimization.
Please ignore any parts of that as long as they don't break other stuff.

This pull request is there to make the multiples functionality available in develop.
Basically it allows to perform blas operations like

y_j = sum_i a_ij x_i + y_j 

with up to 4 input and 2 output vectors.

This pull also adds
* A `warp_reduce` helper function for doing global reductions without needing to know the block size
* Optimizes the pre-existing block reductions
* Significant clean up of the blas generator to minimize boilerplate code using C++ traits
* Use the newly developed block-solver blas technology to optimize the GCR solver and use pipelined GCR for coarse-grid multigrid solves


mathiaswagner:
Please don't delete the block solver branch. I can reuse that.

mathiaswagner:
Looks like we have some Fermi build issues related to use of shuffle in `include/generics/shfl.h`.
Just for the isolation of the issue

lib/multi_blas_core.cuh
```diff
@@ -242,8 +243,8 @@ void multiblasCuda(const Complex *a, const double2 &b, const double2 &c,
       ypp.push_back(&yp[i]); wpp.push_back(&wp[i]);
     }
 
-    multiblasCuda<NXZ, NYW, RegType, StoreType, yType, M, Functor, writeX, writeY, writeZ, writeW>
-       (a, b, c, xpp, ypp, zpp, wpp, length);
+    multiblasCuda<NXZ, RegType, StoreType, yType, M, Functor, writeX, writeY, writeZ, writeW>
+      (a, b, c, xpp, ypp, zpp, wpp, NYW, length);
```
mathiaswagner: This breaks compilation with clang: 
`error: variable length array of non-POD element type 'Spinor<float4, short4, 6, 1, 1>`

maddyscientist: I guess it's time to add CLANG to the Jenkins tests 😉

PR #501, opened 2016-09-08
add 'QUDA_ENABLE_NUMA' environment variable

mathiaswagner:
to  allow for easily disabling QUDA's internal numa affinity setting.

If that is ok remind me to add that to the wiki page with ENV variables.


maddyscientist:
I've updated the wiki page https://github.com/lattice/quda/wiki/QUDA-Environment-Variables

PR #509, opened 2016-09-23
Feature/multi shift cleanup

maddyscientist:
This pull request is minor feature and cleanup of the blas and multi-shift solver
* Reduced allocation / free overheads in CG and multi-shift CG, especially apparent for peer-to-peer systems
* Reduced initalization overhead of multi-blas kernels: vector coefficients can be put in the parameter struct instead of explicit copy to `__constant__`
* Mixed-precision support for multi axpyBzpcx kernel and now supports arbitrary length vectors of fields using recursion
* Slight improvement to mixed-precision stability for CG (reprojection of direction vectors should be done using a complex dot product)
* Blas kernels now called directly on full fields instead of recursing to their parity subsets
* Work around for bug in multi-GPU clover derivative on CUDA 7.5

### DIFF
--- a/include/register_traits.h
+++ b/include/register_traits.h
@@ -83,6 +83,10 @@ namespace quda {
   template<> struct scalar<float3> { typedef float type; };
   template<> struct scalar<float2> { typedef float type; };
   template<> struct scalar<float> { typedef float type; };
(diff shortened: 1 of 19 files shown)


mathiaswagner:
Took me a while to come back to this but looks good.

PR #514, opened 2016-09-29
Improve cmake build 

mathiaswagner:
This adds some small improvements to the cmake build:

* make `QUDA_GPU_ARCH` and `CMAKE_BUILD_TYPE` dropdown lists (in ccmake and cmake-gui)
* regenerate `.mod` files when FORTAN INTERFACE is build
* adds `QUDA_BUILD_ALL_TESTS` advanced options to turn exclude tests from default build


cpviolator:
@weinbe2 The culprit is right here!

PR #518, opened 2016-10-04
Fix QUDA_BUILD_ALL_TESTS functionality in cmake build

### LINKED ISSUE #517: Test executables appear to not build
Hey all,

After grabbing a recent version of quda from develop, I've noticed that none of the test executables build after an appropriate ccmake/make -j 16. I didn't have this issue about a week ago (maybe  a3d95c8 is the culprit? not familiar enough with cmake to know). 'libquda.a' still builds and appears in the /lib directory (though I currently have no executable to link against it to attempt to verify it). 

The build is out-of-source. The only QUDA build flags I have 'ON' are QUDA_DIRAC_WILSON, QUDA_INTERFACE_QDP (which has sufficed to get 'invert_test' to automatically build before). The only advanced options I've dug into are changing the default c++ compiler to g++ (from literally c++). I noticed that there's now a new advanced option QUDA_BUILD_ALL_TESTS (which I've left set to the default 'ON'). 

Has anyone else hit this issue?

mathiaswagner:
This fixes the `QUDA_BUILD_ALL_TESTS` introduced in #514 and closes #517.


weinbe2:
Tested both QUDA_BUILD_ALL_TESTS = YES and NO on sm_52.

- YES automatically built all tests.
- NO didn't build tests, but had no issue explicitly running (for ex) 'make dslash_test' to build the test. 'make clean' worked as expected.

This test wasn't exhaustive (in the sense I used a config with only DSLASH_WILSON and INTERFACE_QDP enabled, single node), but based on the minimal changes made to the cmake config I don't think that should make a difference.

@maddyscientist, can I merge this/do I have permissions to merge?

maddyscientist:
@weinbe2 I don't think you have permission to merge, regardless, it looks like this pull has build issues on sm_20, so it can't be merged yet anyway.

maddyscientist:
@mathiaswagner I've retested the CUDA 8.0 / sm_20 build, and it works fine (with clover enabled since it failed on clover_deriv_quda.o).  Any idea what's wrong with this build?  This isn't the first Jenkins build to appear to fail, for reasons I don't understand,

mathiaswagner:
@weinbe2 I think you have rights to merge.
@maddyscientist Not sure what is wrong but I think it is a Jenkins issue. No time to look into this now but I don't think that fixing the typo (i.e. this pull) caused it to happen and we should not hold back merging this pull because of the build issue.

Issue #506, opened 2016-09-20
Implement BiCGStab(l) solver


maddyscientist:
This will provide a faster baseline Wilson solver, and potentially also help with null-space generation for MG.

weinbe2:
Created branch feature/bicgstabl as a starting point.

maddyscientist:
Closed with pull #519

PR #519, opened 2016-10-05
Feature/bicgstabl

weinbe2:
This pull request addresses #506, adding support for BiCGstab-L for arbitrary L. The appropriate modifications have been made to invert_quda.h and the solver factory in solver.cpp to create the BiCGstabl inverter class, and enums have been put in place. I extended invert_test.cpp (and its appropriate dependencies) to test bicgstab-L for a few values of L. 

On an implementation level, I reuse the parameter "nGCRkrylov", used to specify N in GCR(N), to define L for BiCGstab-L. I exposed nGCRkrylov as a command-line argument for the test programs appropriately.

I've only tested the code for a single GPU build against the Wilson operator. Convergence was consistent between BiCGstab and BiCGstag-L with L=1, higher values of L also converged. One hope out of implementing BiCGstab-L is that higher L will lead to smoother (and ideally faster) convergence for ill-conditioned matrices.


mathiaswagner:

mathiaswagner:

maddyscientist:
@weinbe2 This is a good first pull 😄  Hopefully first of many.

Let me do some testing and report back.

maddyscientist:
Ok, there's two issues I can see at the moment:
* While the solver seems to converge, and the CPU number matches the iterated residual, the reported QUDA true residual looks not to have been set.  I suspect you're just not setting this at the end of the solver (check out the epilogue at the end of the standard BiCGStab solver)
```
Residuals: (L2 relative) tol 1e-07, QUDA = 9.3804e-05, host = 9.16105e-08; (heavy-quark) tol 0, QUDA = 0
```
* There's a memory leak in the code, that seems to scale linearly with the size of "L"
```

WARNING: The following internal memory allocations were not freed.

Type    Pointer          Size             Location
----------------------------------------------------------
Device     0x13107c0000         12582912  create(), /home/kate/github/quda-develop/lib/cuda_color_spinor_field.cu:173
Device     0x13113c0000         12582912  create(), /home/kate/github/quda-develop/lib/cuda_color_spinor_field.cu:173
Device     0x1311fc0000         12582912  create(), /home/kate/github/quda-develop/lib/cuda_color_spinor_field.cu:173
Device     0x1312bc0000         12582912  create(), /home/kate/github/quda-develop/lib/cuda_color_spinor_field.cu:173
Device     0x13137c0000         12582912  create(), /home/kate/github/quda-develop/lib/cuda_color_spinor_field.cu:173
Device     0x13143c0000         12582912  create(), /home/kate/github/quda-develop/lib/cuda_color_spinor_field.cu:173
Host         0x13310370             2064  operator new(), /home/kate/github/quda-develop/include/object.h:23
Host         0x13311870             2064  operator new(), /home/kate/github/quda-develop/include/object.h:23
Host         0x13312d70             2064  operator new(), /home/kate/github/quda-develop/include/object.h:23
Host         0x13314270             2064  operator new(), /home/kate/github/quda-develop/include/object.h:23
Host         0x13315770             2064  operator new(), /home/kate/github/quda-develop/include/object.h:23
Host         0x13316c70             2064  operator new(), /home/kate/github/quda-develop/include/object.h:23
```

weinbe2:
@maddyscientist I've fixed the two issues---both were trivial fixes, I really needed to look more closely at and understand the output of invert_test... lesson learned.

I've noticed that the true residual does deviate from the iterated residual... that's something I've never noticed before, so I'm going to hunt into that a bit.

maddyscientist:
In double precision, I'm seeing a very match between iterated and true residual, is this is just something in single precision? (which is probably expected)

weinbe2:
Oh, yes, you're right. The "issue" goes away if I set "--prec double", while it's there if I explicitly set "--prec single". 

I noticed that the implementation of BiCGstab doesn't suffer this issue---is this a consequence of reliable updates (which I don't have in BiCGstab-L [yet])?

maddyscientist:
Yes, reliable updates cures this.

maddyscientist:
On the tests I've done so far (with aniso WIlson 2 flavour lattices), at small quark mass BiCGStab(4) and (8) reduces the iteration count by a small (~10%) amount versus regular BiCGStab.

When running BiCGStab(32), the divergence between iterated and true residuals is significant even in double precision, indicating reliable updates are really needed for large L.

maddyscientist:
Awesome.  How well does this work?  Can you run with large Nkrylov size with good stability?

weinbe2:
It ran and converged, didn't have any issues with single precision. I
didn't have a good configuration to test a large Nkrylov size with though
(just the default random). Do you have a good config? I tried to tweak the
mass from the command line on invert_test to tank the condition number, it
didn't seem to make a difference...

maddyscientist:
Compile with QIO/QMP enabled (you'll find it installed on /usr/local/scidac), and use the 16^3x64 and 24^3x64 lattices in ~kate/lattices.  Parameters are `--anisotropy 2.38 --mass -0.42` for both to run at around critical mass.

weinbe2:
Thanks for the info! I compiled with QIO/QMP and ran tests in both single
and double precision for both the 16^3 and 24^3 configs.

Overall remarks:
* In terms of number of iterations, L = 4 to 8 seems to be optimal.
BiCGstab as-implemented wins in general in terms of time-to-solution
(except for the 24^3, single precision, where using BiCGstab-L, L >= 4, was
necessary to establish good convergence!).
* This is for a few reasons I wager: (i) BiCGstab is better optimized
(fused operations, etc), while BiCGstab is just standard BLAS, so the
reduced iteration count isn't enough to win in time-to-solution, and (ii)
BiCGstab-L includes a modified Gram-Schmidt which scales as L^2 (best noted
on the double precision, 16^3 results, where BiCGstab-6 through -16 had the
same number of iterations, but the time-to-solution increases).
* BiCGstab-L starts to break down for L >~ 8 in single precision, L >~ 24
in double precision. This is unsurprising, as the Gram-Schmidt in
BiCGstab-L is an explicit Gram-Schmidt on the Krylov space, which is known
to be numerically unstable.

Anyway, to numbers!

invertQuda compute time
Quadro M6000, pre-tuned.

16^3 x 64, --anisotropy 2.38 --mass -0.42, Wilson to rel resid 1e-7.
Single precision: --prec single
BiCGstab:      1.437 sec, 512 iter
BiCGstab-1:    1.722 sec, 532 iter
BiCGstab-2:    1.758 sec, 504 iter
BiCGstab-4:    1.999 sec, 496 iter
BiCGstab-6:    2.372 sec, 516 iter
BiCGstab-8:    2.720 sec, 528 iter
BiCGstab-12:   3.553 sec, 564 iter
BiCGstab-16:   5.936 sec, 800 iter
BiCGstab-24:   13.98 sec, 1440 iter
BiCGstab-32:   24.95 sec, 2080 iter

Remarks: The stability of BiCGstab-L's internal Gram-Schmidt and Cholesky
starts to break down around L = 8 or 12.

Double precision: --prec double
BiCGstab:      3.089 sec, 501 iter
BiCGstab-1:    3.446 sec, 498 iter
BiCGstab-2:    3.631 sec, 490 iter
BiCGstab-4:    4.151 sec, 492 iter
BiCGstab-6:    4.536 sec, 480 iter
BiCGstab-8:    5.047 sec, 480 iter
BiCGstab-12:   6.063 sec, 480 iter
BiCGstab-16:   7.070 sec, 480 iter
BiCGstab-24:   9.533 sec, 504 iter
BiCGstab-32:   15.55 sec, 672 iter
BiCGstab-64:   58.85 sec, 1472 iter

Remarks:
* BiCGstab-L does a modified Gram-Schmidt with L vectors, which contributes
a quadratic scaling. This is why BiCGstab-6 through -16 have the same
iteration count but the timing grows. (Maybe some magic in BlockCG could
help here? This would require "re-deriving" BiCGstab-L to see the
Gram-Schmidt and the Cholesky decomp that goes on more explicitly.)
* The stability of BiCGstab-L's internal Gram-Schmidt and Cholesky starts
to break down around L = 24.

24^3 x 64, --anisotropy 2.38 --mass -0.42, Wilson to rel resid 1e-7.
Single precision: --prec single
BiCGstab:      38.33 sec, 4111 iter
BiCGstab-1:    did not converge (maxed at 10000 iter)
BiCGstab-2:    26.78 sec, 2310 iter
BiCGstab-4:    39.77 sec, 2972 iter (stalled around 1e-6 for a long time,
maybe single prec artifact?)
BiCGstab-6:    24.15 sec, 1590 iter
BiCGstab-8:    25.49 sec, 1496 iter
BiCGstab-12:   37.12 sec, 1788 iter
BiCGstab-16:   89.72 sec, 3664 iter
BiCGstab-24:   179.3 sec, 5616 iter

Remarks:
* I think this set of numbers is the best example that L > 1 can stabilize
the solve in terms of iterations.
* The stability of BiCGstab-L's internal Gram-Schmidt and Cholesky starts
to break down around L = 12.

Double precision: --prec double
BiCGstab:      35.59 sec, 1746 iter
BiCGstab-1:    38.60 sec, 1684 iter
BiCGstab-2:    38.65 sec, 1578 iter
BiCGstab-4:    39.99 sec, 1432 iter
BiCGstab-6:    44.28 sec, 1410 iter
BiCGstab-8:    49.10 sec, 1408 iter
BiCGstab-12:   59.30 sec, 1416 iter
BiCGstab-16:   68.86 sec, 1408 iter
BiCGstab-24:   90.71 sec, 1440 iter
BiCGstab-32:   254.3 sec, 3296 iter
BiCGstab-64:   656.2 sec, 4928 iter

Remarks :
* BiCGstab-L does a modified Gram-Schmidt with L vectors, which contributes
a quadratic scaling. This is why BiCGstab-6 through -16 have the same
iteration count but the timing grows. (Maybe some magic in BlockCG could
help here? This would require "re-deriving" BiCGstab-L to see the
Gram-Schmidt and the Cholesky decomp that goes on more explicitly.)
* The stability of BiCGstab-L's internal Gram-Schmidt and Cholesky starts
to break down around L = 24.

maddyscientist:
Thanks Evan.  Looks like it has a larger improvement for larger volumes, so this is good to see.

weinbe2:
MG remark: null vector generation with BiCGstab worked just fine without reliable updates, however, BiCGstab-L with L >= 4 needed reliable updates (I used delta = 0.1) to get meaningful (i.e., actually useful) null vectors.

mathiaswagner:
@weinbe2 I have not looked at the code but using multiblas / multi reduce should help to get a orthogonalization run in close to linear time (cache reuse).

Let me know if you think that would help bicgstabl.

maddyscientist:
@mathiaswagner it has already been multi-blas-ed to death

cpviolator:
@weinbe2 Do you have any twisted mass configs you can try it out on? We're having a bit of trouble with BiCGStab convergence when using multigrid with `--dslash-type twisted-mass --solve-type direct-pc` and any help from BiCGStab(l) would be great. @maddyscientist mentioned that it may perform better with the low complex eigenvalues.

maddyscientist:
@cpviolator, this is one of reasons we've implemented BiCGStab(l)!  I've been wanting to test this algorithm with twisted mass for a long time.

If I give an account on my workstation, can you scp some configs to me for testing both as a standard solver and as part of null generation?

kostrzewa:
@maddyscientist @cpviolator 
I can also provide configs if @cpviolator has access problems. From our coarse physical pion mass ensemble cA2.09.48, I recommend conf.0500, conf.1000, conf.1500, conf.2000 and conf.2500 as these are the configurations that were used to test DDalphaAMG. Configurations on a 64^3 lattice with a slightly re-tuned kappa parameter are also available.

mathiaswagner:
@Jenkins ok to test

weinbe2:
@mathiaswagner There are a few optimizations @maddyscientist and I are still testing, should be done by the end of the day, as is the code will break in a few corner cases.

mathiaswagner:
My above comment was direct to the Jenkins build server to allow the it to automatically build this pr.

weinbe2:
Gotcha. Today I learned!

weinbe2:
Thank you for the suggestion---in fact, your suggestion is a lot of what I
spent my time on during the hackathon with Kate's supervision, including
writing some new blas and multiblas routines!

As a remark (I'm working on some notes on this), we thought about it at the
hackathon, the multiblas as is doesn't bring the orthogonalization to
linear time, but it does significantly help reduce the constant in front of
the quadratic term. The main way to see it is to look at Gram Schmidt:

for i = 1 to nvec
   For j = 1 to j-1
      Ortho V(i) against V(j)
   End for
End for

Which becomes

for i = 1 to nvec
   Block ortho V(i) against all V(j)
End for

Even though you avoid a memory traffic quadratic term from loading V(i)
each time on the inner for loop, you still have to load each V(j), so the
quadratic term is still there.

Now, you could probably kill that by doing something similar to what you
have in block cg, doing an outer product of inner products fused with all
the orthogonalization subtractions, but I also explicitly need some of the
byproducts of the Gram Schmidt later in the algorithm, so I took it as too
much of a headache/compile time killer for now to figure it out ;)

weinbe2:
@maddyscientist I ran into trouble merging your commit where you split blas/reduce and multi-blas/reduce into separate files. Multireduce cDot seems to be the current culprit, can't figure out exactly why. Can you please take a look? It compiled just fine, I've triggered the issue on Maxwell with the command:

`./invert_test --device 0 --load-gauge ~kate/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --sdim 16 --tdim 64 --anisotropy 2.38 --mass -0.42 --niter 10000 --inv-type bicgstab-l --ngcrkrylov 3 --pipeline 3 --prec double --prec-sloppy double --tol 1e-12`

maddyscientist:
@weinbe2 Hang is fixed with d0601e03bce1a407dd0387cd65077447f8d0b735.  In 5e794222edfa13e07336a2dd2a6cc68ad5eb88f9 I've fixed the Fermi compilation, and renamed the files to be consistent with other blas files, and to allow for easy addition of mixed precision support in multi-reductions if needed.

weinbe2:
@maddyscientist Thank you!

maddyscientist:
@weinbe2 I've fixed a couple of issues
* Add support for coarse dslash with worker class
* Fix bug where worker was not reset to `NULL` after `mat` being called which would break subsequent calls to dslash outside the scope of the solver

While I've confirmed multigrid with BiCGStab(l) works ok, I've left the default MG null-space generation with regular BiCGStab.  I will expose the solver type for this as an interface option in my feature/multigrid branch (where I'm working offloading the full setup to the GPU).

@mathiaswagner this is now ready for merging.  Can you review and merge please?

lib/CMakeLists.txt
```diff
@@ -62,7 +61,6 @@ LIST(REMOVE_ITEM QUDA_OBJS ${QUDA_CU_OBJS})
 
 if(BUILD_FORTRAN_INTERFACE)
   LIST(APPEND QUDA_OBJS quda_fortran.F90)
-  set_source_files_properties(quda_fortran.F90 PROPERTIES OBJECT_OUTPUTS ${CMAKE_CURRENT_BINARY_DIR}/quda_fortran.mod)
```
mathiaswagner: Was this removed intentionally?

maddyscientist: Good catch. I guess @weinbe2 this was an accident in 7055588d7c920fbdd4bbb752ff032c94d29d1dae?

I'll restore it.

PR #525, opened 2016-11-21
remove spurious check for compute mode in peer2peer initialization

mathiaswagner:
I left the check for UVA for now. Guess we can remove that as well but did not check in detail.


maddyscientist:
Mering.  We should probably keep the UVA test in there as that will be needed for SHMEM-type deployment as a future optimization.

PR #536, opened 2016-12-12
Memory-pool allocator and MILC interface improvements

maddyscientist:
* Moved the pinned-memory pool allocator in `LatticeField` to separate `pool` namespace, and created macro defined interface similar to pre-existing memory allocation interface, e.g.
```c++
#define pool_pinned_malloc(size) quda::pool::pinned_malloc_(__func__, __FILE__, __LINE__, size)
```
* Created a device-memory pool allocator in that mimics the above pinned-memory pool allocator
* All device-memory allocations used by lattice fields now use this memory-pool to reuse memory allocations: this essentially reduces device memory allocation over heads to zero
* Exposed both the device and pinned memory pool allocator in the MILC interface
```c++
void* qudaAllocatePinned(size_t bytes);
void qudaFreePinned(void *ptr);
```
 to allow MILC to allocate pinned memory for fields to be copied between MILC and the GPU
* Memory pool allocators can be disabled with environment variables `QUDA_ENABLE_DEVICE_MEMORY_POOL=0` and `QUDA_ENABLE_PINNED_MEMORY_POOL=0`
*  Reconstruct type used by HISQ solver can now be set by an environment variable `QUDA_MILC_HISQ_RECONSTRUCT=9/13/18`
* All non-native fields are not aligned on texture boundaries (this fixes a bug for references to external fields for certain problem sizes)
* Fixed unitialized variables that could led to unstable behaviour 
* Removed legacy fat-link kernel code that did not use extended fields
* Fixes potential initialization hazard of auto tuning
* Simplifies enabling autotuning: now enabled by default, but can be disabled with the environment variable `QUDA_ENABLE_TUNING=0`


maddyscientist:
I believe this branch is ready for merging now @mathiaswagner 😄

lib/comm_mpi.cpp
```diff
@@ -144,10 +144,11 @@ void comm_peer2peer_init(const char* hostname_recv_buf)
 	  int canAccessPeer[2];
 	  cudaDeviceCanAccessPeer(&canAccessPeer[0], gpuid, neighbor_gpuid);
 	  cudaDeviceCanAccessPeer(&canAccessPeer[1], neighbor_gpuid, gpuid);
+
 	  if(canAccessPeer[0]*canAccessPeer[1]){
 	    peer2peer_enabled[dir][dim] = true;
 	    if (getVerbosity() > QUDA_SILENT)
-	      printf("Peer-to-peer enabled for rank %d gpu=%d with neighbor %d gpu=%d dir=%d, dim=%d\n",
+	      printf("Peer-to-peer enabled for rank %d (gpu=%d) with neighbor %d (gpu=%d) dir=%d, dim=%d\n",
```
mathiaswagner: Haven't checked yet but should this be replicated to comm_qmp ?

mathiaswagner:
We need to update the Wiki with the env variables.

maddyscientist:
Wiki page updated

PR #537, opened 2016-12-13
Hotfix/cmake libcuda

mathiaswagner:
This is a hotfix that replace linking against the driver api by the CUDA_CUDA_LIBRARY found by cmake.
It also includes a workaround to look for the stubs in cases where libcuda is not installed on the build system.

Requires testing on Titan.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -5,7 +5,9 @@ include_directories(.)
 set(QUDA_TEST_COMMON gtest-all.cc test_util.cpp misc.cpp)
 cuda_add_library(quda_test STATIC ${QUDA_TEST_COMMON})
 
-set(TEST_LIBS quda quda_test ${CMAKE_THREAD_LIBS_INIT} ${QUDA_LIBS} cuda)
+FIND_LIBRARY(CUDA cuda HINTS ${CUDA_TOOLKIT_ROOT_DIR}/lib/ ${CUDA_TOOLKIT_ROOT_DIR}/lib/stubs)
+
+set(TEST_LIBS quda quda_test ${CMAKE_THREAD_LIBS_INIT} ${QUDA_LIBS} ${CUDA})
 
 if(QUDA_QIO)
   LIST(APPEND TEST_LIBS ${QIO_LIB} ${LIME_LIB})
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -5,9 +5,10 @@ include_directories(.)
 set(QUDA_TEST_COMMON gtest-all.cc test_util.cpp misc.cpp)
 cuda_add_library(quda_test STATIC ${QUDA_TEST_COMMON})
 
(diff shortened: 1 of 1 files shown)


maddyscientist:
@evanberkowitz When you've confirmed this patch fixes Titan, can you record here please?

evanberkowitz:
I will.  My build with an explicit `-L/path/to/stubs` worked.  I'll give this a try and report back.

evanberkowitz:
I configured with
```
CC="/ccs/proj/lgt100/software/install/openmpi-2.0.0/bin/mpicc" CFLAGS="-fopenmp -g -O2 -std=c99" CXX="/ccs/proj/lgt100/software/install/openmpi-2.0.0/bin/mpicxx" CXXFLAGS="-fopenmp -g -O2 -std=c++11" LIBS="" QUDA_LIBS="" LDFLAGS="" cmake /ccs/home/berkowit/USQCD/titan/source/quda_hotfix -DQUDA_GPU_ARCH=sm_35 -DQUDA_MPI=ON -DQUDA_QMP=ON -DQUDA_QMPHOME=/ccs/home/berkowit/USQCD/titan/install/qmp -DMPI_C_COMPILER=/ccs/proj/lgt100/software/install/openmpi-2.0.0/bin/mpicc -DMPI_CXX_COMPILER=/ccs/proj/lgt100/software/install/openmpi-2.0.0/bin/mpicxx -DQUDA_DIRAC_WILSON=ON -DQUDA_DIRAC_DOMAIN_WALL=ON
```

which doesn't have any `-L` or `-l` explicit (and which was failing earlier!).  Compiled without a hitch, and ran the tests.  `deflation_test` failed because I didn't compile `MAGMA`.  For `dslash_test` and `staggered_dslash_test` I got some fails at 1e-06 and finer, but ultimately passed.  Ran `invert_test` and `staggered_invert_test` successfully.  For `su3_test` I had some fails at 1e-08 and finer. 

`pack_test` failed.  The entire output was

```
ERROR: Unsupported field order 6
 (rank 0, host nid00823, /ccs/home/berkowit/USQCD/titan/source/quda_hotfix/lib/color_spinor_util.cu:128 in genericSource())
```

But, the main point is that it compiled, so the `cmake` fix seems to have worked!

maddyscientist:
Excellent.  Closing and merging this pull.

(Test behaviour is as expected.   Most tests default to single precision, so errors at 1e-6/1e-7 expect.  "pack_test" is an old test that needs to be either removed or updated.)

Thanks!

evanberkowitz:
Now when I compile chroma it complains about now -lcuda.  Is there a way for QUDA to tell chroma to look for the stubs too?  Or this must be done by hand?

mathiaswagner:
Unfortunately we have not managed to generate something like a pkgconfig file or anything similar for QUDA to provide a way to communicate LDFlags to application. I think we have an issue open for that for quite some time.
So for now this needs to be fixed from the Chroma side.

evanberkowitz:
OK.  Adding `-L/path/to/stubs` to chroma's `LDFLAGS` upon configuration seems to have done the trick.

PR #540, opened 2016-12-19
Fixed a bug in the Stout smearing routines and optimised them

cpviolator:
using single prec math. Added an ErrorSU3 function that returns
the L2 norm of a given SU3. Added APE and Stout smearing tests
to su3_test with hardcoded parameters: the plaquette action is
calculated on APE and Stout smeared lattices and should be
equal up to small error.

### DIFF
--- a/lib/quda_matrix.h
+++ b/lib/quda_matrix.h
@@ -992,12 +992,13 @@ namespace quda {
       // Equation numbers in the paper are referenced by [eq_no].
 
       //Declarations
-      float inv3 = 1.0/3.0;
-      
-      float c0, c1, c0_max, Tr_re;
-      float f0_re, f0_im, f1_re, f1_im, f2_re, f2_im;
-      float theta;
-      float u_p, w_p;  //u, w parameters.
+      typedef decltype(Q(0,0).x) undMatType;
+
+      undMatType inv3 = 1.0/3.0;      
+      undMatType c0, c1, c0_max, Tr_re;
+      undMatType f0_re, f0_im, f1_re, f1_im, f2_re, f2_im;
+      undMatType theta;
+      undMatTypeu_p, w_p;  //u, w parameters.
       Matrix<T,3> temp1;
       Matrix<T,3> temp2;
       //[14] c0 = det(Q) = 1/3Tr(Q^3)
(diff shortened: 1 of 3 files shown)


mathiaswagner:

mathiaswagner:

maddyscientist:
@cpviolator thanks for this.  Good to see the STOUT smearing getting a shake out.  A couple of points:
* The error checking code is explicitly commented out which means it requires hand editing to enable it.  I presume it's only need for debugging, so how about guarding it with `ifdef HOST_DEBUG` (lines 195-225)?
* The `exponentiate_iQ` function is explicitly float as opposed to a templated type?  Won't this lose precision when doing double-precision smearing?
* For optimizing division routines, the compiler will likely produce faster code with this style:
```c++
denorm _inv = 1.0/denorm;
a = b* denorm_inv;
c = d* denorm_inv;
e = f* denorm_inv;
```

cpviolator:
* I wasn't sure about the error code, I'll use the `#ifdef HOST_DEBUG` flag for it.
* I'll change the precision in exp(iQ) to double. I was going for speed after it was clear that the routine worked, but the speed up is only ~5%, probably not worth the prec loss.
* I'll rewrite those divisions too

maddyscientist:
Why not just make exp(iQ) templated on the type (use `decltype` to get the underlying type of the Matrix)?

cpviolator:
Always steep learning curves in QUDA development! I think I did that decltype right.

cpviolator:
Also, those suggestions on division really helped, APE and stout now solve the same problem at the same time to within 1%, with plaquette difference <0.1% (4 K20m GPUs, L=24) Thank you!

maddyscientist:
One other thing on precision, is that both single and double both use sinf/cosf/sqrtf single precision functions.  You should instead use sin/cos/sqrt etc., since in C++ CUDA overloads these, such they will compile into their single precision equivalents when a single precision argument is passed (http://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html#group__CUDA__MATH).

lib/quda_matrix.h
```diff
@@ -1005,22 +1028,22 @@ namespace quda {
...
+      undMatType cos_w = cos(w_p);
+      undMatType sinc_w;
+      undMatType hj_re = 0.0;
+      undMatType hj_im = 0.0;
   
       //[33] Added one more term to the series given in the paper.
       if (w_p < 0.05 && w_p > -0.05) {      
 	//1 - 1/6 x^2 (1 - 1/20 x^2 (1 - 1/42 x^2(1 - 1/72*x^2)))
-	sinc_w = 1.0-1.0/6.0*w_sq*(1-0.05*w_sq*(1-1.0/42.0*w_sq*(1-1.0/72.0*w_sq)));
+	sinc_w = 1.0 - 1.0/6.0*w_sq*(1 - 0.05*w_sq*(1 - (1.0/42.0)*w_sq*(1 - (1.0/72.0)*w_sq)));
```
maddyscientist: While the order of evaluation may be well defined in C++, it would make it easier to visually parse if brackets were used to clearly delineate the division by 6.0 and subsequent multiplication by `w_sq`.

maddyscientist:
@mathiaswagner this all looks good, but Jenkins doesn't appear to have done any build testing.  I won't merge this until you've got Jenkins to test this.

cpviolator:
No, thank you! New `decltype` superpowers acquired 👍

mathiaswagner:
@Jenkins ok to test

lib/quda_matrix.h
```diff
@@ -958,45 +958,59 @@ namespace quda {
     return true;
   }
 
+  template<class Cmplx>
+  __device__ __host__
+    double ErrorSU3(const Matrix<Cmplx,3>& matrix)
+    {
+      const Matrix<Cmplx,3> identity_comp = conj(matrix)*matrix;
+      double error;
```
maddyscientist: Was just about to merge this, then I noticed a bug here: `error` hasn't been initialized so could lead to nonsense results....

PR #544, opened 2016-12-22
 removed staggered_oprod_new.o from Makefile since it broke non CMake

### LINKED ISSUE #543: Presence of staggered_oprod_new.o in lib/Makefile prevents QUDA building without CMake
I pulled QUDA this morning and it complained that it couldn't make staggered_oprod_new.o 
Indeed there was no .cu file there in 'develop' and this file didn't seem to appear in CMakeLists.txt.
I have removed this file from lib/Makefile and am doing a test build. Once that is finished, I will pull the fix and create a pull req.

bjoo:
Fixes #543


alexstrel:
this branch is ready for merge, if there are no objects.

Issue #548, opened 2017-01-02
HOSTDEBUG and DEBUG targets fail to compile due to quda_generated_gauge_stout.cu


kostrzewa:
I'm trying to invest some more time into figuring out why I see the issues with twisted mass and our tmLQCD interface: #527 #474 

Upon compilation of the latest commit of the develop branch (ccbf13f40b85c9fe23fac325de0d128295168805), I noticed that I cannot compile the DEBUG or HOSTDEBUG targets. Suspecting that it might be related to my using ```make.inc -> make``` rather than cmake, I finally got around to figuring out how to build QUDA using cmake on Jureca. However, this did not resolve the problem. The RELEASE and DEVEL targets build just fine. Sadly, building with cmake also doesn't magically resolve the aforementioned issues...

Could anyone please confirm if this is indeed currently broken?

I'm not quite sure why the stout routines are even being built, since I've disabled the gauge stuff. I guess that enabling clover 

```
lib/quda_matrix.h(971): error: no operator "-" matches these operands
            operand types are: const quda::complex<float> - double
          detected during:
            instantiation of "double quda::ErrorSU3(const quda::Matrix<Cmplx, 3> &) [with Cmplx=quda::complex<float>]" 
/homec/hbn28/hbn288/code/quda-develop-dynamic-clover/lib/gauge_stout.cu(217): here
            instantiation of "void quda::computeSTOUTStep(quda::GaugeSTOUTArg<Float, GaugeOr, GaugeDs>) [with Float=float, GaugeOr=quda::gauge::FloatNOrder<float, 18, 2, 18>, GaugeDs=quda::gauge::FloatNOrder<float, 18, 2, 18>]" 
/homec/hbn28/hbn288/code/quda-develop-dynamic-clover/lib/gauge_stout.cu(253): here
            instantiation of "void quda::GaugeSTOUT<Float, GaugeOr, GaugeDs>::apply(const cudaStream_t &) [with Float=float, GaugeOr=quda::gauge::FloatNOrder<float, 18, 2, 18>, GaugeDs=quda::gauge::FloatNOrder<float, 18, 2, 18>]" 
/homec/hbn28/hbn288/code/quda-develop-dynamic-clover/lib/gauge_stout.cu(247): here
[...]

PR #549, opened 2017-01-04
Make sure quda_fortran.mod is recreated

### LINKED ISSUE #545: CMake not robust when building Fortran interface
If CMake is used to build QUDA, if after being build, the lib/quda_fortran.mod file is manually deleted, when `make` is rerun this file is not rebuilt.  The only recourse I have found is to nuke the cache file and rebuild from scratch.

mathiaswagner:
This should fix #545.


maddyscientist:
Merging (debug build failure is expected due to #548).

PR #558, opened 2017-01-30
Add support for native CUDA support in cmake and create an install target

mathiaswagner:
This pull request adds

* native CUDA support in cmake (to come with cmake 3.8)
* compatible with current FindCUDA approach (used for cmake <= 3.7.x)
* first iteration of exported cmake targets to easily include quda in other cmake projects
* first iteration on make install

I am putting this pull request here to have the build run through Jenkins and to get some feedback.

@bjoo Not sure whether you looked into cmake for Chroma but the exported target should make it easier to add QUDA support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,8 @@ endif()
 #
 # PROJECT is QUDA
 #
-project("QUDA")
+project("QUDA" VERSION 0.9.0)
+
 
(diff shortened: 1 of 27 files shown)


mathiaswagner:
@robertmaynard FYI: I merged your changes into the feature/cmake_install branch of QUDA.

PR #570, opened 2017-02-27
Add GPU_ARCH to GITVERSION string

mathiaswagner:
Should now print something like `QUDA 0.9.0 (git v0.8.0-1570-g7a8a8ba-sm_52)`


mathiaswagner:
I am also using this PR to test updates to the Jenkins build.

PR #573, opened 2017-03-06
Remove difference between QUDA_TWIST_PLUS and MINUS, put it in the sign of mu

AlexVaq:
Replaced QUDA_TWIST_{PLUS/MINUS} with QUDA_TWIST_SINGLET. If we want a difference between the plus and the minus in the mu, we can just change the sign of mu (like in the rest of the libraries), and that should do.

### DIFF
--- a/include/enum_quda.h
+++ b/include/enum_quda.h
@@ -362,8 +362,7 @@ extern "C" {
   } QudaDWFPCType; 
 
   typedef enum QudaTwistFlavorType_s {
-    QUDA_TWIST_MINUS = -1,
-    QUDA_TWIST_PLUS = +1,
+    QUDA_TWIST_SINGLET = 1,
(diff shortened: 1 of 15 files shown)


cpviolator:
@AlexVaq @maddyscientist Awesome. Just to clarify, the flavour checks still present in, for example, `lib/dirac_twisted_clover.cpp`
```c++
if (in.TwistFlavor() != out.TwistFlavor()) 
     errorQuda("Twist flavors %d %d don't match", in.TwistFlavor(), out.TwistFlavor());
```
will always pass now as the member function `TwistFlavor()` will always be set to `QUDA_TWIST_SINGLET`. Is this correct?

AlexVaq:
No, you could try to mix the doublet with a single flavor, and you'd get an error.

Alex

PR #575, opened 2017-03-07
add QUDA_MAGMAHOME cmake variable

kostrzewa:
to allow manual configuration of MAGMA directories
if not set, fall back to pkg-config


cpviolator:
While you're doing this, can you change line 240 so that it reads as:
```cmake
message("${MAGMA_INCLUDEDIR}")
```
Without the quotes, CMake throws an error.

kostrzewa:
Ah, the quotes were the problem, cheers Dean!

mathiaswagner:
Looks good. I assume that this works for you so I am happy to merge it and whenever we need a better solution can go on from here.

CMakeLists.txt
```diff
@@ -230,12 +232,18 @@ endif()
...
-  find_library(MAGMA ${MAGMA_LIBRARIES} PATH ${MAGMA_LIBRARY_DIRS})
+  if("${QUDA_MAGMAHOME}" STREQUAL "")
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(MAGMA  magma)
+    include_directories(SYSTEM ${MAGMA_INCLUDEDIR})
+    message("${MAGMA_INCLUDEDIR}")
+    find_library(MAGMA ${MAGMA_LIBRARIES} PATH ${MAGMA_LIBRARY_DIRS})
+  else()
+    # prefer static library
+    FIND_LIBRARY(MAGMA_LIB libmagma.a magma ${QUDA_MAGMAHOME}/lib/)
```
kostrzewa: @maddyscientist @mathiaswagner 
Sorry, I messed up here, this should be ```MAGMA``` and not ```MAGMA_LIB```. Will submit hotfix.

PR #578, opened 2017-03-08
add 'one-size-fits-all' doxygen input file

kostrzewa:
This may be useful to others, please comment if you (dis-)like the fact that it creates another directory...


mathiaswagner:
Eventually would be great to get that into cmake somehow ...

mathiaswagner:
https://vicrucann.github.io/tutorials/quick-cmake-doxygen/ has some good hints for cmake integration.

There is also https://cmake.org/cmake/help/v3.0/module/FindDoxygen.html.

kostrzewa:
okay, will do, so you'd like the docs to be created in ```build``` ?

mathiaswagner:
Guess we can leave that up for discussion. It was my first reaction to always keep the build dir clean but I haven't given it any more thought yet.

rbabich:
For comparison, the doxygen input file that we've traditionally used is here: https://github.com/lattice/quda-ref/tree/master

mathiaswagner:
It's been a long time but I still like the doxygen idea. @kostrzewa, @maddyscientist  Where do we stand on this? Do we want better cmake integration here?

maddyscientist:
Yes, we want good cmake integration with doxygen is very relevant would be good.  Possible for the 0.9.0 release you think?

mathiaswagner:
Pulling this into a branch of the github repository. I will add a follow up pull request for merge into develop.

PR #589, opened 2017-03-17
Hotfix/single twisted nullspace

cpviolator:
This eliminates the need for generating two separate null spaces for twisted mass. The first MG instance (up or down) calculates the null space, and then passed through the MG routines so that the second (down or up) MG instance can reference the null space. Coarse grid calculations then follow as normal.


maddyscientist:
The million dollar question is how well this works?

cpviolator:
With a segfault... for the time being.

maddyscientist:
@cpviolator I'm closing this pull, since these multigrid developments should be merged into feature/multigrid rather than develop for now.  So can you reopen a pull into feature/multigrid instead.  We'll then do one merge of all the latest multigrid developments into develop from there.

cpviolator:
Sorry, didn't mean to push to develop... or push at all!

Single nullspace, multiple preconditioners is trickier than I first thought. I can get one preconditioner to reference a previous instance's level one null vectors, and each performs a separate coarse grid set-up. I perform a down type and up type inversion, switching between preconditioners, and both converge well with only slightly different iteration counts (only non-physical lattices tested so far.) However, I think destroying the preconditioners causes a segfault because as soon as one is deleted, the other's nullspace no longer exists.

An awful lot of messy gymnastics were involved here, such as passing a void pointer to an MG preconditioner through several redefined functions, and changing the protected twist variable for each null vector. I can use the QUDA branch that has QUDA_TWIST_SINGLET to solve the twist issue, but two MG preconditioners sharing a null space that is supposed to be intrinsic will be problematic for memory clean-up. I'd like to create a new structure in the MG class whose job is to calculate and store null vectors. Each multigrid_solver can reference it, and it can have its own destructor (Just delete the B[i]) That way, each mg instance can generate from new, load from host, or reference existing. In that last option, no new memory is allocated for additional mg instances.

I'll update this branch for inspection, but it's nowhere near a finished job.

maddyscientist:
@cpviolator no harm done 😄 You didn't push anything, just your pull request needs to be directed towards feature/multigrid and not develop.

How about instead of two preconditioners, each preconditioner contains multiple coarse operators?  For a given set of null space vectors, we could have multiple coarse operators associated with that null space, each operator having its own mu, mass, etc.?

maddyscientist:
Though perhaps if I just added the ability to switch the mu and mass parameters on a given preconditioner this functionality become moot....

cpviolator:
Something in MG has to be split to do this cleanly. Here's the workflow as
I did it haphazardly:

**********
Declare an UP type MG preconditioner. (preconditionerUP)
On creation, pass to it a NULL pointer to a multigrid_solver instance (mg_)
Detect that mg_ is NULL, allocate space for B vectors and generate(load) an
UP type null space.
Compute the coarse grid(s) with those level 0 null vectors.

Declare a DOWN type MG preconditioner. (preconditionerDN)
O creation, pass to it a valid pointer to preconditionerUP.
Detect that preconditionerUP is not NULL. Reference its B vectors.
Compute the coarse grid(s) with the level 0 null vectors from
preconditionerUP.

In invertQuda, run two solvers. One with UP type spinors and a
preconditioner pointing to preconditionerUP.
(tricky bit) Change the flavour of the source and solution spinors
(currently protected in feature/multigrid)
(tricky bit) Change the flavour of the B vectors (currently protected in
feature/multigrid) so checks are passed.
invert with DOWN types.

[This works well and both solvers converge on two and three levels]

Destroy both preconditioners, segfault ensues.
**********

Now, if one has a single preconditioner instance and a single set of null
vectors, there should be a simple structure added to contain all the coarse
grid information from different parameters so that one can point to them
with ease. There's a lot of stuff going on down there, not too sure how to
implement it best. Just declare two MG *coarse in the constructor? Would
everything coarse related need a duplicate? It should be easy enough to
start the split on level 0, then just keep things linear for lower levels.

Anyway, this works in principle, but some code restructure is required. It
will be easier once QUDA_TWIST_SINGLET is incorporated into
feature/multigrid. I'll keep strategising.

Dean

maddyscientist:
(I've just merged in develop into feature/multigrid)

cpviolator:
Awesome, that's one thing takes care of! :) I'll look at what needs to be in a container for coarse grids. When they are all in one place, and can be referenced, we just need two instances at level 0 and then both instances continue linearly. This will clean up a lot of things.

PR #592, opened 2017-03-21
Feature/multireduce

maddyscientist:
This should be fine to merge in, but since I had to make a couple of fixes to inv_cg_quda.cpp when merging in the latest feature/blocksolver, best to check first.


maddyscientist:
@weinbe2 and @mathiaswagner Looks like all the build tests fail in the feature/multireduce branch due to Eigen dependencies.  Is this the case with the feature/blocksolver branch as well?  If not, @weinbe2 can you fix this before we close this pull?

mathiaswagner:
While the builds fail I think this looks good.

PR #593, opened 2017-03-21
Multiple setup iterations

sbacchio:
I've started to work on some improvements to the setup phase inspired by DDalphaAMG. 

So far I just did the easy part (multiple setup iterations); but the goal should be the adaptive setup a la DDalphaAMG.


lib/multigrid.cpp
```diff
@@ -783,6 +783,7 @@ namespace quda {
     DiracMdagM mdagmSloppy(diracSloppy);
     if(solverParam.inv_type == QUDA_CG_INVERTER) {
       solverParam.maxiter = 2000;
+      solverParam.compute_true_res = false; // computing the true residual makes xmyNorm throw error on the coarse levels
```
maddyscientist: Can you tell what happens here?  Is this something that needs to be fixed?

sbacchio: I had the problem that calling CG on the coarse level with compute_true_res was throwing the following error:
```
MG level 2 (GPU): ERROR: Not supported (rank 0, host nid00016, /ccs/home/sbacchio/src/quda/lib/reduce_core.h:29 in re
duceCuda())
MG level 2 (GPU):        last kernel called was (name=N4quda4blas8xmyNorm2Id6float2S2_EE,volume=6x4x4x2,aux=vol=192,s
tride=192,precision=4,TwistFlavour=1)
```
My understanding is that xmyNorm is calling ReduceCuda which does not support a generic number of color. But I don't understand why the errors only appear for the xmyNorm in compute_true_residual..

PR #596, opened 2017-03-23
small fix to CmakeLists

mathiaswagner:
Just to behave as recommended in Cmake for requiring C++11.


maddyscientist:
What does this do exactly?

mathiaswagner:
https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED 

Just initialize a boolean variable with a boolean value (and not 11). Probably no functional effect.

PR #599, opened 2017-03-24
dependency checks for QMP, QIO and LIME

kostrzewa:
add QUDA_LIMEHOME and add CMake conditionals for QIO, QMP and LIME, such that QUDA_QMPHOME, QUDA_QIOHOME and QUDA_LIMEHOME are required to be set in order to configure with QUDA_QMP and/or QUDA_QIO set

I'm not sure if this is generally useful, but I find that I get linking errors and/or ```lime.h``` not found if I don't give an extra path for LIME. (maybe the github clone of the QIO repo is weird compared to any tarball people might have lying around).

Since QUDA_QMPHOME and QUDA_QIOHOME need to be set, I added some error conditions to ```CMakeLists.txt``` reflecting this fact.


cpviolator:
That GitHub QIO / QMP set up is a pain. I had trouble with it on Stampede
and Piz-Daint. I'll let you know how this works on both machines.

PR #601, opened 2017-03-24
Generation of Gaussian distributed initial momenta for HMC

hsandmeyer:
Add a routine that generates the initial Gaussian distributed elements of the algebra. This is necessary for generating the momenta in each start of a trajectory in the HMC.


mathiaswagner:

mathiaswagner:
@jenkins ok to test

mathiaswagner:

lib/interface_quda.cpp
```diff
@@ -5463,6 +5467,41 @@ void set_kernel_pack_t_(int* pack)
...
+  if (!gaugePrecise)
+    errorQuda("Cannot generate Gauss GaugeField as there is no resident gauge field");
+
+  cudaGaugeField *data = NULL;
+  data = gaugePrecise;
+
+  profileGauss.TPSTOP(QUDA_PROFILE_INIT);
+
+  profileGauss.TPSTART(QUDA_PROFILE_COMPUTE);
+  std::cout << "VOLUME " << data->Volume() << " " << data->X()[0] << std::endl;
```
mathiaswagner: looks like a leftover development / debug output

maddyscientist:
@hsandmeyer This looks like a good addition to QUDA.  Some comments:
* Can you add the doxygen markup for the spinor gaussian field generation (`spinorGauss`)
* While it's officially deprecated, since it's easy to do, can you make sure that this code compiles with configure / make rather than cmake?  Should be a trivial addition to lib/Makefile adding the filenames.
* I only see an interface function for the gauge field variant and not for the quark field.  Is that intentional?
* spinor_gauss.cu line 156 is a copy of the error message on line 143, and isn't the right message
* spinor_gauss.cu is hard-coded for `nColor=3` and `nSpin=1`, but isn't querying these field properties to see if they are valid.  Can you add these checks, and give the appropriate error message?  I presume it's trivial to add support for Wilson-type fermions?

hsandmeyer:
@maddyscientist 

- Doxygen markup is implemented
- It does compile with configure
- I did not need a interface function to call the Gaussian spinor, as I call this routine directly from our code.
- The Gaussian spinor is now possible also for Wilson fermions. 
- I removed those error messages in lines 143 and 156, since the check for GPU_STAGGERED_DIRAC does not make sense any more.

maddyscientist:
@hsandmeyer Sorry for letting this slip through the cracks.  I see the current version of your pulls is failing some of the build checks but not others.  Do you see the problem here?  Will get this merged once this issue is sorted out.
```
/home/jenkins/.jenkins/jobs/QUDAcmake/workspace/CUDAVERSION/8.0/GPUARCH/sm_60/lib/spinor_gauss.cu: In function ‘void quda::spinorGauss(quda::ColorSpinorField&, quda::RNG&)’:
/home/jenkins/.jenkins/jobs/QUDAcmake/workspace/CUDAVERSION/8.0/GPUARCH/sm_60/lib/spinor_gauss.cu:137:149: error: too many arguments for format [-Werror=format-extra-args]
```

maddyscientist:
@mathiaswagner can you look why one of the Jenkins build tests failed (Fermi build)?  I checked this manually, and it's fine.  Is the failure caused by treating warnings as errors (Fermi build lead to warnings owing to Fermi being a deprecated target with CUDA 8.0).

Either way, this pull is good, and I'm merging.

mathiaswagner:
Yes. Looks like we need to add `-Wno-deprecated-gpu-targets` to the strict build type.

maddyscientist:
@mathiaswagner I've added the `-Wno-deprecated-gpu-targets` to all builds, since we are (re)supporting QUDA on Fermi for the foreseeable future, it probably makes sense that the compiler shouldn't issues warnings all over the place, even if compiling with CUDA 8.0.  This is in the feature/rdma branch: makes sense to put it there, since that's the branch that fixes Fermi support.

PR #602, opened 2017-03-24
CGNE and CGNR

sbacchio:
This add the solvers CGNE and CGNR as derived class from CG.

This will clean most of the if statement we are adding in MG when CG is used as a solver.

Indeed CGNE and CGNR can be initialized with the direct operator (instead of defining the squared one) and they return directly the wanted solution without requiring more treatments.


maddyscientist:
Excellent addition!

kostrzewa:
@sbacchio @maddyscientist I guess that this will, when the changes you refer to are made, replace the solve types ```QUDA_NORMERR_[_PC_]_SOLVE``` and ```QUDA_NORMOP_[_PC_]_SOLVE```, correct?

maddyscientist:
Yes, that's the plan.  For the near term the old interface will still be supported but deprecated , e.g., for 0.9 release, and removed for 1.0 release.

PR #607, opened 2017-04-05
New Over-Improved-Symanzik stout smearing

cpviolator:
This routine implements the Over-Improved stout smearing outlined in https://arxiv.org/pdf/0801.1165.pdf. Topological charge calculations have been added after each link smearing routine in `su3_test` for quick comparisons.

### DIFF
--- a/include/gauge_tools.h
+++ b/include/gauge_tools.h
@@ -34,6 +34,17 @@ namespace quda {
 		  const GaugeField& dataOr,
 		  double rho);
 
+  /**
+     Apply Over Improved STOUT smearing to the gauge field
+
+     @param dataDs Output smeared field
+     @param dataOr Input gauge field
+     @param rho smearing parameter
+  */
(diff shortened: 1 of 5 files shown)


cpviolator:
Not sure why that build failed, all is good my end...

maddyscientist:
@mathiaswagner This looks like a Jenkins / cmake issue.  Can you see what went wrong with these build tests?

@cpviolator The code itself looks good.  Will just wait until the build test issue is cleared up before we get this merged.  I presume performance is good with this?

cpviolator:
Performance is as expected. 3D stout has 2x4 = 8 SU(3) multiplications per
link, but over improved 4D has 2x6 + 3x4x6 = 84. It is about 8 times slower
per step, so a touch faster than the expected 10.5 due to recycling links
once communicated.

mathiaswagner:
@maddyscientist Looks like I screwed up the Jenkins configuration. CUDA was not found. Hope to have that fixed now.

PR #618, opened 2017-06-01
Respect qmp-config ldflags and libs

evanberkowitz:
Addresses issue #617 via the suggestions of @cpviolator.
Adds ${QMP_LDFLAGS} and ${QMP_LIBS} to QUDA_QMP and QUDA_QDPJIT.

I have verified that the QUDA_QMP (non-QDPJIT) changes work for my compilation on Titan.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -109,11 +109,11 @@ if(QUDA_QIO)
 endif()
 
 if(QUDA_QDPJIT)
-  target_link_libraries(quda ${QDP_LIB} ${QIO_LIB} ${LIME_LIB} ${QMP_LIB} ${MPI_CXX_LIBRARIES})
+  target_link_libraries(quda ${QDP_LIB} ${QIO_LIB} ${LIME_LIB} ${QMP_LIB} ${QMP_LDFLAGS} ${QMP_LIBS} ${MPI_CXX_LIBRARIES})
 endif()
 
 if(QUDA_QMP)
-  target_link_libraries(quda ${QMP_LIB} ${MPI_CXX_LIBRARIES})
+  target_link_libraries(quda ${QMP_LIB} ${QMP_LDFLAGS} ${QMP_LIBS} ${MPI_CXX_LIBRARIES})
 endif()
 
 if(QUDA_MPI)


mathiaswagner:

mathiaswagner:

cpviolator:
Can you do the same for the CFLAGS too?

evanberkowitz:
I can and meant to, but to be honest I've never used CMAKE before, so I couldn't figure out from a quick google where to stick them!

mathiaswagner:

cpviolator:
Yeah, CMake is really, really nice, but there's definitely a significant learning curve! We've never known anybody to use a non standard QMP, so that's why this got missed.

cpviolator:
When you use `target_link_libraries` it's just like the old `LIB += $LDFLAGS $LIBS`

evanberkowitz:
For a little context, I'm integrating [mpi_jm](https://www.olcf.ornl.gov/wp-content/uploads/2016/10/2017-05-Oak-Ridge-Mpi-Jm-1.pdf), which we hope to release sooner rather than later into QMP.

mathiaswagner:
I feel like this is a somewhat hacky solution. I'd like to evaluate whether there is a cleaner way to do it. I don't like having to many hacks in the CMakeLists.

Anyway, @evanberkowitz do you mind pushing that to a branch (like hotfix/cmake_qmp ) in the quda repository? This way we can work from here. It looks like this lives in your branch now. 

Can you also somehow share your modified QMP for testing?

evanberkowitz:
I can share my qmp-config no problem.  My QMP is https://github.com/callat-qcd/qmp, I'm compiling the mpi_jm branch with `--with-mpi-jm=/path/to/mpi_jm/install`, but it requires mpi_jm, which is currently a private repo.  @kenmcelvain, @gustav-jansen, @walkloud

cpviolator:
I guess one could just

```cmake
LIST(APPEND QMP_LIB QMP_LIBS)
```
after `QMP_LIB` is set, but then the use would have to ensure, with no prompts from CMake, that the library paths are in `$LD_LIBRARY_PATH`

evanberkowitz:
@mathiaswagner is it possible to change this pull request to a hotfix branch?  Or do I need to push to my repo and create a new PR?

mathiaswagner:
The bigger picture here is to have an install target for QUDA which correctly export it's dependencies. 
That's why we use target_link_libraries and add the dependencies to QUDA. You can see that the individual tests inherit the dependencies and then properly link. Other projects then could use QUDA and correctly get dependencies and flags like the tests already do now.

This is an on and off project for me. If @evanberkowitz is fine using his solution for now I will have a look at a cleaner solution in the next days. There is some build / testing work that has piled up and maybe I can get some of it done before the Lattice conference.

evanberkowitz:
I'm OK for now.  I can't seem to [change the base branch of this PR](https://stackoverflow.com/a/24159161/2966190) to hotfix/cmake_qmp

cpviolator:
Yeah, I saw that upgrade to `tests/CMakeLists.txt`, really nice. This is relevant to us because rather than hacking QUDA, we want to use it as a dependency, while also using the QUDA style as much as possible.

evanberkowitz:
I think the issue is that `hotfix/cmake_qmp` doesn't exist and I don't have write permissions to this repo, so I can't create it.  @mathiaswagner do you want me to leave this PR as-is?

mathiaswagner:
I modified  this to be pulled into a hotfix/cmake_qmp. I will merge this and we can then create a new PR. I think it is not able to change the source of the pull request without creating a new one.
@evanberkowitz I will also check and grant your permissions to the quda repo.

mathiaswagner:
@evanberkowitz You should have write access to the quda repository. Did you try to push into develop? That won't work. We only allow pull requests to develop, not direct push. But you should be able to create a branch and push to that.

evanberkowitz:
I was trying to change my pull request to point to a new branch rather than develop, but it wouldn't work.  I also tried creating a new PR targeting a new branch and that wouldn't work either.  So... who knows.

azrael417:
Thank you!
Spark by Readdle

PR #623, opened 2017-06-15
added QDPJIT_LDFLAGS and QDPJIT_LIBS onto the CMAKE_EXE_LINKER_FLAGS

bjoo:
Currently the CMakeFile explicitly looks for some of the QDP Libraries
(libqdp, libqio, libc-lime etc), however, it doesn't look for all
the libraries on the QDP++ link line. This can cause issues when
QDP-JIT is being used, as the malloc is redirected to a QDPJIT implementation.
Hence all the QDP-JIT libs are neeeded especially in the case of an LLVM
based build. In this hotfix, I just add those libs onto the linker line.

However, do beware: The libraries are shared libs, so they still need to
be on the user's LD_LIBRARY_PATH for now. This could be solved with
an Rpath which we could configure in CMakeLists however, it would not be
needed for the QDP-JIT/PTX build. So until QDP-JIT/PTX is obsoleted,
the user needs to take care of it, both here and in any other client code
(e.g. chroma)


bjoo:
Ah, this didn't work properly. I have a new one. 
I will close this.

PR #632, opened 2017-07-12
Feature/multigrid circlestag

weinbe2:
Updated the spin_mapper to take in parity, as well as spin block sizes to handle fine spin = 1, which is in prep of coarsening staggered fermions. Since I haven't implemented the base components of Alexei's code to build a coarsened staggered operator, this is technically untested, but I verified that this doesn't break Wilson, clover, or twisted clover on both a two- and three- level solve (which covers all current combos of fine and coarse spin) in various permutations, coarsening both direct and direct-pc. Did a quick test of emulated multi-GPU.


mathiaswagner:

mathiaswagner:

mathiaswagner:

include/multigrid_helper.cuh
```diff
@@ -8,15 +8,17 @@ namespace quda {
...
 
     /**
        Return the coarse spin coordinate from the fine spin coordinate
        @param s Fine spin coordinate
+       @param parity fine parity, for staggered
        @return Coarse spin coordinate
      */
-    __device__ __host__ constexpr inline int operator()( int s ) const
-    { return s / (spin_block_size > 0 ? spin_block_size : 1); }
+    __device__ __host__ constexpr inline int operator()( int s, int parity ) const
```
maddyscientist: While not give the parity argument a default parameter since it only applies to staggered?

lib/transfer_util.cu
```diff
@@ -704,23 +702,24 @@ namespace quda {
 
   template<typename Float>
   void BlockOrthogonalize(ColorSpinorField &V, int Nvec,
```
maddyscientist: Is an extra template parameter needed here (spinBlockSize)?  Could one not just create a trait that maps from the nSpin template, to avoid having to pass this extra template around?

include/multigrid_helper.cuh
weinbe2: This change to the spin_mapper lets me reuse existing prolongator, restrictor, and transfer code as is, agnostic to Wilson-type or staggered fermions. I just updated spin_map(s) -> spin_map(s,parity) everywhere as part of this. I don't think there'd be a Wilson-only or staggered-only kernel that'd use the spin mapper?

lib/transfer_util.cu
weinbe2: Maybe I'm misunderstanding your question, but there isn't a perfect one to one mapping between nSpin and spinBlockSize. 

Maybe this helps: the ultimate culprit line that I thought required this extra template parameter is:

(644:) constexpr int coarseSpin = (nSpin == 4 || nSpin == 2 || spinBlockSize == 0) ? 2 : 1;

Which used to just check (nSpin == 4 || nSpin == 2) for Wilson-like -> coarse and coarse -> coarse, respectively. nSpin == 1 wouldn't be unique to staggered (lets say if you wanted to coarsen the Laplace operator). If we want to assume nSpin == 1 is uniquely staggered, then I can throw all of this extra stuff out the window.

Odds are I'm just misunderstanding what a trait could be used for, though!

include/multigrid_helper.cuh
maddyscientist: You're probably right that there's no Wilson-only part, so my comment is irrelevant.

lib/transfer_util.cu
maddyscientist: Ok, I see my trait suggestion breaks down because of the presence of the Laplace operator.  

One further comment here is to include the Laplace option here in the `GPU_STAGGERED_DIRAC` macro, since the Laplace operator is only built if `GPU_STAGGERED_DIRAC` is defined (since that's the macro that enables nSpin=1 blas routines).  To avoid horrible compile time when doing staggered-only multigrid, perhaps it makes sense to also guard the nSpin=4 and nSpin=2 cases  with `GPU_WILSON_DIRAC` which is enabled for all Wilson-like operators.

weinbe2: Added!

maddyscientist: Excellent.  Probably worthwhile adding them to prolongator and restrictor as well I guess...

weinbe2: Added, verified things still compile and work! Thanks for noticing it could be done there too.

PR #634, opened 2017-07-13
fix clang compilation of inv_cg_quda

mathiaswagner:
array of non-POD is not allowed in C++, only gcc compiles that.

Not sure how to test with Np set.

### DIFF
--- a/lib/inv_cg_quda.cpp
+++ b/lib/inv_cg_quda.cpp
@@ -1,6 +1,13 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cmath>
+#include <sys/time.h>
+#include <limits>
+#include <memory>
+
+#ifdef BLOCKSOLVER
+#include <Eigen/Dense>
+#endif
 
 #include <quda_internal.h>
 #include <color_spinor_field.h>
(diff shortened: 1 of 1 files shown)


lib/inv_cg_quda.cpp
```diff
@@ -399,11 +399,11 @@ namespace quda {
 	  } else {
 
 	    if ( (j+1)%Np == 0 ) {
-	      Complex alpha_[Np];
+	      auto alpha_ = std::unique_ptr<Complex[]>(new Complex[Np]);
```
maddyscientist: Memory leak?

mathiaswagner: Why? It is a unique_ptr.

maddyscientist: you are correct of course 😄

maddyscientist:
To test with Np, you just run invert_test with `--solution-pipeline N`.  And now I see that I failed to add this to the command line help list....  Perhaps you can add that here while testing? 😉

mathiaswagner:
Will do.

mathiaswagner:
`--solution-pipeline N` is already present in line of 1664 of test_util.cpp for me.

PR #635, opened 2017-07-13
Hotfix/improve eigen download

mathiaswagner:
Download Eigen already during cmake.

Note this currently downloads the eigen tarball into the source directory. Extraction is done into the build directory. 

Pro:
* Download eigen only once for multiple build dirs.
* If the automatic download fails users can copy the tarball into the right location into the source directory. (If the file is present and the SHA sum matches the download is skipped)
* error if download not possible already during cmake run - not during build.

Con:
* Doesn't keep the source dir clean (but externals is ignored in git)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ profile.tsv
 config.log
 CMakeCache.txt
 CMakeFiles
+externals
+.tags*
+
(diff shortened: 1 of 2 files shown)


mathiaswagner:
The native cmake build above fails because of a cmake issue (only shows up when using native CUDA support in cmake >= 3.8 and using Ninja as generator instead of make). I reported that to kitware. Should not stop us from going on.

maddyscientist:
If we merge this, won't this now mean that our native cmake Jenkins builds will now all fail then?

mathiaswagner:
I have changed it to build using the makefile generator.

PR #636, opened 2017-07-13
fix wrong include directory for downloaded eigen

mathiaswagner:
Fixes issues with #635.


maddyscientist:
@alexstrel can you check this fixes your issue?

alexstrel:
@maddyscientist yes, I have already checked this fix, we can merge.

PR #638, opened 2017-07-13
Added passing the mass to "createCoarseOp" to the base Dirac op in di…

weinbe2:
At the end of the day, this is a very trivial change: it just added "mass" as a parameter that gets carried around everywhere and is ignored for Wilson-type fermions, but it'll be needed for staggered. This isn't intended to add support for both kappa and mass normalizations for Wilson-type fermions, nor for being able to add a mass-like term to the coarse op (though that could be done).

That said, it involves tiny changes, but for every Wilson-type fermion, so it required a lot of testing just to be thorough, so I think it's best to get this merged sooner as opposed to later. It'd probably be safer to explicitly build the mass into a coarse clover, though.

Tests I performed:

`./multigrid_invert_test --prec double --prec-sloppy single`
`./multigrid_invert_test --prec double --prec-sloppy single --mg-levels 3 --mg-block-size 0 3 3 3 2 --mg-block-size 1 2 2 2 2 --solve-type direct-pc --dslash-type clover`
`./multigrid_invert_test --prec double --prec-sloppy single --mg-levels 3 --mg-block-size 0 3 3 3 2 --mg-block-size 1 2 2 2 2 --solve-type direct-pc --dslash-type twisted-mass`
`./multigrid_invert_test --prec double --prec-sloppy single --mg-levels 2 --mg-block-size 0 3 3 3 2 --solve-type direct-pc --dslash-type twisted-clover`

### DIFF
--- a/include/dirac_quda.h
+++ b/include/dirac_quda.h
@@ -161,11 +161,12 @@ namespace quda {
      * @param X[out] Coarse clover field
      * @param T[in] Transfer operator defining the coarse grid
      * @param kappa Kappa parameter for the coarse operator
+     * @param mass Mass parameter for the coarse operator (gets explicitly built into clover, hard coded to zero for non-staggered ops)
      * @param mu TM mu parameter for the coarse operator
      * @param mu_factor multiplicative factor for the mu parameter
      */
     virtual void createCoarseOp(GaugeField &Y, GaugeField &X, const Transfer &T,
-				double kappa, double mu=0., double mu_factor=0.) const
+				double kappa, double mass=0., double mu=0., double mu_factor=0.) const
     {errorQuda("Not implemented");}
 
   };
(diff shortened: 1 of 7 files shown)


mathiaswagner:

mathiaswagner:

alexstrel:
I just wonder why we do need yet another implementation of the coarse staggered, is the already checked one from stag-mg-esw not appropriate?

mathiaswagner:

weinbe2:
@alexstrel Your implementation of coarsening the staggered operator was correct for what we thought we needed a year+ ago, but it's not correct anymore (i.e., we don't want to enforce that terms on the block diagonal are zero, we do need preconditioning support, both things that were enforced in your code). There are also a lot of pieces of your code that would need to be updated to match the optimizations and features Kate has put into the current feature/multigrid branch, so it's ultimately better to start from an up-to-date version of that branch. The work you did is all useful and is being put into this branch as appropriate, which is why the stag-mg-esw shouldn't be deleted as there's plenty that needs to be transferred from there. 

In the interest of there not being too many cooks in the kitchen, I'm taking point on this new branch.

alexstrel:
@weinbe2 ok thanks, understood.

PR #642, opened 2017-08-03
Feature/twist clover cleanup

maddyscientist:
This pull request consists of a complete rewrite of dslash_quda.cu, and also includes multiple framework improvements to facilitate this rewrite.
* Fixes compilation for CUDA 9.0 by updating the use of warp synchronous intrinsics to use the new `_sync` variations introduced with CUDA 9.0 and updated to the latest version of CUB
* Volta optimization for dslash kernels: prefer large shared-memory carve out  
* Add texture support to `colorspinor::FloatNOrder` and `clover::FloatNOrder`accessors for fast half-precision data reading
* Add texture support to `gauge::FloatOrder` accessor, though for now not enabled (will be revisited)
* Add new `HMatrix` class which wraps around triangular packed Hermitian matrices (e.g., clover) and provides support for basic arithmetic operations that are optimal
* Added new `linalg` namespace, which includes `linalg::cholesky`, `linalg::forward`, and `linalg::backward` which are highly optimized for solving dense Hermitian matrix systems (e.g., solving for the Clover matrix)
* Complete rewrite of all kernels in dslash_quda.cu to use high-level framework.  This dramatically simplifies the kernels.  No performance regressions, and in many cases, new code is significantly faster (e.g., dynamic clover inversion using `linalg::cholesky`)
* Removal of now redundant core files in dslash_core
* Cleanup of backup and restore methods to simplify writing `Tunable::preTune` and `Tunable::postTune`
* Removes legacy `GPU_DIRECT` build option (closes #124)
* Other small cleanup and fixes

Asking for @alexstrel and @AlexVaq to review these changes since this affects kernels written by you folks.

PR #646, opened 2017-10-03
Feature/gauge comms cleanup

maddyscientist:
(This pull requires that first #642 is merged)

This pull request mainly addresses the enablement of peer-to-peer communication and GPU Direct RDMA for the extended gauge field exchange.
* Move common communications functionality from `ColorSpinorField` into `LatticeField`
* Add support for peer-to-peer and RDMA exchange for extended gauge field creation
* Remove legacy `FaceBuffer` class
* Remove legacy `bufferPinned` and `bufferDevice` static allocations
* Use the clover matrix element symmetry to reduce memory traffic in asymmetric clover dslash
* Remove clover_rho.cu (which added a `rho` constant to the clover diagonal) and replace this functionality with simply adding `rho` to the diagonal when apply clover dslash
* Optimize long-link computation
* Add Fortran helper interface for pinning externally declared memory
* When using BQCD gauge fields, use a hash comparison to check if gauge field needs to be refreshed 
* Ancillary cleanup and optimization


mathiaswagner:
After merging #642 I merged develop into this (without conflicts) to get a cleaner diff.

mathiaswagner:
Just a question: What is the rationale behind having the interface function for pinning/register only in Fortran?
Looks good otherwise.

maddyscientist:
@mathiaswagner I exposed an interface for pinning / unpinning memory specifically for BQCD.  BQCD uses persistent memory allocations, and so it makes sense to pin it's copy of the gauge field once it has been allocated.  I initially looked into an approach akin to `qudaAllocatePinned`, `qudaFreePinned` as is done in the C interface, however, this proved to be troublesome due to assigning a Fortran array to a C pointer not being a well-defined operation.  I think in the long term we can remove this, and have QUDA internally pin all buffers that are passed to it (if not already pinned), and only free them when `endQuda` is called.  This would simplify application code (obviously this approach would only be for applications that have persistent memory allocations, and it would be opt in).

Make sense?

mathiaswagner:
Yes, sounds like a good plan.

PR #650, opened 2017-10-19
Feature/milc site optimize

maddyscientist:
This pull request is focussed upon reducing the overhead of the MILC-QUDA interface and fixes some other issues
* Introduce a new gauge field ordering type `QUDA_MILC_SITE_GAUGE_ORDER` which corresponds to the `gauge::MILCSiteOrder` field accessor.  This is for mapping directly to MILC's internal array of sites and facilitates QUDA being able to read and write directly from this array.
* Updates the `qudaUpdate`, `qudaGaugeForce` and `qudaUnitarizeSU3` interface functions to use the direct site mapping for greatly improved throughput.
* Introduce a new `vector_store` helper that does streaming stores for generic vector types.  This ensures that all stores from `gauge::FloatNOrder`, `clover::FloatNOrder` and `colorspinor::FloatNOrder` are vectorized.
* Adds an explicit check for all texture sizes, to ensure that we don't attempt to bind a texture larger than supported by the device 

This should not be merged until we have confirmed that the partner pull request for MILC will also be merged.

### DIFF
--- a/lib/cuda_gauge_field.cu
+++ b/lib/cuda_gauge_field.cu
@@ -525,33 +525,47 @@ namespace quda {
 	qudaMemcpy(gauge, buffer, bytes, cudaMemcpyHostToDevice);
 	pool_pinned_free(buffer);
       } else { // else on the GPU
-	void *buffer = create_gauge_buffer(src.Bytes(), src.Order(), src.Geometry());
-	size_t ghost_bytes[8];
-	int srcNinternal = src.Reconstruct() != QUDA_RECONSTRUCT_NO ? src.Reconstruct() : 2*nColor*nColor;
-	for (int d=0; d<geometry; d++) ghost_bytes[d] = nFace * surface[d%4] * srcNinternal * src.Precision();
-	void **ghost_buffer = (nFace > 0) ? create_ghost_buffer(ghost_bytes, src.Order(), geometry) : nullptr;
 
-	if (src.Order() == QUDA_QDP_GAUGE_ORDER) {
-	  for (int d=0; d<geometry; d++) {
-	    qudaMemcpy(((void**)buffer)[d], ((void**)src.Gauge_p())[d], src.Bytes()/geometry, cudaMemcpyHostToDevice);
+	if (src.Order() == QUDA_MILC_SITE_GAUGE_ORDER) {
+	  // special case where we use zero-copy memory to read/write directly from MILC's data
+	  void *src_d;
+	  cudaHostGetDevicePointer(&src_d, const_cast<void*>(src.Gauge_p()), 0);
+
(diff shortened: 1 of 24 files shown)


mathiaswagner:
Since https://github.com/milc-qcd/milc_qcd/pull/17 is merged we should be fine when building against the development version of MILC. Remaining question is where do we point people that still need the old interface and how to reduce the number of people running into issues with having either version wrong ?

Just live with it and hope it will be communicated clear enough?

lib/cuda_gauge_field.cu
```diff
@@ -525,33 +535,47 @@ namespace quda {
...
-	for (int d=0; d<geometry; d++) ghost_bytes[d] = nFace * surface[d%4] * srcNinternal * src.Precision();
-	void **ghost_buffer = (nFace > 0) ? create_ghost_buffer(ghost_bytes, src.Order(), geometry) : nullptr;
 
-	if (src.Order() == QUDA_QDP_GAUGE_ORDER) {
-	  for (int d=0; d<geometry; d++) {
-	    qudaMemcpy(((void**)buffer)[d], ((void**)src.Gauge_p())[d], src.Bytes()/geometry, cudaMemcpyHostToDevice);
+	if (src.Order() == QUDA_MILC_SITE_GAUGE_ORDER) {
+	  // special case where we use zero-copy memory to read/write directly from MILC's data
+	  void *src_d;
+	  cudaHostGetDevicePointer(&src_d, const_cast<void*>(src.Gauge_p()), 0);
```
mathiaswagner: Until I had enough coffee just let me ask: What if `cudaHostGetDevicePointer` goes wrong?

I might need to look at the code more closely but I thought that with UVA we can get by without the `cudaHostGetDevicePointer` if we used cudaHostRegister / cudaHostMalloc.

maddyscientist: I've added an error on the return of this function to check that this mapping is valid.  You are correct with UVA the API call isn't needed, but no harm in having this there since it would then work on non-UVA systems (Windows anyone 😉  ?).  Also, since there's now an explicit error check on this, we'd catch now with a sensible error message if a regular pointer was passed in as opposed to a CUDA registered one (asopposedd to a random segmentation fault).

mathiaswagner: I added the check also in another place. Using `cudaHostGetDevicePointer` to make sure the memory has been registered is a reasonable thing.

mathiaswagner:
Just wondering whether we can add some compile time error checking for the Interface / MILC compatibility.

Like in `quda.h` add 
```
#define MILC_INTERFACE_V2
```

and on the MILC side (where `quda.h` is include) something like

```
#ifndef MILC_INTERFACE_V2
#error "This version of MILC requires ..."
```
However I don't see an obvious way to check the other direction and give a good error message.

PR #651, opened 2017-10-23
Hotfix/arpack cleaning

alexstrel:
Cleaned arpack header file. Fixed bug in the arpack matvec functor.


lib/quda_arpack_interface.cpp
```diff
@@ -117,7 +117,7 @@ namespace quda{
       QudaMatvec(DiracMatrix &matEigen, QudaPrecision prec, ColorSpinorField &meta) : matEigen(matEigen), matPrecision(prec) { 
          ColorSpinorParam csParam(meta);
          //csParam.extendDimensionality();//4d -> 5d
-         csParam.fieldOrder = QUDA_FLOAT2_FIELD_ORDER;
+         csParam.fieldOrder = (meta.Nspin() == 1 || matPrecision == QUDA_DOUBLE_PRECISION) ? QUDA_FLOAT2_FIELD_ORDER : QUDA_FLOAT4_FIELD_ORDER;
```
maddyscientist: @alexstrel I'm confused by this edit.  The `setPrecision` method is called here which should take care of setting the field order.  Was this not working?

alexstrel: @maddyscientist  csParam is created  here from a non-native 'meta' field, since arpack wants cpu objects, which may have,e.g.,  space-spin-color field order. So setPrecision() will treat it as non-native as well, and will not convert to float2 or float4 field order automatically.

maddyscientist: Ok, can you put a comment in there to this effect then, so it isn't accidentally deleted in the future?

Also, any reason to keep this line there:
```c++
 //csParam.extendDimensionality();//4d -> 5d
```

PR #652, opened 2017-10-24
Hotfix/multi gpu p2p bugfix

maddyscientist:
* Important bug fix for a latent race condition in the dslash: was triggered if in a given dimension, the MPI communication in a given direction completed before the local peer-to-peer communication  completed in the other direction in the same dimension.  This scenario only really applies to systems with first-class GPU RDMA support where MPI could complete in comparable time as peer-to-peer.
* Ensure that split kernel dslash halo kernels run in order T->X to prevent any possible race conditions if lower dimensions complete before high dimensions.  This ensures reproducibility of dslash computation. 
* Added option to environment variable `QUDA_ENABLE_P2P=-1`: if the process topology is length 2, then enabling this will only allow half of the network traffic to traverse peer-to-peer and force the other half to traverse MPI.  Useful to extrapolating performance at high node counts. 
* Improved reporting statistics for bi-directional bandwidth of dslash_test
* Targeted kernel and precision testing of blas_test, since this now respects `--test` and `--prec` flags to overwrite default behaviour

### DIFF
--- a/lib/comm_common.cpp
+++ b/lib/comm_common.cpp
@@ -142,6 +142,101 @@ void comm_destroy_topology(Topology *topo)
 }
 
 
+static bool peer2peer_enabled[2][4] = { {false,false,false,false},
+                                        {false,false,false,false} };
+static bool peer2peer_init = false;
+void comm_peer2peer_init(const char* hostname_recv_buf)
+{
+  if (peer2peer_init) return;
+
+  bool disable_peer_to_peer = false;
+  char *enable_peer_to_peer_env = getenv("QUDA_ENABLE_P2P");
+  if (enable_peer_to_peer_env && strcmp(enable_peer_to_peer_env, "0") == 0) {
+    if (getVerbosity() > QUDA_SILENT) printfQuda("Disabling peer-to-peer access\n");
+    disable_peer_to_peer = true;
+  }
+
+  // disable peer-to-peer comms in one direction if QUDA_ENABLE_P2P=-1
+  // and comm_dim(dim) == 2 (used for perf benchmarking)
+  bool disable_peer_to_peer_bidir = false;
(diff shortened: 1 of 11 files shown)


mathiaswagner:
Wondering whether we should merge this before #650 to get this fix also in for the old MILC interface.

maddyscientist:
@mathiaswagner just use 84a072f70e2a7b1e407e2557433fdee59862bd42 for old MILC: this has all the fixes without the feature/milc-site-optimize branch merged in.  Tag this and make that the pre0.9 release.

mathiaswagner:
Re-merged develop as github complained about merge conflicts.

lib/comm_common.cpp
```diff
@@ -142,6 +142,101 @@ void comm_destroy_topology(Topology *topo)
...
+	    cudaDeviceGetP2PAttribute(&accessRank[1], cudaDevP2PAttrPerformanceRank, neighbor_gpuid, gpuid);
+	  }
+#endif
+
+	  // enable P2P if we can access the peer or if peer is self
+	  if (canAccessPeer[0]*canAccessPeer[1] || gpuid == neighbor_gpuid) {
+	    peer2peer_enabled[dir][dim] = true;
+	    if (getVerbosity() > QUDA_SILENT) {
+	      printf("Peer-to-peer enabled for rank %d (gpu=%d) with neighbor %d (gpu=%d) dir=%d, dim=%d, performance rank = (%d, %d)\n",
+		     comm_rank(), gpuid, neighbor_rank, neighbor_gpuid, dir, dim, accessRank[0], accessRank[1]);
```
mathiaswagner: I guess accessRank is undefined here if gpuid ==neighbor_gpuid.

maddyscientist: No, it's initialized on line 199.

PR #654, opened 2017-10-28
Enable performant GPU Direct RDMA for asymmetric topologies

### LINKED ISSUE #643: Enable selective GDR support for asymmetric node topologies
I've been thinking about how to enable GDR on the FNAL K40 system, which is asymmetric with respect to NIC and GPU placement.  E.g.,
```
CPU 0 - - - - - - QPI - - - - - - CPU 1
* GPU 0                         * GPU 2
* GPU 1                         * GPU 3
* NIC
```
How about I expose yet another env arg

```bash
export QUDA_ENABLE_GDR=1
export QUDA_ENABLE_GDR_RANK="0,1"
```

This would only enable GDR for the processes listed in the `QUDA_ENABLE_GDR_RANK` env arg.  The default, if `QUDA_ENABLE_GDR_RANK` is not enabled, would be to enable GDR for all ranks (if `QUDA_ENABLE_GDR` is set).

Internally in QUDA, when it autotunes the policy, when a process that has GDR disabled, it would just run the corresponding non-GDR dslash policy; I don't see any reason why we couldn't have this asymmetric policy like this enabled since all each process would see would be a MPI send/recv.

Thoughts from anyone on this?  Hopefully, node topologies like this will be a thing of the past, but it would be good to maximize performance on existing systems.

maddyscientist:
This pull request is for improving the performance and robustness of using GPU Direct RDMA (GDR)
* Add new environment variable `QUDA_ENABLE_GDR_BLACKLIST`.  Any process with GPU device ordinal equal to a value listed in this environment variable will not use GDR for the dslash neighbor exchange and will instead explicitly route all exchange through CPU memory (excluding peer-to-peer exchange).  This enables performant support for GDR on systems with asymmetric topologies, e.g., where only a subset of GPUs are on the same PCIe bus as the NIC (closes #643).
* Ensure that all processes use the same policy with an explicit broadcast after policy tuning is complete.  This is strictly required for identical reproducibility between runs.  Moreover, without this fix, there exists a danger that one process will run a dslash policy with t-dimension kernel packing, and another does not: this will lead to invalid results being generated until the tunecache is resynchronized (typically when the next kernel is tuned).
* Add new command-line parameter `--rank-order <col/row>` for unit tests to swap between column-major order and row-major order of the processes.  E.g., we have a process topology `[Pt][Pz][Py][Px] `, the default is column major where t runs fastest, and the alternative row major has x running fastest.
* When using `QUDA_BQCD_GAUGE_ORDER` gauge fields, we now do zero-copy reading/and writing of the gauge field.  This avoids the need to copy down the padded gauge field, reducing the copy overhead.
* Significant (but not exhaustive) cleanup of dslash_policy.cuh to reduce boilerplate and simplify the code.
* Adds support for peer-to-peer and GDR exchange in `GaugeField::exchangeGhost()` (e.g., accelerates `loadGaugeQuda()`)
* Fixed various corner cases with multi-GPU communication (e.g., latent race condition in dslash communication when using pure peer-to-peer communication with T partitioning only, dslash policy autotuning robustness, etc. )  

I have thoroughly tested the blacklisting feature on both Pascal and Kepler systems.  The benefit on systems tested so far is minor, since they have very weak networks (FDR IB shared between 4 or 6 GPUs) and the NICs were not a PLX switch with the GPUs which is required for performant GDR.  However, the benefit from blacklisting GPUs that are not on the same PCIe bus as the NIC is clear.

The results below were taken from a two quad Pascal nodes, where CPU0 has 1x P100 and CPU1 has 3xP100s and a NIC.  Here we are testing performance in GFLOPS of single precision Wilson dslash, over local volumes 4^4 - 32^4 with a 1x1x1x8 process topology, and are restricting to the following policies using `QUDA_ENABLE_DSLASH_POLICY`
* `staging` all transfers go through CPU memory
* `p2p` add a peer-to-peer communication where possible
* `gdr` for GPU Direct RDMA for all GPUs (e.g, MPI manages all inter-GPU transfers)
* `p2p/gdr` as `gdr` with peer-to-peer where possible
* `p2p/gdr/blacklist` set `QUDA_ENABLE_GDR_BLACKLIST=0`, to exclude GPU 0 from trying to use GDR and route through CPU memory

```
 L     staging      p2p      gdr    p2p/gdr   p2p/gdr/blacklist
================================================================
 4        4.11     4.19     4.41       4.42                4.06
 8       48.91    48.52    56.07      56.30               48.49
 12     170.52   171.47   158.52     157.80              172.25
 16     370.44   373.55   256.86     262.35              373.25
 20     605.41   606.68   391.42     390.03              608.84
 24     776.86   771.58   478.44     479.45              777.71
 28     886.95   879.88   563.77     561.25              887.12
 32     914.79   915.54   641.01     645.73              917.89
```
Forcing GDR for all GPUs has a catastrophic effect on performance, since GPU0 is being forced to use GDR which has horrid performance.  Adding explicit P2P doesn't really help, even pure staging with no P2P  is better.  Finally, once we blacklist GPU0 from using GDR, we see that performance increases dramatically, and we achieve performance parity with, or even sometimes better than the usual P2P and staging policy.

The implementation isn't perfect, since GDR requires kernel packing in the t dimension, so the GPUs that are not running GDR also must use kernel packing for correctness.  The other issue is that when GDR-enabled GPUs are communicating with non-GDR GPUs in the node, we rely on MPI to do the staging rather than QUDA's manual staging.  Both of these effects can have a small effect on performance, but the benefit of GDR where available will outweigh this.  Also note performance will never be worse, since the dslash policy auto tuner will just choose a non-GDR policy if it's faster.

I have updated the documentation in https://github.com/lattice/quda/wiki/Multi-GPU-Support#asymmetric-topologies with an example run script on how to achieve optimal performance.

This pull should be completed after #653.


maddyscientist:
@alexstrel it would be good to test this on FNAL's systems (assuming you have a recent OFED driver there that support GPU Direct RDMA).

alexstrel:
@maddyscientist on FNAL cluster  we have:
 
mlx4_core: Mellanox ConnectX core driver v3.1-1.1.0 (07 Dec 2015)
<mlx4_ib> mlx4_ib_add: mlx4_ib: Mellanox ConnectX InfiniBand driver v3.1-1.1.0 (07 Dec 2015)
mlx4_en: Mellanox ConnectX HCA Ethernet driver v3.1-1.1.0 (07 Dec 2015)

which is not the recent one but it does support GPU RDMA , the oldest driver is 2.1, e.g.
 
https://community.mellanox.com/community/support/software-drivers/rdma-software-for-gpu/blog/2014/01/27/using-gpudirect-rdma-with-mpi

maddyscientist:
Yes Alexei, I think that version of the driver should be fine.  You need also to make sure that nvidia-peer-memory (v1.0.5) is installed on the system as well (sys admin will need to do this).  Ideally, the gdrcopy library should also be installed (https://github.com/NVIDIA/gdrcopy).

Can you log into one of the node at FNAL, and tell me the output of "nvidia-smi topo -m"?  This will allow me to make the appropriate run script (though this is quite trivial - see https://github.com/lattice/quda/wiki/Multi-GPU-Support#asymmetric-topologies).

alexstrel:
@maddyscientist I had already a look at Wiki page, thanks, it's very helpful. Here is the output from nvidia-smi : 
GPU0     X      PHB     SOC     SOC     SOC     0-7
GPU1    PHB      X      SOC     SOC     SOC     0-7
GPU2    SOC     SOC      X      PHB     PHB     8-15
GPU3    SOC     SOC     PHB      X      PHB     8-15
mlx4_0  SOC     SOC     PHB     PHB      X  
so I blacklisted gpu 0 and gpu 1. So far I did not see performance difference , although I did not play with dslash policies. My environment is pretty old , e.g.,  cuda-7.5 openmpi 1.10.3. Maybe this is the reason. 
We have nv_peer_mem installed,  just quickly checked lsmod:
Module                  Size  Used by
nvidia_uvm            726837  0 
....
nvidia_drm              1150  0 
nvidia              11933738  2 nvidia_uvm,nv_peer_mem 
I cannot check its version though.

alexstrel:
One more question, what happened if the user has already tuned the library, for some reason, without GDR tricks and generated a valid tunecache file, and then will decide to enable the topology-aware optimizations in the next runs? I guess this will not re-tune the application automatically so the user should do this procedure manually.  Is it possible to warn the user in such a situation?

maddyscientist:
@alexstrel Good question.  Well, the autotuner will force a retune if either of the environmental variables `QUDA_ENABLE_P2P` or `QUDA_ENABLE_GDR` are changed between the original tuning run and the subsequent run.  It won't detect a change in the `QUDA_ENABLE_GDR_BLACKLIST` though: I don't think there is a good way to detect such a change though.  Since the blacklist values can be different for different nodes, e.g., node 0 could have its NIC on CPU0 but node 1 could have its NIC on CPU1, we would want to set a different value of the envarg for each node (which could easily be done in the run script).

mathiaswagner:
I guess we can just document well enough that the blacklist option does not force a retune and that care should be taken when using it. It is quite an advanced feature. Users striving for this optimizations should be able to understand its implications.

maddyscientist:
Yes, I think that will suffice.

There a few improvements I'm still working on here to this.  Having done a bit more benchmarking, I see that OpenMPI is really bad at host staging, and for the intra-node traffic to non-p2p GPUs, this could easily be a significant bottleneck.  I'm working on code cleanup here, and will at the same time work out how to enable QUDA's host staging for this.

mathiaswagner:
I guess I am waiting for an explicit go before reviewing this.

lib/dslash_policy.cuh
```diff
@@ -1816,14 +1836,12 @@ struct DslashFactory {
...
 
 	   // check this is a valid policy choice
-	   if ( (dslash_policy == QUDA_GPU_COMMS_DSLASH || dslash_policy == QUDA_FUSED_GPU_COMMS_DSLASH) && !comm_gdr_enabled() ) {
+	   if ( (dslash_policy == QUDA_GDR_DSLASH || dslash_policy == QUDA_FUSED_GDR_DSLASH ||
+		 dslash_policy == QUDA_GDR_RECV_DSLASH || dslash_policy == QUDA_FUSED_GDR_RECV_DSLASH)
+		&& !comm_gdr_enabled() ) {
 	     errorQuda("Cannot select a GDR policy %d unless QUDA_ENABLE_GDR is set", dslash_policy);
 	   }
 
-	   if ((dslash_policy == QUDA_ZERO_COPY_DSLASH || dslash_policy == QUDA_FUSED_ZERO_COPY_DSLASH) && comm_peer2peer_enabled_global()) {
```
mathiaswagner: the dependence of the relation of the value written in the tunecache for the policy on the enabled Dslashes still gives me some headaches. We can leave it like that for now but when cleaning up I'd opt for something unique.

maddyscientist:
Single-GPU compilation fixed.

lib/dslash_policy.cuh
maddyscientist: I'll have a think about this tonight for a more absolute and less relative method.

maddyscientist:
That is a dumb bug in the lattice_field.cpp, and I'm glad this has been caught.  Undefined behaivour that could lead to a halo kernel being launched before the p2p memcpy is in flight, but I could see how it would be unlikely to happen (the forwards direction isn't buggy so it would have to be in flight before any halo kernel is initiated).  Fix for that incoming.

mathiaswagner:
Credit for this goes to @j-stephan. Good catch!

lib/dslash_policy.cuh
mathiaswagner: I have a first implementation in #666

lib/copy_gauge_helper.cuh
```diff
@@ -161,13 +163,13 @@ namespace quda {
...
-	      arg.out.Ghost(d+arg.offset, parity, x, i, j) = arg.in.Ghost(d+arg.offset, parity, x, i, j);
+	      arg.out.Ghost(d+arg.out_offset, parity, x, i, j) = arg.in.Ghost(d+arg.in_offset, parity, x, i, j);
 #else
 	  RegTypeIn in[length];
 	  RegTypeOut out[length];
-	  arg.in.loadGhost(in, x, d+arg.offset, parity); // assumes we are loading
+	  arg.in.loadGhost(in, x, d+arg.out_offset, parity); // assumes we are loading
 	  for (int i=0; i<length; i++) out[i] = in[i];
-	  arg.out.saveGhost(out, x, d+arg.offset, parity);
+	  arg.out.saveGhost(out, x, d+arg.in_offset, parity);
```
mathiaswagner: is the offset right here?

maddyscientist: Thanks.  Fix pushed,

maddyscientist:
Ok, this pull request is now ready.  I've updated the descriptor in the pull request with the latest updates.

lib/dslash_quda.cu
```diff
@@ -89,14 +88,15 @@ namespace quda {
...
-      cudaEventCreate(&packEnd[i], cudaEventDisableTiming);
       cudaEventCreate(&gatherStart[i], cudaEventDisableTiming);
       cudaEventCreate(&gatherEnd[i], cudaEventDisableTiming);
       cudaEventCreateWithFlags(&scatterStart[i], cudaEventDisableTiming);
       cudaEventCreateWithFlags(&scatterEnd[i], cudaEventDisableTiming);
     }
-    cudaEventCreateWithFlags(&dslashStart, cudaEventDisableTiming);
-    cudaEventCreateWithFlags(&dslashEnd, cudaEventDisableTiming);
+    for (int i=0; i<2; i++) {
+      cudaEventCreate(&packEnd[i], cudaEventDisableTiming);
```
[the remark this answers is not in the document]
mathiaswagner: Ok, seems like the `CudaEventCreate(cudaEvent_t* event, unsigned int  flags)` is in `cuda_runtime.h` and not in `cuda_runtime_api.h`. 
But as it just does
```
static __inline__ __host__ cudaError_t cudaEventCreate(
  cudaEvent_t  *event,
  unsigned int  flags
)
{
  return ::cudaEventCreateWithFlags(event, flags);
}
```
they seem to be equivalent apart from mixing the C++ and the C API.

Maybe its a good time to choose one of them.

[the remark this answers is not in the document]
maddyscientist: Fixed (to make explicit use of `cudaEventCreateWithFlags`).

PR #659, opened 2017-11-23
Add doxygen support

mathiaswagner:
This is to cleanup #578 with cmake support and then merge it in.

I will need to add this cleanup first and let you know once this is done.


mathiaswagner:
Little to-do list here:

**remove generating doc for external header**

- [ ] gtest
- [ ] cub

**add cmake support**

- [ ] add FindDoxygen
- [ ] add CMakeLists.txt in doc
- [ ] modify doxyfile to write to binary dir

mathiaswagner:
Should be good to test

PR #661, opened 2017-11-23
Remove autoconf

mathiaswagner:
This PR 

* removes the `configure` script from the quda main directory as well as the `configure.something` examples. Note that autoconf can still be used by
```
1. autoconf
2. ./configure
3. make -f Makefile.LastResort old
```
* Remove all references to autoconf in the README
* Renames README to README.md to make it show up on the GitHub


alexstrel:
@maddyscientist no, EigCG does not need MAGMA and can be run out-of-box with Eigen routines. I overlooked the README file was not updated. I wonder, since we have Wiki , we may just refer to the corresponding Wiki page?

maddyscientist:
@alexstrel ok, I have updated the README to clarify this and also add a link to the wiki.  The wiki is good for the details, but the README needs to have the basics correct!

mathiaswagner:
@maddyscientist I saw you approved this. I am good with it so whenever you are good we can merge. I guess we will have at least one PR once we have branched of 0.9 to update the NEWS file.

PR #662, opened 2017-11-27
remove legacy CUDA_TOOLKIT_ROOTDIR dependency from FindCUDALibs.

mathiaswagner:
CUDA_TOOLKIT_ROOT_DIR was used by FindCUDA.cmake. When using native cuda support in cmake we use FindCUDAlibs.cmake which is based on FindCUDA to detect the cuda libraries.

We now detect libraries based on the values of `${CMAKE_CUDA_COMPILER}` _(path to nvcc)_ and cmake's `${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}` (same mechanism used to also detect implicit link directories for other languages).

### DIFF
--- a/cmake/FindCUDALibs.cmake
+++ b/cmake/FindCUDALibs.cmake
@@ -51,6 +51,7 @@
 #   CUDA_nvcuvid_LIBRARY  -- CUDA Video Decoder library.
 #                            Only available for CUDA version 3.2+.
 #                            Windows only.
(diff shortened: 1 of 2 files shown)


mathiaswagner:
Note that this is not yet cleaned up. It is mostly out for testing. If everything looks good I can just clean it up a bit before we merge.

mathiaswagner:
I did run some tests and they look good and I think this is good to be merged from my side.

PR #665, opened 2017-12-06
Always build deflated solvers

mathiaswagner:
we no longer need the DEFLATEDSOLVER ifdef

### DIFF
--- a/include/deflation.h
+++ b/include/deflation.h
@@ -84,9 +84,6 @@ namespace quda {
     /** TimeProfile for this level */
     TimeProfile profile;
 
-    /** Prefix label used for printf*/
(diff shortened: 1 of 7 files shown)


include/deflation.h
```diff
@@ -85,7 +85,7 @@ namespace quda {
     TimeProfile profile;
 
     /** Prefix label used for printf*/
-    char prefix[128];
+    // char prefix[128];
```
maddyscientist: What this here for?

alexstrel:
Removed the obsolete code . Ran a few tests with deflated solvers, no issues so far. I think it can be merged.

PR #666, opened 2017-12-07
rewrite handling of selected dslash_policies for tunecache to make the value in the…

mathiaswagner:
The value in the tunecache now always represents the same policy, independent of enabled policies.

This is a suggestion to extend #654 before merging into develop.


maddyscientist:
Running `mpirun -np 3 dslash_test --gridsize 1 1 1 3 --dim 8 8 8 8` fails with this branch when running with tuning, though the post-tune run is fine (other partitionings are fine).  Problem does not occur with gdr-blacklist branch.  I guess something wrong with the labelling of the policies?

mathiaswagner:
I could not get it to fail yet. Can you point me to your build?

PR #667, opened 2017-12-14
Static assignment of aux.x values to Dslash policies for coarse Dslash

mathiaswagner:
This should work but has not seen extensive testing yet

### DIFF
--- a/lib/dslash_coarse.cu
+++ b/lib/dslash_coarse.cu
@@ -902,7 +902,7 @@ namespace quda {
 
 #endif // GPU_MULTIGRID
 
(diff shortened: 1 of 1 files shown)


maddyscientist:
Build errors in Jenkins
`/scratch/jenkins/.jenkins/jobs/QUDAcmake/workspace/CUDAVERSION/8.0/GPUARCH/sm_60/lib/dslash_coarse.cu:1074:165: error: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘quda::DslashCoarsePolicy’ [-Werror=format=]
`

PR #672, opened 2018-01-13
Hotfix/cmake sharedlib fixes

mathiaswagner:
allow quda to build as a shared library with QDP_JIT and bypass QDP pool allocator if `QDP_USE_CUDA_MANAGED_MEMORY` is set 
(currently only available in @bjoo's  fork of qdpjit)


mathiaswagner:
Updated to a newer version to see whether that builds fine with older cmake.

PR #673, opened 2018-01-18
Feature/ghost precision

maddyscientist:
* Add new members `LatticeField::ghost_precision` `LatticeFieldParam::ghost_precision` for separate halo precision and field precision specification
* Both `LatticeFieldParam::precision` is now protected forcing the use of `setPrecision` and `Precision` methods for setting and getting the precision (this ensures consistent setting of `ghost_precision`)
* Dimension and direction parallelization in `genericPackGhost` (with the former autotuned)
* Optionally specify the halo precision in `ColorspinorField::exchangeGhost` 
* Add support for block-float format to fine-grained colorspinor accessor (enabled if field is floating point but halo precision is fixed point)
* Add support for block-float 16-bit halos from single-precision fields in `genericPackGhost` function through cooperative computation of the site max
* Dslash coarse now uses the precision of the link field to specify its halo precision (with the result that 16-bit link fields result in 16-bit halo exchange)

### DIFF
--- a/include/blas_cublas.h
+++ b/include/blas_cublas.h
@@ -7,6 +7,16 @@ namespace quda {
   namespace cublas {
 
     /**
+       @brief Create the CUBLAS context
+    */
+    void init();
+
+    /**
+       @brief Destroy the CUBLAS context
+    */
+    void destroy();
+
+    /**
        Batch inversion the matrix field using an LU decomposition method.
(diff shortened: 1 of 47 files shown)


bjoo:
Hi All, I have had some issues working with the multishift after having tried a merge myself. I got Nan'a and wrong answers there. Kate is working on this. Please let us not merge for the moment.

PR #676, opened 2018-02-06
update of CMakeLists for build as shared lib

mathiaswagner:
Also includes some ifdefs to switch to pinned allocations for QUDA if QDP uses managed memory


mathiaswagner:
Sorry, the fixed I tried to make is actually only implemented in cmake 3.9. So for old cmake a more verbose solution is necessary ...

mathiaswagner:
I think this is fine now. Shared lib will require cmake 3.8+ now.

PR #680, opened 2018-02-15
Feature/full gpu mg setup

maddyscientist:
This pull enables the entire MG setup to be on the GPU
* Use CURAND for generating noise fields on the GPU
* The `QudaMultigridParam::setup_location[]` parameter determines where the `B` fields are allocated, so if setup is done on the GPU then these fields will be allocated on GPU and kept on the GPU
* Packed null-space components are allocated lazily on the "other" memory space (e.g., only allocated if required)
* If MG setup verification is requested then CPU-GPU transfers will be invoked

@bjoo can you test this branch with your standard baseline HMC test before we merge this in?
@mathiaswagner good to have another eye on this


maddyscientist:
Verified by Balint on Summit.

PR #682, opened 2018-03-01
Feature/milc optimize

maddyscientist:
This pull is a represents a rewrite of the HISQ force computation to use much cleaner code, reduce memory footprint and reduce communiucation
* Complete rewrite of the HISQ fermion force kernels to use accessors instead of macro hell
* Fields are only extended in the partitioned dimensions
* Add support for copying from non-extended CPU gauge fields directly into extended CUDA gauge fields
* Removal of obsolete single-GPU only ASQTAD force for R algorithm
* Cleanup of hisq_path_force_test.cpp (far from perfect)


maddyscientist:
Ok, I think this is effectively done.  I've done some more cleanup and reduced the amount of boilerplate.

I've realized that kernel fusion of the fat-link force kernels will require a significant rewrite as they are presently written in a scatter-centric fashion: we would need to rewrite the kernels in a gather-centric fashion to allow this.  Far beyond the scope of this pull.

maddyscientist:
As discussed with @mathiaswagner, I have decided to "finish the job" on the HISQ force.  Latest push further optimizes the HISQ force:
* Fuse the outer product and subsequent force computation into a single QUDA offload function (`computeHISQForceQuda)
* This reduces the memory traffic between CPU and GPU between these two calls and at the same time simplifies the surrounding code in MILC that calls these routines.

This will require coordination on the MILC end from @detar before we merge.

mathiaswagner:
Do we have coordination with MILC for this? I don't see a PR in the milc repository.

detar:
Hi Mathias,

What is this about?

Thanks,

Carleton

PR #683, opened 2018-03-09
Support for quarter precision coarse halo exchanges

weinbe2:
This commit includes the necessary changes to enable quarter precision halo exchanges for the coarse operator. This is a rather large addition which lays the groundwork for using quarter precision any/everywhere within QUDA. In particular, this includes copying between quarter and all other precisions, and updating the accessor routines to support quarter precision. The interface exposing reduced precision halos has not yet been implemented (@maddyscientist is planning on implementing them). I have intentionally left the preprocessor defines at the top of "dslash_coarse.cu" that I used while debugging to indicate the changes that will need to be made.

Other additions that have come along for the ride:

- A new test, "copy_test.cpp", which is a stripped down version of "blas_test.cpp" which only checks the copy routines.
- An implementation of cudaColorSpinorField::PrintVector, analogous to the cpuColorSpinorField version.
- Support for quarter precision BLAS routines and tests thereof in "blas_test.cpp". (Mixed-precision should be a trivial addition.)

Adding quarter precision more ubiquitously throughout QUDA, such as all of the necessary pieces for quarter precision sloppy solves, is left for future work.


mathiaswagner:

mathiaswagner:

mathiaswagner:

maddyscientist:
Ok, I have verified that this is working well for Chroma HMC.  On the 16^3x48 test on my workstation there is no difference in the convergence when using 8-bit halos versus 16-bit halos.

PR #688, opened 2018-04-24
Fix bug introduced in dd7621f

maddyscientist:
* Fixes the recent Fermi patch which while fixing Fermi, managed to break every other GPU architecture

@mathiaswagner this is likely the cause of issues you were seeing


maddyscientist:
@mathiaswagner silly bug fixed (missing `#ifdef` guard).

PR #690, opened 2018-05-04
Feature/p2p zero copy

maddyscientist:
This pull request is intended to maximize performance on dense-node systems, especially those that lack GDR.
* Adds support for p2p to Dslash zero-copy policies (previously these were disabled for these policies)
* Adds support for the packing kernel to directly write to their p2p remotes instead of using copy engines
* In dslash policy autotuner we now have an outer loop over all the policies where we test whether to use copy engines or direct writes
* When we use dslash policies that utilize zero-copy writing to host memory we use grid-size tuning and limit the size of the grid to one quarter of the GPU to ensure better overlap of any interior compute kernels


mathiaswagner:
Is this ready for review now?

maddyscientist:
Not yet.  I'm still testing it, and I want to disable the remote p2p writes with an envarg (for future performance and debugging).  I'm thinking of extending the meaning of `QUDA_ENABLE_P2P` from the present binary enable/disable to a 2-bit scheme
```
QUDA_ENABLE_P2P=0 # disable all p2p transfers
QUDA_ENABLE_P2P=1 # enable only copy engines
QUDA_ENABLE_P2P=2 # enable only remote writing
QUDA_ENABLE_P2P=3 # enable both copy engines and remote writing (e.g., 3 = 2+1)
```
Thoughts?

mathiaswagner:
I think the env var is good. If I do find some cycles I will check out that branch and try to run some tests myself.

PR #695, opened 2018-05-24
Feature/p2p write optimize

maddyscientist:
This pull request improves performance of the dslash packer when writing halos to remote peer-to-peer GPUs using the zero-copy policies.  For large problem sizes, the prior implementation suffered from the fact that all active threads would be working on the same direction simultaneously, resulting in poor NVLink saturation.  This pull request fixes that problem by distributing the active thread blocks throughout the problem set, ensuring that we simultaneously utilize all available off-chip connections.

This is enabled by default (undefining the macro `STRIPED` in dslash_pack.cu will revert to prior behaviour).


mathiaswagner:
Any idea why the builds all fail?

maddyscientist:
It is because Ubuntu updated its compiler to GCC 5.5, and this breaks CUDA :sad:  

Can you redirect Jenkins to use  GCC 6 and/or GCC 4?

mathiaswagner:
Urgh. Reminds me to work on containerized build environments for Jenkins again. 
I'll look into this / switching out GCC in the next days.

PR #698, opened 2018-06-01
Feature/eigensolver

cpviolator:
Interface with ARPACK stand alone eigensolver tested with double and single prec for wilson. MG comparison checks tested in single prec with wilson. Code is in dire need of templetisation and consolidation, but seems to be working OK. For larger machines, we'll need to all some extra OMP and LAPACK flags to the ARPACK CMake flags to enable OMP acceleration across multiple MPI nodes. An SVD routine can easily be written by supplying MdagM to the eigensolver for the right singular vectors and generating left singular vectors by M multiplication. Looking at null vector overlap with the low SVD space might be illuminating.


mathiaswagner:

weinbe2:
This is really nice work, @cpviolator. I may not be able to get to a more thorough review before Monday. A few quick suggested additions/thoughts:
* You should add support for half/quarter precision. This is an easy addition: if the precision of the gauge field/spinor field is half/quarter, you explicitly make the cpu fields single precision, use the single precision arpack routines, and the ColorSpinorField copy routines do the conversion work for you. The only thing you need to be careful about is not setting the tolerance that tightly, especially for quarter precision. (This actually leads to interesting tests: you can make some perturbative arguments that reduced precision really messes up the low eigenvalues, so this would be a way to test it explicitly.)
* There should be an analogous `staggered_arpack_host_test` for staggered operators. I'd actually like to try my hand at doing this early next week, both for the experience, and also to see first-hand if there would be any useful small tweaks to the interface.
* At some point, you should add a page to the wiki explaining how to use the test routine and some heuristics for the parameters you expose.

All this aside, I can't underscore how useful this contribution will be. Thank you for working on this!

cpviolator:
Thanks @weinbe2. I'll also add some load/save functions for the eigenvectors. HALF and SINGLE prec is proving tricky. I'm getting a hang in the function

```C++
deviation = sqrt( xmyNorm(*gpuTemp, *tmp2) / norm2(*gpuTemp) );
```

where gpuTemp is a `HALF/QUARTER` prec spinor, and tmp2's precision is hardcoded to be the same as `prec_sloppy` as it's params are the same as the null vector's at the start of the `verify()` routine. @maddyscientist does this hang ring any bells for you?

For now, I've hardcoded single prec ARPACK and direct solves only.

weinbe2:
A comment and a proposed feature I forgot about before:
* If you try to generate eigenvectors as near-null vectors for the free field case, it'll fail/be very ill-conditioned (you'll get prolongators/restrictors that aren't full rank) if you request more than 6 = (Nc = 3)*(Nspin = 4)/(Nchiral/coarse spin = 2 eigenvectors. This is because of the automatic chiral doubling that happens: you'll end up with an overcomplete basis, and after the Gram-Schmidt you'll get (again, up to tolerance) ill-conditioned prolongators. I created a way to generate the exact low-mode near-null vectors which is aware of this and creates near-null vectors appropriately. You may want to put in a heuristic to check for the free field case and complain---if, say, the lowest two eigenvalues are "exactly" degenerate and purely real, you're probably in the free field case unless some cosmic coincidence happens.
* You should expose the option the use only the eigenvectors corresponding to eigenvalues which have a zero or positive [arbitrarily] imaginary part. Due to preserving gamma_5 hermiticity, you'll get the negative imaginary eigenvalues for free (though I have a remark in the staggered MG paper that this doesn't mean you get the negative eigenvectors for free). This tends to work just as well, if not better, then grabbing an equal number of positive and negative eigenvalues.

cpviolator:
@alexstrel Will do. Right now I'm wrestling with with polynomial operation, but I will write the simple usage instructions today.

cpviolator:
@alexstrel Let me know if I missed anything: https://github.com/lattice/quda/wiki/ARPACK-Driven-Eigensolver

cpviolator:
Chebyshev acceleration now works, the wiki has been updated with a tutorial on its use. Once the GPU port is complete this pull should be ready for a closer review.

CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
 # add a directory for cmake modules
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
-
+set(CMAKE_CXX_COMPILER g++-6)
```
mathiaswagner: Don't do that here. It breaks build tests. Please use `CC=gcc-6 CXX=g++-6 cmake (all other options)`

CMakeLists.txt
```diff
@@ -541,10 +558,16 @@ enable_language(C)
 #define FORTRAN FLAGS
 set(CMAKE_F_FLAGS -std=c99 CACHE STRING "Fortran Flags")
 
+enable_language(Fortran)
```
mathiaswagner: Can this move back in the if below?

CMakeLists.txt
```diff
@@ -541,10 +558,16 @@ enable_language(C)
...
+if(QUDA_MPI OR QUDA_QMP)
+  set(CMAKE_Fortran_COMPILER mpifort)
```
mathiaswagner: again don't set compilers here

cpviolator: The problem I was having here (related to the above comment too) was that when I performed a QMP or MPI build, gfortran was selected as the fortran compiler. Changing this forced me to delete the cache and start over:

```
You have changed variables that require your cache to be deleted.
 Configure will be re-run and you may have to reset some variables.
 The following variables have changed:
 CMAKE_Fortran_COMPILER=mpifort
```

Is there any way one can make CMake recognize a multi GPU build and select a default MPI fortran compiler?

mathiaswagner:
Can we safely change the base to develop now that feature/multigrid is fully merged into develop?

cpviolator:
Sure thing.

maddyscientist:
Superseded by #765 so closing this.

PR #700, opened 2018-06-11
Fix minor dslash tuning bug causing unnecessary double tuning 

maddyscientist:
`fillAux` must be called after `dslashParam.ghostDim` is set in DslashCuda constructor.


maddyscientist:
Pushed a minor fix to CUDA API wrapper
* missing d2d async copy wrapper
* make some arguments pass by reference instead of pass by value

PR #704, opened 2018-06-26
[DONOTMERGE] Add a QudaEnv singleton class to manage all the environment variables in one place

mathiaswagner:
Motivation is:
* have a central place for all environment variable
* allow to control the features from within code (e.g. set different dslash policies for the unit tests) without needing to hack that with run scripts

This is not ready for merge but I put it out for suggestions


mathiaswagner:
Closing that but leaving the branch there if we want to pick up on that again.

PR #705, opened 2018-06-26
Enable ctest functionality to allow automated run time tests

mathiaswagner:
This pull adds 
* more fine grained tests based on gtest for the dslash and blas tests
* it supports running them through ctest
* fixes a build issue with cmake 3.11
* updates gtest
* fixes a bug in packing for staggered fermions


mathiaswagner:
I guess that looks good now.
* ndeg_twisted_mass fails for some of the tests but I think this is not related to the changes in this pull and only was discovered by having better test coverage and should be handled separately.
* there are a lot of ideas to improve the tests but I think this is also not for 0.9.0

mathiaswagner:
latest version passed the test on sm_70.

PR #706, opened 2018-06-28
Saving gauge fields + standalone heatbath

weinbe2:
This pull request adds in routines to save gauge fields in single or double position. It additionally adds a standalone heatbath executable in the test directory which can optionally load and, more importantly, save generated gauge fields. 

The following heatbath parameters are exposed on the command line: the value of beta; the number of thermalization and measurement (plaquette+topology) steps; the number of heatbath and overrelaxation hits per step; and whether to use a cold or hot start, or load from a file.

I'll work on documenting the new flags on the wiki momentarily.

### DIFF
--- a/include/qio_field.h
+++ b/include/qio_field.h
@@ -4,12 +4,19 @@
 #ifdef HAVE_QIO
 void read_gauge_field(const char *filename, void *gauge[], QudaPrecision prec, const int *X,
 		      int argc, char *argv[]);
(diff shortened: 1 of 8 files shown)


mathiaswagner:

mathiaswagner:
@jenkins add to whitelist
@jenkins ok to test

maddyscientist:
@mathiaswagner Jenkins failed here since it looks to be still configured for Cmake 3.7, but we now need 3.8.

PR #710, opened 2018-07-09
Enable testing of GDR policies

mathiaswagner:
if QUDA_ENABLE_GDR is set when executing cmake.

Note that policy-tuning already tuned for GDR policies but explicit testing of GDR policies was not included. This pull request fixes that.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -148,11 +148,10 @@ add_test(NAME blas_test COMMAND blas_test --sdim 16 --tdim 16 --gtest_output=xml
 
 # loop over Dslash policies
 if(QUDA_CTEST_SEP_DSLASH_POLICIES)
+  set(DSLASH_POLICIES 0 1 6 7 8 9 10 11 -1)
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -42,9 +42,9 @@ if(QUDA_DIRAC_WILSON OR QUDA_DIRAC_CLOVER OR QUDA_DIRAC_TWISTED_MASS OR QUDA_DIR
 endif()
 
 if(QUDA_DIRAC_WILSON OR QUDA_DIRAC_CLOVER OR QUDA_DIRAC_TWISTED_MASS OR QUDA_DIRAC_TWISTED_CLOVER OR QUDA_DIRAC_DOMAIN_WALL OR QUDA_DIRAC_STAGGERED)
```
maddyscientist: not related to this pull specifically, but I guess this if statement should also include non-degenerate twisted mass as well.

mathiaswagner: good catch. That has gone unnoticed for quite some time now. Fixed it now.

PR #711, opened 2018-07-09
Hisq stencil unit test

weinbe2:
This PR creates a new unit test, `hisq_stencil_test`, which constructs the full HISQ stencil on both the GPU and the CPU, supporting both double and single precision. Unlike other unit tests considering just fat/long link construction and link unitarization, it includes the exact HISQ coefficients used for each step of HISQ smearing. 

As part of writing the unit test, I added a new command line parameter, `--epsilon-naik`, which takes a single real number which gets used as the non-relativistic correction parameter "epsilon" described in the original HISQ paper. When "epsilon" is non-zero, there are three steps to building the HISQ stencil: (1) the W field construction (fat7+unitarization), (2) constructing the epsilon corrections, and (3) constructing+adding in the X and long link fields. The second step can be skipped if "epsilon" is zero (the default), which the code will correctly do, including avoiding allocating any memory for the step.

As a remark, the GPU part of the test directly uses the MILC routines, which copy to and from the GPU for each step. Keeping it on the GPU the entire time is a straightforward extension (and would be a quick, small optimization win), but it's not on my critical path at the moment so I'm skipping it for now.

This PR doesn't address creating the fat and long links from a gauge field loaded from a file, nor does it update `staggered_dslash_test` and `staggered_invert_test` to use correctly generated fat and long links. For that reason, please leave this branch open after merging it in! Those features are next on my critical path.

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -3760,12 +3760,7 @@ void computeKSLinkQuda(void* fatlink, void* longlink, void* ulink, void* inlink,
   gParam.create = QUDA_ZERO_FIELD_CREATE;
   gParam.link_type = QUDA_GENERAL_LINKS;
   gParam.reconstruct = QUDA_RECONSTRUCT_NO;
-<<<<<<< HEAD
-  gParam.order = QUDA_FLOAT2_GAUGE_ORDER;
-  gParam.setPrecision(param->cuda_prec);
-=======
   gParam.setPrecision(param->cuda_prec, true);
->>>>>>> feature/multigrid
   gParam.ghostExchange = QUDA_GHOST_EXCHANGE_NO;
   cudaGaugeField *cudaFatLink = new cudaGaugeField(gParam);
   cudaGaugeField *cudaUnitarizedLink = ulink ? new cudaGaugeField(gParam) : nullptr;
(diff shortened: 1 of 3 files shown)


maddyscientist:
@weinbe2 As part of closing this pull, can you also update the wiki describing this?  I believe we have a conversation on slack that would be close to copy and paste to fulfill this....

weinbe2:
@maddyscientist On it!

mathiaswagner:
Can we get this to make proper use of gtest / ctest like the (staggered_)dslash_ctest ?

mathiaswagner:
Why does this go to feature/multigrid and not develop?

maddyscientist:
@mathiaswagner Well this is new stuff, so I don't think it needs to go in develop.  I think we've essentially frozen develop haven't we?

@weinbe2 Thoughts on adding ctest support?  If this is too much of a distraction since I know you need to get on with the staggered multigrid stuff, then I think we can skip that for now.

weinbe2:
@maddyscientist @mathiaswagner I can get it into `ctest` sort of for free once I get to putting correct long link/fat link construction into `staggered_dslash_(c)test`, which is the next step. Does that sound like a good compromise?

mathiaswagner:
Sounds good.

PR #717, opened 2018-08-17
HISQ stencils plus general staggered features/cleanup

weinbe2:
This PR contains a comprehensive update to the test files `staggered_dslash_ctest.cpp`, `staggered_dslash_test.cpp`, and `staggered_invert_test.cpp`, and additionally adds proper tadpole coefficient support to building the HISQ stencil, both in `hisq_stencil_test.cpp` and elsewhere.

All tests now support loading an external gauge field (when QIO support is enabled), properly applying phases, and properly constructing the fat/long links in the case of the HISQ operator. This is automatically enabled when passing in a gauge field using the existing flag `--load-gauge [filename]`. 

In the case of the HISQ stencil, correct fat/long links are constructed by passing in the flag `--compute-fat-long true` (default false). The tadpole coefficient can be passed in via the flag `--tadpole-coeff [value]`, defaulting to a value of 1.0. In the case of charm (heavy) quarks, the non-perturbative correction "epsilon" can be passed in with the flag `--epsilon-naik [value]`, default zero. When a non-zero epsilon is passed in, `hisq_stencil_test.cpp` constructs and tests both the original fat/long links and the corrected links. The three other tests will only construct and test the non-perturbatively corrected fields when a non-zero epsilon is passed in.

In addition, this PR re-enables testing and verifying the full staggered/HISQ/Laplace operator in both the `dslash` and `invert` tests (`--test 2`). 

For various reasons, some combinations of precisions and tests do not work for legitimate reasons--for example, the HISQ fat/long construct doesn't support half precision. These cases are marked across the various test files, where testing certain invalid combinations generally errors out verbosely. One exception to this is in `staggered_dslash_ctest` where the code intelligently sets reconstruct values as appropriate---for example, in the case of the HISQ stencil, reconstruct 12 is converted to no reconstruct for the fat links and reconstruct 13 for the long links (in the case where there's no epsilon correction).

At one point or another I've tested every sane combination of options for multi-GPU builds, and took a "Monte Carlo" approach to testing combinations of options for single-GPU builds.

There's some code commented out related to computing the plaquettes of loaded fields when the code's compiled against gauge field tools/algs; it worked at some point, then at some point it broke. I'd like to figure it out at some point, but in the meantime it hasn't contributed to any unit tests failing and is largely orthogonal to the point of this branch. 

...I think that covers it all.


lib/cuda_gauge_field.cu
```diff
@@ -28,7 +28,7 @@ namespace quda {
 	int minimum_pad = nFace*surfaceCB[i] * (geometry == QUDA_COARSE_GEOMETRY ? 2 : 1);
 	if (pad < minimum_pad) pad_check = false;
 	if (!pad_check)
-	  errorQuda("cudaGaugeField being constructed with insufficient padding (%d < %d)\n", pad, minimum_pad);
+	  errorQuda("cudaGaugeField being constructed with insufficient padding in dim %d (%d < %d)\n", i, pad, minimum_pad);
```
maddyscientist: 👍

maddyscientist:
@weinbe2 This looks good.  Did you mean to include staggered_invert_test.bak?

weinbe2:
@maddyscientist Fixed!

weinbe2:
This shouldn't be merged until https://github.com/lattice/quda/pull/720 has been merged into develop -> feature/multigrid

maddyscientist:
It also can't be merged until the build failures are fixed 😜 (cmake37 failures are expected, cmake311 are not).
```
llfat_reference.cpp:(.text+0x2323): undefined reference to `quda::unitarizeLinksCPU(quda::cpuGaugeField&, quda::cpuGaugeField const&)'
```

weinbe2:
Working on it now :)

maddyscientist:
@Jenkins test this please

PR #719, opened 2018-08-24
Use refinement_sloppy precision for multi shift refinement 

mathiaswagner:
This introduces another precision (refinement_sloppy) to be used as sloppy precision in the refinement step of multi-shift cg. This allows

* using a pure double (or single) precision multi shift with a mixed-precision refinement step
* using a mixed precision multi shift (double-single) and a different mixed-precision refinement (double-half).

Note that for Shift 0 the Krylov space is reused and once we only iterator on shift 0 we can switch to a CG (with a more aggressive mixed-precision).


include/invert_quda.h
```diff
@@ -427,7 +430,10 @@ namespace quda {
     CG(DiracMatrix &mat, DiracMatrix &matSloppy, SolverParam &param, TimeProfile &profile);
     virtual ~CG();
 
-    void operator()(ColorSpinorField &out, ColorSpinorField &in);
+    void operator()(ColorSpinorField &out, ColorSpinorField &in){
```
maddyscientist: Doxygen markup needed here for these `operator()` member functions.

Also I'd suggest we rename `pin` to `p_in` (I first read this as "pin" and was trying to workout if the field was somehow "pinned").

lib/inv_cg_quda.cpp
```diff
@@ -324,16 +325,33 @@ namespace quda {
...
       csParam.create = QUDA_COPY_FIELD_CREATE;
-      for (auto &pi : p) pi = ColorSpinorField::Create(rSloppy, csParam);
+      if(pin){
+        for (auto &pi : p) pi = ColorSpinorField::Create(*pin, csParam);        
+      } else {
+        for (auto &pi : p) pi = ColorSpinorField::Create(rSloppy, csParam);
+      }
     } else {
-      for (auto &pi : p) *pi = rSloppy;
+      if(pin){
```
maddyscientist: Why not just write this as?
```c++
for (auto &p_i : p) *p_i = p_in ? *p_in : rSloppy;
```
Or at least why not just use `=` instead of `blas::copy` for the first case?

Also, style police: for consistency  `if( ){` -> `if () {`

lib/inv_multi_cg_quda.cpp
```diff
@@ -179,12 +180,22 @@ namespace quda {
     }
     
     // this is the limit of precision possible
-    const double prec_tol = pow(10.,(-2*(int)b.Precision()+1));
 
-    double *zeta = new double[num_offset];
-    double *zeta_old = new double[num_offset];
-    double *alpha = new double[num_offset];
-    double *beta = new double[num_offset];
+    bool exitearly = false;
```
maddyscientist: `exitearly` -> `exit_early`

lib/inv_multi_cg_quda.cpp
```diff
@@ -200,7 +211,7 @@ namespace quda {
       if (param.tol_offset[j] < param.delta) reliable = true;
 
 
-    cudaColorSpinorField *r = new cudaColorSpinorField(b);
+    auto *r = new cudaColorSpinorField(b);
```
maddyscientist: This reminds me, I need to update the multi-shift solver to use generic `ColorSpinorField` fields (out of scope for this pull).

maddyscientist: But I like the use of auto 👍

lib/inv_multi_cg_quda.cpp
```diff
@@ -235,12 +246,12 @@ namespace quda {
 	x_sloppy[i] = new cudaColorSpinorField(*x[i], csParam);
     }
   
-    std::vector<ColorSpinorField*> p;
+    // std::vector<ColorSpinorField*> p;
```
maddyscientist: Just delete this line?

lib/inv_multi_cg_quda.cpp
```diff
@@ -440,15 +454,21 @@ namespace quda {
...
       // happen to converge when the un-shifted system converges
-      if ( (convergence(r2, stop, num_offset_now) ||  k == param.maxiter) && aux_update == true) {
+      if ( (convergence(r2, stop, num_offset_now) || exitearly|| k == param.maxiter) && aux_update == true) {
 	if (getVerbosity() >= QUDA_VERBOSE) 
 	  printfQuda("Convergence of unshifted system so trigger shiftUpdate\n");
 	
 	// set worker to do all updates at once
 	shift_update.updateNupdate(1);
-	shift_update.apply(0);
+	shift_update.apply(nullptr);
```
maddyscientist: Since the function signature is `void apply(const cudaStream &stream)` I'm not sure if passing a `nullptr` is syntactically what we should be doing.  It works of course, since a `cudaStream_t` is defined to be a pointer
```c++
typedef CUstream_st * cudaStream_t
```
and that definition won't change.  I don't care that strongly, but it seems funny to make this change here where `0` is used everywhere else in QUDA.

maddyscientist:
This looks like great work, and I'm looking forward to seeing this in action.  You can see my various comments in the code, mostly style and documatation.

mathiaswagner:
Thanks for the comments. Good points.

One thing I thought while implementing that is the use shared_ptr for the gauge fields instead of having all the manual pointer comparisons.
But then again I thought to first get this done and we can discuss how to clean that up later.

PR #720, opened 2018-08-28
Hotfix/profile naik epsilon

maddyscientist:
A couple of bugs and minor additions that need to be fixed sooner rather than later
* Fixes long-link reconstruction with HISQ fermions with non-zero Naik epsilon relativistic correction
* In the MILC interface, the environment variable `QUDA_MILC_HISQ_RECONSTRUCT` is now supported by `qudaDslash`, `qudaInvert`, `qudaInvertMsrc` and `qudaEigCGInvert`
* Fixes an issue where some kernels were wrongly being counted as policies in profile_async.tsv instead of kernels in profile.tsv
* I've also updated the QUDA_VERSION for 0.9.1, which is appropriate here since the MILC interface functions have changed (in case there is a desire to maintain compatibility from 0.9.0 to 0.9.1) from within MILC.

This pull request will require an update to MILC (I'll file a pull request for that shortly) since the interface functions need to be changed to accomodate the passing of the naik_epsilon factor.

### DIFF
--- a/include/quda_milc_interface.h
+++ b/include/quda_milc_interface.h
@@ -40,6 +40,8 @@ extern "C" {
     int make_resident_solution; /** Make the solution resident and don't copy back */
     int use_resident_solution; /** Use the resident solution */
     QudaInverterType solver_type; /** Type of solver to use */
+    double tadpole; /** Tadpole improvement factor */
+    double naik_epsilon; /** Naik epsilon parameter (HISQ fermions only).*/
   } QudaInvertArgs_t;
 
   /**
@@ -197,7 +199,6 @@ extern "C" {
    * @param inv_args Struct setting some solver metadata
    * @param milc_fatlink Fat-link field on the host
    * @param milc_longlink Long-link field on the host
-   * @param tadpole Tadpole improvement facter
    * @param source Right-hand side source field
    * @param solution Solution spinor field
    */
@@ -206,7 +207,6 @@ extern "C" {
 		  QudaInvertArgs_t inv_args,
 		  const void* const milc_fatlink,
 		  const void* const milc_longlink,
-		  const double tadpole,
 		  void* source,
 		  void* solution,
 		  int* num_iters);
@@ -264,7 +264,6 @@ extern "C" {
    * @param target_relative_residual Target Fermilab residual
    * @param milc_fatlink Fat-link field on the host
    * @param milc_longlink Long-link field on the host
-   * @param tadpole Tadpole improvement facter
    * @param source Right-hand side source field
    * @param solution Solution spinor field
    * @param final_residual True residual
@@ -279,7 +278,6 @@ extern "C" {
 		  double target_fermilab_residual,
 		  const void* const milc_fatlink,
 		  const void* const milc_longlink,
-		  const double tadpole,
 		  void* source,
 		  void* solution,
 		  double* const final_resid,
@@ -300,7 +298,6 @@ extern "C" {
    * @param target_relative_residual Target Fermilab residual
    * @param milc_fatlink Fat-link field on the host
    * @param milc_longlink Long-link field on the host
-   * @param tadpole Tadpole improvement facter
    * @param source array of right-hand side source fields
    * @param solution array of solution spinor fields
    * @param final_residual True residual
@@ -316,7 +313,6 @@ extern "C" {
                       double target_fermilab_residual,
                       const void* const fatlink,
                       const void* const longlink,
-                      const double tadpole,
                       void** sourceArray,
                       void** solutionArray,
                       double* const final_residual,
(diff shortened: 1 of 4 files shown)


maddyscientist:
MILC PR is here: https://github.com/milc-qcd/milc_qcd/pull/21

@detar, @stevengottlieb

detar:
Hi Kate,

Thank you for the changes.  I have merged your changes.  Just curious -- 
what was the previous default setting for three-link reconstruction?  
Fortunately, we have not used QUDA for quarks that were heavy enough to 
require a nonzero Naik epsilon, but we are preparing to do so, so the 
change is timely.  The calculations we have done with heavy HISQ quarks 
were not done on GPUs.

Best,

Carleton

include/quda_milc_interface.h
```diff
@@ -40,6 +40,8 @@ extern "C" {
     int make_resident_solution; /** Make the solution resident and don't copy back */
     int use_resident_solution; /** Use the resident solution */
     QudaInverterType solver_type; /** Type of solver to use */
+    double tadpole; /** Tadpole improvement factor */
```
weinbe2: I'd add a comment saying that for HISQ fermions in particular this should be set to 1.0 b/c the tadpole coefficient for the original gauge fields gets baked into the fat/long link construct.

maddyscientist:
@detar The default is to use no reconstruction, so it would have worked fine by default previously. What broke was if the environment variable `QUDA_MILC_HISQ_RECONSTRUCT=13` was set, then the long link would pick up the wrong scaling. It now works as it should regardless of naik-epsilon value.

@weinbe2 I'll add your suggested comment

mathiaswagner:
One idea I had in mind when doing reconstruct and mixed precision is to only use reconstruct on the sloppy long links. Performance impact should be negligible but any issue with reconstruction should be correct through the not reconstructed precise links.

Of course if we don't use a mixed precision solver we could still define mixed precision as double_r18/ double_r13(9) or so but currently QUDA is not set up for this. But I'd suggest that we use mixed precision as double_r18 / (single|half)_r(13|9).

maddyscientist:
Yeh, mixing reconstructions but not precisions would be a way to ensure correct convergence.  A counterpoint why we might not always want to do that is memory footprint.  For example, APEX Large fits on 32 GPUs with recon-13 used for all solvers throughout, however, if we use recon-18 at all in any of the outer double precision solves then it no longers fits without disabling the pool memory manager.

Anyway, I think QUDA is already to support mixing reconstruct types but not precisions.  E.g. in `loadGaugeQuda` you can see that both precision and reconstruct are queried to determine whether to create a unique sloppy or preconditioner gauge field.  E.g.,
```c++
  // switch the parameters for creating the mirror sloppy cuda gauge field                                                                                                                                                                                                      
  gauge_param.reconstruct = param->reconstruct_sloppy;
  gauge_param.setPrecision(param->cuda_prec_sloppy, true);
  cudaGaugeField *sloppy = NULL;
  if (param->cuda_prec != param->cuda_prec_sloppy ||
      param->reconstruct != param->reconstruct_sloppy) {
    sloppy = new cudaGaugeField(gauge_param);
    sloppy->copy(*precise);
  } else {
    sloppy = precise;
  }
```

mathiaswagner:
Agree about the memory footprint. Story of my work in Bielefeld.

I am not sure it is a convincing argument for not making no_reconstruct the default on the precise links and think about another way to opt-in to more aggressive memory savings.

About mixing reconstructs with the same precision: You are right about the gauge fields (came to my mind while making my coffee). The other point would be the trigger of reliable updates in the solvers in that case but the more I think about it I guess we are good also in that respect.

maddyscientist:
I don’t follow @mathiasw: the default is to have `QUDA_RECONSTRUCT_NO` for the outer links and you have to opt in for reconstruct (whether it be 9 or 13).  There isn’t a mechanism to opt in to different reconstructs for the sloppy but that’s outside the scope of this big fix.

Also, reconstruct-13 (when working properly) will not introduce any precision loss - unlike reconstruct-9.  I would suggest that once we get reconstruct-9 working properly, then we need to be opt in to different inner and outer reconstructs.

maddyscientist:
Worth noting that this setting for reconstruct = reconstruct_sloppy = long_reconstruct isn’t introduced in this pull request.  It was already there, just not uniformly applied to all solver interfaces.  This pull applies it to all.

I could quickly add a new envarg that allows the user to set a sloppy reconstruct separately, and if not, default to the outer one.

One thing: how would this work with alternative reliable updates?

mathiaswagner:
Agree. We already had reconstruct, just not everywhere. 

Maybe `QUDA_MILC_HISQ_RECONSTRUCT_SLOPPY` is the way to go?
Full control.

Alternative reliable updates need some testing. My guess is as they do not care about reconstruct and do all error estimates just based on precision it should be ok for a double_r18/double_r9 solve. The correction step might come a tad late but as the error in reconstruct should not be huge and it is anyway just an error estimate ...

 As they are not the default choice I'd suggest we worry about that later.

maddyscientist:
Ok, change is pushed and tested with su3_rhmd_hisq.  Sloppy and outer reconstructs can be set independently, and if `QUDA_MILC_HISQ_RECONSTRUCT_SLOPPY` is unset, it will default to the outer reconstruct type.

weinbe2:
Pending the build checks completing correctly, everything looks good to me!

PR #721, opened 2018-08-28
Move to C++14 and cmake 3.8

mathiaswagner:
This PR switches the required C++ Standard to C++14.

Note that `integer_sequence.hpp` currently is only minimally modified to allow compilation. Guess it can probably be removed with the use of C++14.

Some smaller things:
* It does some example clean-up in interface_quda.cpp using clang-tidy.
* It also does remove some of the pre 3.8 cmake code.


maddyscientist:
Yes, you can just remove integer_sequence.hpp now.

mathiaswagner:
If you build on the wrong machine. Should be fine now.

weinbe2:
I hit zero issues building and running, CUDA 9.2, g++ 5.4, sm_70, CMake 3.9. `multigrid_evolve_test` with a gauge field loaded via QIO worked like a charm.

mathiaswagner:
Good point. Will do.

mathiaswagner:
@Jenkins test this please

mathiaswagner:
We can now switch the standard using QUDA_CXX_STANDARD to be either 11 or 14. 
Should give us an easy switch to CXX 14 once there is nothing blocking us.

PR #722, opened 2018-08-30
Hotfix/gcc 5.5

maddyscientist:
This pull request fixes compilation with GCC 5.5 with CUDA 9.x.  There is an incompatibility due to AVX header changes in GCC 5.5 versus prior versions of GCC 5.x (this does not affect GCC 6.x or 7.x at all).  The only place these headers are included is when the `<algorithm>` header is included in a CUDA source code file - hence this PR removes all use of `<algorithm>` from such files
* For `std::swap` we can get this with `<utility>` instead of `<algorithm>`
* Replace `std::stable_sort` with `thrust::stable_sort`
* Replace use of `std::max` with hand-coded variant
* gtest.h included `<algorithm>` but doesn't seem to use it at all so I've just commented this inclusion out

With these changes made, we can now compile with GCC 5.5 without issues.

I've also fixed  a couple of other issue:
* Fortran interface was broken in 3820720 (changes made to quda.h not propagated to quda_fortran.F90)
* Remove stray profile stop command

### DIFF
--- a/include/quda_arpack_interface.h
+++ b/include/quda_arpack_interface.h
@@ -7,7 +7,6 @@
 #include <color_spinor_field.h>
 #include <dirac_quda.h>
 #include <vector>
-#include <algorithm>
 
 //#ifdef PRIMME_LIB
 //#include "primme.h"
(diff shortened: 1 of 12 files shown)


maddyscientist:
(closing since this should go into release/0.9.x not develop

PR #723, opened 2018-08-31
Hotfix/qmp rank order

### LINKED ISSUE #712: QIO and QUDA disagree on process topology
When QUDA uses "column" as the default Cartesian rank ordering (`--rank-order col`), QIO assumes the rank order is "row" ordered `--rank-order row`.  This means that I/O gives junk if we partition more than 1 dimension, unless we explicitly pass `--rank-order row` to QUDA.  We need to set QIO's rank order to match QUDA's.

maddyscientist:
Fixes issue when using QMP with QUDA's internal tests.  Issue was that logical topology was declared using `QMP_declare_logical_topology` which is ignorant of the ordering QUDA is using.  Instead use `QMP_declare_logical_topology_map` and pass in the appropriate map.

Closes #712

### DIFF
--- a/lib/dslash_policy.cuh
+++ b/lib/dslash_policy.cuh
@@ -296,7 +296,7 @@ namespace {
 	if (!gdr_recv && !zero_copy_recv) { // Issue CPU->GPU copy if not GDR
 
 	  if (async) {
-#if (CUDA_VERSION >= 8000)
+#if (CUDA_VERSION >= 8000) && 0
 	    // this will trigger the copy asynchronously
 	    *((volatile cuuint32_t*)(commsEnd_h+2*dim+dir2)) = 1;
 #else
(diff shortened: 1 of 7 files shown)


weinbe2:
This looks good to me. @maddyscientist, did you prevent compilation of the async policies because GDR isn't working at the moment?

maddyscientist:
I disabled async policies from being compiled, but they were never included in the policy tuning.  @detar reported linkage issues at FNAL with the async policies (which I couldn't reproduce), so I just `#if 0....endif`-ed them for now.

weinbe2:
Sounds good! I haven't had a chance to link this against FUEL yet (I need to set up a build), would you like me to get that all spun up and tested before we merge this, or should I keep that an orthogonal effort?

maddyscientist:
Orthogonal

weinbe2:
@maddyscientist is there any reason you want to keep hotfix/qmp-rank-order alive, or should I delete it?

PR #724, opened 2018-09-04
Update Eigen to 3.3.5

mathiaswagner:
Just to make sure we pick up bugfixes in Eigen. See http://eigen.tuxfamily.org/index.php?title=ChangeLog#Eigen_3.3.5

Might also help with xlc compatibility


maddyscientist:
This definitely fixes this the original culprit with XLC as I see my bug in the release notes, though whether it fixes the issue completely I don't know.....Regardless, pull is all good.

PR #725, opened 2018-09-05
Feature/jitify

### LINKED ISSUE #362: Employ nvrtc for run-time compilation of QUDA kernels
As of CUDA 7.0, run-time compilation is possible using the nvrtc library.  If we rewrite QUDA to use nvrtc, then the benefits are significant:
- Massive reduction in up-front compilation time, with all kernels only compiled on demand.  This will also have the benefit of reducing the memory resources needed for compilation.  This will significantly aid development as well as deployment, where currently tricks must be played in order to compile on systems with weak front ends.
- Improved code generation: all kernels that use nvrtc can easily use the `__launch_bounds__` qualified since the block size can be made a compile-time variable.  Experiments have shown this can improve performance of many kernels, especially double-precision ones (e.g., on a K40 DP plaquette performance goes from 300 GFLOPS to 350 GFLOPS).
- Enables compile-time autotuning: as well as the block size as mentioned above, compile-time autotuning could also include the use of texture or not, data ordering, cache intrinsics, etc.  Where CUB is deployed, CUB's policies can be swept over on a per kernel per unique parameter set.
- Reduced CPU overhead: the use of nvrtc requires the use of the CUDA driver API, which while more verbose and awkward to use, comes with the benefit of reduce CPU overhead.  This will improve strong scaling where we can be CPU bound.

The effort in deploying this is significant, though I have rough idea of how to do this.  There are multiple obstacles that have to be overcome:
- This is going to be a very intrusive change to QUDA, so to begin with we should only apply this to kernels where it will have the most impact.  That was we can refine the approach until we have something fairly clean that can be applied library wide.
- It would be nice to retain a non-nvrtc mode, which would allow QUDA to be compiled on on pre CUDA 7.0 machines.  This is not a requirement: the large Cray machines will be upgraded to CUDA 7.0, so compatibility should not be a problem.
- We need to cache the results of the compilation, so that compilation is not repeated over and over again when a kernel is returned to.  I think this would be easy to do by adding the compiled kernel pieces `CUmodule` and `CUfunction` to each instance of `std::map<TuneKey, TuneParam>`.  So when we come across a kernel that has been tuned, as well as looking up the launch parameters we look up the kernel itself.  When doing the auto tuning, for each block size we have to compile a different kernel, and only cache the best one.

The major issue that has prevented me from pursuing this prior to now has been the lack of support for compiling a templated kernel in nvrtc.  While templates within the kernel are supported, template parameters to the kernel are not supported.  Moreover, once compiled the CUDA driver API has to be able to look up the name of the kernel; if the kernel is a mangled C++ name, this can be very hard to do.  The solution here is to change the templated kernel, e.g.

``` c++
template<class T>
__global__ void testKernel(T *data)
{
  // do something
}
```

into the following

``` c++
template<class T>
__device__ void testKernel(T *data)
{
  // do something
}

extern "C" __global__ void testNew(TYPE *data) {  testKernel<TYPE>(data); }
```

and pass the compile flag `-DTYPE=type` to nvrtc to set the templated type, where `type` could be `int`, `float` etc.  While this allows us to have have effectively dynamic templates, it requires a lot of boilerplate code in the host code, since we would end up with statements like the following when compiling the kernel

``` c++
template <typename T>
void compileToPTX(const char *fname, const char *source) 
{
  const int n_flags = 1;
  char flags[n_flags][128];
  if (typeid(T) == typeid(float) {
    sprintf(flags[0], "-DTYPE=float");
  } else if (typeid(T) == typeid(int)) {
    sprintf(flags[0], "-DTYPE=int");
  }

  // compile to PTX                                                                                                                                                                                                                                                                  
  nvrtcProgram prog;
  nvrtcCreateProgram(&prog, source, fname, 0, NULL, NULL));
  nvrtcResult res = nvrtcCompileProgram(prog, n_flags, (const char**)flags);
}
```

In the above example this is manageable since we have simple types.  If, however, the argument to the kernel is some templated object then we would require a ridiculous amount of boilerplate building up the string to pass to nvrtc.

The simple work around is to employ `type_info::name` to build the string to pass to the preprocessor.  There's a gotcha here though, that being that this returns the C++ mangled name of the type and not the human-readable type that we need to be able to compile.  Fortunately, we can employ the `cxxabi` which provides a portable method to accessing the demangled name.

``` c++
#include <cxxabi.h> // for compiling with clang on Ubuntu, the libc++abi-dev package is required
#include <string>
#include <typeinfo>
std::string demangle(const char* mangled)
{
  using std;
  int status;
  unique_ptr<char[], void (*)(void*)> result(abi::__cxa_demangle(mangled, 0, 0, &status), free);
  return result.get() ? string(result.get()) : "error occurred";
}
```

This makes it very easy to build to compile templates on demand, and this should work for arbitrary templated types.

``` c++
template <typename T>
void compileToPTX(const char *fname, const char *source)
{
  const int n_flags = 1;
  char flags[n_flags][128];
  sprintf(flags[0], "-DTYPE=%s", demangle(typeid(T).name()).c_str()); // magic is here!

  // compile to PTX                                                                                                                                                                                                                                                                 
  nvrtcProgram prog;
  nvrtcCreateProgram(&prog, source, fname, 0, NULL, NULL));
  nvrtcResult res = nvrtcCompileProgram(prog, n_flags, (const char**)flags);
}
```

maddyscientist:
This pull requests adds the ability to use jitify run-time compilation for QUDA kernels.
* Enabled with `QUDA_JITIFY=ON` (default is off).
* Jitify framework is contained in jitify.hpp which is taken directly from https://github.com/NVIDIA/jitify
* For inclusion in QUDA kernels, the jitify_helper.cuh should be used.
* To allow for both offline and run-time compilation, kernels should be defined in separate include files.  I have created the directory include/kernels for this purpose.
* This will significantly reduce the compilation time for various kernels, specifically those that involve a lot of template instantiation.
* Closes #362 

For now I have got block_orthogonalize.cu and gauge_plaq.cu jitify-enabled.  I'll likely add more before we close this pull request.


mathiaswagner:
Without looking at the code: How do we test the jitify path? At least make sure the kernels get compiled?

Should we add tests that just make sure that the kernels are jitted for the the most important instantiations ?

Nothing that should block this great thing but want to prevent hard to reproduce issues. Especially that in the wild the driver version now decides about the jit compiler and covering different driver versions in tests doesn't sound like fun.

maddyscientist:
Good question @mathiaswagner.  For testing jitify-enabled QUDA this will require that we get the automated run-time testing infrastructure up and running since the compilation takes place at run time.  I think with run-time testing, the jitify testing is a trivial extension.

I've just pushed an update: now restrictor.cu is jitify enabled.  This has the nice side effect that now since both block orthogonalization and the restrictor are run-time compiled, the block size (multigrid aggregate size) can now be any value and we don't need to rely on a pre-baked constant in a switch statement from hell.

maddyscientist:
On the testing, the other thing to note is that for the moment, I mainly see jitify QUDA to be a developer productivity tool.  It will stop us going insane (at least the ones not already insane 😜) in compile-debug cycle.  For deploying a package, e.g., on Summit for production, I think the offline route will remain the way to go for now.

mathiaswagner:
Well, to first order we would not need to go through the full optimization processes. Just making sure that the code is valid (i.e. not typos etc). 
Anyway, guess we need to give that some more thought offline.

maddyscientist:
Ok, @mathiaswagner and @weinbe2 I've now got everything in there that I want to for my first pass at this.  I've done a fair bit of testing of this, and all seems like it's working as it should be.

One thing that should be noted, when running it, you have to set any compiler options using the `JITIFY_OPTIONS` envarg.  There is an example `jitify.sh` script included showing this.

Also, I've noticed that libnvrtc gets confused if there are older versions of CUDA ahead in `LD_LIBRARY_PATH` or set in ld.conf.  So always make sure that the version you want to run with is foremost in your `LD_LIBRARY_PATH`.

mathiaswagner:
@Jenkins test this please

mathiaswagner:
Just ran into:

```
[100%] Built target su3_test
[100%] Built target pack_test
[100%] Built target covdev_test
[100%] Built target multigrid_benchmark_test
[100%] Built target deflated_invert_test
[100%] Built target invert_test
[100%] Built target multigrid_invert_test
[100%] Built target copy_test
[100%] Built target dslash_ctest
[100%] Built target dslash_test
[100%] Built target blas_test
[100%] Built target staggered_invert_test
[100%] Built target staggered_dslash_test
[100%] Built target staggered_dslash_ctest
mathiasw@hpcwur:~/qudabuild/jitify/build$
mathiasw@hpcwur:~/qudabuild/jitify/build$ ./tests/multigrid_benchmark_test
Rank order is column major (t running fastest)
running the following test:
S_dimension T_dimension Nspin Ncolor
  8 /  8 /   8     8      0     0
Grid partition info:     X  Y  Z  T
                         0  0  0  0
QUDA 0.9.1 (git v0.9.0-645-g4c1f78c-sm_70)
CUDA Driver version = 10000
CUDA Runtime version = 10000
Found device 0: Quadro GV100
Found device 1: Quadro GV100
Using device 0: Quadro GV100
WARNING: Data reordering done on GPU (set with QUDA_REORDER_LOCATION=GPU/CPU)
WARNING: Environment variable QUDA_RESOURCE_PATH is not set.
WARNING: Caching of tuned parameters will be disabled.

Benchmarking single precision with 100 iterations...

terminate called after throwing an instance of 'std::runtime_error'
  what():  Source not found: kernels/copy_gauge.cuh
Aborted (core dumped)
```

Long story short: How do we handle the kernels dir? Can we set is using an env-variable? Shall we think about ways to have cmake handle that (also for installs ?)

Guess we also need an error message that provides some guidance.

maddyscientist:
@mathiaswagner my plan for dealing with the verbosity was to extend jitify to allow for run-time specification of verbosity, together with the current compile-time option.  If this wasn't the case, then I'd say the cmake specification of verbosity would be fine, but since this will fixed, e.g., `QUDA_VERBOSE` will enable `JITIFY_PRINT_LAUNCH` and `QUDA_DEBUG_VERBOSE` will enable all the others, I think it doesn't matter at the moment.

maddyscientist:
And if it wasn't clear - the run-time specification of verbosity is something that will happen in my next jitify pull..... 😉

PR #727, opened 2018-09-12
Feature/multigrid

maddyscientist:
This pull request brings all the latest developments for multigrid into the main development branch.  Too many features to list here, but a small subset includes
* A multitude of multigrid optimizations and robustness improvements
* CA-GCR solver
* Support fixed-point precision halos with fine-grain accessors 
* Framework for low-precision halo communication
* 8-bit precision (can be used for halo communication in preconditioner or smoother)
* Memory footprint optimizations
* Multigrid setup now completely offloaded to GPU (e.g., including RNG)
* Optimization of block-orthogonalization into a single kernel


weinbe2:
@maddyscientist at this rate, should we wait for feature/hisq-unit to get merged in? Or should we let that get more of a trial-by-fire outside of develop first?

mathiaswagner:
I think we agree on basically just merging that and fix bugs and issues as we go? There is no way of doing a proper review on that. After all it is going into develop.

maddyscientist:
Yep @mathiaswagner we are in agreement.  I'd just like to have CUDA 8.0 and older arch build tests running first (e.g., on newer cmake) before hitting merge though, since at the moment only CUDA 9.2 and Pascal/Volta are tested.

PR #728, opened 2018-09-13
Hotfix/multishift

mathiaswagner:
Fixes some issues with sloppy refinement of multi shift and compute_true_res

Also serves as test for Jenkins


lib/inv_multi_cg_quda.cpp
```diff
@@ -497,49 +497,53 @@ namespace quda {
...
-	param.iter_res_offset[i] = sqrt(r2[i]/b2);
-	param.true_res_hq_offset[i] = sqrt(blas::HeavyQuarkResidualNorm(*x[i], *r).z);
-      } else
-      {
-        param.iter_res_offset[i] = sqrt(r2[i]/b2);
-      }
+      reliable ? y[1] : (tmp2.Precision() == x[0]->Precision() && &tmp1 != tmp2_p) ? tmp2_p : ColorSpinorField::Create(csParam);
+
+      for (int i = 0; i < num_offset; i++) {
+        // only calculate true residual if we do not need to
```
maddyscientist: I suspect we shouldn't only be computing the true residual if we don't need to 😉

lib/inv_multi_cg_quda.cpp
```diff
@@ -509,11 +509,9 @@ namespace quda {
 
       for (int i = 0; i < num_offset; i++) {
         // only calculate true residual if we need to:
-        // 1.) for higher shifts if requested tolerance is larger than
-        // the precision of the sloppy solve, i.e. we might not need
-        // to refine
-        // 2.) we did not exit early and shift is 0
-        if ( (i>0 and not mixed) or (i == 0 and not exit_early) ) {
+        // 1.) For shift 0 if we did not exit early  (we went to the full solution)
```
maddyscientist: @mathiaswagner comments got switched around here to not match the order of the if statement

mathiaswagner: fixed.

maddyscientist:
@mathiaswagner build failure on this pull.  Is that expected?

Also, you fixed the order of comments, but inserted a tab indent 👎  😜

mathiaswagner:
Guess I should not have used the GitHub editor to fix that. I added a few more configs to Jenkins. Seems the clang build fails which is not unexpected.

maddyscientist:
Ok, I can fix some of the warning that are induced on clang, though one of them is coming from something block CG related that I'm not sure what to do.  Looks like block CG is using `solve` rather than `operator()` for its solver interface and this interferes with the `solve` methods that various solvers are using for solving dense linear systems (on the host using eigen).  Any reason why block CG is using this interface?  And can you fix this to make it more consistent with the other solvers?

mathiaswagner:
Yes. I’ll look into that.

maddyscientist:
@mathiaswagner great, thanks.  In the meantime, I think I've fixed the other compiler warnings that clang is complaining about.

lib/block_orthogonalize.cu
```diff
@@ -15,7 +15,10 @@
 #include <launch_kernel.cuh>
 
 #include <jitify_helper.cuh>
+
+#ifdef GPU_MULTIGRID
```
maddyscientist: Just curious: why bother adding `#ifdef GPU_MULTIGRID` around the kernel include files?  They won't be compiled if GPU_MULTIGRID isn't enabled anyway since the kernels won't be instantiated so the additional compilation time should be trivial.  And long term, I hope we can do away with as many `#ifdef` conditional as possible.

PR #729, opened 2018-09-17
fix compilation of multigrid_benchmark_test

mathiaswagner:


mathiaswagner:
That was a hiccup I created with an experiment in the Jenkins configuration. Should be fine. Anyway
@Jenkins test this please

PR #732, opened 2018-09-20
Remove spurious print statements, plus bug fixes and feature additions to test files.

weinbe2:
NOTE: _None_ of these issues effect the correctness of the QUDA library. It just adds spurious print statements and fixes bugs in the test executables.

* Removed remnant debugging print statements when prepare/reconstruct is called for improved staggered fermions.
* Found that eps_naik was commented out in the GPU link build in `staggered_invert_test`. This inconsistency causes an issue if reconstruct 13 is used since the GPU and CPU fields end up with a different scale.
* In the process, I discovered `staggered_invert_test` never actually tests reconstruct 13. Fixed this by adding the behavior that, if reconstruct 12 is specified, reconstruct 13 gets used for the long links. Reconstruct 13 cannot be passed in directly because the code will also try to use it for building fat and long links, and it'll throw an error. `staggered_dslash_(c)test` doesn't have this issue because it explicitly tests reconstruct 8 and 12 for link building, and then correctly checks and uses reconstruct 13 instead for the dslash application.
* Discovered a bug when `--epsilon-naik` is set on the command line but `--compute-fat-long` is not set _and_ reconstruct 13 is used. This gets triggered because `gauge_param.scale` still gets set with the epsilon corrections, but gauge links that use a correct epsilon scaling never actually get created, only links with |det(U)| = 1. This is addressed by catching this scenario and setting `eps_naik` to 1 before the gauge scaling is set. _Remark:_ this is irrelevant for the tadpole correction.

This branch can be deleted after merging.

### DIFF
--- a/tests/staggered_invert_test.cpp
+++ b/tests/staggered_invert_test.cpp
@@ -492,7 +492,7 @@ invert_test(void)
       computeHISQLinksGPU(qdp_fatlink, qdp_longlink,
                           (n_naiks == 2) ? qdp_fatlink_naik_temp : nullptr,
                           (n_naiks == 2) ? qdp_longlink_naik_temp : nullptr,
-                          qdp_inlink, &gaugeParam, act_paths, 0.0 /* eps_naik */);
+                          qdp_inlink, &gaugeParam, act_paths, eps_naik);
 
       
       if (n_naiks == 2) {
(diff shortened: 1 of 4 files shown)


mathiaswagner:
changed to be merged into develop.

tests/staggered_invert_test.cpp
```diff
@@ -607,8 +607,8 @@ invert_test(void)
   if (dslash_type == QUDA_ASQTAD_DSLASH) {
     gaugeParam.type = QUDA_ASQTAD_LONG_LINKS;
     gaugeParam.ga_pad = link_pad;
-    gaugeParam.reconstruct= link_recon;
-    gaugeParam.reconstruct_sloppy = link_recon_sloppy;
+    gaugeParam.reconstruct= (link_recon == QUDA_RECONSTRUCT_12 || link_recon == QUDA_RECONSTRUCT_8) ? QUDA_RECONSTRUCT_13 : link_recon;
```
mathiaswagner: I don't get why we need this. Isn't specifying `--recon 13` on the command line good enough?

weinbe2: I made that choice based on precedence---in `staggered_dslash_(c)test`, you specify (or, for `ctest`, it loops over) reconstruct 18, 12, or 8, and it has a similar conditional when uploading the long links.

Precedence aside, there are other places `link_recon` is used in the code where recon 13 can't be used. For example, it's used to define the reconstruct used on the original links for building the fat/long links, and in that case recon 13 gives you an error that reads something along the lines of "link recon 13 is only supported for staggered long links."

In that regard, it's an unavoidable choice, which is why I went with precedence.

weinbe2:
> changed to be merged into develop.

Thanks for that catch---I forgot the full convention change was merged into `develop` with the `feature/multigrid` merge.

tests/staggered_invert_test.cpp
mathiaswagner: Let's just go with that then.

PR #734, opened 2018-09-27
Hotfix/jitify options

mathiaswagner:
missed a commit here. This weirdly goes into multi shift but has the fix for include/externals already in.


mathiaswagner:
I ran the multigrid benchmark test and that completed fine.

Re quotes: http://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

But probably more a policy/ style question I don’t care about deeply.

maddyscientist:
@mathiaswagner testing this pull, but I get these warnings from cmake.  Looks related to clang additions.
```
 CMake Warning (dev) at lib/CMakeLists.txt:135:
   Syntax Warning in cmake code at column 164

   Argument not separated from preceding token by whitespace.
 This warning is for project developers.  Use -Wno-dev to suppress it.
```

PR #735, opened 2018-10-03
Hotfix/tunecrash

maddyscientist:
* Initialized `Tunable::aux` string with `NUL` in `Tunable` constructor to ensure no undefined behaviour
* Remove some boilerplate of setting compile-type string and location in tuner


mathiaswagner:
Looks good.

PR #742, opened 2018-10-15
Add dumpMultigridQuda interface

maddyscientist:
This pull request adds `dumpMultigridQuda` for dumping the null-space vectors to disk from a pre-existing MG preconditioner.   The output file prefix is defined from the `QudaMultigridParam::vec_outfile` string, and each level is dumped out separately.  Requires QUDA be compiled with QIO/QMP enabled.


mathiaswagner:
@jenkins rebuild this please

PR #748, opened 2018-10-29
Hotfix/p9 milc

maddyscientist:
Some fixes and cleanup in this pull
* When allocating mapped memory, use `cudaAllocHost` instead of `malloc` and `cudaHostRegister` (work around for issue on POWER9 systems)
* Improved test coverage of `unitarizeLinkTest` to use mapped memory matching what the actual interface does
* Fixed bug in link unitarization (only first parity was being unitarized) and some performance optimization
* Fix bug in HISQ force outer product when including a non-zero Naik term in the action
* In MILC interface, when using QMP communication, do not supply a rank-to-coordinate functor, allowing QMP to specify the logical topology
* Remove legacy some `volatile` declarations

### DIFF
--- a/lib/malloc.cpp
+++ b/lib/malloc.cpp
@@ -278,13 +278,22 @@ namespace quda {
   void *mapped_malloc_(const char *func, const char *file, int line, size_t size)
   {
     MemAlloc a(func, file, line);
+
+#if 1
+    void *ptr;
+    cudaError_t err = cudaHostAlloc(&ptr, size, cudaHostRegisterMapped | cudaHostRegisterPortable);
+    if (err != cudaSuccess) {
+      printfQuda("ERROR: cudaHostAlloc failed of size %zu (%s:%d in %s())\n", size, file, line, func);
+      errorQuda("Aborting");
+    }
+#else
     void *ptr = aligned_malloc(a, size);
-    
     cudaError_t err = cudaHostRegister(ptr, a.base_size, cudaHostRegisterMapped);
     if (err != cudaSuccess) {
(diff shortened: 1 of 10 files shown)


lib/malloc.cpp
```diff
@@ -269,6 +269,7 @@ namespace quda {
     return ptr;
   }
 
+#define HOST_ALLOC // this needs to be set presently on P9
```
mathiaswagner: should this be defined by cmake?

maddyscientist: I don't think so.  Making it a cmake parameter means a global rebuild when changing it, where as for debugging purposes you really want to be able to make a local change and quick rebuild.  This isn't something we ever need to change, but I left the old code in there for future testing.

mathiaswagner: Ok. Was not sure whether this should be enabled only on Power.

lib/malloc.cpp
```diff
@@ -363,26 +373,35 @@ namespace quda {
...
+      free(ptr);
     } else if (alloc[MAPPED].count(ptr)) {
+#ifdef HOST_ALLOC
+      cudaError_t err = cudaFreeHost(ptr);
+      if (err != cudaSuccess) {
+	printfQuda("ERROR: Failed to free host memory (%s:%d in %s())\n", file, line, func);
+	errorQuda("Aborting");
+      }
+#else
       cudaError_t err = cudaHostUnregister(ptr);
```
mathiaswagner: do we miss a free(ptr) here if HOST_ALLOC is not defined?

maddyscientist: Good catch.  Will push a fix

Issue #749, opened 2018-10-31
Runtime error on develop branch with DWF invert


cpviolator:
I'm hooking up QLUA with QUDA's DWF inverter. Compiling the latest QUDA develop and running wilson/clover inverter are fine. The invert test for a mobius domain wall action gives the following runtime error.

```
[howarth@scc-k02 tests]$ ./invert_test --sdim 16 --tdim 16 --dslash-type mobius --Lsdim 6
Enabling peer-to-peer copy engine and direct load/store access
Disabling GPU-Direct RDMA access
Rank order is column major (t running fastest)
running the following test:
prec    prec_sloppy   multishift  matpc_type  recon  recon_sloppy S_dimension T_dimension Ls_dimension   dslash_type  normalization
single   single          0        even-even     18     18          16/ 16/ 16      16          6               mobius     kappa
Grid partition info:     X  Y  Z  T
                         0  0  0  0
QUDA 0.9.1 (git v0.9.0-703-g419f7b0-dirty-sm_60)
CUDA Driver version = 9020
CUDA Runtime version = 8000
Found device 0: Tesla P100-PCIE-16GB
Using device 0: Tesla P100-PCIE-16GB
WARNING: Data reordering done on GPU (set with QUDA_REORDER_LOCATION=GPU/CPU)
WARNING: Using device memory pool allocator
WARNING: Using pinned memory pool allocator
Loaded 32 sets of cached parameters from /projectnb/isingmag/howarth/QLUA/quda_build/tests/tunecache.tsv
Computed plaquette is 1.223368e-01 (spatial = 1.224580e-01, temporal = 1.222156e-01)
Tuned block=(32,1,1), grid=(1,1,1), shared=0,  giving 0.00 Gflop/s, 8.62 GB/s for cudaMemcpyHostToDevice with loadSpinorField,cuda_color_spinor_field.cu,599
Tuned block=(64,1,1), shared=0,  giving 0.00 Gflop/s, 492.89 GB/s for N4quda15CopyColorSpinorIfdLi4ELi3ENS_18CopyColorSpinorArgINS_11colorspinor11FloatNOrderIfLi4ELi3ELi4ELb0EEENS2_21SpaceSpinorColorOrderIdLi4ELi3EEEEEEE with out_stride=196608,in_stride=196608,NonRelBasis
[...]

PR #750, opened 2018-10-31
Fixed tuning error in DWF dslash

cpviolator:
Added a constraint to the DWF tuning routines so that that thread hardware limits are obeyed.


cpviolator:
Addresses #749

PR #759, opened 2018-12-10
Feature/tifr dec2018

maddyscientist:
* If the TIFR interface is built, compile staggered fermions with direct support for the TIFR phase convention to allow for native gauge reconstruct support 
* Include the `CUDA_VISIBLE_DEVICES` information in the `comm_topology_string` to ensure that unique policy tunings are found if the device order is changed
* In multi-shift solver, ensure that prec_tol is set to be no looser than the expected precision limit
* Logic clean up of resident solutions in `invertQuda`: only use a resident solution if explicitly requested
* Clean up of host-side ghost field allocation - use seperate arrays for the send and receive buffers
* Clean up of gauge_phase.cu
* Fix bug in staggered dslash when host-mapped ghosts are used in combination with peer-to-peer communication
* Fix link error when compiling staggered fermions without `GAUGE_TOOLS` enabled
* Fix bug in padded color spinor accessor when using full fields

### DIFF
--- a/include/color_spinor.h
+++ b/include/color_spinor.h
@@ -101,6 +101,13 @@ namespace quda {
       */
       __device__ __host__ inline const complex<Float>& operator()(int idx) const { return data[idx]; }
 
+      __device__ __host__ void print() {
+        for (int s=0; s<Ns; s++) {
+          for (int c=0; c<Nc; c++) {
+            printf("s=%d c=%d %e %e\n", s, c, data[s*Nc+c].real(), data[s*Nc+c].imag());
+          }
+        }
+      }
     };
 
   /**
@@ -614,7 +621,7 @@ namespace quda {
       *this = a;
     }
 
-  __device__ __host__ void print() {
+    __device__ __host__ void print() {
       for (int s=0; s<Ns; s++) {
 	for (int c=0; c<Nc; c++) {
 	  printf("s=%d c=%d %e %e\n", s, c, data[s*Nc+c].real(), data[s*Nc+c].imag());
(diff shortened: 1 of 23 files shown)


weinbe2:
Everything looks good after a code review and compile with some extra CMake flags on my own machine. A few suggestions related to adding `CUDA_VISIBLE_DEVICES` info to the topology string:

- When `CUDA_VISIBLE_DEVICES` isn't set, can we assume the devices are in the default order? For ex, on a 3 GPU machine, it's effectively `0,1,2`? If so, should we just add `,order=012` (or the analogous) by default if `CUDA_VISIBLE_DEVICES` is unset?
- In the context of MG, the standard topology string doesn't seem to be appended to coarse spinor packing routines. Should the order be stored independently of and in addition to the topology string so it can be appended to the coarse spinor routines?

To be clear, these suggestions are independent of the correctness of anything in this PR.

maddyscientist:
Regarding @weinbe2's two com,ents:
* The decision to leave the `order` unlabelled when `CUDA_VISIBLE_DEVICES` is unset was intentional, as this provides more debug information (e.g., that the variable wasn't set), compared to assuming the default ordering and labelling it as such.  So I'll leave that as it is.
* I've added the string to the generic color spinor packer.  It was an oversight that `comm_dim_topology_string` wasn't included in the `tuneKey` there.

weinbe2:
@maddyscientist I support your reasoning on leaving `order` unset when `CUDA_VISIBLE_DEVICES` is unset, so all good there.

I verified everything's ship-shape with the generic packing routines. Pending the rest of the checks completing, this gets my sign-off.

PR #761, opened 2018-12-15
Feature/tifr dec2018

maddyscientist:
Some minor fixes and optimizations
* Remove check for valid device ordinal being less than number of devices, since this breaks when `CUDA_VISIBLE_DEVICES` is set to a subset of the number of devices
* Only do autotuning in copy-gauge when using GPU
* Fix bugs in how the labelling of the copy-gauge kernel are recorded as being on CPU or GPU
* Add support for device-reorder support when using padded TIFR fields: read/write directly to host memory to avoid needing a temporary on the GPU
* Remove unused blas functions
* Alias some blas functions that have common access patterns to remove unnecessary compilation
* Remove legacy blas_cpu.cpp file


lib/blas_mixed_core.h
```diff
@@ -4,10 +4,14 @@ namespace mixed {
...
     void blasCuda(const double2 &a, const double2 &b, const double2 &c,
 		  ColorSpinorField &x, ColorSpinorField &y,
-		  ColorSpinorField &z, ColorSpinorField &w) {
+		  ColorSpinorField &z, ColorSpinorField &w,
+                  ColorSpinorField &v) {
+
+    checkPrecision(x, z, w);
+    checkPrecision(y, v);
 
     if (checkLocation(x, y, z, w) == QUDA_CUDA_FIELD_LOCATION) {
```
weinbe2: `checkLocation(x, y, z, w, v)`

lib/blas_quda.cu
```diff
@@ -97,8 +97,8 @@ namespace quda {
       const Float2 a;
       const Float2 b;
       axpbyz_(const Float2 &a, const Float2 &b, const Float2 &c) : a(a), b(b) { ; }
-      __device__ __host__ void operator()(FloatN &x, FloatN &y, FloatN &z, FloatN &w)
-      { z = a.x*x + b.x*y; }
+      __device__ __host__ void operator()(FloatN &x, FloatN &y, FloatN &z, FloatN &w, FloatN &v)
+      { v = a.x*x + b.x*y; } // use w not z to ensure same precision as y
```
weinbe2: Should be `use v not z` in the comment

weinbe2:
Once @maddyscientist has tested this in a build of TIFR+QUDA, and assuming the build tests go through, it's good to go as far as I can tell!

Issue #762, opened 2019-01-22
Use CheckCXXCompilerFlag for potentially unsupported compiler flags


mathiaswagner:
See https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html

This might be useful to check whether flags are supported on a specific arch or by a specific compiler.

PR #763, opened 2019-01-22
Hotfix/cmake power

mathiaswagner:
For `RELEASE` build type use

`-mcpu / -mtune -Ofast` flags on POWER systems as `-march` is not supported.

This is just a quick workaround and we should look into a better solution, see #762.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,13 +235,28 @@ include_directories(SYSTEM ${EIGEN_INCLUDE_DIRS})
 ## Now we hopefully found some way to get eigen to work
 
 
+
+# Linux: CMAKE_HOST_SYSTEM_PROCESSOR "x86_64"
+# Mac: CMAKE_HOST_SYSTEM_PROCESSOR "x86_64"
+# Power: CMAKE_HOST_SYSTEM_PROCESSOR "ppc64le"
+
+## We need to use different optimization flags depending on whether we are on x86 or power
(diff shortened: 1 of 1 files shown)


maddyscientist:
@walkloud you're assigned to review this pull request.  Are you ok with it?

walkloud:
@mathiaswagner in the fix, you have the following
```-mcpu=native -mtune=native```
while in my build, I have
```-mcpu=power9 -mtune=power9```
have you tried the `native`?  I've been meaning to (but haven't yet) checked if that works

mathiaswagner:
I just tried and it seems to compile fine (at least the compiler does not complain about the flags) and in that process realized that I also needed to fix my check.
Should be fine now but our CI is purely x86 right now so any additional checks on other machines are highly appreciated.

But we can always improve the fix later if is not yet perfect.

walkloud:
@mathiaswagner @maddyscientist the code is compiling with those changes, `-mcpu=native -mtune=native`, would you like me to run a performance test as well with one of the tests?

mathiaswagner:
Thanks for checking @walkloud. I don't think the code will not run if you can compile it but of course a quick firing up of one of the quad internal tests (or whatever is least effort for you) is appreciated. Just to be sure ...

walkloud:
OK - I’ll do an MDWF test tomorrow

I wasn’t thinking of working or not, but if the `native` gets the same optimization as `power9`

PR #766, opened 2019-01-31
Feature/ca-cg

weinbe2:
This PR adds support for both power basis and Chebyshev basis CA-CG, as well as a wrapper implementation of CA-CGNE and CA-CGNR. There is full support for mixed precision and reliable updates. (As a by-product, it fixes a bug in mixed precision CGNE and CGNR.) Adding support for additional bases is relatively straightforward. (There are reasons to use a further customized basis: while using the Chebyshev basis instead of a power basis alleviates the issue of the vectors blowing up and low modes getting completely lost in the power iterations, it doesn't completely eliminate the issue of co-linearity in practice.) 

The Chebyshev basis implementations requires conservative estimates of the smallest and largest eigenvalues to define the window for the rescaled Chebyshev polynomials. These can be passed in explicitly; alternatively the default behavior is to assume smallest eigenvalue is 0 (which, again, because it's just used to define a window is perfectly fine), and to perform 100 steps of power iterations on the first application of the solver to get an estimate of the largest eigenvalue, which is rescaled by a factor of 1.1. 

This PR exposes several additional parameters on the command line for specifying the "s"-step size, which basis to use, and the minimum+maximum eigenvalues. These can be supplied for an outer solver, as well as for near-null vector generation and the coarsest solve. (In practice, using the normal operator for the smoother has never been useful, so I didn't expose it for the smoother. If someone finds it useful it'd be a trivial addition.) I'll get these parameters listed on the wiki page soon. 

There are still some tentative memory optimizations, in particular for generating the Chebyshev polynomial (we'd need a multi-`caxy` as opposed to just the `caxpy`, or a `caxpbypczw`; for now we do a separate `zero` and multi-caxy`). 

Note: there is a quirk in how residuals are reported for CA-CG where the printed residual is actually from one s-step previous. This is because the residual of the previous s-step is a by-product of a given s-step iteration, otherwise an extra reduction is required per s-step which defeats the purpose of a communication avoiding (well, mitigating) algorithm. This does mean the algorithm in general requires one additional s-step before convergence is detected, however for a large number of iterations this is negligible. The implementation of CA-CG _does_ support detecting when a fixed number of iterations are requested, and in this case the extra s-step doesn't occur.

Here's a reference command which covers everything for an invert test:

`./invert_test --inv-type ca-cgne --verbosity verbose --ngcrkrylov 4 --prec double --prec-sloppy half --ca-basis-type chebyshev --cheby-basis-eig-min 0 --cheby-basis-eig-max 2.0533007`


alexstrel:
@weinbe2 a minor issue: the restart counter at 
https://github.com/lattice/quda/blob/feature/ca-cg/lib/inv_ca_cg.cpp#L500 
is unused, e.g.,  should be incremented somewhere in the reliable update block.

weinbe2:
@alexstrel aye, @maddyscientist pointed out to me that restarts are currently broken in CA-CG, and I also broke them in CA-GCR, That's on my to-do list for this PR.

To have a full record here, I also need to replace the internal basis enums in CA-GCR with the "global" enums I use in CA-CG. CA-GCR doesn't have Chebyshev basis support (yet), so my planned approach is to just have it warning out if it's requested and switch to the power basis.

maddyscientist:
It looks good @weinbe2.  I've done some testing of it, and one thing I did change was that I noticed that the `QudaInvertParam::ca_lambda_min` and `QudaInvertParam::ca_lambda_max` values were not being updated at the end of the solver meaning that back-to-back solves would recompute the Rayleigh quotient unnecessarily.  This is fixed now, and I moved the Rayleigh quotient under the `init` timing category, since this is where it belongs.

Just pushed this, so it will be doing its build tests overnight.  Happy for you @weinbe2 to merge this in the morning, or @mathiaswagner can do so 😄

PR #767, opened 2019-02-04
This adds a clang format file to QUDA

mathiaswagner:
Style is based on webkit

Should result in minimal changes to files, but some points need discussion.

C.f.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

Open points:
- [ ] ColumnLimit _ Currently not set to avoid a lot of rewrapping of code_
- [ ] Spacing around operators _I guess it is fine the way it is but still ..._
- [ ] BreakBeforeBraces  (Attach or Linux)  _We seem to be a bit inconsistent here_
- [ ] ConstructorInitializerAllOnOneLineOrOnePerLine  _WIthout ColumLimit the OneLine version puts everything on a single line_
- [ ] Reflow Comments _This leads to a lot of changes, but if we enforce a column limit_

There are a couple of other options not currently other than in the webkit default.
The webkit default settings (i.e. the values used for those properties not specified in the file) can be obtained using `clang-format -style=llvm -dump-config` and the full version proposed here with
`clang-format -dump-config` somewhere in the quad directory.

I suggest just reformatting a few files and run a diff to see what has changed.

Later on we can probably use the __Script for patch reformatting__ suggested here: https://clang.llvm.org/docs/ClangFormat.html.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,14 +1,17 @@
 ﻿---
 BasedOnStyle: Webkit
+IndentWidth: 2
 AllowShortBlocksOnASingleLine: true
 AllowShortCaseLabelsOnASingleLine : true
 AllowShortIfStatementsOnASingleLine: true
 AllowShortLoopsOnASingleLine: true
-BreakBeforeBraces: Attach
+BreakBeforeBraces: Linux
 ConstructorInitializerAllOnOneLineOrOnePerLine: true
 Cpp11BracedListStyle: true
 FixNamespaceComments: true
 NamespaceIndentation: All
 SortIncludes: false
 SpaceBeforeAssignmentOperators: true
+CommentPragmas:  '^\\.+'
+UseTab: Never
 ...
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,14 @@
+﻿---
+BasedOnStyle: Webkit
+AllowShortBlocksOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine : true
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: true
+BreakBeforeBraces: Attach
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+Cpp11BracedListStyle: true
+FixNamespaceComments: true
+NamespaceIndentation: All
+SortIncludes: false
+SpaceBeforeAssignmentOperators: true
+...
--- a/.clang-format
+++ b/.clang-format
@@ -1,15 +1,19 @@
-﻿---
+---
 BasedOnStyle: Webkit
 IndentWidth: 2
+AlignTrailingComments: true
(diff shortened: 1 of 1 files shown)


rbabich:
> BreakBeforeBraces  (Attach or Linux)  _We seem to be a bit inconsistent here_

At least originally, I think we mostly followed the K&R style, where there's a break before braces that starts a function (and classes/structs) but not before those that start a conditional or loop, e.g.,
```
void doSomething(bool flag, int n)
{
  for (int i = 0; i < n; i++) {
    if (flag) {
      doThis();
    } else {
      doThat();
    }
  }
}
```
clang-format calls this `BS_Linux`.

mathiaswagner:
Yeah, I don't have a strong preference here. Just trying to come up with something that does match the style as close as possible and then allows us to enforce that.

mathiaswagner:
@hummingtree I saw you used clang-format. We have started work on an official style for QUDA. Still WIP but maybe you have comments for this PR.

@maddyscientist @weinbe2 Ping.

maddyscientist:
Putting this here for some reading homework
https://engineering.mongodb.com/post/succeeding-with-clangformat-part-1-pitfalls-and-planning

maddyscientist:
Ok, pushed some changes
* To respond to @rbabich comment about BreakBeforeBraces: I've switched it to `Linux` (which means `BS_Linux`).
* Set 2 space indents (not the default 4)
* Never use tabs

Not to say that it's perfect (yet).  I think imposing a column limit is probably a good thing since presently it's removing the wrap around, e.g., for member instantiation in constructors.

**before**
```c++
  cudaColorSpinorField::cudaColorSpinorField(const ColorSpinorParam &param) :
    ColorSpinorField(param), alloc(false), init(true), texInit(false),
    ghostTexInit(false), ghost_field_tex{nullptr,nullptr,nullptr,nullptr}
  {
```

**after**
```c++
  cudaColorSpinorField::cudaColorSpinorField(const ColorSpinorParam& param)
      : ColorSpinorField(param), alloc(false), init(true), texInit(false), ghostTexInit(false), ghost_field_tex {nullptr, nullptr, nullptr, nullptr}
  {
```

mathiaswagner:
The removal of the wrap for constructors is controlled with `ConstructorInitializerAllOnOneLineOrOnePerLine: true`. However it seems that it literally is _one per line_ or `AllOnOne` and no option to leave it as it is. 
Not sure what we prefer. I think _AllOnOneLine_ wraps if there is a column limit but I am not a big fan of enforcing wrapping, but maybe it is time to reconsider.

mathiaswagner:
What about doing a wide column limit then? I think the default `80` is quite restrictive with the long type names. What about doing something like `160` ? 
With that I can easily view the full width on 15" and even two files next to each other on my 27".
Or maybe be less aggressive and do `120` ?

PR #768, opened 2019-02-05
initCommsGridQuda: returns early if comms_initialized

jxy:
Repeated calls of `initCommsGridQuda`, such as is done in
`milc_interface.cpp`, would overwrite the comms map that has been
setup from previous calls, leading to an unintended comms map.

This commit makes the exposed interface `initCommsGridQuda` return
early if `comms_initialized`.  As it sets `comms_initialized` upon
successful return, we create the restriction that only the first
call of `initCommsGridQuda` has an effect.


mathiaswagner:

PR #770, opened 2019-02-13
Add support for a user-defined MPI communicator in place of MPI_COMM_…

detar:
…WORLD

### DIFF
--- a/include/mpi_comm_handle.h
+++ b/include/mpi_comm_handle.h
@@ -1,11 +1,15 @@
 #ifndef _COMM_HANDLE_H
 #define _COMM_HANDLE_H
 
+#if defined(QMP_COMMS) || defined(MPI_COMMS)
 #include <mpi.h>
 extern MPI_Comm MPI_COMM_HANDLE;
+#endif
 
+#ifdef QMP_COMMS
 extern "C"{
 QMP_status_t QMP_get_mpi_comm(QMP_comm_t comm, void** mpicomm);
 }
+#endif
 
 #endif /* _COMM_HANDLE_H */
--- a/include/mpi_comm_handle.h
+++ b/include/mpi_comm_handle.h
@@ -1,0 +1,11 @@
+#ifndef _COMM_HANDLE_H
+#define _COMM_HANDLE_H
+
+#include <mpi.h>
+extern MPI_Comm MPI_COMM_HANDLE;
+
+extern "C"{
+QMP_status_t QMP_get_mpi_comm(QMP_comm_t comm, void** mpicomm);
+}
+
+#endif /* _COMM_HANDLE_H */
--- a/include/quda.h
+++ b/include/quda.h
@@ -582,6 +582,12 @@ extern "C" {
   typedef int (*QudaCommsMap)(const int *coords, void *fdata);
 
   /**
+   * @param mycomm User provided MPI communicator in place of MPI_COMM_WORLD
+   */
+
+  void qudaSetCommHandle(void *mycomm);
+
+  /**
    * Declare the grid mapping ("logical topology" in QMP parlance)
    * used for communications in a multi-GPU grid.  This function
    * should be called prior to initQuda().  The only case in which
@@ -607,6 +613,7 @@ extern "C" {
    *
    * @see QudaCommsMap
    */
+
   void initCommsGridQuda(int nDim, const int *dims, QudaCommsMap func, void *fdata);
 
   /**
@@ -1195,6 +1202,8 @@ extern "C" {
    * Free resources allocated by the deflated solver
    */
   void destroyDeflationQuda(void *df_instance);
+
+  void qudaSetMPICommHandle(void *mycomm);
 
 #ifdef __cplusplus
 }
--- a/include/quda.h
+++ b/include/quda.h
@@ -1203,7 +1203,7 @@ extern "C" {
    */
   void destroyDeflationQuda(void *df_instance);
 
-  void qudaSetMPICommHandle(void *mycomm);
+  void setMPICommHandleQuda(void *mycomm);
 
 #ifdef __cplusplus
 }
--- a/lib/comm_mpi.cpp
+++ b/lib/comm_mpi.cpp
@@ -5,7 +5,7 @@
 #include <csignal>
 #include <quda_internal.h>
 #include <comm_quda.h>
-#include <comm_handle.h>
+#include <mpi_comm_handle.h>
 
 #define MPI_CHECK(mpi_call) do {                    \
   int status = mpi_call;                            \
--- a/lib/comm_mpi.cpp
+++ b/lib/comm_mpi.cpp
@@ -5,7 +5,7 @@
 #include <csignal>
 #include <quda_internal.h>
 #include <comm_quda.h>
-
+#include <comm_handle.h>
 
 #define MPI_CHECK(mpi_call) do {                    \
   int status = mpi_call;                            \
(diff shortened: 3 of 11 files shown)


mathiaswagner:

mathiaswagner:
Thanks, @detar !

@jenkinks Ok to test.

mathiaswagner:
I have resolved a merge conflict in 
commit https://github.com/lattice/quda/pull/770/commits/cdd1dec63bae4fe6af9050edc894c04998232f6f.

@detar There still seems to be a header (or two) missing from your commit. Can you check that

`mpi_comm_handle.h` and /or 
`comm_handle.h`

Could you please add them?

mathiaswagner:
From Carleton:

> I have pushed my changes to feature/set-comm-handle after merging your changes with mine.  I seem to have fixed the qudaLoadCloverField bug in milc_interface.cpp, but I am not happy with the hack.  I suspect we are the only ones using the MILC interface, and we were bitten because we haven't used the QUDA clover inverter for a long time.  The problem arises when calling qudaCloverInvert.  This routine maps the gauge and clover fields to QUDA objects and then calls the QUDA clover inverter.  In the process, it needs to map MILC values to the QUDA invertParam class.  They are mapped only after qudaLoadCloverField is called.  But for some reason qudaLoadCloverField defines its own local invertParam values before calling the main QUDA loadClover.  The latter checks the invertParam values before using them, and that check failed because qudaLoadCloverField didn't define all the irrelevant values.  I would need a better understanding of the need to have two instantiations of invertParam.  It would seem better to instantiate it in qudaCloverInvert before calling qudaLoadCloverField, and then passing it the fully defined invertParam.  So, instead, I simply hacked qudaLoadCloverField's local version of invertParam, giving dummy values to the missing parameters irrelevant to loadClover, just so it passes the checker.  This seems to have fixed the problem.

include/mpi_comm_handle.h
```diff
@@ -0,0 +1,11 @@
+#ifndef _COMM_HANDLE_H
+#define _COMM_HANDLE_H
+
+#include <mpi.h>
+extern MPI_Comm MPI_COMM_HANDLE;
+
+extern "C"{
```
mathiaswagner: This needs some `#ifdef` to make sure it builds for MPI, QMP, and single GPU. Currently the MPI based CI test fail as they don't use QMP.

include/quda.h
```diff
@@ -614,6 +614,12 @@ extern "C" {
    */
   typedef int (*QudaCommsMap)(const int *coords, void *fdata);
 
+  /**
+   * @param mycomm User provided MPI communicator in place of MPI_COMM_WORLD
+   */
+
+  void qudaSetCommHandle(void *mycomm);
+
```
mathiaswagner: what is the difference between `qudaSetCommHandle` and `setMPICommHanfleQuda`?

lib/comm_qmp.cpp
```diff
@@ -2,6 +2,7 @@
 #include <csignal>
 #include <quda_internal.h>
 #include <comm_quda.h>
+#include <mpi_comm_handle.h>
```
mathiaswagner: Maybe the inclusion of `mpi_comm_handle.h` should be done in `comm_quda.h`.

lib/interface_quda.cpp
```diff
@@ -361,12 +362,52 @@ static int qmp_rank_from_coords(const int *coords, void *fdata)
 #endif
 
 
+// Provision for user control over MPI comm handle
+// Assumes an MPI implementation of QMP
+
+MPI_Comm MPI_COMM_HANDLE;
```
mathiaswagner: This will require some `#ifdef` to not break compilation if we compile without MPI.

lib/comm_mpi.cpp
```diff
@@ -5,7 +5,7 @@
 #include <csignal>
 #include <quda_internal.h>
 #include <comm_quda.h>
-
+#include <comm_handle.h>
```
mathiaswagner: Should this be `mpi_comm_handle.h` ?

detar:
Hi Mathias,

Comments below.  No, I have not tested a build with only MPI and on a 
single GPU.  I'll see about fixing the problems you have noted.

Best,

Carleton

On 2/19/19 1:24 PM, Mathias Wagner wrote:
>
> *@mathiaswagner* requested changes on this pull request.
>
> This looks like a good addition, but we need to address some things to 
> make sure nothing breaks.
> I have added a few comments. Note that currently compilation still 
> fails on the MPI build, so not sure whether other errors might pop up 
> later.
>
> ------------------------------------------------------------------------
>
> In include/mpi_comm_handle.h 
> <https://github.com/lattice/quda/pull/770#discussion_r258211922>:
>
> > @@ -0,0 +1,11 @@
> +#ifndef _COMM_HANDLE_H
> +#define _COMM_HANDLE_H
> +
> +#include <mpi.h>
> +extern MPI_Comm MPI_COMM_HANDLE;
> +
> +extern "C"{
>
> This needs some |#ifdef| to make sure it builds for MPI, QMP, and 
> single GPU. Currently the MPI based CI test fail as they don't use QMP.
>
My understanding is that QUDA assumes the MPI implementation of QMP.  
But, yes, it needs protection against a single GPU build.
> ------------------------------------------------------------------------
>
> In include/quda.h 
> <https://github.com/lattice/quda/pull/770#discussion_r258212569>:
>
> > @@ -614,6 +614,12 @@ extern "C" {
>      */
>     typedef int (*QudaCommsMap)(const int *coords, void *fdata);
>   
> +  /**
> +   * @param mycomm User provided MPI communicator in place of MPI_COMM_WORLD
> +   */
> +
> +  void qudaSetCommHandle(void *mycomm);
> +
>
> what is the difference between |qudaSetCommHandle| and 
> |setMPICommHanfleQuda|?
>
qudaSetCommHandle is the MILC interface function, following the QUDA 
naming conventions.  It appears only in milc_interface.cpp. 
setCommHandleQuda is the main QUDA function and the one for non -MILC 
interfacing.  It is called by qudaSetCommHandle.
> ------------------------------------------------------------------------
>
> In lib/comm_qmp.cpp 
> <https://github.com/lattice/quda/pull/770#discussion_r258213186>:
>
> > @@ -2,6 +2,7 @@
>   #include <csignal>
>   #include <quda_internal.h>
>   #include <comm_quda.h>
> +#include <mpi_comm_handle.h>
>
> Maybe the inclusion of |mpi_comm_handle.h| should be done in 
> |comm_quda.h|.
>
> ------------------------------------------------------------------------
>
> In lib/interface_quda.cpp 
> <https://github.com/lattice/quda/pull/770#discussion_r258213492>:
>
> > @@ -361,12 +362,52 @@ static int qmp_rank_from_coords(const int *coords, void *fdata)
>   #endif
>   
>   
> +// Provision for user control over MPI comm handle
> +// Assumes an MPI implementation of QMP
> +
> +MPI_Comm MPI_COMM_HANDLE;
>
> This will require some |#ifdef| to not break compilation if we compile 
> without MPI.
>
> ------------------------------------------------------------------------
>
> In lib/comm_mpi.cpp 
> <https://github.com/lattice/quda/pull/770#discussion_r258213860>:
>
> > @@ -5,7 +5,7 @@
>   #include <csignal>
>   #include <quda_internal.h>
>   #include <comm_quda.h>
> -
> +#include <comm_handle.h>
>
> Should this be |mpi_comm_handle.h| ?
>
Hmm.  I'll check.

detar:
Hi Mathias,

I made the changes that you suggested and pushed to 
feature/set-comm-handle.   Could you try the test again?

Thanks,

Carleton

weinbe2:
It looks like there are a few remaining references to `MPI_COMM_WORLD` (as opposed to `MPI_COMM_HANDLE`):

* `lib/nvtx_pmpi.c` and `lib/generate/nvtx.w` --- I can't say I have familiarity with the files, but I imagine they need the appropriate include+replace
* `tests/test_util.cpp`

I'm guessing these are just straightforward changes?

Do we want to add a test executable that demonstrates using split communicators for this pull, or is that something we should add as an "Issue" and get to later just to get this merged in so we're at feature parity with MILC's develop branch?

detar:
Hi Evan,

I thought I had found them all.  I'll look again and fix them. What do 
you suggest for a spit comm  test?   A call to some appropriate QUDA 
function from the separated communicators?  What is your favorite function?

Thanks,

Carleton

weinbe2:
I'm not very familiar with split communicator workflows, but my impression is any function would be as good as any other, so perhaps a simple staggered dslash application? 

Something (personally) illuminating could just be using split communicators to achieve naive parallelism, for ex, having, say, 4 separate communicators for an 8 rank job -> 2 ranks per communicator? Is that a worthwhile test case? (How does autotuning get handled there? Calling on your experience, Carleton, do you first autotune a 2 rank, one communicator job, then use that `tunecache` for a "real" 8 rank, four communicator job?)

detar:
I haven't paid much attention to the autotuning sequence.  I our case 
the split jobs used identical code and should get the same autotuning 
parameters.  I guess in most cases I first ran an unsplit job on the 
smaller partition, which produced the first autotuning file, and then 
ran a bundled job on a proportionately scaled number of nodes where each 
job read and, possibly, modified the autotuning file.  So the last job 
part to write the tunecache file got to leave its mark.

Best,

Carleton

weinbe2:
Thank you for explaining, Carleton. 

In the interest of getting your pull merged sooner as opposed to later, let's postpone coming up with a QUDA test executable. Once the `MPI_COMM_WORLD` references are fixed I'm happy getting this merged ASAP.

maddyscientist:
This looks ready now for merging.  Thanks @weinbe2 for catching the last few `MPI_COMM_WORLD` instances.  @detar I'll merge this once the build tests all complete as verified.

PR #775, opened 2019-02-21
Feature/rewrite staggered

mathiaswagner:
This is the rewrite for staggered. Some things to take into consideration:

Some issues still to be addressed, but not super critical for all as this does not go into develop yet.

**Must fix**
- [x] cleanup and probably some comments
- [x] guarding off old code with #ifdefs to allow compilation without TEX
- [ ] as standard staggered is now recon 13 make sure that is reflected in the interface

**Fix later**
- [ ] multiple rhs (including swizzle)
- [ ] TIFR phases
- [ ] proper performance sweep on multi GPU
- [ ] check whether reported FLOPS for the individual kernels come out right


include/gauge_field_order.h
```diff
@@ -56,8 +58,8 @@ namespace quda {
 	 @param[in] x_cb Checkerboarded space-time index we are accessing
 	 @param[in] parity Parity we are accessing
        */
-      __device__ __host__ inline gauge_wrapper<Float,T>(T &gauge, int dim, int x_cb, int parity)
-	: gauge(gauge), dim(dim), x_cb(x_cb), parity(parity) { }
+      __device__ __host__ inline gauge_wrapper<Float,T>(T &gauge, int dim, int x_cb, int parity, Float phase=10.0)
```
maddyscientist: any reason why 10.0 is the default here?

include/gauge_field_order.h
```diff
@@ -1456,26 +1529,22 @@ namespace quda {
...
     }
 
-    template<typename I>
+       template<typename I>
     __device__ __host__ inline void Unpack(RegType out[18], const RegType in[8], int idx, int dir,
                                            const RegType phase, const I *X, const int *R,
-                                           const Complex scale=Complex(static_cast<RegType>(1.0),
-                                                                       static_cast<RegType>(1.0))) const {
+                                           const Complex scale,
+                                           const Complex u) const {
```
maddyscientist: trailing whitespace

lib/copy_gauge_inc.cu
```diff
@@ -13,33 +13,40 @@ namespace quda {
...
-	if (typeid(FloatOut)==typeid(short) && out.LinkType() == QUDA_ASQTAD_FAT_LINKS) {
-	  copyGauge<short,FloatIn,length>
-	    (FloatNOrder<short,length,2,19>(out, (short*)Out, (short**)outGhost, override), inOrder, out, in, location, type);
-	} else {
-	  typedef typename gauge_mapper<FloatOut,QUDA_RECONSTRUCT_NO>::type G;
-	  copyGauge<FloatOut,FloatIn,length>
-	    (G(out, Out, outGhost, override), inOrder, out, in, location, type);
-	}
+      	if (typeid(FloatOut)==typeid(short) && out.LinkType() == QUDA_ASQTAD_FAT_LINKS) {
+      		copyGauge<short,FloatIn,length>
```
maddyscientist: tab indents here

lib/gauge_field.cpp
```diff
@@ -49,8 +49,10 @@ namespace quda {
     if (link_type != QUDA_WILSON_LINKS && fixed == QUDA_GAUGE_FIXED_YES)
       errorQuda("Temporal gauge fixing only supported for Wilson links");
 
-    if(link_type != QUDA_ASQTAD_LONG_LINKS && (reconstruct ==  QUDA_RECONSTRUCT_13 || reconstruct == QUDA_RECONSTRUCT_9))
-      errorQuda("reconstruct %d only supported for staggered long links\n", reconstruct);
+    // if(link_type != QUDA_ASQTAD_LONG_LINKS && (reconstruct ==  QUDA_RECONSTRUCT_13 || reconstruct == QUDA_RECONSTRUCT_9))
+    //   errorQuda("reconstruct %d only supported for staggered long links\n", reconstruct);
+       
+    if (link_type == QUDA_ASQTAD_MOM_LINKS) scale = 1.0;
```
maddyscientist: I think I removed this line since `scale` now defaults to 1.0

include/gauge_field_order.h
mathiaswagner: debugging artifact. I think I caught all places where I used that now.

include/gauge_field_order.h
mathiaswagner: clang-format (I will just apply the latest working version to my changes and we can use that as a test)

lib/copy_gauge_inc.cu
mathiaswagner: Another clang-format ...

mathiaswagner:
Comments should be addressed now (for now).

include/kernels/dslash_staggered.cuh
```diff
@@ -0,0 +1,219 @@
...
+    case 1: sign = (coords[3] + coords[0]) % 2 == 0 ? static_cast<Float>(1.0) : static_cast<Float>(-1.0); break;
+    case 2: sign = (coords[3] + coords[1] + coords[0]) % 2 == 0 ? static_cast<Float>(1.0) : static_cast<Float>(-1.0); break;
+    case 3: sign = (coords[3] + dir >= X[3] || coords[3] + dir < 0) ? tboundary : static_cast<Float>(1.0); break;
+    default: sign = static_cast<Float>(1.0);
+    }
+    return sign;
+  }
+
+  /**
+       Applies the off-diagonal part of the Laplace operator
```
maddyscientist: Laplace ? 😜

lib/dirac_improved_staggered.cpp
```diff
@@ -51,11 +51,16 @@ namespace quda {
 			      const QudaParity parity) const
   {
     checkParitySpinor(in, out);
-
     if (checkLocation(out, in) == QUDA_CUDA_FIELD_LOCATION) {
+#ifdef USE_LEGACY_DSLASH
```
maddyscientist: One of the improvements from the new code vs the old code is that the dslash kernels can be applied to full fields.  So you can replace the two Deo and Doe calls in `DiracStaggered::M` and `DiracImprovedStaggered::M` with a single kernel call.  See what I did in dirac_wilson.cpp as an example.

mathiaswagner: fixed in 
https://github.com/lattice/quda/pull/775/commits/f631634ac5f15344b9d60423e0732d41551bcb26

Issue #333, opened 2015-07-30
(Staggered) mixed precision inverter is broken with reconstruction 9


mathiaswagner:
I tried using the staggered mixed precision inverter in MILC with reconstruction and saw convergence issues. I can reproduce the issue with staggered_invert_test on a single node.

For reference pure double:

```
test_invert_double18_double18.txt:CG: Convergence at 1825 iterations, L2 relative residual: iterated = 9.865356e-08, true = 9.865356e-08
```

Double single without reconstruction and pure double with 9 reconstruct:

```
test_invert_double18_single18.txt:CG: Convergence at 1829 iterations, L2 relative residual: iterated = 9.975623e-08, true = 9.975623e-08
test_invert_double9_double9.txt:CG: Convergence at 1832 iterations, L2 relative residual: iterated = 9.997816e-08, true = 9.997816e-08
```

And finally mixed precision with reconstruction:

```
test_invert_double9_single18.txt:CG: Convergence at 715 iterations, L2 relative residual: iterated = 1.437793e+01, true = 1.437793e+01
test_invert_double9_single9.txt:CG: Convergence at 717 iterations, L2 relative residual: iterated = 1.443085e+01, true = 1.443085e+01
```

The name of the output is `test_invert_<prec><recon>_<prec_sloppy><recon_sloppy>.txt`

There are more combinations in this parameter space but something is going wrong when combining reconstruction with mixed precision.

mikeaclark:
Is there a way in MILC you can set the boundary conditions to be all periodic?  I wonder if it's related to boundary conditions getting screwed up.

mikeaclark:
This is the develop branch isn't it?

mathiaswagner:
This is the branch for the milc baseline (essentially develop + NVTX).
The results I copied above are from the internal QUDA tests:
```
./staggered_invert_test --recon 9 --recon_sloppy 9 --prec double --prec_sloppy single
```

mathiaswagner:
Adding more datapoints:
```
test_invert_double13_single13.txt:CG: Convergence at 1841 iterations, L2 relative residual: iterated = 9.950236e-08, true = 9.950236e-08
test_invert_double13_single18.txt:CG: Convergence at 1878 iterations, L2 relative residual: iterated = 9.979013e-08, true = 9.979013e-08
```

13 reconstruct seems to be fine.

mikeaclark:
[...]

Issue #692, opened 2018-05-07
"Attempting to bind too large a texture"


kostrzewa:
It seems that I'm having some issues running on PizDaint on few nodes when device memory utilisation is high and local lattice edge lengths are large due to errors of the following kind:

```
# QUDA: ERROR: Attempting to bind too large a texture 143327232 > 134217728 (rank 0, host nid06365, /users/bartek/code/quda_feature_multigrid/lib/
cuda_gauge_field.cu:145 in createTexObject())
# QUDA:        last kernel called was (name=N4quda14ExtractGhostExIdLi18ELi4ELi3ENS_5gauge11FloatNOrderIdLi18ELi2ELi18EL20QudaStaggeredPhase_s0ELb
1EEEEE,volume=48x48x48x26,aux=prec=8,stride=1437696,extract=0,dimension=3,geometry=4)
```

This is a simple double-half mixed CG run on a 48c96 lattice (twisted clover with dynamic clover inversion).

The corresponding run on 9 nodes reports device memory usage below 5 GB, which would mean that the problem should just about fit on 3 P100, yet it only seems to fit on 6 nodes and above.

```
# QUDA: CG: Convergence at 29733 iterations, L2 relative residual: iterated = 2.733731e-10, true = 2.733731e-10
# QUDA: Done: 29733 iter / 187.209 secs = 8298.42 Gflops
# QUDA: time spent in reorder_spinor_fromQuda: 0.040400 secs
# Inversion done in 29733 iterations, squared residue = 1.162918e-11!
# Inversion done in 1.91e+02 sec. 
# QUDA: WARNING: Environment variable QUDA_PROFILE_OUTPUT_BASE is not set; writing to profile.tsv and profile_async.tsv
# QUDA: Saving 64 sets of cached parameters to /users/bartek/local/quda_resources/PizDaint-dynamic_clover-b87de9877010807be613550aac919bf7419d2b24_gdr1/profile_0.tsv
# QUDA: Saving 9 sets of cached profiles to /users/bartek/local/quda_resources/PizDaint-dynamic_clover-b87de9877010807be613550aac919bf7419d2b24_gdr1/profile_async_0.tsv
# QUDA: 
               initQuda Total time = 1.48045 secs
# QUDA:                   init     = 1.480448 secs (   100%), with        2 calls at 7.402240e+05 us per call
# QUDA:      total accounted       = 1.480448 secs (   100%)
# QUDA:      total missing         = 0.000000 secs (     0%)
[...]

Issue #583, opened 2017-03-10
Investigate double-half CG solver for twisted clover


maddyscientist:
Twisted clover fermions with double-half CG solver is very unstable compared to clover or regular twisted-mass fermions.  Perhaps related to #582.

mathiaswagner:
New or old reliable updates?

maddyscientist:
It's broken either way

maddyscientist:
Note for self, use these lattices for testing:

/scratch/ETMC_gauge_configurations/tmclover_nf2/cA2.60.32

with kappa = 0.1373, mu = 0.006, clover coeff = 1.57551

kostrzewa:
It seems that ```QUDA_DYNAMIC_CLOVER``` helps (i.e., it works then). This is an improvement from a few months ago. 

Tested with 65a2b78b5c3efd041e19d1b88367faf5f0a2dac8 of feature/multigrid

kostrzewa:
@maddyscientist had an argument for this, noting that storing and loading in half precision, leads to some loss of precision. This could be especially severe for twisted-clover because we have [ (1+C_ee)^2 + \mu^2 ]^-1, rather than just the inverse itself.

kostrzewa:
stability confirmed up to 64c128

maddyscientist:
@kostrzewa can you test this with the develop branch enabling @mathiaswagner's new reliable update scheme (enabled with `QUDA_ALTRELIABLE=ON`).  Would be good to see iteration count differences for pure double, double-single and double-half using the old mechanism and the new one.

kostrzewa:
will do

maddyscientist:
Closed with #776

PR #776, opened 2019-02-27
Complete Dslash rewrite using accessors and extensive integration with Jitify

maddyscientist:
This pull request is a complete rewrite of the dslash framework
* All dslash operators are re-implemented in the C++ accessor framework, no macro hell and no python generation
* QUDA can now be compiled without textures at all, removing the texture limit as a problem size limiter (set cmake option `-DQUDA_TEX=OFF`)
* All clover formulations now support dynamic clover inversion
* Increased kernel fusion for improved performance
* All dslash kernels, blas operations and other top time consuming kernels are now supported with jitify compilation
* Improvements to the regression testing using `ctest` when running on multiple processes 
* Removes all use of texture references from the library (texture objects remain)

Before we merge, here is the outstanding list of things to be done:
- [x] as standard staggered is now recon 13 make sure that is reflected in the interface
- [x] proper performance sweep on multi GPU
- [x] ~~`dslash4_*_staggered_preconditioned.cu` files need a static assert that `nParity==1` in the `launch` function, see `dslash4_wilson_clover_preconditioned.cu` for reference~~ No longer relevant, `kernels/dslash_staggered.cuh` appropriately handles even, odd, and full parity.
- [x] enable quarter precision support for staggered-type operators, edit `staggered_dslash_ctest` appropriately, see: https://github.com/lattice/quda/pull/776/commits/6cae2a3b8151733278697323da196c0d9e10ad38
- [x] go through `MWTODO` comments in the code and see what should be fixed now vs put off until after the merge

And here's the list of apps we should test against
- [x] tmLQCD
- [x] Chroma (covers both wilson-clover and dwf)
- [x] TIFR
- [x] MILC

There will be some features we need to add after the merge that will almost certainly not be included in this merge.  Until we get all missing features added, the plan will be to keep the legacy code path in place (enabled with the cmake option `-DQUDA_LEGACY_DSLASH=ON`.  The following should be created as new issues (assuming they do not make this pull)
- [x] TIFR phases (or enable true recon-9/13 support for "dynamic phasing" - @mathiaswagner)
- [ ] rewrite contraction kernels that use the old macro framework (@AlexVaq)
- [ ] multi-RHS support for all dslash types
- [ ] some missing experimental optimization features (swizzle and spatial blocking) 
- [ ] remove all legacy Dslash kernels
- [ ] work out how to set the shared memory carve out when using Jitify
- [ ] update wiki as needed (for ex, new advanced flags in `cmake`, support for U(3) links in recon 9/13)

Closes issues #333, #583, #692, #779


alexstrel:
a couple minor cosmetic points while glancing through the code:
1. there is no reason to keep the condition 
https://github.com/lattice/quda/blob/feature/dwf-rewrite/include/dslash.h#L92
2. setPackComms already invokes setPackComms2, no need for another call in 
https://github.com/lattice/quda/blob/feature/dwf-rewrite/include/dslash.h#L217  ?
And setPackComms2 does not have MULTI_GPU macro control.

maddyscientist:
Thanks for the comments @alexstrel.  I've addressed these points:
1.) the condition should have been if Volta or above, e.g., `deviceProp >= 7`, since on prior architectures this doesn't do anything (except a wasteful CUDA API call that will cause unnecessary latency)
2.) I've cleaned up the packing kernels: the legacy code is only compiled when legacy is enabled, and new packing is only compiled when new code is enabled.  I've removed the `setPackComms2` function, and instead we have just a single `setPackComms` function (same name is used for both legacy and rewrite).  We don't need the `MULTI_GPU` macro since on a single GPU the comm dim array will always be all zeros.  We should be removing unnecessary macro like this where we can I think.

I've also updated the pull description above, as per our discussions on slack earlier.

bjoo:
Hi @maddyscientist, I pulled this into develop for testing and ran Chroma's t_symm_prec which checked
clover BiCGStab, MultiShift CG, and multi-grid. All the tests in that passed. I can try a larger scale test, e.g. on an actual short gauge generation?

maddyscientist:
Thanks @bjoo for the testing with Chroma.  If you don't mind doing a full HMC test that would be most appreciated.  Likely your testing already suffices, but an HMC test would be more definitive.

@kostrzewa From your last communications with on #692 it sounds like tmLQCD is all good with this code now, and the solvers are working well for you?  (Beyond the orthogonal issue of GDR on Cray having issues?)

I've added an application testing list for this pull in the description above: tmLQCD, Chroma, MILC and TIFR.  I'll test MILC and TIFR in the next few days.

include/color_spinor_field_order.h
```diff
@@ -861,28 +861,31 @@ namespace quda {
   }
 
   __device__ __host__ inline void save(const RegType v[length], int x, int parity=0) {
-    RegType scale = 0.0;
     RegType tmp[length];
 
     if (sizeof(Float)==sizeof(short) || sizeof(Float)==sizeof(char)) {
```
weinbe2: Use `if_fixed` trait?

lib/dslash4_wilson.cu
```diff
@@ -40,7 +45,7 @@ namespace quda {
     const bool xpay;      // whether we are doing xpay or not
 
     WilsonArg(ColorSpinorField &out, const ColorSpinorField &in, const GaugeField &U,
-              Float kappa, const ColorSpinorField &x, int parity, bool dagger)
+              double kappa, const ColorSpinorField &x, int parity, bool dagger)
```
weinbe2: `real`, not `double`

include/dslash_helper.cuh
```diff
@@ -12,4 +18,359 @@ namespace quda {
...
+      if (kernel_type == INTERIOR_KERNEL) strcat(aux[kernel_type],ghost_str);
+      strcat(aux[kernel_type],aux_base);
+    }
+
+    bool tuneGridDim() const { return false; }
+    unsigned int minThreads() const { return arg.threads; }
+
+    template <typename T, typename Arg>
+    inline void launch(T *f, const TuneParam &tp, Arg &arg, const cudaStream_t &stream) {
+      if (1) { // test on Volta
```
weinbe2: Was this tested?

include/dslash_helper.cuh
```diff
@@ -1,12 +1,16 @@
...
 #include <register_traits.h>
 #include <index_helper.cuh>
 #include <tune_quda.h>
+#include <dslash_quda.h>
 
 namespace quda {
 
-  void setPackComms(const int *);
+  void setPackComms(const int *); // original packing kernel
+  void setPackComms2(const int *); // rewriten packing kernel
```
weinbe2: Did this get an updated name?

lib/dslash4_wilson.cu
weinbe2: Convinced myself it makes sense that this is a `double`.

lib/coarse_op.cuh
```diff
@@ -72,6 +73,7 @@ namespace quda {
       case COMPUTE_CONVERT:
       case COMPUTE_RESCALE:
       case COMPUTE_CLOVER_INV_MAX: // FIXME
+      case COMPUTE_TWISTED_CLOVER_INV_MAX: // FIXME
```
weinbe2: DId this get fixed?

lib/read_clover.h
```diff
@@ -187,40 +187,40 @@
   C1.y += param.rho; C2.x += param.rho; C2.y += param.rho;
 
 #define READ_CLOVER_SINGLE_TEX(clover, chi)			\
```
weinbe2: Broad file-wide comment---this is missing "QUARTER" versions of these routines, and presumably wherever these routines get called are also missing quarter versions.

maddyscientist: This is a legacy macro file that will be deleted.

include/dslash_helper.cuh
maddyscientist: There are two `setPackComms` functions.  One is defined in the legacy packer and one in the new packer.  Only one is compiled at once (originally the second variant was used for the new packer).

include/dslash_helper.cuh
maddyscientist: This code has been gone for a while, so not sure why you're looking at this out of date file.

lib/read_clover.h
weinbe2: :+1:

lib/dslash_pack2.cu
```diff
@@ -44,17 +45,22 @@ namespace quda {
...
     int_fastdiv swizzle;
     int sites_per_block;
 
-    PackArg(void **ghost, const ColorSpinorField &field, int nFace, bool dagger, int parity)
+    PackArg(void **ghost, const ColorSpinorField &field, int nFace,
+            bool dagger, int parity, double a, double b)
       : field(field, 1, nullptr, nullptr, reinterpret_cast<Float**>(ghost)),
         nFace(nFace), dagger(dagger), parity(parity), nParity(field.SiteSubset()),
-        dc(field.getDslashConstant())
+        dc(field.getDslashConstant()), a(a), b(b), twist(a != 0.0 && b != 0.0)
```
weinbe2: Should this be `a != 0.0 || b != 0.0`? I imagine it doesn't make sense to call this op without having both non-zero, but still, it could happen, and it at least should throw a warning/error (assuming it doesn't already somewhere...)

include/color_spinor_field_order.h
maddyscientist: Again, were you looking at an old file somehow here?

lib/coarse_op.cuh
maddyscientist: This is a minor FIXME, and isn't worth worrying about there I think.  It's just the flops for computing the max aren't being computed: mainly because there is a the question whether we special case for dynamic clover or not.  Regardless, the `bytes` is the important metric here, since this will be a bandwidth bound kernel.

include/kernels/dslash_ndeg_twisted_mass_preconditioned.cuh
```diff
@@ -0,0 +1,144 @@
...
+  template <typename Float, int nDim, int nColor, int nParity, bool dagger, bool asymmetric, bool xpay, KernelType kernel_type, typename Arg>
+  __global__ void ndegTwistedMassGPU(Arg arg)
+  {
+    int x_cb = blockIdx.x*blockDim.x + threadIdx.x;
+    if (x_cb >= arg.threads) return;
+
+    // for this operator flavor can be spread onto different blocks
+    int flavor = blockIdx.y*blockDim.y + threadIdx.y;
+
+    // for full fields set parity from y thread index else use arg setting
```
weinbe2: typo --- `z` thread index. this probably got copy+pasted from other files.

lib/dslash_pack2.cu
maddyscientist: Fixed 78a9149

include/kernels/dslash_ndeg_twisted_mass_preconditioned.cuh
maddyscientist: Fixed in d871493

lib/staggered_dslash.cu
```diff
@@ -21,53 +21,80 @@ namespace quda {
...
-        case 0:
-        index = ((x[0]+nFace-1)*X[4]*X[3]*X[2]*X[1] + x[4]*X[3]*X[2]*X[1] + x[3]*(X[2]*X[1])+x[2]*X[1] + x[1])>>1;
-    // index = ((x[0])*X[4]*X[3]*X[2]*X[1] + x[4]*X[3]*X[2]*X[1] + x[3]*(X[2]*X[1])+x[2]*X[1] + x[1])>>1;
-        break;
-        case 1:
-        index = ((x[0]-X[0]+nFace)*X[4]*X[3]*X[2]*X[1] + x[4]*X[3]*X[2]*X[1] + x[3]*(X[2]*X[1]) + x[2]*X[1] + x[1])>>1;
-        break;
-      }  
+      case 0:
+ index = ((x[0]+nFace-1)*X[4]*X[3]*X[2]*X[1] + x[4]*X[3]*X[2]*X[1] + x[3]*(X[2]*X[1])+x[2]*X[1] + x[1])>>1;
```
weinbe2: This may get addressed in a later pull, but just in case, @mathiaswagner can math like this which includes `nFace` get put into a helper routine in `dslash_helper`? Routines like this for other ops live there, iirc

lib/staggered_dslash.cu
```diff
@@ -250,14 +253,15 @@ namespace quda {
...
+// printf("NEW1 Out cb %i %i \t %f %f %f %f %f %f\n",x_cb, my_spinor_parity, out.data[0].real(),out.data[0].imag(),out.data[1].real(),out.data[1].imag(),out.data[2].real(),out.data[2].imag());
+   
     //MWTODO: clean up here
     // if (xpay) {
     //   Vector x = arg.x(x_cb, my_spinor_parity);
     //   out = arg.a * x -out ;
     // }
     if (dagger){
-      out = Float(-1)*out;
+      out = real(-1)*out;
```
weinbe2: `static_cast<real>(*)`

lib/dirac_twisted_clover.cpp
```diff
@@ -234,32 +292,27 @@ namespace quda {
...
+#else
+        TwistCloverInv(*tmp1, in, parity[1]);
+        DiracWilson::Dslash(out, *tmp1, parity[0]);
+        TwistCloverInv(*tmp1, out, parity[0]);
+        DiracWilson::DslashXpay(out, *tmp1, parity[1], in, kappa2);
+#endif
       }
     } else { //Twist doublet
-      errorQuda("Non-degenerate DiracTwistedCloverPC is not implemented \n");
+      errorQuda("Non-degenerate operator is not implemented");
```
weinbe2: Can this error be more instructive? i.e., mentioning that the user should explicitly use the non-degenerate operator instead of the regular twisted operator? This comment is relevant other places for twisted mass, too.

maddyscientist: Well this non-degenerate twisted clover is something that  is going to fill in at some point, and noone is using  it at the moment, so I don't see this as something to really worry about.

lib/staggered_dslash.cu
```diff
@@ -103,11 +104,12 @@ namespace quda {
     const GU U;            // the gauge field
     const GL L;            // the long gauge field
 
-    const Float a;
+    const real a;
+    const real fat_link_max; //MWTODO: FIXME: This a workaround should probably be in the accessor ...
```
weinbe2: Did this get addressed later?

weinbe2: Looks like it was! (I can only see `outdated` after positing a comment...)

lib/dirac_twisted_clover.cpp
weinbe2: Sounds good!

lib/staggered_dslash.cu
weinbe2: Found it, all good!

maddyscientist:
I have created milestone [20](https://github.com/lattice/quda/milestone/20), which covers the things left to do after we merge this pull request.  The issues listed on that milestone need to be completed before we delete the legacy kernels.

weinbe2:
Is there any reason we replaced `kappa` (Wilson-esque lingo) with `a` (dwf-esque lingo)? Specifically, this commit: https://github.com/lattice/quda/pull/776/commits/eaf3e59a0f4a92b991829d8b75b879a1ed19a740

I know it gets documented in code (i.e., things that mention it's the Wilson scale factor), but it's still a bit confusing from a physics standpoint. The convention change also didn't propagate its way into the multigrid code.

maddyscientist:
@weinbe2 The reason for the `kappa` -> `a` change was to minimize confusion.  Specifically in the Wilson kernel, the scale factor can sometimes be `-kappa`, `+kappa`, `-kappa^2` depending on where it is deployed.  Hence it makes it more clear if in the kernel we give it a different name.

maddyscientist:
@weinbe2 @AlexVaq @alexstrel are you happy to sign off on this?  This pull is awaiting your official approval 😉

weinbe2:
@maddyscientist did this:

> as standard staggered is now recon 13 make sure that is reflected in the interface

Get covered? I think we've hit multi-GPU (the other checkbox) has been covered "well enough".

alexstrel:
@maddyscientist since textures are enabled by default , would it be helpful to put a warning message to users about this mode? I guess the reason for the default mode is that many applications are still run  on pre-volta systems, still it might be a good idea to disable textures for better utilization of the compute nodes.

maddyscientist:
@alexstrel not using textures is still considered experimental I think, so I don't think this should be enabled by default.  There are a couple of performance regressions that I will plan to look at after this merge.  Also note that not using textures is not a Volta specific optimization, it works fine on older architectures as well.  So I think leaving this as an advanced option for now is fine, and we'll promote to a production feature, or even the default after we remove all the legacy code paths after this pull.

maddyscientist:
I just closed #779 since I was tracking down issues with the staggered phase factors so just decided to finish this up.

maddyscientist:
@weinbe2 Now that TIFR is verified to be working along with MILC, and since both these are the only applications that use the gauge reconstruction optimization, I think we're now covered for any changes needed with the convention change for 8->9 and 12->13 reconstruct for staggered fermions.  So now we're waiting for you to officially approve this pull so we can get it merged. 😜

maddyscientist:
Thanks @alexstrel!

PR #796, opened 2019-05-01
jitified and cleaned up the APE routine

cpviolator:
Jitify the APE routines.

### DIFF
--- a/include/kernels/gauge_ape.cuh
+++ b/include/kernels/gauge_ape.cuh
@@ -1,0 +1,145 @@
+#include <gauge_field_order.h>
+#include <index_helper.cuh>
+#include <quda_matrix.h>
+#include <su3_project.cuh>
+
+namespace quda
+{
+
+  template <typename Float, typename GaugeOr, typename GaugeDs> struct GaugeAPEArg {
+    int threads; // number of active threads required
+    int X[4];    // grid dimensions
+    int border[4];
+    GaugeOr origin;
+    const Float alpha;
+    const Float tolerance;
+
+    GaugeDs dest;
+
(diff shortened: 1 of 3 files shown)


cpviolator:
@mathiaswagner That's right. The Stout code shifted the GPU kernel code from the `.cu` file to the `kernels` file. I just did the same.

Now that I've done this I'm seeing a lot of boiler plate code that can be shared among all the gauge algorithms. I'm working on porting a gauge fixing routine to the same style as the smearing routines, then we can think about a class that holds the boiler plate for all gauge algos.

I developed this on summit. Let me pull it to a machine with a `clang-format` executable.

cpviolator:
Not having a good time with the class-format script:

```

cpviolator:
@mathiaswagner Is this good to go? I want to update all the gauge field algorithms from this branch.

weinbe2:
Due-diligence check, just because I've forgotten to do this in the past and it's come back to bite me: have you verified this works for both a single- and multi-GPU compile, jitify and non-jitify?

Also, it's not worth holding up this specific PR, but it'd be nice if there were CPU verify routines in the test files, at minimum for smearing and computing the topological charge, similar to the CPU dslash implementations. I know implementing CPU checks is a massive PITA (and they're slow, so I don't really see the need to implement a CPU gauge fix, for ex), but it'd be nice to have an independent confirmation on correctness.

weinbe2:
@cpviolator, if you haven't already, can you do a quick check with a single GPU build? No need to roll back this merge, if there's an issue just do a quick hotfix branch and we'll get it plugged in quickly.

cpviolator:
Turning on jitify breaks all the tests! (blas_test, gauge_alg_test, heatbath_test, pack_test, su3_test) Is this a known issue?

Issue #791, opened 2019-04-11
Support clang address sanitizer


maddyscientist:
It would be nice to support clang's address sanitizer.  While this would only work with the CPU code compiled with clang, this would nevertheless help track down errors.  Integrating this will require integrating this with our cmake build system.

mathiaswagner:
For reference: https://blog.kitware.com/ctest-cdash-add-support-for-new-dynamic-analysis-tools/

PR #799, opened 2019-05-03
Hotfix/milc bnl

### LINKED ISSUE #744: Add MG verify test that checks correctness of Schur operator
At the moment the MG verification tests do not check for the correctness of the even-odd preconditioned operator that is constructed.  We should add a verification test for this, since this is a significant gap in testing coverage.

As noted by @weinbe2 we can just compare the constructed operator against a manual construction of the preconditioned operator by applying the relevant hopping terms and coarse clover inverses.

Assigning to @weinbe2.....

maddyscientist:
This is primarily bug-fix pull request that improves the robustness of QUDA that have been found as a result of improved testing infrastructure
* Communicator destruction and allocation is now done after a global barrier to ensure that all communication is complete, removing risk of potential undefined behaviour
* Coarse dslash now encodes the p2p/gdr/policy env args in its `TuneKey`, so that a global retune is not forced when these environment variables change
* Fix bug with zero-copy variant of `cudaColorSpinorField::saveSpinorField`
* Add plaquette computation to multigrid_invert_test
* Fix host residual computation in multigrid_invert_test for all fermion types and solve types
* Fix bug with changing precision of the fat sloppy gauge fields when non-trivial reconstruction is enabled for long links
* Fix single-GPU bug with `setMPICommHandleQuda` when used with the MILC interface
* Remove legacy `tex_id` from Texture object
* Fix minor issues with covariant derivative
* `spinorNoise` function now works on CPU fields (opaquely copying from GPU behind the scenes)
* Deploy hypercubic RNG for quark field creation in invert_test, multigrid_invert_test and `MG` class ensuring we have reproducible RNGs regardless of rank layout
* Add new environment variable `QUDA_DETERMINISTIC_REDUCE=1` which ensures that MPI reductions are deterministic regardless of rank order (helpful for debugging)
* Fixes for some valgrind and address sanitizer warnings
* Fix bug in Dirichlet boundary conditions when using a fused preconditioned operator linear operator
* Improve multigrid verification testing run after the setup completes: run top-down to ensure that the first logical failure is identified first, and added new test for preconditioned operator correctness
* Add support for Address and Undefined sanitizers to QUDA.  Enabled through setting the Cmake build target as `SANITIZE` - enabling identified various bugs and undefined behaviour that would not have been found otherwise
* Fix bug with peer-to-peer copy engine communication in the temporal dimension for full-field operators
* Use `cuPointerGetAttributes` instead of `cudaPointerGetAttributes` to silence false positives from `cuda-memcheck`

Closes #744, #791


CMakeLists.txt
```diff
@@ -602,6 +619,8 @@ set(CMAKE_CUDA_FLAGS_DEVICEDEBUG "${QUDA_NVCC_FLAGS} -G" CACHE STRING
   "Flags used by the C++ compiler during device-debug builds.")
 set(CMAKE_CUDA_FLAGS_DEBUG "${QUDA_NVCC_FLAGS} -g -DHOST_DEBUG -G" CACHE STRING
   "Flags used by the C++ compiler during full (host+device) debug builds." )
+set(CMAKE_CUDA_FLAGS_SANITIZE "${QUDA_NVCC_FLAGS} -g -lineinfo -DHOST_DEBUG -Xcompiler -fsanitize=address,-fsanitize=undefined" CACHE STRING
```
mathiaswagner: Shouldn't the host compiler flags be propagated from the CXX_FLAGS here, like we do for the other flags?

maddyscientist: Yes that's true.  If the propagate option is disable though then it wouldn't be.  Remind me, what was the rationale in having the option to not propagate?

mathiaswagner: Not sure. Might be a legacy thing as the old _FindCUDA_ module before native CUDA support had that. Only other reason I can think of right now is if one CXX flag causes trouble with CUDA compilation to disable it and have the option to set it manually.

lib/color_spinor_util.cu
```diff
@@ -434,10 +433,9 @@ namespace quda {
     for (int s = 0; s < Ns; s++) {
       printfQuda("x = %u, s = %d, { ", i, s);
```
weinbe2: In light of the below changes, should this also be a vanilla `printf`? Also, do we want to do one "big" `sprintf` build then `printf` so we can guarantee everything for a single coordinate/color/spin gets printed together (for multi-rank cases)?

mathiaswagner: Also, any good argument for using `printf` here and `std::out` above ?

lib/spinor_noise.cu
```diff
@@ -186,12 +162,27 @@ namespace quda {
...
   }
 
-  void spinorNoise(ColorSpinorField &src, RNG& randstates, QudaNoiseType type)
+  void spinorNoise(ColorSpinorField &src_, RNG& randstates, QudaNoiseType type)
   {
-    switch (src.Precision()) {
-    case QUDA_DOUBLE_PRECISION: spinorNoise<double>(src, randstates, type); break;
-    case QUDA_SINGLE_PRECISION: spinorNoise<float>(src, randstates, type); break;
-    default: errorQuda("Precision %d not implemented", src.Precision());
+    // if src is a CPU field then create GPU field
```
mathiaswagner: does this implicitly transform a CPU field into a GPU field?

lib/comm_common.cpp
```diff
@@ -432,12 +432,18 @@ int comm_coord(int dim)
...
+  CUmemorytype memType;
+  void *attrdata[] = {(void *)&memType};
+  CUpointer_attribute attributes[2] = {CU_POINTER_ATTRIBUTE_MEMORY_TYPE};
+  CUresult err = cuPointerGetAttributes(1, attributes, attrdata, (CUdeviceptr)buffer);
+  if (err != CUDA_SUCCESS) {
+    const char *str;
+    cuGetErrorName(err, &str);
+    errorQuda("cuPointerGetAttributes returned error %s", str);
+  }
+  return (memType != CU_MEMORYTYPE_DEVICE);
```
mathiaswagner: While we currently only use device and host, for unified memory this would return true.

lib/comm_qmp.cpp
```diff
@@ -301,7 +333,29 @@ void comm_allreduce_min(double* data)
 
 void comm_allreduce_array(double* data, size_t size)
 {
-  QMP_CHECK( QMP_sum_double_array(data, size) );
+  if (!deterministic_reduce) {
```
mathiaswagner: is there a simple way to avoid the code duplication when we fall back to MPI?

lib/comm_qmp.cpp
```diff
@@ -340,8 +394,23 @@ void comm_abort(int status)
   QMP_abort(status);
 }
 
-const char* comm_dim_partitioned_string() {
-  return partition_string;
+static char partition_override_string[16];
+
+const char* comm_dim_partitioned_string(const int *comm_dim_override)
```
mathiaswagner: should this move to comm_common ?

lib/copy_quda.cu
```diff
@@ -137,47 +137,47 @@ namespace quda {
 	  }
 	} else if (dst.Precision() == QUDA_DOUBLE_PRECISION && src.Precision() == QUDA_SINGLE_PRECISION) {
 	  if (src.Nspin() == 4){
-	    Spinor<float4, float4, 6, 0, 0> src_tex(src);
+	    Spinor<float4, float4, 6, 0> src_tex(src);
 	    Spinor<float4, double2, 6, 1> dst_spinor(dst);
-	    CopyCuda<float4, 6, Spinor<float4, double2, 6, 1>, Spinor<float4, float4, 6, 0, 0> >
-	      copy(dst_spinor, src_tex, src.Volume(), partitions);
+	    CopyCuda<float4, 6, decltype(dst_spinor), decltype(src_tex)>
```
mathiaswagner: :-)

lib/color_spinor_util.cu
rbabich: Unless things have changed, `printfQuda()` should be used for all output, which ensures that the "prefix" and "outfile" options of `setVerbosityQuda()` are respected.

CMakeLists.txt
maddyscientist: So what do you think I should do here?  I'm happy to remove the explicit passing of the sanitize flags since they'll be propagated by default anyway.

mathiaswagner:
Leave it for now. We can fix that up later. Maybe just add a TODO. SANITIZE is probably a option most will never use.

There are a couple of things to clean up at some point but not urgent enough currently.


lib/color_spinor_util.cu
maddyscientist: @weinbe2 Good catch, thank you.

@rbabich The motivation here was that `printfQuda` always only prints from process 0.  In this specific case we want to printing from all processes since it's a specific debugging routine.  However, looking at the source code now, I see that it is possible to achieve what we want with `getRankVerbosity`.  I have no memory of adding that 😄 , but it appears that I did a few years ago - with this in mind I'll change this routine back to using `printfQuda` leaving the runtime opt-in if multiple processes with the `QUDA_RANK_VERBOSITY` env.

lib/spinor_noise.cu
maddyscientist: Not really - it creates a temporary GPU field, runs the hypercubic RNG, copies back into the original CPU field and then deletes the GPU field.

lib/multigrid.cpp
```diff
@@ -187,32 +177,35 @@ namespace quda {
...
     postTrace();
+    pushVerbosity(param.mg_global.verbosity[level]);
+    setOutputPrefix(prefix);
   }
 
-  void MG::destroySmoother() {
+  void MG::popLevel(int level) const
+  {
+    popVerbosity();
+    setOutputPrefix("");
```
weinbe2: Why does this get set to the empty string instead of the previous output prefix? Should we have a `pushOutputPrefix`/`popOutputPrefix` for MG?

lib/comm_common.cpp
maddyscientist: Yes, for now I'll add an explicit check for unified memory and through an error if this is encountered.  While we may want to support unified memory on fields, we should *never* user unified memory for MPI buffers.

mathiaswagner:
Agreed.


lib/multigrid.cpp
```diff
@@ -599,22 +609,28 @@ namespace quda {
...
+        for (int x=0; x<x_coarse->Volume(); x++) tmp1->PrintVector(x);
 
-    if (getVerbosity() >= QUDA_VERBOSE) printfQuda("Vector norms Emulated=%e Native=%e ", norm2(*x_coarse), norm2(*r_coarse));
+        if (getVerbosity() >= QUDA_VERBOSE) printfQuda("actual\n");
+        for (int x=0; x<r_coarse->Volume(); x++) tmp2->PrintVector(x);
+      }
+      comm_barrier();
+    }
+    setOutputPrefix(prefix);
+#endif
```
weinbe2: Perhaps we should have two versions of `printVector`: one that takes a checkerboard index and prints the values for just this site, and then one that takes zero arguments and prints all of the sites using this pattern (i.e., loops over comm ranks, inner loop over all checkerboard sites). We could also give it a different name to avoid confusion. (`printEntireVector`?)

tests/invert_test.cpp
```diff
@@ -350,7 +345,7 @@ int main(int argc, char **argv)
     read_gauge_field(latfile, gauge, gauge_param.cpu_prec, gauge_param.X, argc, argv);
     construct_gauge_field(gauge, 2, gauge_param.cpu_prec, &gauge_param);
   } else { // else generate a random SU(3) field
-    construct_gauge_field(gauge, 1, gauge_param.cpu_prec, &gauge_param);
+    construct_gauge_field(gauge, 0, gauge_param.cpu_prec, &gauge_param);
```
weinbe2: While we're here, should we just add support for a GPU-generated random gauge field?

lib/multigrid.cpp
maddyscientist: Well this was consistent with prior behaviour, but I'll add new `pushOutputPrefix` and `popOutputPrefix` functions to address this and simplify things.

lib/multigrid.cpp
maddyscientist: Definitely for a subsequent pull.  @weinbe2 can you add an issue for this?

tests/invert_test.cpp
maddyscientist: For a subsequent pull: #804 tracks this.

lib/comm_qmp.cpp
maddyscientist: I've halved the amount of duplication in my next commit.

lib/comm_qmp.cpp
```diff
@@ -301,10 +220,16 @@ int comm_query(MsgHandle *mh)
   return (QMP_is_complete(mh->handle) == QMP_TRUE);
 }
 
+template <typename T>
```
mathiaswagner: this is not used with QMP or do I miss that somewhere? Maybe an incomplete commit?

weinbe2: Yeah, this looks like an incomplete commit, the `std::accumulate`s haven't been replaced further in the file.

lib/multigrid.cpp
weinbe2: Done, #807

lib/comm_qmp.cpp
maddyscientist: Fixed

weinbe2:
All looks good to me!

PR #803, opened 2019-05-10
jitified the qcharge routine, removed spurious location param from ga…

cpviolator:
…uge field routines, renames some files


include/kernels/gauge_qcharge.cuh
```diff
@@ -0,0 +1,50 @@
+#include <gauge_field_order.h>
+#include <index_helper.cuh>
+#include <quda_matrix.h>
+#include <su3_project.cuh>
+
+#include <cub/cub.cuh>
```
maddyscientist: We should not be including cub headers directly, just the cub_helper.cuh should be sufficient.

lib/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ set (QUDA_OBJS
   dirac_domain_wall_4d.cpp dirac_mobius.cpp dirac_twisted_clover.cpp
   dirac_twisted_mass.cpp tune.cpp
   llfat_quda.cu gauge_force.cu gauge_random.cu
-  field_strength_tensor.cu clover_quda.cu dslash_quda.cu
+  gauge_field_strength_tensor.cu clover_quda.cu dslash_quda.cu
```
maddyscientist: Why the change in file name?

lib/interface_quda.cpp
```diff
@@ -5548,10 +5547,8 @@ void performAPEnStep(unsigned int nSteps, double alpha)
   GaugeFieldParam gParam(*gaugeSmeared);
   auto *cudaGaugeTemp = new cudaGaugeField(gParam);
 
-  if (getVerbosity() == QUDA_VERBOSE) {
-    double3 plq = plaquette(*gaugeSmeared, QUDA_CUDA_FIELD_LOCATION);
-    printfQuda("Plaquette after 0 APE steps: %le %le %le\n", plq.x, plq.y, plq.z);
-  }
+  double3 plaq = plaquette(*gaugeSmeared);
+  printfQuda("Plaquette after 0 APE steps: %le %le %le\n", plaq.x, plaq.y, plaq.z);
```
maddyscientist: These `printQuda` statements need to have verbosity guards around them since if the verbosity is `QUDA_SILENT` then QUDA should not print anything.  Though the prior ones do look buggy since they didn't use `>=`, perhaps `if (getVerbosity() >= QUDA_SUMMARIZE) { ... }` ?

lib/interface_quda.cpp
```diff
@@ -5835,7 +5822,7 @@ void contract(const cudaColorSpinorField x, const cudaColorSpinorField y, void *
   }
 }
 
-double qChargeCuda ()
+double qChargeQuda()
```
maddyscientist: I think you missed multigrid_evolve_test uses this function.

lib/gauge_qcharge.cu
```diff
@@ -0,0 +1,120 @@
...
+        arg.result_h[0] = 0.;
+        TuneParam tp = tuneLaunch(*this, getTuning(), getVerbosity());
+#ifdef JITIFY
+        using namespace jitify::reflection;
+        jitify_error = program->kernel("quda::qChargeComputeKernel")
+                         .instantiate(Type<Float>(), Type<Arg>())
+                         .configure(tp.grid, tp.block, tp.shared_bytes, stream)
+                         .launch(arg);
+#else
+        // DMH: Hardcoded 32 blockSize for reduction kernel
```
maddyscientist: Curious: why not keep the prior `LAUNCH_KERNEL` here which instantiates all block sizes.

include/kernels/gauge_qcharge.cuh
cpviolator: Relic from the original code. Fixed the other instances too.

lib/CMakeLists.txt
cpviolator: Exactly. I'll put all the gauge algorithms that use the same instantiation structure in files prepended with `gauge_` like the `dslash_` ones. Then they'll all be contiguous in the GitHub page too.

lib/interface_quda.cpp
cpviolator: Thanks for the catch.

lib/gauge_qcharge.cu
cpviolator: Relic from an attempt create a function that returned individual q(x) values. Thanks for the catch.

lib/interface_quda.cpp
cpviolator: Changed to >= QUDA_SUMMARIZE.

maddyscientist:
Looks all good to me now.  Thanks @cpviolator.  We can just wait until build tests finish before merging.

cpviolator:
My bad. That last push was just clang formatting :/

mathiaswagner:
I fixed the jenkins issues so builds look good.

PR #809, opened 2019-05-16
Use NVTX3 if available

mathiaswagner:
This is header only and avoid and external dependency on libnvToolsExt.
Requires CUDA 10 but will just fall back to old NVTX behavior if no `nvtx3/` headers are found.


lib/milc_interface.cpp
```diff
@@ -61,19 +66,20 @@ using namespace quda::fermion_force;
...
-    printfQuda("QUDA_MILC_INTERFACE: %s (called) \n", func);
-    PUSH_RANGE(func, 1);
-  } else {
-    printfQuda("QUDA_MILC_INTERFACE: %s (return) \n", func);
-    POP_RANGE;
+  if (verb >= QUDA_VERBOSE) {
+    if (start) {
+      printfQuda("QUDA_MILC_INTERFACE: %s (called) \n", func);
+    } else {
+      printfQuda("QUDA_MILC_INTERFACE: %s (return) \n", func);
```
maddyscientist: this looks like it breaks the behaviour of the interface mark up so that a interface routine

mathiaswagner: No, let me fix that ...

PR #814, opened 2019-05-21
Topological charge density function

cpviolator:
This PR adds some functionality to the topological charge routines. One can now pass an array to the function, and QUDA passes back the individual q(x)/(4xPixPi) values in Even/Odd order. Useful for \eta prime calculations.


cpviolator:
Good to go here?

PR #816, opened 2019-05-23
Feature/deflated mg

cpviolator:
A stable Thick Restarted Lanczos, with preliminary support for deflated multigrid.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -374,7 +374,7 @@ if(QUDA_QMP)
       GIT_TAG qmp2-5-1
       GIT_SHALLOW YES
       PREFIX usqcd
-      CONFIGURE_COMMAND <SOURCE_DIR>/configure --with-qmp-comms-type=MPI --prefix=<INSTALL_DIR> CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} CFLAGS=-std=c99 CXXFLAGS=-std=c99
+      CONFIGURE_COMMAND <SOURCE_DIR>/configure --with-qmp-comms-type=MPI --prefix=<INSTALL_DIR> CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} CFLAGS=-std=c99
       BUILD_COMMAND  make
       INSTALL_COMMAND make install
       LOG_INSTALL ON
@@ -425,7 +425,7 @@ if(QUDA_QIO)
       GIT_TAG qio2-5-0
       GIT_SHALLOW YES
       PREFIX usqcd 
-      CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} CFLAGS=-std=c99 CXXFLAGS=-std=c99 <SOURCE_DIR>/configure --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
+      CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} CFLAGS=-std=c99 <SOURCE_DIR>/configure --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
       BUILD_COMMAND  make
       INSTALL_COMMAND make install
       DEPENDS QMP
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -414,76 +414,37 @@ if(QUDA_MAGMA)
   endif()  
 endif(QUDA_MAGMA)
 
+# This selects arpack or parpack for Multi GPU
 if(QUDA_ARPACK)
+  enable_language(Fortran)
   add_definitions(-DARPACK_LIB)
   find_package(PkgConfig REQUIRED)
+  enable_language(Fortran)
   
-  # Check if the library is installed already
+  # We always need the serial library
   pkg_check_modules(ARPACK QUIET arpack)
-  if(QUDA_MPI OR QUDA_QMP)
-    pkg_check_modules(PARPACK QUIET parpack)
-  endif()
-  
-  if(QUDA_ARPACK_HOME)
-    # User has specified a path, use it
+  if(NOT ARPACK_FOUND OR QUDA_ARPACK_HOME)
     find_library(ARPACK arpack PATH ${QUDA_ARPACK_HOME})
-    if(QUDA_MPI OR QUDA_QMP)
-      find_library(PARPACK parpack PATH ${QUDA_ARPACK_HOME})
-    endif()
-  endif()
-    
-  if(ARPACK_FOUND AND NOT QUDA_ARPACK_HOME)
-    # ARPACK is installed, user did not give specific path
+  else()
     find_library(ARPACK ${ARPACK_LIBRARIES} PATH ${ARPACK_LIBRARY_DIRS})
-    if(QUDA_MPI OR QUDA_QMP)
-      find_library(PARPACK ${PARPACK_LIBRARIES} PATH ${PARPACK_LIBRARY_DIRS})
-    endif()
   endif()
-  
-  if(NOT ARPACK_FOUND AND NOT QUDA_ARPACK_HOME)
-    # Download and install (P)ARPACK
-    set(QUDA_ARPACK_CONFIG CC=${C_COMPILER} CXX=${CXX_COMPILER} F77=${FORTRAN_COMPILER} --prefix=<INSTALL_DIR>)
-    if(QUDA_MPI OR QUDA_QMP)
-      set(QUDA_ARPACK_CONFIG CC=${C_COMPILER} CXX=${CXX_COMPILER} F77=${FORTRAN_COMPILER} MPICC=${C_COMPILER} MPICXX=${CXX_COMPILER} MPIF77=${MPI_FORTRAN_COMPILER} --enable-mpi --prefix=<INSTALL_DIR>)
-    endif()
-    ExternalProject_Add(ARPACK-NG
-      GIT_REPOSITORY https://github.com/opencollab/arpack-ng.git
-      GIT_TAG 3.7.0
-      GIT_SHALLOW YES
-      PREFIX arpack-ng
-      CONFIGURE_COMMAND <SOURCE_DIR>/configure ${QUDA_ARPACK_CONFIG}
-      BUILD_COMMAND  make
-      BUILD_IN_SOURCE ON
-      LOG_MERGED_STDOUTERR ON
-      LOG_INSTALL ON
-      LOG_BUILD ON
-      LOG_DOWNLOAD ON
-      LOG_OUTPUT_ON_FAILURE ON
-      )        
-  endif()
-
-  ExternalProject_Add_Step(ARPACK-NG bootstrap
-    COMMAND ./bootstrap
-    WORKING_DIRECTORY <SOURCE_DIR>
-    DEPENDERS configure
-    DEPENDEES download
-    )  
-  
-  ExternalProject_Get_Property(ARPACK-NG INSTALL_DIR)
-  set(QUDA_ARPACK_HOME ${INSTALL_DIR})
-  set(ARPACK "-L${INSTALL_DIR}/lib -larpack")
-  set(PARPACK "-L${INSTALL_DIR}/lib -lparpack")
   include_directories(SYSTEM ${QUDA_ARPACK_HOME}/SRC)
   include_directories(SYSTEM ${QUDA_ARPACK_HOME}/UTIL)
-
-  # Include the parallel library if required
+  
+  
+  # Link the parallel library if required
   if(QUDA_MPI OR QUDA_QMP)
+    pkg_check_modules(PARPACK QUIET parpack)
+    if(NOT PARPACK_FOUND OR QUDA_ARPACK_HOME)
+      find_library(PARPACK parpack PATH ${QUDA_ARPACK_HOME})
+    else()
+      find_library(PARPACK ${PARPACK_LIBRARIES} PATH ${PARPACK_LIBRARY_DIRS})
+    endif()
     include_directories(SYSTEM ${QUDA_ARPACK_HOME}/PARPACK/SRC/MPI)
     include_directories(SYSTEM ${QUDA_ARPACK_HOME}/PARPACK/UTIL/MPI)
   endif()
   
 endif(QUDA_ARPACK)
-
 
 # set which precisions to enable
 add_definitions(-DQUDA_PRECISION=${QUDA_PRECISION})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -374,7 +374,7 @@ if(QUDA_QMP)
       GIT_TAG qmp2-5-1
       GIT_SHALLOW YES
       PREFIX usqcd
-      CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER}  <SOURCE_DIR>/configure --with-qmp-comms-type=MPI --prefix=<INSTALL_DIR>
+      CONFIGURE_COMMAND <SOURCE_DIR>/configure --with-qmp-comms-type=MPI --prefix=<INSTALL_DIR> CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} CFLAGS=-std=c99 CXXFLAGS=-std=c99
       BUILD_COMMAND  make
       INSTALL_COMMAND make install
       LOG_INSTALL ON
@@ -425,7 +425,7 @@ if(QUDA_QIO)
       GIT_TAG qio2-5-0
       GIT_SHALLOW YES
       PREFIX usqcd 
-      CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} <SOURCE_DIR>/configure --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
+      CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} CFLAGS=-std=c99 CXXFLAGS=-std=c99 <SOURCE_DIR>/configure --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
       BUILD_COMMAND  make
       INSTALL_COMMAND make install
       DEPENDS QMP
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -414,30 +414,70 @@ if(QUDA_MAGMA)
   endif()  
 endif(QUDA_MAGMA)
 
-# This selects arpack or parpack for Multi GPU
 if(QUDA_ARPACK)
   add_definitions(-DARPACK_LIB)
   find_package(PkgConfig REQUIRED)
-
-  # We always need the serial library
+  
+  # Check if the library is installed already
   pkg_check_modules(ARPACK QUIET arpack)
-  if(NOT ARPACK_FOUND OR QUDA_ARPACK_HOME)
-    find_library(ARPACK arpack PATH ${QUDA_ARPACK_HOME})
-  else()
-    find_library(ARPACK ${ARPACK_LIBRARIES} PATH ${ARPACK_LIBRARY_DIRS})
+  if(QUDA_MPI OR QUDA_QMP)
+    pkg_check_modules(PARPACK QUIET parpack)
   endif()
+  
+  if(QUDA_ARPACK_HOME)
+    # User has specified a path, use it
+    find_library(ARPACK arpack PATH ${QUDA_ARPACK_HOME})
+    if(QUDA_MPI OR QUDA_QMP)
+      find_library(PARPACK parpack PATH ${QUDA_ARPACK_HOME})
+    endif()
+  endif()
+    
+  if(ARPACK_FOUND AND NOT QUDA_ARPACK_HOME)
+    # ARPACK is installed, user did not give specific path
+    find_library(ARPACK ${ARPACK_LIBRARIES} PATH ${ARPACK_LIBRARY_DIRS})
+    if(QUDA_MPI OR QUDA_QMP)
+      find_library(PARPACK ${PARPACK_LIBRARIES} PATH ${PARPACK_LIBRARY_DIRS})
+    endif()
+  endif()
+  
+  if(NOT ARPACK_FOUND AND NOT QUDA_ARPACK_HOME)
+    # Download and install (P)ARPACK
+    set(QUDA_ARPACK_CONFIG CC=${C_COMPILER} CXX=${CXX_COMPILER} F77=${FORTRAN_COMPILER} --prefix=<INSTALL_DIR>)
+    if(QUDA_MPI OR QUDA_QMP)
+      set(QUDA_ARPACK_CONFIG CC=${C_COMPILER} CXX=${CXX_COMPILER} F77=${FORTRAN_COMPILER} MPICC=${C_COMPILER} MPICXX=${CXX_COMPILER} MPIF77=${MPI_FORTRAN_COMPILER} --enable-mpi --prefix=<INSTALL_DIR>)
+    endif()
+    ExternalProject_Add(ARPACK-NG
+      GIT_REPOSITORY https://github.com/opencollab/arpack-ng.git
+      GIT_TAG 3.7.0
+      GIT_SHALLOW YES
+      PREFIX arpack-ng
+      CONFIGURE_COMMAND <SOURCE_DIR>/configure ${QUDA_ARPACK_CONFIG}
+      BUILD_COMMAND  make
+      BUILD_IN_SOURCE ON
+      LOG_MERGED_STDOUTERR ON
+      LOG_INSTALL ON
+      LOG_BUILD ON
+      LOG_DOWNLOAD ON
+      LOG_OUTPUT_ON_FAILURE ON
+      )        
+  endif()
+
+  ExternalProject_Add_Step(ARPACK-NG bootstrap
+    COMMAND ./bootstrap
+    WORKING_DIRECTORY <SOURCE_DIR>
+    DEPENDERS configure
+    DEPENDEES download
+    )  
+  
+  ExternalProject_Get_Property(ARPACK-NG INSTALL_DIR)
+  set(QUDA_ARPACK_HOME ${INSTALL_DIR})
+  set(ARPACK "-L${INSTALL_DIR}/lib -larpack")
+  set(PARPACK "-L${INSTALL_DIR}/lib -lparpack")
   include_directories(SYSTEM ${QUDA_ARPACK_HOME}/SRC)
   include_directories(SYSTEM ${QUDA_ARPACK_HOME}/UTIL)
 
-
-  # Link the parallel library if required
+  # Include the parallel library if required
   if(QUDA_MPI OR QUDA_QMP)
-    pkg_check_modules(PARPACK QUIET parpack)
-    if(NOT PARPACK_FOUND OR QUDA_ARPACK_HOME)
-      find_library(PARPACK parpack PATH ${QUDA_ARPACK_HOME})
-    else()
-      find_library(PARPACK ${PARPACK_LIBRARIES} PATH ${PARPACK_LIBRARY_DIRS})
-    endif()
     include_directories(SYSTEM ${QUDA_ARPACK_HOME}/PARPACK/SRC/MPI)
     include_directories(SYSTEM ${QUDA_ARPACK_HOME}/PARPACK/UTIL/MPI)
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,10 @@
 # basic setup for cmake
-cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.11 FATAL_ERROR)
+
+if(POLICY CMP0074)
+  cmake_policy(SET CMP0074 NEW)
+endif()
+
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_INCLUDE_DIRECTORIES_PROJECT_BEFORE ON)
 set(CMAKE_COLOR_MAKEFILE ON)
@@ -21,7 +26,7 @@ find_package(Git)
 if(DEFINED ENV{QUDA_BUILD_TYPE})
   set(DEFBUILD $ENV{QUDA_BUILD_TYPE})
 else()
-  set(DEFBUILD "RELEASE")
+  set(DEFBUILD "DEVEL")
 endif()
 
 # IF(GIT_FOUND)
@@ -171,6 +176,7 @@ set(QUDA_MULTIGRID OFF CACHE BOOL "build multigrid solvers")
 set(QUDA_BLOCKSOLVER OFF CACHE BOOL "build block solvers")
 set(QUDA_USE_EIGEN OFF CACHE BOOL "use EIGEN library (where optional)")
 set(QUDA_DOWNLOAD_EIGEN ON CACHE BOOL "Download Eigen")
+set(QUDA_DOWNLOAD_USQCD OFF CACHE BOOL "Download USQCD software as requested by QUDA_QMP / QUDA_QIO")
 
 
 option(QUDA_GENERATE_DOXYGEN "generate doxygen documentation")
@@ -203,9 +209,7 @@ mark_as_advanced(QUDA_CXX_STANDARD)
 
 mark_as_advanced(QUDA_JITIFY)
 
-#######################################################################
-# options that are not exposed at all because only one option exists
-set(CPU_ARCH ${CMAKE_HOST_SYSTEM_PROCESSOR})
+
 
 
 #######################################################################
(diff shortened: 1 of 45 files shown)


CMakeLists.txt
```diff
@@ -413,23 +418,30 @@ endif(QUDA_MAGMA)
 if(QUDA_ARPACK)
   add_definitions(-DARPACK_LIB)
   find_package(PkgConfig REQUIRED)
-  if(QUDA_QMP OR QUDA_MPI)
+
```
maddyscientist: Can you explain why the changes to CMake here with respect to `ARPACK`?

CMakeLists.txt
```diff
@@ -589,7 +601,7 @@ endif(QUDA_COVDEV)
 set(CMAKE_F_FLAGS -std=c99 CACHE STRING "Fortran Flags")
 
 # derive whether we need to build the fortran interface
-if(QUDA_INTERFACE_TIFR OR QUDA_INTERFACE_BQCD)
+if(QUDA_INTERFACE_TIFR OR QUDA_INTERFACE_BQCD OR QUDA_ARPACK)
```
maddyscientist: Why does `QUDA_ARPACK` being enabled mean we build QUDA's Fortran interface?

include/eigensolve_quda.h
```diff
@@ -0,0 +1,224 @@
...
+    double *residua;
+
+    // Device side vector workspace
+    std::vector<ColorSpinorField *> r;
+    std::vector<ColorSpinorField *> d_vecs_tmp;
+
+    Complex *Qmat;
+
+public:
+    EigenSolver(QudaEigParam *eig_param, TimeProfile &profile);
```
maddyscientist: doxygen needed for the constructor and other member functions

include/enum_quda.h
```diff
@@ -129,11 +129,26 @@ extern "C" {
   } QudaInverterType;
 
   typedef enum QudaEigType_s {
-    QUDA_LANCZOS, //Normal Lanczos eigen solver
-    QUDA_IMP_RST_LANCZOS, //implicit restarted lanczos solver
-    QUDA_INVALID_TYPE = QUDA_INVALID_ENUM
+    QUDA_LANCZOS,           // Lanczos solver
```
maddyscientist: This is probably a good chance to clean up the enums here.  Perhaps `QUDA_EIG_LANCZOS`, QUDA_EIG_ARNOLDI`, etc. here.

Also, I'd prefer `QUDA_SPECTRUM_SR_EIG` over `QUDA_SR_EIG_SPRECTRUM`.  Not all enums follow this convention of increasing specificity as we move from left to right, but that was a poor early design choice.  Since this is a new type, let's get this one correct 😉

include/enum_quda_fortran.h
```diff
@@ -126,9 +126,12 @@
...
 
 #define QudaEigType integer(4)
-#define QUDA_LANCZOS 0 //Normal Lanczos eigen solver
-#define QUDA_IMP_RST_LANCZOS 1 //implicit restarted lanczos solver
-#define QUDA_INVALID_TYPE QUDA_INVALID_ENUM
+#define QUDA_LANCZOS 0           // Lanczos solver
+#define QUDA_IMP_RST_LANCZOS 1   // Implicitly restarted lanczos solver
+#define QUDA_THICK_RST_LANCZOS 2 // Thick restarted lanczos solver
+#define QUDA_ARNOLDI 3           // Arnoldi solver
+#define QUDA_IMP_RST_ARNOLDI 4   // Implicitly restarted Arnoldi solver
```
maddyscientist: Do we need the `SPECTRUM` enums defined in the Fortran interface?

include/invert_quda.h
```diff
@@ -964,17 +1055,55 @@ namespace quda {
...
+public:
+    PreconditionedSolver(Solver &solver, const Dirac &dirac, SolverParam &param, TimeProfile &profile,
+                         const char *prefix) :
+      Solver(param, profile),
+      solver(&solver),
+      dirac(dirac),
+      prefix(prefix)
+    {
+
+      // DMH
```
maddyscientist: Big commented out block here?

include/invert_quda.h
```diff
@@ -1139,16 +1267,22 @@ namespace quda {
...
 
-  public:
-
+public:
     IncEigCG(DiracMatrix &mat, DiracMatrix &matSloppy, DiracMatrix &matPrecon, SolverParam &param, TimeProfile &profile);
 
     virtual ~IncEigCG();
 
+    /**
+       @breif Expands deflation space.
```
maddyscientist: `brief`

include/multigrid.h
```diff
@@ -320,7 +303,7 @@ namespace quda {
     /**
        @brief Dump the null-space vectors to disk.  Will recurse dumping all levels.
     */
-    void dumpNullVectors() const;
+    void dumpNullVectors();
```
maddyscientist: Why did this routine lose its `const` qualifier?

include/multigrid.h
```diff
@@ -362,17 +345,34 @@ namespace quda {
...
+       @brief Load the null space vectors in from file
+       @param B Loaded null-space vectors (pre-allocated)
+    */
+    void loadVectors(std::vector<ColorSpinorField *> &B);
+
+    /**
+       @brief Save the null space vectors in from file
+       @param B Save null-space vectors from here
+    */
+    void saveVectors(const std::vector<ColorSpinorField *> &B);
```
maddyscientist: Lost `const` qualifier as well

include/quda_arpack_interface.h
```diff
@@ -7,10 +7,7 @@
 #include <color_spinor_field.h>
 #include <dirac_quda.h>
 #include <vector>
-
-//#ifdef PRIMME_LIB
-//#include "primme.h"
-//#endif
+#include <algorithm>
```
maddyscientist: Can you move these `std` headers to where they are needed rather than blanket inclusion in this header?

lib/check_params.h
```diff
@@ -452,14 +475,15 @@ void printQudaInvertParam(QudaInvertParam *param) {
   }
 #endif
 
-
 #ifdef INIT_PARAM
   P(use_init_guess, QUDA_USE_INIT_GUESS_NO); //set the default to no
-  //P(compute_null_vector, QUDA_COMPUTE_NULL_VECTOR_NO); //set the default to no
+  // P(deflated_guess, QUDA_DEFLATED_GUESS_NO); //set the default to no
```
maddyscientist: any reason these are not deleted since they are commented out?

include/ritz_quda.h
```diff
@@ -10,6 +10,8 @@
 #include <dirac_quda.h>
 #include <typeinfo>
 
+#if 0
```
maddyscientist: Looks like this file can be deleted

CMakeLists.txt
cpviolator: In the old code, if QMP or MPI was deleted, only `libparpack.a` would be linked. ARPACK needs both `libparpack.a` and `libarpack.a` to work with MPI, so we always link with `libarpack.a`, and additionally link with `libparpack.a` for parallel builds.

CMakeLists.txt
cpviolator: Probably doesn't need it. I'll test it out without.

include/quda.h
```diff
@@ -814,8 +870,28 @@ extern "C" {
...
+
+  /**
+   * Perform the eigensolve using ARPACK. The problem matrix is defined by the
+   * invert param, the mode of solution is specified by the eig param.
+   * It is assumed that the gauge field has already been loaded via loadGaugeQuda().
+   * @param h_els  Host side eigenvalues
+   * @param h_evs  Host side eigenvectors
+   * @param param  Contains all metadata regarding the type of solve.
+   */
+  void eigensolveARPACK(void *h_evals, void *h_evecs, QudaEigParam *param);
```
maddyscientist: If this is a public function to be called from applications then this should be `eigensolveARPACKQuda` to follow the usual naming convention.  Perhaps it should just be subsumed into `eigensolveQuda` though as an option to use `ARPACK` instead of native QUDA?

lib/solver.cpp
```diff
@@ -239,6 +240,117 @@ namespace quda {
     }
   }
 
+  /*
```
maddyscientist: what's the point of this (since it's commented out)?

lib/inv_eigcg_quda.cpp
```diff
@@ -21,189 +21,209 @@
...
-//special types needed for compatibility with QUDA blas:
-   using RowMajorDenseMatrix = Matrix<Complex, Dynamic, Dynamic, RowMajor>;
-
-   static int max_eigcg_cycles = 4;//how many eigcg cycles do we allow?
-
-
-   enum  class libtype {eigen_lib, magma_lib, lapack_lib, mkl_lib};
-
-   class EigCGArgs{
+  using namespace blas;
```
maddyscientist: This file, inv_eigcg_quda.cpp has lots of changes here, but this pull isn't really focused on EigCG.  What is being changed in the algorithm here?

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,1510 @@
...
+      sum += s;
+      blas::caxpy(-s, *vecs[i], *rvec[0]);
+    }
+    time_ += clock();
+    time_mb += time_;
+    return sum;
+  }
+
+  // Orthogonalise r against V_[j]
+  Complex EigenSolver::blockOrthogonalise(std::vector<ColorSpinorField *> vecs, std::vector<ColorSpinorField *> rvec,
```
maddyscientist: As much as it pains me, probably should be consistent in naming with other uses of orthogonalize in the library

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,1510 @@
...
+
+    // 2. Perform block caxpy: V_i * (L_i)^{-1} * A_i
+    for (int i = 0; i < n_defl; i++) { s[i] /= evals[i].real(); }
+
+    // 3. Accumulate sum vec_defl = Sum_i V_i * (L_i)^{-1} * A_i
+    blas::zero(*vec_defl[0]);
+    blas::caxpy(s, eig_vecs_ptr, vec_defl);
+
+    // Orthonormality check in deflation.
+    for (int i = 0; i < n_defl; i++)
```
maddyscientist: Looks like this can be deleted?

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,1510 @@
...
+    // Revert to MdagM (or MMdag) mat vec
+    eig_param->use_norm_op = QUDA_BOOLEAN_YES;
+
+    delete tmp[0];
+  }
+
+  // ARPACK INTERAFCE ROUTINES
+  //--------------------------------------------------------------------------
+
+#ifdef ARPACK_LIB
```
maddyscientist: What's the merit in having the ARPACK stuff in here as opposed to in its own file as before?

include/enum_quda_fortran.h
cpviolator: I don't know, I just put them there because the the old enums were there. Should they go?

lib/solver.cpp
cpviolator: This was an attempt to move the vector IO to the solver class. I moved them to the Eigensolver class instead.

lib/inv_eigcg_quda.cpp
cpviolator: I think I made some attempts and integrating the file, and clang formatting changed a lot of stuff. I'm going to checkout the develop version.

include/enum_quda_fortran.h
maddyscientist: For now I would suggest putting the `QUDA_SPECTRUM_XXX` defines in the enum_fortran_fortan.h file as well, for consistency if nothing else.  Eventual plan is to auto generate the `enum_quda_fortran.h` file from the the `enum_quda.h` one once I figure a good way to do that....

mathiaswagner:
@Jenkins test this please

maddyscientist:
@weinbe2 before we merge this, do you want to give it a once over?  I've tested this fairly broadly in the context of prolongation and coarse-grid deflation.

include/multigrid.h
```diff
@@ -445,7 +442,7 @@ namespace quda {
 
   /**
      @brief Calculate preconditioned coarse links and coarse clover inverse field
-     @param Yhat[out] Preconditioned coarse link field
+     @param Yhat[out] Preconidtioned coarse link field
```
weinbe2: Typo! (I'm helping, right?)

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,948 @@
...
+    // b_j = ||r||
+    beta[j] = sqrt(blas::norm2(*r[0]));
+
+    // Prepare next step.
+    // v_{j+1} = r / b_j
+    blas::zero(*v[j + 1]);
+    blas::axpy(1.0 / beta[j], *r[0], *v[j + 1]);
+  }
+
+  void TRLM::reorder(std::vector<ColorSpinorField *> &kSpace)
```
weinbe2: Not that I'd hold up the merge with this, but can we accomplish this with one of the standard C++ algorithms?

maddyscientist: What I think is needed here is to create a `std::pair` of eigenvalues and eigenvectors and then do `std::sort` on the eigenvalues.  As noted, not something to delay the pull on.

include/multigrid.h
maddyscientist: Pushed a fix.

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,948 @@
...
+    residua = (double*)safe_malloc(nKr * sizeof(double));
+    for (int i = 0; i < nKr; i++) { residua[nKr] = 0.0; }
+
+    // Quda MultiBLAS freindly array
+    Qmat = (Complex*)safe_malloc(nEv * nKr * sizeof(Complex));
+
+    // Part of the spectrum to be computed.
+    spectrum = strdup("SR"); // Initialised to stop the compiler warning.
+
+    if (eig_param->use_poly_acc) {
```
weinbe2: In the case of polynomial acceleration: do we want to make anything that's not `SR` throw an error? (except maybe `LR`, which may be worth a warning instead---it's highly suboptimal but it's not really wrong if I remember correctly). 

There's a big asterisk here in that iirc Chebyshev acceleration can be proven to work for spectra that aren't purely real as long as they lie in a well-behaved ellipse in the right half plane (i.e., I believe you can get it to work on a Wilson-clover spectrum with enough experimenting), so that'd motivate just letting this be for now...

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,948 @@
...
+  // Orthogonalise r against V_[j]
+  Complex EigenSolver::orthogonalize(std::vector<ColorSpinorField *> vecs, std::vector<ColorSpinorField *> rvec, int j)
+  {
+    time_ = -clock();
+    Complex s(0.0, 0.0);
+    Complex sum(0.0, 0.0);
+    for (int i = 0; i < j; i++) {
+      s = blas::cDotProduct(*vecs[i], *rvec[0]);
+      sum += s;
+      blas::caxpy(-s, *vecs[i], *rvec[0]);
```
weinbe2: Can we plug in the hybrid modified Gram-Schmidt/Gram-Schmidt ortho here?

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,948 @@
...
+      sigma_old = sigma;
+    }
+    blas::copy(out,*tmp2);
+
+    delete tmp1;
+    delete tmp2;
+  }
+
+  // Orthogonalise r against V_[j]
+  Complex EigenSolver::orthogonalize(std::vector<ColorSpinorField *> vecs, std::vector<ColorSpinorField *> rvec, int j)
```
weinbe2: Is there any reason we have this around since we have `blockOrthogonalize`?

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,948 @@
...
+
+    return;
+  }
+
+  void EigenSolver::loadFromFile(const DiracMatrix &mat, std::vector<ColorSpinorField *> &kSpace,
+                                 std::vector<Complex> &evals)
+  {
+
+    if (getVerbosity() >= QUDA_SUMMARIZE) printfQuda("Loading eigenvectors\n");
+    loadVectors(kSpace, eig_param->vec_infile);
```
weinbe2: This comment would be better put in `qio_field.cpp`, but anyway: I think we should have a more constructive error when a file doesn't exist, especially that loading/saving eigenvectors is becoming a lot more important in QUDA with all these additions. Right now the only error message in `qio_field.cpp` is that "Open file failed!", and I think we should say something more useful (if QIO gives us that info), i.e., if we can discriminate "file doesn't exist" from "file isn't properly formatted", that's a little thing that goes a long way.

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,948 @@
...
+    kSpace.push_back(ColorSpinorField::Create(csParam));
+    csParam.create = QUDA_ZERO_FIELD_CREATE;
+    r.push_back(ColorSpinorField::Create(csParam));
+    //---------------------------------------------------------------------------
+
+    // Convergence and locking criteria
+    double mat_norm = 0.0;
+    double epsilon = DBL_EPSILON;
+    QudaPrecision prec = kSpace[0]->Precision();
+    if (prec == QUDA_DOUBLE_PRECISION) {
```
weinbe2: We should turn this `if` chain into ` switch` statement, and further, we should add `#ifdef` guards that respect the precision `cmake` variable (and throw an error if it's an invalid precision).

lib/eigensolve_quda.cpp
```diff
@@ -0,0 +1,948 @@
...
+        kSpace.push_back(ColorSpinorField::Create(csParam));
+      }
+    }
+
+    for (int i = 0; i < iter_keep; i++) {
+      int k = offset + i;
+      *r[0] = *kSpace[num_locked];
+      blas::ax(ritz_mat[dim * i], *r[0]);
+      *kSpace[k] = *r[0];
+      for (int j = 1; j < dim; j++) { blas::axpy(ritz_mat[i * dim + j], *kSpace[num_locked + j], *kSpace[k]); }
```
weinbe2: Not that I want to add more multi-blas routines, but for performance reasons would it ultimately be a net win to replace this with a batch `caxpy` as opposed to multiple `axpy`s? (Bandwidth savings compensating for unnecessary arithmetic costs)

weinbe2: At least be a net win*

lib/interface_quda.cpp
```diff
@@ -2403,155 +2404,137 @@ void cloverQuda(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity
...
-
-  // download source
-  ColorSpinorParam cudaParam(cpuParam, *param);
-  cudaParam.create = QUDA_COPY_FIELD_CREATE;
-  r = new cudaColorSpinorField(*h_r, cudaParam);
-  Apsi = new cudaColorSpinorField(*h_Apsi, cudaParam);
+  ColorSpinorParam cudaParam(cpuParam);
+  cudaParam.location = QUDA_CUDA_FIELD_LOCATION;
+  cudaParam.create = QUDA_ZERO_FIELD_CREATE;
+  eig_param->cuda_prec_ritz == QUDA_DOUBLE_PRECISION ? cudaParam.fieldOrder = QUDA_FLOAT2_FIELD_ORDER :
```
weinbe2: Is this strictly correct? Isn't it always `FLOAT2` for staggered fields in general?

lib/inv_cg_quda.cpp
```diff
@@ -310,10 +346,30 @@ namespace quda {
     // compute initial residual
     double r2 = 0.0;
     if (param.use_init_guess == QUDA_USE_INIT_GUESS_YES) {
-      mat(r, x, y, tmp3);
-      r2 = blas::xmyNorm(b, r);
-      if (b2 == 0) b2 = r2;
-      blas::copy(y, x);
+
+      // Just replace any initial guess with a deflated RHS
```
weinbe2: This is a bad idea: I've seen cases in MILC where sometimes CG is called to "polish" a solution (i.e., check and make sure it's a solution, and if it's not polish it up so it is, which is normally just a few iterations). In cases like that, completely nuking the existing initial guess is suboptimal. We should work out the logic to avoid this problem.

weinbe2: That said, I don't know if there's an intersection of deflation and polishing solutions in practice in general, but still, this shouldn't be too difficult to work out.

lib/multigrid.cpp
```diff
@@ -894,9 +968,8 @@ namespace quda
     } else { // do the coarse grid solve
 
       ColorSpinorField *out=nullptr, *in=nullptr;
-
+      // Deflate EVEN ODD here
```
weinbe2: Is this comment still relevant?

lib/multigrid.cpp
```diff
@@ -1429,4 +1387,53 @@ namespace quda
...
+    // This is the vector precision used by matResidual
+    csParam.setPrecision(param.mg_global.invert_param->cuda_prec_sloppy, QUDA_INVALID_PRECISION, true);
+
+    for (int i = 0; i < nKr; i++) B_evecs.push_back(ColorSpinorField::Create(csParam));
+
+    // before entering the eigen solver, lets free the B vectors to save some memory
+    ColorSpinorParam bParam(*param.B[0]);
+    for (int i = 0; i < (int)param.B.size(); i++) delete param.B[i];
+
+    DiracMdagM *mdagm = new DiracMdagM(*diracResidual);
```
weinbe2: Do we want to expose the option of using `DiracMMdag` instead? (In practice I don't think it makes sense to use left singular vectors, so this isn't a hard request.)

lib/quda_arpack_interface.cpp
```diff
@@ -1,384 +1,541 @@
...
+    // Construct parameters and memory allocation
+    //---------------------------------------------------------------------------------
+    double time_ar = 0.0; // time in ARPACK
+    double time_mv = 0.0; // time in QUDA mat vec + data transfer
+    double time_ev = 0.0; // time in computing Eigenvectors
+
+    // MPI objects
+    int *fcomm_ = nullptr;
+#if (defined(QMP_COMMS) || defined(MPI_COMMS))
+    MPI_Fint mpi_comm_fort = MPI_Comm_c2f(MPI_COMM_WORLD);
```
weinbe2: This needs to be `MPI_COMM_HANDLE` w/regards to the multiple communicator support.

weinbe2: In lieu of adding this comment anywhere else (without me checking if it actually does appear anywhere else), just run a `grep -r "MPI_COMM_WORLD"` and make sure it doesn't show up anywhere except for setting the default value of `MPI_COMM_HANDLE` :)

lib/quda_arpack_interface.cpp
```diff
@@ -1,384 +1,541 @@
...
+    iparam_[6] = 1;
+
+    // ARPACK problem type to be solved
+    char howmny = 'P';
+    char bmat = 'I';
+    char *spectrum;
+    spectrum = strdup("SR"); // Initialsed just to stop the compiler warning...
+
+    if (eig_param->use_poly_acc) {
+      if (eig_param->spectrum == QUDA_SPECTRUM_SR_EIG)
```
weinbe2: See comment I've put elsewhere about throwing warnings/errors about polynomial acceleration for non-normal ops

lib/eigensolve_quda.cpp
weinbe2: Update: it looks like this has already been worked out in `lib/quda_arpack_interface.cpp`, let's grab it from there.

tests/test_util.cpp
```diff
@@ -1799,11 +1846,21 @@ void usage(char** argv )
   printf("    --mass-normalization                      # Mass normalization (kappa (default) / mass / asym-mass)\n");
   printf("    --matpc                                   # Matrix preconditioning type (even-even, odd-odd, even-even-asym, odd-odd-asym) \n");
   printf("    --solve-type                              # The type of solve to do (direct, direct-pc, normop, normop-pc, normerr, normerr-pc) \n");
+  printf("    --solution-type                           # The solution we desire (mat (default), mat-dag-mat, mat-pc, "
+         "mat-pc-dag-mat-pc (default for multi-shift))\n");
```
weinbe2: Except for staggered!

weinbe2: iirc The `--solve-type` is `direct-pc` and the `--solution-type` is `mat-pc`, has anyone made sure this isn't broken by the default changes? I _think_ `staggered_invert_test` hardcodes a bunch of these things if you request the multishift tests.

CMakeLists.txt
```diff
@@ -326,9 +326,14 @@ if(QUDA_MPI OR QUDA_QMP)
     message(STATUS "Found environment variable MPICC. Using it for MPI detection: $ENV{MPICC}")
     set(MPI_C_COMPILER $ENV{MPICC})
     set(mpimessage True)
+  endif()
+  if((NOT MPI_FORT_COMPILER) AND DEFINED ENV{MPIFORT})
+    message(STATUS "Found environment variable MPIFORT. Using it for MPI detection: $ENV{MPIFORT}")
+    set(MPI_C_COMPILER $ENV{MPICC})
```
mathiaswagner: This is not what we want (MPI_C_COMPILER) - but I am thinking about dropping support for that anyway.

CMakeLists.txt
```diff
@@ -411,28 +416,36 @@ endif(QUDA_MAGMA)
 
 # This selects arpack or parpack for Multi GPU
 if(QUDA_ARPACK)
+  enable_language(Fortran)
```
mathiaswagner: duplicated. But I think for FindMPI to work properly this needs to come before FindMPI anyway

lib/eigensolve_quda.cpp
maddyscientist: I'll fix this in my next push.  However, I don't agree with respect to cmake variable since this code doesn't involve template instantiation.

lib/eigensolve_quda.cpp
maddyscientist: In the `axpy`, can the indices `k` and `num_locked + j` be equal?  Just wondering hos easy this is multi-blas-ify

lib/interface_quda.cpp
maddyscientist: Fixed in next push

lib/quda_arpack_interface.cpp
maddyscientist: Fixed in next push

tests/test_util.cpp
maddyscientist: Yes, staggered hard codes a bunch of things, and has done for ages (outside of this pull).  For reasons I can't recall, staggered_invert_test uses `--test` to set these things.  This does need to be fixed for staggered to make it consistent.  Not on this pull.

lib/inv_cg_quda.cpp
maddyscientist: I agree, this should be fixed.  It would also come if we were at some point to combine deflation with multi-shift solvers, e.g., where we run deflated staggered multi shift in mixed precision and then want to refine using deflated CG.

lib/eigensolve_quda.cpp
cpviolator: This can multi-blas-ified. `num_locked + j` is never greater than `dim`, which is never greater than `nKr`. `k` is always equal to `nKr + 1`, so the vectors never alias.

lib/eigensolve_quda.cpp
cpviolator: Ok, leaving this alone for the moment.

lib/eigensolve_quda.cpp
cpviolator: Nope, it has been banished :)

lib/eigensolve_quda.cpp
cpviolator: Is this a general comment about how our QIO code handles errors, or about how I use it in the eigensolver? Shall we fix the QIO `stdout` on this pull?

lib/multigrid.cpp
cpviolator: If it doesn't make sense, then I won't expose the option. We use Lanczos here not to extract any particular information about the spectrum of a particular operator (as we would in the stand alone solver) but to simply deflate a CG solve.

lib/quda_arpack_interface.cpp
cpviolator: Again, leaving this alone. For normal operators checks, see here: https://github.com/lattice/quda/blob/feature/deflatedMG/lib/interface_quda.cpp#L2470

lib/multigrid.cpp
maddyscientist: It does make sense to expose the option.  When doing CGNE, you solve using MMdag not MMdag.

lib/inv_cg_quda.cpp
maddyscientist: You form new residual equation, e.g., `r = b - Ax_0`, which is what you do normally with an initial guess.  Now the system you deflate is `A y = r`, and the total solution vector is given by `x = x0 + y`.

include/invert_quda.h
```diff
@@ -1139,16 +1236,22 @@ namespace quda {
...
-  public:
-
+public:
     IncEigCG(DiracMatrix &mat, DiracMatrix &matSloppy, DiracMatrix &matPrecon, SolverParam &param, TimeProfile &profile);
 
     virtual ~IncEigCG();
 
+    /**
+       @brief Expands deflation space.
+       @param V container of new eigenvectors
```
mathiaswagner: How is V a container? Is it using a Composite field?

lib/deflation.cpp
```diff
@@ -125,33 +125,34 @@ namespace quda {
     std::vector<ColorSpinorField*> res;
     res.push_back(r);
 
-    for(int i = 0; i < nevs_to_print; i++)
-    {
-       zero(*r);
+    for (int i = 0; i < nevs_to_print; i++) {
+      zero(*r);
```
mathiaswagner: not new in this pull but the blank lines confused me.

maddyscientist:
The addition of a staggered eigenvector is set to #821 and will for a later pull.

lib/eigensolve_quda.cpp
maddyscientist: Ok, while that specific print statement doesn't belong there since there is another equivalent statement in `EigenSolver::loadVectors` which called from this function (and so I'll remove it on my next push), I do see that the info from QIO is actually not bad.  E.g., if QIO fails to find a file, it will print the following
```
QIO_create_reader(0): cannot open conf_stout_1e4.1000__level_0_nvec_6 as SINGLEFILE; trying PARTFILE
QIO_create_reader(0): cannot open conf_stout_1e4.1000__level_0_nvec_6.vol0000 as PARTFILE; trying PARTFILE_DIR
QIO_create_reader(0): cannot open vol0000/conf_stout_1e4.1000__level_0_nvec_6 as PARTFILE_DIR; QIO_create_reader FAILED
MG level 0 (GPU): open_test_input(0): QIO_open_read returns NULL.
MG level 0 (GPU): ERROR: MG level 0 (GPU): ERROR: Open file failed
```
This seems plenty informative.

lib/deflation.cpp
maddyscientist: Fixed in my next push

include/invert_quda.h
maddyscientist: Yes, this is a composite field.  Made this clear in my next push.

lib/quda_arpack_interface.cpp
maddyscientist: Marking this as resolved...

CMakeLists.txt
maddyscientist: Fixed in #820

CMakeLists.txt
maddyscientist: Fixed in #820

lib/multigrid.cpp
maddyscientist: Noting this is fixed now

lib/interface_quda.cpp
```diff
@@ -2403,155 +2404,157 @@ void cloverQuda(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity
...
-  if (getVerbosity() >= QUDA_VERBOSE) {
-    cpu = blas::norm2(*h_r);
-    gpu = blas::norm2(*r);
-    printfQuda("r vector CPU %1.14e CUDA %1.14e\n", cpu, gpu);
-    cpu = blas::norm2(*h_Apsi);
-    gpu = blas::norm2(*Apsi);
-    printfQuda("Apsi vector CPU %1.14e CUDA %1.14e\n", cpu, gpu);
+  // If you use polynomial acceleration on a non-symmetric matrix,
+  // the solver will fail.
+  if (eig_param->use_poly_acc && !eig_param->use_norm_op && !(inv_param->dslash_type == QUDA_LAPLACE_DSLASH)) {
```
weinbe2: We need a special check for staggered/asqtad PC operators, since they're normal.

weinbe2: ^and therefore safe

maddyscientist:
@cpviolator and I have done some further cleanup and more testing.  Given approval from @weinbe2 already, we're ready to merge.

PR #818, opened 2019-05-24
Placed error on dir in the proper place

cpviolator:


include/kernels/laplace.cuh
```diff
@@ -159,7 +161,8 @@ namespace quda
                                                                          active);
       break;
     default:
-      errorQuda("Unsupported laplace direction %d given", arg.dir);
+      applyLaplace<Float, nDim, nColor, nParity, dagger, kernel_type, -1>(out, arg, coord, x_cb, parity, idx,
```
mathiaswagner: Can we just do something like:
```C++
case 3:
  apply < ...., 3>... 
  break;
case 4:
default:
  apply <...., -1>
```

here?

cpviolator: Yes, that's much neater.

maddyscientist:
Looks like the switch on `arg.dir` is a run-time check not a compile time check.  This should be changed to a compile time switch, e.g., make `dir` a template parameter rather than a member of the argument struct.

cpviolator:
@maddyscientist Similar to the compile time switch on dir for covDev?

https://github.com/lattice/quda/blob/develop/lib/covDev.cu#L340

maddyscientist:
Yeh, it probably doesn't matter that much.  I'll just merge this as is, and if we ever need 1% more performance out of Laplace, we'll know where to get it 😉

PR #819, opened 2019-05-24
Add ability to download and build QIO and QMP to cmake

mathiaswagner:
This is still a prototype but you should be able to try something like
```
cmake ~/quda -DQUDA_QMP=ON -DQUDA_QIO=ON -GNinja
```

This will fetch QIO and QMP when `QUDA_QIOHOME` and `QUDA_QMPHOME` are not specified.
Note, I don't do anything fancy and there is certainly room to improve the logic on when to download and build QIO/QMP. Also the build may just fail, I did not do a lot of testing.

And I forgot to mention that Jenkins will love it. We should be able to easily do QMP builds.


maddyscientist:
Ok, I've done some quick testing.  It fails with this error if you only request QMP and not QIO:
```
-- Configuring done
CMake Error at tests/CMakeLists.txt:19 (add_dependencies):
  The dependency target "QIO" of target "quda_test" does not exist.
```

And it also fails if you only request QIO and not QMP:
```
-- Detecting C compile features
-- Detecting C compile features - done
CMake Error at /opt/cmake-3.13.4-Linux-x86_64/share/cmake-3.13/Modules/ExternalProject.cmake:2835 (get_property):
  get_property could not find TARGET QMP.  Perhaps it has not yet been
  created.
Call Stack (most recent call first):
  /opt/cmake-3.13.4-Linux-x86_64/share/cmake-3.13/Modules/ExternalProject.cmake:3108 (_ep_add_configure_command)
  CMakeLists.txt:389 (ExternalProject_Add)


-- The CUDA compiler identification is NVIDIA 10.1.94
-- Check for working CUDA compiler: /usr/local/cuda-10.1/bin/nvcc
-- Check for working CUDA compiler: /usr/local/cuda-10.1/bin/nvcc -- works
```

And then if you do try with with QMP and QIO, then the build fails
```
[  2%] Performing build step for 'QMP'
 cd /scratch/kate/quda-cyprus/usqcd/src/QMP && automake-1.14 --gnu
/bin/bash: line 4: automake-1.14: command not found
Makefile:299: recipe for target '/scratch/kate/quda-cyprus/usqcd/src/QMP/Makefile.in' failed
make[3]: *** [/scratch/kate/quda-cyprus/usqcd/src/QMP/Makefile.in] Error 1
CMakeFiles/QMP.dir/build.make:110: recipe for target 'usqcd/src/QMP-stamp/QMP-build' failed
make[2]: *** [usqcd/src/QMP-stamp/QMP-build] Error 2
CMakeFiles/Makefile2:946: recipe for target 'CMakeFiles/QMP.dir/all' failed
make[1]: *** [CMakeFiles/QMP.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
```

So looks like some work to go there yet.....  😉

mathiaswagner:
Yes, I also realized some more issues. 
Just wanted to put it out so it get's some more testing and I am expecting that this might take a while to settle - hence the draft status on the pull request.

maddyscientist:
Also with QIO, are doing a recursive update to ensure that LIME is also grabbed?  Since the build fails before it gets to QIO I can't see if QIO will build or not.

And what does `UPDATE_COAMMD ""` do? Should this be `UPDATE_COMMAND ""`

mathiaswagner:
Yes, it does the proper git submodule. The `UPDATE_COMMAND` is indeed a typo.

I am currently looking at some issues with getting the dependencies to work as I want them.
A bit tricky as QIO/QMP are only built at build time hence detection of libs, flags, calling `rio-config` at configure (cmake) time does not work.

maddyscientist:
Ok. this is looking much better now.  A couple of points:
* still get an unintuitive error when asking for `QIO` but no `QMP`
```
CMake Error at /opt/cmake-3.13.4-Linux-x86_64/share/cmake-3.13/Modules/ExternalProject.cmake:2835 (get_property):
  get_property could not find TARGET QMP.  Perhaps it has not yet been
  created.
Call Stack (most recent call first):
  /opt/cmake-3.13.4-Linux-x86_64/share/cmake-3.13/Modules/ExternalProject.cmake:3108 (_ep_add_configure_command)
  CMakeLists.txt:401 (ExternalProject_Add)
```
* should the `usqcd` directory be in the `externals`?

cpviolator:
This is great. Automatic download and build of QMP and QIO will save a lot of time for new developers.

In order to sidestep @maddyscientist's first point, can one add a bool, say `QUDA_DOWNLOAD_USQCD_IO`? If `YES` the it ignores all `QMP` and `QIO` flags and just downloads and builds the libraries.

mathiaswagner:
Yes, we could add a bool. I think having the usqcd in the build dir is the proper place. We might think about putting the download in the source dir. But unlike Eigen it is an optional directory. 

About QIO only build: Is it possible to build QIO without QMP? And what about the ‘-single’ builds? Do we need QMP-single? 

Currently I also have hard coded ‘CC=mpicc’ and this also needs a more _cmake-y_ solution. This will always stay hacky until USQCD gives up qinstall / autoconf and picks up cmake. 

Anyway I think we can make it work (at least for 80% of cases) with pure QUDA. For applications I don’t think QUDA`` should provide QIO/QMP. 

Happy to hear your suggestions.

cpviolator:
This is why I advocate a catch-all bool for `QMP` and `QIO`. `QIO` is built on top of `QMP` (Lime can work with MPI or `QMP`) so it's easier to just build the whole stack or not at all. Edge cases where one needs one and not the other will be so rare, that users will not need an automatic download/build option.

@mathiaswagner is right that a quick option to build the full IO stack will cater for the 80%. The other 20% will be fine using explicit build parameters.

maddyscientist:
Yeh, I'm happy with a single catch all option.

My comment about `QIO` without `QMP` was simply that the error message isn't very intuitive.  Simple solution is just to have error out if `QIO` is enabled but `QMP` is not (with a sensible error message).

mathiaswagner:
There is now a `QUDA_DOWNLOAD_USQCD` cmake option. It is `OFF` be default to not break existing builds.

cpviolator:
@mathiaswagner Having trouble with the `CMake` variable `LOG_MERGED_STDOUTERR` using `cmake version 3.13.4` Which version do you use?

maddyscientist:
@cpviolator @mathiaswagner that’s a cmake >= 3.14 option, so we can’t use that.

mathiaswagner:
All the ‘LOG_’ stuff can be commented. Just there to reduce verbosity. I fix that soon.

mathiaswagner:
After the bump to cmake 3.11 I need to update some of our Jenkins builds.

mathiaswagner:
@Jenkins test this please

mathiaswagner:
Sure. I disabled the cmake 3.8 checks in Jenkins but that won’t remove them from the PR. I will bump the cmake version for the failed checks later.

maddyscientist:
Disabling cmake 3.8 doesn't mean killing the CUDA 8.0 builds for Jenkins I assume?  Just noting we still need that since a few clusters still use this, e.g., Bonn.

mathiaswagner:
No. I will update these builds to use cmake 3.11.

PR #820, opened 2019-05-28
Download arpack-ng

mathiaswagner:
This is a prototype implementation for downloading arrack-ng.

Might be broken for a couple if things.

One thing I noticed is that arrack-ng seems to miss some logging features hence the modifications in `quda_arpack_interface.cpp`.

TODO before merging:

- [x] verify that results are correct
- [x] switch to using the ISO-C bindings from arpack-ng when we use arpack-ng


cpviolator:
@mathiaswagner This compiles and links perfectly. However, the ARPACK solver does not behave as expected. A GPU problem that converged in fewer than 100 iterations does not converge at all in ARPACK.. checking what has changed.

mathiaswagner:
That’s not what should happen 😔. 
Did you try ‘aprpack-ng’ before? Wondering whether the ‘-ng’ or the automatic build breaks this.

cpviolator:
I'm not 100% sure. I'm in the middle of deflated-MG build so I can try two code paths WRT this issue. Use the manually compiled ARPACK and use the auto built `NG` version. We shall soon see where this fault lies.

maddyscientist:
Ok, this branch should be retested with the latest `develop` merged in.  I fixed a few ARPACK issues in #816 which could have fixed the badness with ARPACK here.

cpviolator:
I'll merge the develop branch in and iron out anything I find.

mathiaswagner:
Would suggest we try to use the ISO-C bindings from arpack-ng. This is probably more stable. I added some TODOs above.

cpviolator:
@mathiaswagner Absolutely yes on the ISO-C bindings 👍 I will test our existing interface and switch once I've confirmed the results.

cpviolator:
Built on power9 with scalar and MPI. Works great.

mathiaswagner:
Let me do some checks and we can get that ready for review soon.

maddyscientist:
Ok, this pull is for the deflatedMG branch which is extinct.  I guess this pull needs to be retargeted at develop.

mathiaswagner:
I created #834 for the iso-c bindings.

lib/quda_arpack_interface.cpp
```diff
@@ -88,18 +96,17 @@ namespace quda
...
+      // Smallest eig requested by use, largest will requested from ARPACK
+      // due to poly acc
+      reverse = true;
       spectrum[0] = 'L';
-    } else if (strncmp("L", spectrum, 1) == 0 && eig_param->use_poly_acc) {
-      reverse = false;
-      spectrum[0] = 'S';
     }
-
+
```
maddyscientist: is this a left over debug statement @cpviolator ?

cpviolator: I realised that this is the only time one needs to intervene to get the eigenvalues in ascending order with respect to the sorted Ritz values. For LR and SR with no poly-acc, the Ritz and eigenvalues will be in the same order. With poly acc and requesting SR, the Ritz are descending and the eigenvalues are ascending. Just a statement to let the user know what's going on.

maddyscientist: Ok, well I would suggest that this be guarded with `>= QUDA_VERBOSE`.

mathiaswagner:
IIRC there is one thing I still wanted to clean in the CMakeLists. I'l check that today and give it a go then.

lib/quda_arpack_interface.cpp
cpviolator: I removed it, no need for it really.

CMakeLists.txt
```diff
@@ -492,45 +540,100 @@ if(QUDA_MAGMA)
...
+        find_library(PARPACK parpack PATH ${QUDA_ARPACK_HOME})
+      else()
+        find_library(PARPACK ${PARPACK_LIBRARIES} PATH ${PARPACK_LIBRARY_DIRS})
+      endif()
     endif()
-    include_directories(SYSTEM ${QUDA_ARPACK_HOME}/PARPACK/SRC/MPI)
-    include_directories(SYSTEM ${QUDA_ARPACK_HOME}/PARPACK/UTIL/MPI)
+  endif(QUDA_DOWNLOAD_ARPACK)
+
+   include_directories(SYSTEM ${QUDA_ARPACK_HOME}/SRC)
```
mathiaswagner: @cpviolator Do you think we need these include to `SRC` and `UTIL`? All I get from them are warnings. And removing them does not break compilation for me.

mathiaswagner:
On my Ubuntu workstation I get link errors when using them system arpack / parpack packages. Probably because it is a `-NG` version that is installed by default.
```
../lib/libquda.a(quda_arpack_interface.cpp.o): In function `quda::arpack_solve(std::vector<quda::ColorSpinorField*, std::allocator<quda::ColorSpinorField*> >&, std::vector<std::complex<double>, std::allocator<std::complex<double> > >&, quda::DiracMatrix const&, QudaEigParam_s*, quda::TimeProfile&)':
/home/mathiasw/quda/lib/quda_arpack_interface.cpp:338: undefined reference to `finilog_'
/home/mathiasw/quda/lib/quda_arpack_interface.cpp:156: undefined reference to `initlog_'
/home/mathiasw/quda/lib/quda_arpack_interface.cpp:159: undefined reference to `pmcinitdebug_'
collect2: error: ld returned 1 exit status
``` 

If we download arpack that part is guarded with `#ifdef ARPACK-NG`. Should we always guard this and make it an advanced CMake preference `QUDA_ARPACK_LOGGING` ? I guess that is more a developer feature. And I would prefer better build success in the wild.

maddyscientist:
Just conferred with @cpviolator on the state of this pull.  While desirable, ISO C binding is for another day.  All immediate issues are taken care of, so merging this.

mathiaswagner:
Nobody else saw the link errors and/or include warnings? ISO-C had been taken out already.

maddyscientist:
Sorry, not following what you mean?  ISO C is a separate issue, yes I saw that, but what do you mean with respect to warnings?

mathiaswagner:
See my comment above (5 days ago): `include_directories(SYSTEM ${QUDA_ARPACK_HOME}/SRC)` caused issues for me. I don't think we need them but since I did not put them in was not sure.
When we use a pre-installed arpack or download arppack-ng there is no `SRC` subdirectory in `QUDA_ARPKAC_HOME`.

cpviolator:
I pushed a fix for ARPACK logging, but I missed the one about SRC and UTIL
directory includes. I have no issues either way regarding removing them.

mathiaswagner:
I'll do a quick fix then.

PR #822, opened 2019-05-29
Feature/multi blas fast compile

maddyscientist:
This pull request removes the template specialization from the multi-blas and multi-reduce kernels.  This was required with older CUDA compilers for performance, but results in quadratic compile time.  With modern CUDA (9.0 and up) this is no longer needed.  A work around is in place to avoid the performance cliff with older CUDA.

End result is less code, and much faster compilation for multi-blas and multi-reduce.

### DIFF
--- a/include/kernels/multi_blas_core.cuh
+++ b/include/kernels/multi_blas_core.cuh
@@ -22,7 +22,12 @@ namespace quda
     static signed char *Bmatrix_h;
     static signed char *Cmatrix_h;
 
+#if CUDA_VERSION < 9000
+    // as a performance work around we put the argument struct into
+    // __constant__ memory to prevent the compiler from spilling
+    // registers on older CUDA
     static __constant__ signed char arg_buffer[MAX_MATRIX_SIZE];
+#endif
 
     /**
        @brief Parameter struct for generic multi-blas kernel.
--- a/include/kernels/multi_blas_core.cuh
+++ b/include/kernels/multi_blas_core.cuh
@@ -22,6 +22,8 @@ namespace quda
     static signed char *Bmatrix_h;
     static signed char *Cmatrix_h;
 
+    static __constant__ signed char arg_buffer[MAX_MATRIX_SIZE];
+
     /**
        @brief Parameter struct for generic multi-blas kernel.
        @tparam NXZ is dimension of input vectors: X,Z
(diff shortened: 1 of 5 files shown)


tests/blas_test.cu
```diff
@@ -26,7 +26,6 @@ extern int tdim;
 extern int gridsize_from_cmdline[];
 extern int niter;
 
-extern bool verify_results;
```
mathiaswagner: curious - do we always check now or was it stale? Have only looked at the diff so far.

include/kernels/multi_reduce_core.cuh
```diff
@@ -25,6 +25,8 @@ namespace quda
     static signed char *Bmatrix_h;
     static signed char *Cmatrix_h;
 
+    static __constant__ signed char arg_buffer[MAX_MATRIX_SIZE];
```
mathiaswagner: can this be guarded for CUDA_VERSION?

include/kernels/multi_blas_core.cuh
```diff
@@ -22,6 +22,8 @@ namespace quda
     static signed char *Bmatrix_h;
     static signed char *Cmatrix_h;
 
+    static __constant__ signed char arg_buffer[MAX_MATRIX_SIZE];
```
mathiaswagner: can this be guarded for CUDA_VERSION?

tests/blas_test.cu
maddyscientist: blas_test was changed so that the performance sweep is also a google test, so no verifying the result actually meant nothing is done.  I figured the best thing to do is just get rid of the `verify_results` option here.

include/kernels/multi_blas_core.cuh
maddyscientist: Will fix this.

include/kernels/multi_reduce_core.cuh
maddyscientist: Fixed

mathiaswagner:
With the compile being faster ( I have will look at the profile later - just complain right now) is it worth revisiting the default for `QUDA_MAX_MULTI_BLAS_N` ?

maddyscientist:
I thought about this, but I think no.  Anything we can do to maximize Jenkins throughput is a good thing.

PR #824, opened 2019-05-30
fix dependency on QMP/QIO for tests

mathiaswagner:
This should fix the QMP build in Jenkins.

Also fixes unnecessary rebuild with `QUDA_DOWNLOAD_USQCD=ON`.
Adds support for `cmake-format`


mathiaswagner:
Some more goodies in here

maddyscientist:
What is `cmake_format`?

mathiaswagner:
https://github.com/cheshirekow/cmake_format

PR #826, opened 2019-06-02
clean up MPI Fortran compiler detection

mathiaswagner:
Cleans up MPI Fortran compiler detection.

The more important question is: Do we use MPI_Fortran stuff at all? Then I would opt for removing it for simplicity.


maddyscientist:
Cleaning up MPI Fortran detection is a good thing, I am all in favour of it.  We need it when we compile with `PARPACK` enabled since it uses MPI.

One issue I have been having is linking in (P)ARPACK to QUDA:  for whatever reason @cpviolator's build of QUDA includes the necessary linking flags `-lgfortran -lmpifort` but my build wasn't and I had to hack it.  Any ideas what's needed to ensure the correct linkage is passed when compiled with (P)ARPACK?

PR #827, opened 2019-06-03
Feature/low prec lanczos

maddyscientist:
Low-hanging fruit improvements for Lanczos solver
* Add half and quarter precision support
* In doing so this adds low-precision support to the `spinorNoise` routine (using a single-precision temporary)
* eigensolve_test now respects the `--solution` flag, which is used to switch between full or pc matrices
* Cleanup of handling of `spectrum` in eigensolver and ARPACK interface
* Add `MAX_MULTI_BLAS_N` macro value to multi-reduce tile-size tuner `tuneKey` to prevent incorrect policies being used when this macro changes value
* Fix minor memory leak in eigensolver class (`spectrum` pointer not being freed)

### DIFF
--- a/include/eigensolve_quda.h
+++ b/include/eigensolve_quda.h
@@ -17,12 +17,12 @@ protected:
 
     // Problem parameters
     //------------------
-    int nEv;        // Size of initial factorisation
-    int nKr;        // Size of Krylov space after extension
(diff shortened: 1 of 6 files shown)


cpviolator:
Spinning up on Summit.

cpviolator:
Looks good from here!

PR #828, opened 2019-06-03
Fixed path in lib/CMakeLists to install quda_define.h

bjoo:


mathiaswagner:

PR #830, opened 2019-06-03
Place hard error on Lanczos if it fails to converge the requeste numb…

cpviolator:
…er of eigenpairs

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -406,7 +406,7 @@ extern "C" {
 
     /** If true, the solver will error out if the convergence criteria are not met **/
     QudaBoolean require_convergence;
-    
+
     /** Which part of the spectrum to solve **/
     QudaEigSpectrumType spectrum;
 
--- a/include/quda.h
+++ b/include/quda.h
@@ -404,6 +404,9 @@ extern "C" {
     /** Performs an MdagM solve, then constructs the left and right SVD. **/
     QudaBoolean compute_svd;
 
+    /** If true, the solver will error out if the convergence criteria are not met **/
+    QudaBoolean require_convergence;
+    
     /** Which part of the spectrum to solve **/
     QudaEigSpectrumType spectrum;
 
--- a/lib/check_params.h
+++ b/lib/check_params.h
@@ -152,6 +152,7 @@ void printQudaEigParam(QudaEigParam *param) {
   P(use_dagger, QUDA_BOOLEAN_NO);
   P(use_norm_op, QUDA_BOOLEAN_NO);
   P(compute_svd, QUDA_BOOLEAN_NO);
+  P(require_convergence, QUDA_BOOLEAN_YES);
   P(spectrum, QUDA_SPECTRUM_LR_EIG);
   P(nEv, 0);
   P(nKr, 0);
@@ -173,6 +174,7 @@ void printQudaEigParam(QudaEigParam *param) {
   P(use_dagger, QUDA_BOOLEAN_INVALID);
   P(use_norm_op, QUDA_BOOLEAN_INVALID);
   P(compute_svd, QUDA_BOOLEAN_INVALID);
+  P(require_convergence, QUDA_BOOLEAN_INVALID);
   P(nEv, INVALID_INT);
   P(nKr, INVALID_INT);
   P(nConv, INVALID_INT);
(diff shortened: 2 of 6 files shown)


weinbe2:
@cpviolator I don't think we should hide the error inside a verbosity check---it should have the same behavior independent of the verbosity setting. 

I also don't necessarily think there should be an error in any case inside the eigensolver itself---contrived example, but if I requested 1024 eigenvectors, and the eigensolve only ended up finding 1023 before hitting the max iters, I'd consider that "okay" and wouldn't want it to hard error out.

I think it'd be better to have it still print that it didn't converge as many eigenvectors as were requested, return from the call with the info indicating that, then have the code calling the eigensolver decide if it's an error or not. In my contrived example above, it would be fine, and I'd just need to replace factors of 1024 with 1023 in whatever my code is doing and life would go on. In the case of using eigenvectors as my near-null vectors, I by definition can't go on if I request 32 eigenvectors and only get 31, so it'd error out then.

weinbe2:
Clarifying comment: there shouldn't be an error for the case of not generating the requested number of eigenvectors specifically, obviously there could be other truly error-worthy cases :)

maddyscientist:
@weinbe2 are you happy with this pull.  Should we merge this?

weinbe2:
No, since besides changing a `printfQuda` to a `warningQuda` and taking it out of a verbosity guard it doesn't actually fix anything.

Points of pain:
* In `multigrid.cpp`, when generating eigenvectors to be used as near-null vectors, it should error out if the eigensolver didn't converge the requested number of vectors.
* In the various deflated solvers, so far as I can tell the eigensolver returns garbage for some of the eigenvectors if not all of them converged. I can verify this by looking in `eigensolve_quda.cpp` in the `deflate/featureSVD` branch, which as of now is the most up-to-date version: whether or not `num_converged >= nConv`, the eigensolver is more than happy to compute the "eigenvalues" and deflate with a full `nConv` "eigenvectors".

It's up to debate if it should be a warning or error if the eigensolver doesn't generate the requisite number of eigenvectors (or it could be a flag), but it's outright wrong to try to deflate using eigenvectors that haven't converged. (SVD is mathematically fine, but it's still a bit of a lie to say the algorithm is doing what it says it does.)

mathiaswagner:
Can we make that a param? like `require_convergence`? Set it to true in multigrid, false else but expose it to the user for deflated solvers and just print it warning?

cpviolator:
So, to be clear: we add a hard requirement `require_convergence` on converge for the solver. If requested, and not explicitly met, the solver will error out. If `require_convergence` in not set, we give a warning to the user, but press on.

weinbe2:
Yes, I think that's a reasonable implementation of that parameter.

cpviolator:
Some sample behaviour.

Convergence:
```
0036 converged eigenvalues at restart iter 0003
TRLM computed the requested 24 vectors in 3 restart steps and 82 OP*x operations.
RitzValue[0000]: (+2.2164320058451834e-01, +0.0000000000000000e+00) residual 3.9704910971514979e-40
```

Near convergence:
```
0012 converged eigenvalues at restart iter 0002
ERROR: TRLM failed to compute the requested 24 vectors with a 24 search space and 48 Krylov space in 2 restart steps. Exiting. (/scratch/CPviolator/work/QUDA_HOTFIX/quda/lib/eigensolve_quda.cpp:605 in operator()())

mathiaswagner:
Have not checked: Anything that needs to got to the Fortran interface?

cpviolator:
ARPACK has it's own error checking which QUDA respects. If ARPACK returns an error code, we display the message and error out.

cpviolator:
@weinbe2 Good to go here?

weinbe2:
I still think, in the case where not all of the eigenvalues converge (but `require_convergence` is false), deflated CG (for ex) should still behave correctly, i.e., it should only try to deflate the number of converged eigenvalues. More specifically, in `void EigenSolver::deflate`, `int n_defl` should not be `eig_param->nConv` since that is the number of converged eigenvalues that were requested, not necessarily the number that actually did converge. Instead, it could be `min(eig_param->nConv, num_converged)`, where `num_converged` is as defined in `TRLM::operator()`. 

That said, that can be for another day.

mathiaswagner:
@weinbe2 Should your above comment be turned into an issue?
Apart from that I think we are fine to merge.

maddyscientist:
Yes, I think any outstanding improvements can be handled in subsequent work, and @weinbe2 can make an issue for specifying what he wants exactly.

mathiaswagner:
@Jenkins test this please

PR #833, opened 2019-06-05
Repeated block orthonormalization

weinbe2:
This PR adds the ability to apply Gram-Schmidt more than once during the block ortho-normalization stage of an MG setup. The default is still to only block orthonormalize once. 

Block orthonormalization is exposed on the command line via the flag `--mg-n-block-ortho [level] [n]`, where `[level]` refers to the fine level, i.e., the level being used to _generate_ near-null vectors, and `[n]` refers to the number of times Gram-Schmidt is applied.

This has come up as useful when using singular vectors as near-null vectors for staggered fermions: in practice, there is a good deal of block co-linearity of near-null vectors (as expected via local coherence arguments). I've seen cases where a single pass of Gram-Schmidt isn't good enough to get the `(1 - P^dag P) v_k` test to pass, while two passes works without an issue.


weinbe2:
If someone would be willing to apply `clang-format`, I'd be grateful; I can't seem to get it working at the moment and have some higher priority things I'd like to work on.

maddyscientist:
All looks as expected.  What would be interesting would be to see perf vs n-sweep data.  Does it scale linearly or is there any data reuse going on making subsequent sweeps cheaper than the first?

cpviolator:
```
git diff -U0  --no-color  develop...feature/deflateSVD  ':(exclude)tests/googletest/*' ':(exclude)lib/generate/*' ':(exclude)lib/dslash_core/*' | clang-format-diff-7 -i -p1 -v

weinbe2:
> ```
> git diff -U0  --no-color  develop...feature/deflateSVD  ':(exclude)tests/googletest/*' ':(exclude)lib/generate/*' ':(exclude)lib/dslash_core/*' | clang-format-diff-7 -i -p1 -v
> ```

I'm aware of the documentation @cpviolator, it wasn't working for unrelated reasons on my machine.

maddyscientist:
FWIW: the script linked to on the wiki doesn't work for me either.  For nvsocal I installed `clang-format-diff` from `apt-get` and that does work (after hacking for `.cu` and `.cuh` files support).

weinbe2:
> All looks as expected. What would be interesting would be to see perf vs n-sweep data. Does it scale linearly or is there any data reuse going on making subsequent sweeps cheaper than the first?

Looks like there's a slight bit of data reuse, but otherwise it's just scaling linearly. (Also, thank you for reminding me to update `flops()` and `bytes()` with your question!)

PR #843, opened 2019-06-12
Feature/plaq test

### LINKED ISSUE #835: Fix plaquette calculation for recon 9 and 13 in staggered tests
Right now, passing in `--recon 13` breaks the staggered inversion tests because the plaquette tests aren't compiled with recon 13 support. There are two possible trivial fixes: either template in recon 13 and 9 in the plaquette calculation code, or have the gauge field get loaded via recon 18 for the plaquette check then re-loaded in 13 afterwards.

This is more a reminder to myself than anything else.

maddyscientist:
This commit is a minor feature and bug fix pull:
* Fix hypercubic RNG to ensure it works for both single and dual parity fields
* Gaussian generator now generates anti-Hermitian Lie algebra elements
* `gaussGaugeQuda` now create group elements in the resident gauge field and the underlying generator can be used for either group or Lie algebra generation
* Added simple plaq_test that tests the hypercubic RNG Gaussian gauge generator
* Fix bug in QMP `col` ordering
* Adds support for recon-13/recon-9 to plaquette (closes #835)

### DIFF
--- a/include/gauge_field.h
+++ b/include/gauge_field.h
@@ -30,7 +30,7 @@ namespace quda {
     // whether we need to compute the fat link maxima
     // FIXME temporary flag until we have a kernel that can do this, then we just do this in copy()
     // always set to false, requires external override
-    bool compute_fat_link_max; 
+    bool compute_fat_link_max;
 
     /** The staggered phase convention to use */
     QudaStaggeredPhase staggeredPhaseType;
@@ -60,7 +60,7 @@ namespace quda {
       anisotropy(1.0),
       tadpole(1.0),
       gauge(h_gauge),
-      create(QUDA_REFERENCE_FIELD_CREATE), 
+      create(QUDA_REFERENCE_FIELD_CREATE),
       geometry(QUDA_VECTOR_GEOMETRY),
       compute_fat_link_max(false),
       staggeredPhaseType(QUDA_STAGGERED_PHASE_NO),
@@ -74,7 +74,7 @@ namespace quda {
 
   GaugeFieldParam(const int *x, const QudaPrecision precision, const QudaReconstructType reconstruct,
 		  const int pad, const QudaFieldGeometry geometry,
-		  const QudaGhostExchange ghostExchange=QUDA_GHOST_EXCHANGE_PAD) 
+		  const QudaGhostExchange ghostExchange=QUDA_GHOST_EXCHANGE_PAD)
     : LatticeFieldParam(4, x, pad, precision, ghostExchange),
       location(QUDA_INVALID_FIELD_LOCATION), nColor(3), nFace(0), reconstruct(reconstruct),
       order(QUDA_INVALID_GAUGE_ORDER), fixed(QUDA_GAUGE_FIXED_NO),
@@ -107,11 +107,11 @@ namespace quda {
 	    errorQuda("Error: invalid link type(%d)\n", link_type);
 	  }
 	}
-    
+
     /**
        @brief Helper function for setting the precision and corresponding
        field order for QUDA internal fields.
-       @param precision The precision to use 
+       @param precision The precision to use
      */
     void setPrecision(QudaPrecision precision, bool force_native=false) {
       // is the current status in native field order?
@@ -149,7 +149,7 @@ namespace quda {
   class GaugeField : public LatticeField {
 
   protected:
-    size_t bytes; // bytes allocated per full field 
+    size_t bytes; // bytes allocated per full field
     int phase_offset; // offset in bytes to gauge phases - useful to keep track of texture alignment
     int phase_bytes;  // bytes needed to store the phases
     int length;
@@ -265,12 +265,32 @@ namespace quda {
     const double& LinkMax() const { return fat_link_max; }
     int Nface() const { return nFace; }
 
+    /**
+       @brief This does routine will populate the border / halo region of a
+       gauge field that has been created using copyExtendedGauge.
+       @param R The thickness of the extended region in each dimension
+       @param no_comms_fill Do local exchange to fill out the extended
+       region in non-partitioned dimensions
+    */
+    virtual void exchangeExtendedGhost(const int *R, bool no_comms_fill=false) = 0;
+
+    /**
+       @brief This does routine will populate the border / halo region
+       of a gauge field that has been created using copyExtendedGauge.
+       Overloaded variant that will start and stop a comms profile.
+       @param R The thickness of the extended region in each dimension
+       @param profile TimeProfile intance which will record the time taken
+       @param no_comms_fill Do local exchange to fill out the extended
+       region in non-partitioned dimensions
+    */
+    virtual void exchangeExtendedGhost(const int *R, TimeProfile &profile, bool no_comms_fill=false) = 0;
+
     void checkField(const LatticeField &) const;
 
     /**
        This function returns true if the field is stored in an
        internal field order for the given precision.
-    */ 
+    */
     bool isNative() const;
 
     size_t Bytes() const { return bytes; }
@@ -455,7 +475,7 @@ namespace quda {
 
     /**
        @brief This does routine will populate the border / halo region of a
-       gauge field that has been created using copyExtendedGauge.  
+       gauge field that has been created using copyExtendedGauge.
        @param R The thickness of the extended region in each dimension
        @param no_comms_fill Do local exchange to fill out the extended
        region in non-partitioned dimensions
@@ -514,7 +534,7 @@ namespace quda {
 
     const void* Gauge_p() const { return gauge; }
     const void* Even_p() const { return even; }
-    const void* Odd_p() const { return odd; }	
+    const void* Odd_p() const { return odd; }
 
 #ifdef USE_TEXTURE_OBJECTS
     const cudaTextureObject_t& Tex() const { return tex; }
@@ -579,7 +599,7 @@ namespace quda {
 
     /**
        @brief This does routine will populate the border / halo region of a
-       gauge field that has been created using copyExtendedGauge.  
+       gauge field that has been created using copyExtendedGauge.
 
        @param R The thickness of the extended region in each dimension
        @param no_comms_fill Do local exchange to fill out the extended
@@ -588,13 +608,13 @@ namespace quda {
     void exchangeExtendedGhost(const int *R, bool no_comms_fill=false);
 
     /**
-       @brief This does routine will populate the border / halo region of a
-       gauge field that has been created using copyExtendedGauge.
+       @brief This does routine will populate the border / halo region
+       of a gauge field that has been created using copyExtendedGauge.
        Overloaded variant that will start and stop a comms profile.
        @param R The thickness of the extended region in each dimension
        @param profile TimeProfile intance which will record the time taken
        @param no_comms_fill Do local exchange to fill out the extended
-       region in non-partitioned dimenions
+       region in non-partitioned dimensions
     */
     void exchangeExtendedGhost(const int *R, TimeProfile &profile, bool no_comms_fill=false);
 
@@ -660,7 +680,7 @@ namespace quda {
      @param ghostIn The input ghost buffer (optional)
      @param type The type of copy we doing (0 body and ghost else ghost only)
   */
-  void copyGenericGauge(GaugeField &out, const GaugeField &in, QudaFieldLocation location, 
+  void copyGenericGauge(GaugeField &out, const GaugeField &in, QudaFieldLocation location,
 			void *Out=0, void *In=0, void **ghostOut=0, void **ghostIn=0, int type=0);
   /**
      This function is used for copying the gauge field into an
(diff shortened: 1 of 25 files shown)


include/gauge_field_order.h
```diff
@@ -2998,6 +3010,10 @@ namespace quda {
     typedef gauge::FloatNOrder<char, N, 4, 12, stag, huge_alloc, ghostExchange, use_inphase> type;
   };
   template <int N, QudaStaggeredPhase stag, bool huge_alloc, QudaGhostExchange ghostExchange, bool use_inphase>
+  struct gauge_mapper<char, QUDA_RECONSTRUCT_10, N, stag, huge_alloc, ghostExchange, use_inphase> {
+    typedef gauge::FloatNOrder<char, N, 2, 10, stag, huge_alloc, ghostExchange, use_inphase> type;
```
mathiaswagner: Is that supposed do `<char, N, 2, 10, stag ...>` here? Just looking at the diff and above QUDA_RECONSTRUCT_10 is using 11.

include/lattice_field.h
```diff
@@ -7,6 +7,7 @@
 #include <comm_quda.h>
 #include <util_quda.h>
 #include <object.h>
+#include <cuda_runtime.h>
```
mathiaswagner: Is this the right place to put that? But I don't really care, just curious.

include/quda_matrix.h
```diff
@@ -1309,6 +1308,84 @@ namespace quda {
       return;
     }
 
+    /**
+       Direct port of the TIFR expsu3 algorithm
+    */
+    template <typename Float> __device__ __host__ void expsu3(Matrix<complex<Float>, 3> &q)
+    {
+      typedef complex<Float> Complex;
```
mathiaswagner: I just assume that this has been tested somehow ...

include/random_quda.h
```diff
@@ -19,37 +21,47 @@ typedef struct curandStateMRG32k3a cuRNGState;
...
-    int X[4];
-    /*! @brief allocate curand rng states array in device memory */
-    void AllocateRNG();
-    /*! @brief CURAND array states initialization */
-    void INITRNG(int rng_sizes, int seedin, int offsetin);
+
+  private:
+  cuRNGState *state;        /*! array with current curand rng state */
+  cuRNGState *backup_state; /*! array for backup of current curand rng state */
+  int seed;                 /*! initial rng seed */
```
mathiaswagner: above the seed is long. Just checking whether this should be int here.

lib/eigensolve_quda.cpp
```diff
@@ -455,7 +455,7 @@ namespace quda
       if (kSpace[0]->Location() == QUDA_CPU_FIELD_LOCATION) {
         kSpace[0]->Source(QUDA_RANDOM_SOURCE);
       } else {
-        RNG *rng = new RNG(kSpace[0]->Volume(), 1234, kSpace[0]->X());
+        RNG *rng = new RNG(*kSpace[0], 1234);
```
mathiaswagner: nothing new, but 1234 here seems weird. Should that be singled out in a constexpr with a name. Just for code maintainability.

cpviolator: Is there any way we can pass a global seed to QUDA via `--seed 1234`?

include/lattice_field.h
maddyscientist: Yes, it is needed.  It's needed here because this include file uses CUDA runtime structs, such as events and streams, but this file is included in pure `cpp` code.  I found this issue with some header reorganization, that this include was missing but needed.

include/quda_matrix.h
maddyscientist: It's just been moved from gauge_update_quda directly.  And yes, it's been tested, since this is what is used to exponentiate the Lie algebra field when creating random SU(3) fields from Gaussian a distributed su(3) field.

maddyscientist: I moved the function from the gauge update routine to the header so it could be used in different files.

lib/eigensolve_quda.cpp
maddyscientist: In a subsequent pull we could clean up the seed setting to be supplied at command line like @cpviolator suggests.  If we do that, the question is what to do when we have multiple RNGs active?  E.g., in multigrid we have different RNGs active for different grids.

maddyscientist: Anyway, out of scope for this pull, so marking this as resolved.

include/gauge_field_order.h
maddyscientist: Good catch, this is a bug.  It should be 11 here too.  Will fix this in my next push.

This is something that will eventually be cleaned up, since the `11` is unintuitive (it comes from prior ignorance of C++).

include/random_quda.h
maddyscientist: Good question.  This is code that @nmrcardoso wrote a long time ago.  Let me see if I can work this out.....

maddyscientist: Ok, the interface to `curand_init` is 
```
__device__ void curand_init (unsigned long long seed, unsigned long long sequence,
                             unsigned long long offset, curandState_t *state);
```
So I think all the seeds that we carry around should in fact be `unsigned long long`, and neither `int` or `long`.  I'll fix this, including at the interface level.

maddyscientist:
Just pushed some fixes which address @mathiaswagner's comments above.

mathiaswagner:
Thanks. I’ll complete my review (did not make it through everything earlier) and hope we can merge soon.

include/gauge_field.h
```diff
@@ -107,121 +122,120 @@ namespace quda {
...
+           field order for QUDA internal fields.
+           @param precision The precision to use
+         */
+        void setPrecision(QudaPrecision precision, bool force_native = false)
+        {
+          // is the current status in native field order?
+          bool native = force_native ? true : false;
+          if (precision == QUDA_DOUBLE_PRECISION) {
+            if (order == QUDA_FLOAT2_GAUGE_ORDER) native = true;
+          } else if (precision == QUDA_SINGLE_PRECISION || precision == QUDA_HALF_PRECISION) {
```
weinbe2: Should we add `QUDA_QUARTER_PRECISION` here too? Are there other templates we need to instantiate because of this change?

maddyscientist: Good catch Evan.  I'll add it now.  It won't have any other impact anywhere else.  Likely this was being done manually otherwise, though likely just wasn't noticed since quarter precision is used so sparingly and experimentally.

lib/gauge_random.cu
```diff
@@ -11,196 +11,183 @@
...
+    using Link = Matrix<complex<real>, 3>;
+    int x_cb = threadIdx.x + blockIdx.x * blockDim.x;
+    int parity = threadIdx.y + blockIdx.y * blockDim.y;
+    if (x_cb >= arg.threads) return;
+
+    int x[4];
+    getCoords(x, x_cb, arg.X, parity);
+    for (int dr = 0; dr < 4; ++dr) x[dr] += arg.border[dr]; // extended grid coordinates
+
+    if (arg.group && arg.beta == 0.0) {
```
weinbe2: I don't have a better idea of how to handle this, but it is sort of weird from a physics standpoint because `beta = 0` is infinitely hot fields :) I guess we're abusing beta a bit, though!

lib/gauge_random.cu
```diff
@@ -11,196 +11,183 @@
...
-	  //gf.backup();
-	  arg.rngstate.backup();
+  template <typename Float> void gaugeGauss(GaugeField &U, RNG &rngstate, double beta)
+  {
+    if (U.LinkType() == QUDA_SU3_LINKS) { // generate Gaussian distributed SU(3) field
+      if (getVerbosity() >= QUDA_SUMMARIZE)
+        printfQuda("Creating Gaussian distrbuted gauge field with beta = %e\n", beta);
+      switch (U.Reconstruct()) {
+      case QUDA_RECONSTRUCT_NO: genGauss<Float, QUDA_RECONSTRUCT_NO, true>(U, rngstate, beta); break;
+      case QUDA_RECONSTRUCT_12: genGauss<Float, QUDA_RECONSTRUCT_12, true>(U, rngstate, beta); break;
```
weinbe2: Let's add a `QUDA_RECONSTRUCT_13` and `QUDA_RECONSTRUCT_9` version of this, which would just alias to `QUDA_RECONSTRUCT_12` and `QUDA_RECONSTRUCT_8`, respectively? (Since we're explicitly requesting SU(3) fields, the extra phase would just be `theta = 0`.)

lib/gauge_random.cu
```diff
@@ -11,196 +11,183 @@
...
+  void gaugeGauss(GaugeField &U, RNG &rngstate, double beta)
+  {
+    if (!U.isNative()) errorQuda("Order %d with %d reconstruct not supported", U.Order(), U.Reconstruct());
+    if (U.Ncolor() != 3) errorQuda("Nc = %d not supported", U.Ncolor());
 
-      if (!dataDs.isNative())
-	  errorQuda("Order %d with %d reconstruct not supported", dataDs.Order(), dataDs.Reconstruct());
+    switch (U.Precision()) {
+    case QUDA_DOUBLE_PRECISION: gaugeGauss<double>(U, rngstate, beta); break;
+    case QUDA_SINGLE_PRECISION: gaugeGauss<float>(U, rngstate, beta); break;
```
weinbe2: We should add a version for `QUDA_HALF_PRECISION` and `QUDA_QUARTER_PRECISION` which creates a single precision field then copies it into a half/quarter field, for completeness.

lib/gauge_random.cu
maddyscientist: We can't alias them to recon-12 and recon-8 since then `theta` would not be set to `0`.  We need to use their actual accessors and have the accessors compute `theta=0`.  Happy to make this change though if desired.

lib/gauge_random.cu
maddyscientist: I'm not sure where this will ever be needed.  It's needed for the quark fields since this is used in the MG setup.  Can't see why one would ever need to create a half or quarter precision gauge field.

lib/gauge_random.cu
maddyscientist: Indeed.  We can fix this another day if we have a better idea. 😉

lib/random.cu
```diff
@@ -45,43 +45,68 @@ namespace quda {
...
+     @param size size of the CURAND RNG state array
      @param arg Metadata needed for computing multi-gpu offsets
   */
-  __global__ void kernel_random(cuRNGState *state, int seed, int rng_size, rngArg arg) {
+  __global__ void kernel_random(cuRNGState *state, unsigned long long seed, int size_cb, rngArg arg)
+  {
     int id = blockIdx.x * blockDim.x + threadIdx.x;
-    if (id < rng_size) {
+    int parity = blockIdx.y * blockDim.y + threadIdx.y;
+    if (id < size_cb) {
```
weinbe2: Maybe this is better dealt with when we get to multi-rhs support, but should we make this support 4 and 5-d fields natively?

tests/plaq_test.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  gauge_param.cuda_prec = cuda_prec;
+  gauge_param.reconstruct = link_recon;
+
+  gauge_param.cuda_prec_sloppy = cuda_prec_sloppy;
+  gauge_param.reconstruct_sloppy = link_recon_sloppy;
+
+  gauge_param.gauge_fix = QUDA_GAUGE_FIXED_NO;
+
+#ifdef MULTI_GPU
```
weinbe2: While I remember---were we going to move this into the creation of the gauge field itself so we can remove this boilerplate? Is there an issue for this/did it happen already, I missed it, but it can be removed here?

lib/random.cu
maddyscientist: Leave for multi-rhs support, since there are a bunch of things to be reworked.  Way outside the scope of this pull.

tests/plaq_test.cpp
maddyscientist: There are issues that need to be resolved before the boiler plate can be removed.  Postponed until post 1.0.

lib/gauge_random.cu
weinbe2: I missed that you added `--gaussian-beta` to the command line as the way to expose this, so at least it's consistent (if not maybe a little confusing, but whatever). I guess the better idea would be `sigma`, but it's nothing worth holding this up for :)

lib/gauge_random.cu
weinbe2: May as well while we're here. I'll probably run into it head-first when I port some of these things to the staggered tests, too. 

Related---should we add templating over recon-9 and -13 for the internal plaquette computation routines while we're here? I don't think they've been instantiated.

lib/gauge_random.cu
weinbe2: Ah, fair, agreed!

lib/gauge_random.cu
maddyscientist: Just fixed this, and turned beta into sigma.  Took all of 1 minute.

lib/gauge_random.cu
maddyscientist: Ok, will do both of these.

maddyscientist:
Pushed some fixes to address @weinbe2's comments (recon-13/9 support and changed beta->sigma to avoid confusion).  Will merge once build tests complete.

PR #852, opened 2019-06-26
Feature/managed memory

maddyscientist:
This pull request adds support for managed memory to QUDA, and is designed to quickly enable large problems or memory intensive algorithms to just work, e.g., eigenvector solvers
* managed_malloc / managed_free allocators that are tracked similar to other memory allocations
* if `QUDA_ENABLE_MANAGED_MEMORY=1`, all `device_malloc`/`device_free` requests are directed to to use managed memory

Initial testing shows the expected behaviour: when the memory allocations are smaller than available GPU memory, there is negligible performance impact after allocations have migrated to the GPU.  Performance drops off at large problem sizes, but it just works.

@weinbe2 It might be worth trying this with staggered multigrid, I imagine this will allow you to dramatically reduce the node count with minimal impact on the final solver.

### DIFF
--- a/lib/transfer.cpp
+++ b/lib/transfer.cpp
@@ -1,4 +1,6 @@
+
 #include <transfer.h>
+
 #include <blas_quda.h>
 
 #include <transfer.h>
@@ -254,7 +256,7 @@ namespace quda {
     ColorSpinorField &coarse(*coarse_tmp_h);
 
     // compute the coarse grid point for every site (assuming parity ordering currently)
-    for (int i=0; i<fine.Volume(); i++) {
+    for (auto i=0; i<fine.Volume(); i++) {
       // compute the lattice-site index for this offset index
       fine.LatticeIndex(x, i);
       
--- a/lib/transfer.cpp
+++ b/lib/transfer.cpp
@@ -256,7 +256,7 @@ namespace quda {
     ColorSpinorField &coarse(*coarse_tmp_h);
 
     // compute the coarse grid point for every site (assuming parity ordering currently)
-    for (auto i = 0; i < fine.Volume(); i++) {
+    for (size_t i = 0; i < fine.Volume(); i++) {
       // compute the lattice-site index for this offset index
       fine.LatticeIndex(x, i);
       
--- a/lib/transfer.cpp
+++ b/lib/transfer.cpp
@@ -256,7 +256,7 @@ namespace quda {
     ColorSpinorField &coarse(*coarse_tmp_h);
 
     // compute the coarse grid point for every site (assuming parity ordering currently)
-    for (auto i=0; i<fine.Volume(); i++) {
+    for (auto i = 0; i < fine.Volume(); i++) {
       // compute the lattice-site index for this offset index
       fine.LatticeIndex(x, i);
       
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,7 @@
 ---
 BasedOnStyle: Webkit
 IndentWidth: 2
+AccessModifierOffset: -2
 AlignAfterOpenBracket: Align
 AlignTrailingComments: true
 AllowShortBlocksOnASingleLine: true
--- a/include/blas_helper.cuh
+++ b/include/blas_helper.cuh
@@ -20,13 +20,13 @@ namespace quda
   inline void checkSpinor(const ColorSpinorField &a, const ColorSpinorField &b)
   {
     if (a.Length() != b.Length()) errorQuda("lengths do not match: %lu %lu", a.Length(), b.Length());
-    if (a.Stride() != b.Stride()) errorQuda("strides do not match: %d %d", a.Stride(), b.Stride());
+    if (a.Stride() != b.Stride()) errorQuda("strides do not match: %lu %lu", a.Stride(), b.Stride());
   }
 
   inline void checkLength(const ColorSpinorField &a, const ColorSpinorField &b)
   {
     if (a.Length() != b.Length()) errorQuda("lengths do not match: %lu %lu", a.Length(), b.Length());
-    if (a.Stride() != b.Stride()) errorQuda("strides do not match: %d %d", a.Stride(), b.Stride());
+    if (a.Stride() != b.Stride()) errorQuda("strides do not match: %lu %lu", a.Stride(), b.Stride());
   }
 
 #ifdef QUAD_SUM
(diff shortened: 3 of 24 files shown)


maddyscientist:
Why prohibit on pre-Pascal?  It will still just work, but won't allow over subscription, but it will give a sensible error message when it fails to allocate.

mathiaswagner:
What is the point of using it without oversubscription? I only see potential performance impacts with Kepler UVM.

maddyscientist:
I don't disagree with that.  I just think having unnecessary compile-time dependence on behaviour is good.  Simpler is better.

alexstrel:
>Performance drops off at large problem sizes, but it just works. 

@maddyscientist I guess the prefetch API function could mitigate the performance degradation in this case , so would it be helpful to include it into the PR?

mathiaswagner:
Sure, but would suggest to leave that for further optimization. We need some infrastructure to issue prefetches for Quda fields at the right time.

lib/transfer.cpp
```diff
@@ -254,7 +254,7 @@ namespace quda {
     ColorSpinorField &coarse(*coarse_tmp_h);
 
     // compute the coarse grid point for every site (assuming parity ordering currently)
-    for (int i=0; i<fine.Volume(); i++) {
+    for (unsigned int i = 0; i < fine.Volume(); i++) {
```
mathiaswagner: since volume is `size_t`  this should also be `size_t`.

maddyscientist: Not necessarily.  The volume here is always representable as a 32 bit value.  The move to making it 64 bit was to ensure that expressions involving the volume are computed correctly.

maddyscientist:
how about we just convert all `size_t` variables to `uint64_t`?  A portable and clean solution.

mathiaswagner:
I am not a 100% convinced but also don't care enough about that. Only point above was that we should not use an int index that runs up to a size_t boundary. As long as it is consistent I am fine.

weinbe2:
I think this is a very useful addition, but there are subtleties we should think about around tuning.
* At least the first iteration of a kernel tune may be completely meaningless because of data migration. Should there be a quick "fake" pass before starting actual tuning to ensure data migration is complete?
** This could even break down within a tuning sweep when striding patterns change throughout the tune.
* In lockstep with the previous point, the result of a tuning run may not actually be ideal: during tuning the data may always be resident on the device, but when the kernel is otherwise invoked there may always be data migration happening (just coincidentally due to how the program is laid out).
* Whether or not managed memory is being used isn't reflected in tunecache strings! (Possible solution: embed it into the pre-computed subvolume strings used everywhere.)

In the case where there's no memory oversubscription these points don't really matter, but neither does this PR. Where this PR matters is exactly where we need to think about these things. 

As long as we add `managed` (or something of the sort) to the tunecache string, I'm happy with this being added as a bit of an "advanced mode" tag.

lib/transfer.cpp
cpviolator: a0c68c0

cpviolator:
Works great for large problems on summit. I'll make a mental note of a perf test for large partitions Vs small partitions+managed-mem. @weinbe2's [comment](https://github.com/lattice/quda/pull/852#issuecomment-506411623) is a good idea, and any discrepancies in tuning may manifest in the perf test if using old tunecaches. I also like the idea of making `uint64_t` the standard QUDA C++ integer type.

mathiaswagner:
@weinbe2 I support the idea of adding `managed` to the tune string.

For the tuning we use
```C++
tunable.checkLaunchParam(param);
	  tunable.apply(0); // do initial call in case we need to jit compile for these parameters or if policy tuning
	  if (verbosity >= QUDA_DEBUG_VERBOSE) {
	    printfQuda("About to call tunable.apply block=(%d,%d,%d) grid=(%d,%d,%d) shared_bytes=%d aux=(%d,%d,%d)\n",
		       param.block.x, param.block.y, param.block.z,
		       param.grid.x, param.grid.y, param.grid.z,
		       param.shared_bytes,
		       param.aux.x, param.aux.y, param.aux.z);
	  }

	  cudaEventRecord(start, 0);
	  for (int i=0; i<tunable.tuningIter(); i++) {
	    tunable.apply(0);  // calls tuneLaunch() again, which simply returns the currently active param
	  }
	  cudaEventRecord(end, 0);
	  cudaEventSynchronize(end);
	  cudaEventElapsedTime(&elapsed_time, start, end);```

So there is one warmup that should pull in all memory. Sure, it will mess with the profile when in reality a kernel suffers from a lot of page faults as you realized. However, I think we might want to add a warning there but not add more overhead to actually measure the time for each launch at runtime.

**Comment by weinbe2** [comment]
@mathiaswagner good catch that the initial run is there, thanks!

**Comment by maddyscientist** [comment]
Ok, I have pushed a change that ensures the `,managed` is added to the auxiliary tuning key string when `QUDA_ENABLE_MANAGED_MEMORY=1`.

@weinbe2 It was very much by design that the autotuner tunes after an initial warmup run is done.  This ensures that if the problem fits, then page faulting overheads will not count in the tuned results.  And if the problem doesn't fit, then the page faulting overhead will be present regardless.

@cpviolator Great to hear things are working on Summit.  I think `uint64_t` is for another day or pull though.

@mathiaswagner on the auto iterator type: the compiler doesn't interpret it as integer, since there is no compiler warning about unsigned vs signed comparisons.  I don't know how to infer that it doesn't use `unsigned int` though.

**Comment by maddyscientist** [comment]
🥚 on face moment.  I see the auto type deduction does fail on older compilers, and the signed vs unsigned warning is generated.  Pushing a fix 🤦‍♀

**Comment by mathiaswagner** [comment]
```C++
#include <iostream>
#include <typeinfo>    // for 'typeid'

using namespace std;
int main(int argc, char *argv[]) {
  size_t boundary=2;
  for(auto i=0; i< boundary; i++){
    std::cout << i << " " << typeid(i).name() <<  " " << typeid(boundary).name() << std::endl;
  }
  for(auto i=0*boundary; i< boundary; i++){
    std::cout << i << " " << typeid(i).name() <<  " " << typeid(boundary).name() << std::endl;
  }
  for(decltype(boundary) i=0; i< boundary; i++){
    std::cout << i << " " << typeid(i).name() <<  " " << typeid(boundary).name() << std::endl;
  }
}
```

maddyscientist:
I like the `decltype` solution, we can use that going forward.

PR #853, opened 2019-06-27
Change heavy-quark restart behavior

mathiaswagner:
This pull fixes:
* initialization of heavy-quark restart parameters
* disable mixed-precision solves for heavy-quarks as they are sometimes unstable (regular L2 which dominate are still using mixed precision)
* don't try to iterate L2 norm beyond machine precision when heavy-quark residual is requested


mathiaswagner:
#854 to follow up for mixed precision.
Waiting for Carleton to confirm that current mixed precision workaround is fine.

lib/inv_cg_quda.cpp
```diff
@@ -427,6 +427,7 @@ namespace quda {
     // set this to true if maxResIncrease has been exceeded but when we use heavy quark residual we still want to continue the CG
     // only used if we use the heavy_quark_res
     bool L2breakdown = false;
+    const double L2breakdown_eps = 100. * uhigh;
```
maddyscientist: Where does the factor of 100 come from?

mathiaswagner: Somewhat handwaving. We will never reach machine eps and while we might be able to get with in 10 * uhigh it seemed to tight for the purpose here. So just tried 100 ...

mathiaswagner:
Yes. Only waiting on some feedback for mixed precision workaround.

mathiaswagner:
Passed the MILC test :-)

PR #860, opened 2019-07-17
fixes for resize to negative values

mathiaswagner:
This should fix a comparison that lead to a resize of a std::vector to a negative size.


cpviolator:
This works great with two and three level deflated MG from the command line.

PR #855, opened 2019-06-28
Hotfix/real valued block blas

alexstrel:
This pull request fixes real valued  block dot product. It also introduces a real-valued axpy block-blas routine.

PR #862, opened 2019-07-23
Feature/large multi blas

maddyscientist:
This pull request is a significant improvement to the multi-blas and multi-reduction framework
* Increase the supported size of `MAX_MULTI_BLAS_N` to 32
* Optimization of the multi-blas routines to parallelize the `NXZ` dimension (the accumulation dimension) through warp splitting (degree of splitting picked by the autotuner).  This optimization is only applied for small problem sizes that are parallelism limited.
* One side effect of this dynamic calculation is that in order to keep relatively sane code I've bumped QUDA up to required C++14.  While not yet supported by the CUDA toolkit, I've also added C++17 as an option to cmake for future development.
* Limit of multi-blas problem size is no longer simply given by `MAX_MULTI_BLAS_N`, rather the limit is computed dynamically to allow for larger tall and skinny or short and fat fused kernels.  E.g., while the 2-d square kernel limit is given by `MAX_MULTI_BLAS_N * MAX_MULTI_BLAS_N`, a 1-d multi-blas function can be up to `1x166`
* Regardless of what `MAX_MULTI_BLAS_N` is set to, we always compile all powers of two for the `NXZ` dimension (up to 128 for multi-blas and 16 for multi-reductions) .  Thus for most cases, there is no need to set a large `MAX_MULTI_BLAS_N` value
* Added a `TransposeTune` class that picks the optimum layout for doing 2-d multi-reductions, e.g., is it better to pose the problem as a `n x m` or `m x n` problem?
* When doing 1-d multi reductions, we always pose the problem as a `1 x n` problem, and do not apply any tile-size tuning - simply use the largest tile size possible.  This means that any 1-d multi-reduction can be done using the run-time `NYW` parameter, and not the compile-time `NXZ` parameter.
* Reduced compilation overhead of multi-reductions by not instantiating unnecessary large block sizes that are never picked by the autotuner
* Dramatic reduction in autotuning overheads for streaming blas kernels - these have no dependence on the grid size, so simply set the grid size to be 2x SM count.
* Fix bugs from #855 

The result of this rework is that:
* Coarse-grid deflation is now up to 8x faster.  E.g., when deflating 1024 vectors we would previously have needed to use 64 `multi-axpy` and `multi-dot` with kernels which were massively latency bound on small grids.  However, now these can now be done as 8 kernels instead, e.g., as 128-way fused multi-blas and multi-reductions, coming closer to saturating memory bandwidth and amortizing the launch latency.
* Eigensolvers now run much faster in the tuning phase, with up to a 10x reduction in end-to-end autotuning overhead observed

### DIFF
--- a/include/kernels/multi_blas_core.cuh
+++ b/include/kernels/multi_blas_core.cuh
@@ -83,15 +83,8 @@ namespace quda
        @param[in,out] arg Argument struct with required meta data
        (input/output fields, functor, etc.)
     */
-#ifndef CONSTANT_ARG
     template <typename FloatN, int M, int NXZ, int warp_split, typename Arg> __global__ void multiBlasKernel(Arg arg)
     {
-#else
-    template <typename FloatN, int M, int NXZ, int warp_split, typename Arg> __global__ void multiBlasKernel()
-    {
-      Arg &arg = *((Arg *)arg_buffer);
-#endif
-
       // use i to loop over elements in kernel
       const int k = blockIdx.y * blockDim.y + threadIdx.y;
       const int parity = blockIdx.z;
--- a/include/kernels/multi_blas_core.cuh
+++ b/include/kernels/multi_blas_core.cuh
@@ -3,6 +3,10 @@
 #include <color_spinor_field_order.h>
 #include <blas_helper.cuh>
 
+#if CUDA_VERSION < 9000
+#define CONSTANT_ARG
+#endif
+
 namespace quda
 {
 
@@ -13,7 +17,8 @@ namespace quda
 #include <texture.h>
 
     // storage for matrix coefficients
-#define MAX_MATRIX_SIZE 4096
+#define MAX_MATRIX_SIZE 8192
+#define MAX_ARG_SIZE 4096
     static __constant__ signed char Amatrix_d[MAX_MATRIX_SIZE];
     static __constant__ signed char Bmatrix_d[MAX_MATRIX_SIZE];
     static __constant__ signed char Cmatrix_d[MAX_MATRIX_SIZE];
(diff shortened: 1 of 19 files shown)


maddyscientist:
@mathiaswagner can you update the Jenkins builds to exclude CUDA 8.0?  I guess we now longer support CUDA 8.0 as a result of the jump to C++14.

mathiaswagner:
Somewhat reluctant as we might still need them for fixes that go into release. Need to check if there is a way to only keep them for release.

maddyscientist:
@mathiaswagner no problem.  Let's just not worry about the failing on the older version of CUDA with this pull, and fix Jenkins once 1.0 is out the door.

lib/multi_reduce_quda.cu
```diff
@@ -654,39 +636,43 @@ namespace quda {
...
+            break;
+          case 4:
+            multiReduce<4, doubleN, ReduceType, ReducerDiagonal, writeDiagonal, ReducerOffDiagonal, writeOffDiagonal, false>(
+              tmp_dot, a, b, c, x, y, z, w, i_idx, j_idx);
+            break;
+          case 8:
+            multiReduce<8, doubleN, ReduceType, ReducerDiagonal, writeDiagonal, ReducerOffDiagonal, writeOffDiagonal, false>(
+              tmp_dot, a, b, c, x, y, z, w, i_idx, j_idx);
+            break;
+          case 16:
```
weinbe2: Is there a reason we're only going up to 16 for reduce while we go up to 128 for regular blas? Compile time?

Also, is there a reason we always compile powers of two for blas independent of `MAX_MULTI_BLAS_N`, while we only compile powers of two for `MAX_MULTI_BLAS_N` > 2 in the reduction case?

weinbe2:
I'm a big fan of this PR between the compile time speed-ups and the code clean-up, in particular saving the copy+paste zoo whenever we need to add new multi-whatever functions. I put one small question in the code, otherwise I'm happy. I assume correctness has been tested via the `blas_test` routines? Or is there somewhere else in the code I'm forgetting where non-trivial shapes come up? (I've lost track of the state of blockCG in mainline QUDA vs the SC work we did.)

lib/multi_reduce_quda.cu
weinbe2: Disregard, my mind spaced out and I missed that the ` > 2` line was _removed_.

maddyscientist: > Is there a reason we're only going up to 16 for reduce while we go up to 128 for regular blas? Compile time?
A few reasons:
* Yes, partly compile time
* For 1-d multi-reductions, there is no need to have a high `NXZ` dimension since we always do this as a `1xn` reduction
* Also for 32, the compiler spills likes crazy and performance is bad so I figured only put it in if you really want it (which means let the user explicitly opt in for it, e.g., if they are using 2-d multi-reductions in a big block solver)

maddyscientist:
> I assume correctness has been tested via the blas_test routines? Or is there somewhere else in the code I'm forgetting where non-trivial shapes come up? (I've lost track of the state of blockCG in mainline QUDA vs the SC work we did.)

Yes, the blas_test is passing all tests for all precisions, and for both wilson and staggered.

Regarding block-CG, that's still in an orphaned branch that we need to fix.  That work is deferred until we have a multi-rhs abstraction in the dslash framework.

weinbe2:
Thanks for answering my questions @maddyscientist ! Pending the two failed builds, I'm happy.

maddyscientist:
@weinbe2 as noted above, there's nothing we will do about the failed builds.  These come from CUDA 7.5 and 8.0 which do not support C++14.  Until we finish 1.0, we will just have to put up with the failed builds on these targets.

cpviolator:
I see this aspect of QUDA becoming very useful to end users, to the point where they (as @alexstrel did) can start to write their own functions. We should add explanations on adding MultiBLAS functions, and perhaps devising a standard for doing so. I can think about this.

cpviolator:
I have my old Summit scripts for the deflated MG twisted clover perf tests. I can run them through this version too.

lib/tune.cpp
```diff
@@ -229,10 +229,11 @@ namespace quda {
       strncpy(tmp, key.aux, 13);
       bool is_policy_kernel = strncmp(tmp, "policy_kernel", 13) == 0 ? true : false;
       bool is_policy = (strncmp(tmp, "policy", 6) == 0 && !is_policy_kernel) ? true : false;
+      bool is_nested_policy = (strncmp(tmp, "nested_policy", 6) == 0) ? true : false; // nested policies not included
```
mathiaswagner: Where is nested_policy used?

maddyscientist: 2-d multi reductions use nested policies: I've added a transpose tuner, which tests if it is better to do an `mxn` or `nxm` multireduction.  It's nested of course since below that there is also a policy tuner for the tile size.

include/kernels/multi_blas_core.cuh
```diff
@@ -54,79 +42,173 @@ namespace quda
...
+    {
+#ifdef WARP_SPLIT
+      constexpr int warp_size = 32;
+      if (warp_split > 1) {
+#pragma unroll
+        for (int j = 0; j < M; j++) {
+          // reduce down to the first group of column-split threads
+#pragma unroll
+          for (int offset = warp_size / 2; offset >= warp_size / warp_split; offset /= 2) {
+#if (__CUDACC_VER_MAJOR__ >= 9)
```
mathiaswagner: If we drop CUDA 8 because of C++14 is this needed here at all?

PR #863, opened 2019-07-24
Add missing support for running on Xavier - sm_72

maddyscientist:
This hotfix adds support for running on Xavier's iGPU (sm_72), which would otherwise fail at runtime (`switch-case` statement didn't support Xavier).

### DIFF
--- a/include/tune_quda.h
+++ b/include/tune_quda.h
@@ -162,6 +162,7 @@ namespace quda {
       case 7:
         switch (deviceProp.minor) {
         case 0: return 32;
+        case 2: return 32;
         case 5: return 16;
         }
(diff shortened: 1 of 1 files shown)


mathiaswagner:
Do we also need to add sm_72 to CMakeLists ?

maddyscientist:
@mathiaswagner I don't think we need to add explicit `sm_72` cmake support.  With this pull, QUDA will just work on `sm_72` using `sm_70`.  Similar story when running on Erista (`sm_53`) and Parker (`sm_61`), you can just use `sm_52` and `sm_60`, respectively, and it will work.  Moreover, given that 99.9% of our audience will never run QUDA on a Tegra, and that one can always set the architecture explicitly using `-DQUDA_GPU_ARCH=sm_72`, we lose nothing I think leaving the Tegra support implicit.

PR #864, opened 2019-07-25
Feature/staggered eig test release1p0

cpviolator:
Added a unit test for the staggered eigensolver, consolidated some boiler plate for the staggered link construction.

### DIFF
--- a/lib/check_params.h
+++ b/lib/check_params.h
@@ -723,9 +723,13 @@ void printQudaMultigridParam(QudaMultigridParam *param) {
     }
 
 #ifdef INIT_PARAM
-    if (i < QUDA_MAX_MG_LEVEL) { P(n_vec[i], INVALID_INT); }
+    if (i<QUDA_MAX_MG_LEVEL) {
+          P(n_vec[i], INVALID_INT);
+    }
 #else
-    if (i < n_level - 1) { P(n_vec[i], INVALID_INT); }
+    if (i<n_level-1) {
+      P(n_vec[i], INVALID_INT);
+    }
 #endif
 
 #ifdef INIT_PARAM
--- a/lib/check_params.h
+++ b/lib/check_params.h
@@ -723,13 +723,9 @@ void printQudaMultigridParam(QudaMultigridParam *param) {
     }
 
 #ifdef INIT_PARAM
-    if (i<QUDA_MAX_MG_LEVEL) {
-          P(n_vec[i], INVALID_INT);
-    }
+    if (i < QUDA_MAX_MG_LEVEL) { P(n_vec[i], INVALID_INT); }
 #else
-    if (i<n_level-1) {
-      P(n_vec[i], INVALID_INT);
-    }
+    if (i < n_level - 1) { P(n_vec[i], INVALID_INT); }
 #endif
 
 #ifdef INIT_PARAM
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -16,9 +16,6 @@ if(QUDA_QDPJIT)
 endif()
 
 if(QUDA_QIO AND QUDA_DOWNLOAD_USQCD)
-  add_dependencies(quda_test QIO)
-endif()
-if(QUDA_QMP AND QUDA_DOWNLOAD_USQCD)
   add_dependencies(quda_test QMP)
 endif()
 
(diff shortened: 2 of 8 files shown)


cpviolator:
`staggered_dslash_ctest` passed with:

```
[----------] 384 tests from QUDA/StaggeredDslashTest (1581322 ms total)

[----------] Global test environment tear-down
[==========] 384 tests from 1 test suite ran. (1583315 ms total)
[  PASSED  ] 288 tests.
[  SKIPPED ] 96 tests, listed below:
```

lib/check_params.h
```diff
@@ -723,13 +723,9 @@ void printQudaMultigridParam(QudaMultigridParam *param) {
     }
 
 #ifdef INIT_PARAM
-    if (i<QUDA_MAX_MG_LEVEL) {
-          P(n_vec[i], INVALID_INT);
-    }
+    if (i < QUDA_MAX_MG_LEVEL) { P(n_vec[i], INVALID_INT); }
```
weinbe2: @cpviolator not that I expect this to be controversial at all, but can you please remove this change? It's unrelated to the PR, and since it's a PR against release we may as well keep everything on a narrow path.

tests/CMakeLists.txt
```diff
@@ -16,9 +16,6 @@ if(QUDA_QDPJIT)
 endif()
 
 if(QUDA_QIO AND QUDA_DOWNLOAD_USQCD)
```
weinbe2: Similar comment as above, not sure what this has to do with this PR, though I can see that it's important for functional correctness. @maddyscientist what do you think?

mathiaswagner: If this is going wrong we should fix it. Not quite sure removing the `if(QUDA_QMP)` is the way to fix it as we can have only QMP enabled.

weinbe2: @mathiaswagner for sure, agreed. Ultimately my comment is a splitting-hairs comment of if we want to batch it in with this PR or not. I guess it doesn't make a difference.

weinbe2:
This is fantastic work @cpviolator , thank you for doing it. I have a few requests/questions.

* There should be only three tests: direct, even-even, and odd-odd. The other tests are only important for inversions since you could consider testing both the PC system in isolation and preparing/reconstructing the PC system to test a full solve, so including them is redundant. 
* I recommend (though won't necessarily put a fight up about) throwing an error and exiting for a direct test on the staggered/asqtad operator. This is because the normal op of the staggered operator has a degeneracy of 2 for every eigenvalue, and while you've demonstrated your eigensolver is (impressively) robust for degenerate eigenvalues, I think it's best to avoid poking the bear with it regardless. That aside, I can't think of a workflow that wouldn't only deflate the even-even or odd-odd Schur operator anyway. Also, from an algebraic complexity standpoint, it's faster to do each separately than do the full operator even if the stability of algebraic degeneracy wasn't an issue.
* What did you do to test that there weren't any accidental errors copying the construct routines to a separate file?

tests/CMakeLists.txt
mathiaswagner: I know and I technically agree that it might be cleaner to have a separate PR. Ultimately we just need to make sure that the fix is actually the proper fix.

cpviolator:
@weinbe2 

* As far as I'm concerned, it's neater and more complete to have the eigensolver test routines mirror the invert test routines. If anybody has the remotest interest in the performance of an inverter for a given operator, they may also wish to know about the spectrum of that operator.

* If you won't put up a fight, it stays.

* I ran the [dslash_ctest](https://github.com/lattice/quda/pull/864#issuecomment-514853471) and it passed. The eigensolver passes with data consistent with prec, as does the invert test.

weinbe2:
@cpviolator sounds good on 1 and 2, as for 3 I meant the HISQ link construction. A test I was considering was running `staggered_invert_test` with a loaded gauge field, performing the gauge link construct, comparing the number of iterations in the release branch as its stands with the number of iterations with your branch, purely as a proxy for making sure nothing accidentally went awry.

cpviolator:
Just pinging this PR into 1.0.

maddyscientist:
Hi @cpviolator.  I think we're still waiting on confirmation that the testing that @weinbe2 suggested above is working ok.  E.g., can you check that `staggered_invert_test` loading a real lattice before and after this change works identically?  Just to check the refactor of the HISQ setup is working ok.  Thx.

tests/staggered_invert_test.cpp
```diff
@@ -126,243 +116,155 @@ void constructSpinorField(Float *res) {
...
-  inv_param->reliable_delta = reliable_delta;
-  inv_param->use_alternative_reliable = alternative_reliable;
-  inv_param->use_sloppy_partial_accumulator = false;
-  inv_param->solution_accumulator_pipeline = solution_accumulator_pipeline;
-  inv_param->pipeline = pipeline;
-
-  inv_param->Ls = Nsrc;
-
-  if(tol_hq == 0 && tol == 0){
+  inv_param.inv_type = QUDA_BICGSTAB_INVERTER; // Dummy setting
```
weinbe2: @cpviolator Why did you make this change? The user should be able to specify whichever inverter they want.

PR #867, opened 2019-07-26
added missing header file for signal handling, required when HOSTDEBUG option is set

alexstrel:
A minor fix

### DIFF
--- a/lib/comm_common.cpp
+++ b/lib/comm_common.cpp
@@ -3,6 +3,7 @@
 
 #include <quda_internal.h>
 #include <comm_quda.h>
+#include <csignal>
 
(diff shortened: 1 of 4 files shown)


mathiaswagner:
Can you elaborate on what was going wrong without the header?

alexstrel:
Compiler emits an error (undefined functions, e.g., raise() that is decorated by HOST_DEBUG). Got this during compilation on Kate's machine.

mathiaswagner:
Thanks. I missed that when moving the `raise` around. I think we can now remove the `signal` header from `comm_single/mpi/qmp.h`

alexstrel:
Done

PR #870, opened 2019-08-08
Added a routine to compute plaquette of staggered fields

weinbe2:
Contains some wrapper code to properly manage reconstruct settings, dealing with staggered phases, etc. This is only called in `staggered_invert_test`. When using HISQ links, it also computes the fat link plaquette.

This pull also adds proper random spinor field creation to `staggered_invert_test`. 

There's nothing particularly controversial keeping it from being merged into `release/1.0.x` instead, but I don't see a pressing need, either. It's easy enough to change that, though.

I verified that the test works correctly for staggered, hisq, and the Laplace operator, including multishift.


maddyscientist:
Shouldn't this also be in the staggered eigen solver test?

weinbe2:
@maddyscientist good call, I'll take care of that in a moment.

weinbe2:
@maddyscientist done, snuck a few tiny eigensolver usability things in there too.

Issue #875, opened 2019-08-15
Consider removing the legacy sign convention for staggered-type operators


weinbe2:
Staggered operators have the convention that `Dslash` actually applies `-D_{eo}` or `-D_{oe}`, while `DslashXpay` applies `D_{eo} + m I_{ee}`, `D_{oe} + m I_{oo}`. Under the hood, this actually triggers by checking if `m == 0.` or not, which causes an issue if `DslashXpay` is called with zero mass.

There's currently a workaround (right now testing it, but it'll be there soon) which catches this case whenever `Apply(Improved)Staggered` is called, but I think in the long term a better idea may be making this convention consistent. This would have implications in downstream apps (likely MILC) which may assume this sign convention. For this reason, it may be better to _not_ change the quirky sign convention, so this issue is ultimately here as a point of discussion for now.

PR #876, opened 2019-08-15
Fix to zero mass Dirac(Improved)StaggeredPC sign quirk

weinbe2:
This hotfix catches an issue where the preconditioned staggered operator comes out with the wrong sign if the mass is zero. The origin of the issue is the quirky sign convention between Dslash and DslashXpay. Technically there is still an issue if you prepare/reconstruct, but in the zero mass case that is ill defined anyway, so eh.

We may want to ultimately address this sign convention qurk as noted in #875, but I don't think this is the time or place for that change, since it requires upstream changes in MILC and elsewhere.

We can put this into `develop` later by merging `release/1.0.x` back in.

### DIFF
--- a/lib/dirac_improved_staggered.cpp
+++ b/lib/dirac_improved_staggered.cpp
@@ -68,10 +68,10 @@ namespace quda {
       } else {
         ApplyImprovedStaggered(out, in, fatGauge, longGauge, 0., x, parity, QUDA_DAG_YES, commDim, profile);
       }
-      flops += 1146ll*in.Volume();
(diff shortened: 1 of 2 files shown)


weinbe2:
@cpviolator I haven't---do you mind making the magic happen? I still haven't sat down and got it working on my end...

maddyscientist:
> @cpviolator I haven't---do you mind making the magic happen? I still haven't sat down and got it working on my end...

FYI, on nvsocal2 I put a helper script for clang-formatting.  
```
clang-format-quda-8 branch_name
```

PR #894, opened 2019-09-06
Hotfix/blas svd deflate

maddyscientist:
This is primarily a bugfix pull
* Fix SVD deflation what was broken by #855 
* Fix silent blas failure when using quarter-precision fields with float-2 ordering
* Add support for quarter precision to alternative reliable updates
* `EigenSolver` methods `loadVector` and `saveVector` should be static - fixes undefined behaviour from calling a method on a `nullptr`
* Robustness and cleanup of multi-blas framework.  Fixes an ASAN warning.

### DIFF
--- a/include/color_spinor_field.h
+++ b/include/color_spinor_field.h
@@ -156,7 +156,6 @@ namespace quda {
 
       if (!pc_solution) {
         siteSubset = QUDA_FULL_SITE_SUBSET;
-        ;
       } else {
         x[0] /= 2; // X defined the full lattice dimensions
         siteSubset = QUDA_PARITY_SITE_SUBSET;
@@ -219,8 +218,8 @@ namespace quda {
         component_id(0)
     {
       siteSubset = cpuParam.siteSubset;
-      fieldOrder = (precision == QUDA_DOUBLE_PRECISION || nSpin == 1) ? QUDA_FLOAT2_FIELD_ORDER : QUDA_FLOAT4_FIELD_ORDER;
-      }
+      fieldOrder = (precision == QUDA_DOUBLE_PRECISION || nSpin == 1 || nSpin == 2) ? QUDA_FLOAT2_FIELD_ORDER : QUDA_FLOAT4_FIELD_ORDER;
+    }
 
     /**
        If using CUDA native fields, this function will ensure that the
--- a/include/color_spinor_field.h
+++ b/include/color_spinor_field.h
@@ -175,8 +175,8 @@ namespace quda {
       }
 
       if (inv_param.dirac_order == QUDA_INTERNAL_DIRAC_ORDER) {
-        fieldOrder
-          = (precision == QUDA_DOUBLE_PRECISION || nSpin == 1 || nSpin == 2) ? QUDA_FLOAT2_FIELD_ORDER : QUDA_FLOAT4_FIELD_ORDER;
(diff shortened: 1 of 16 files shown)


weinbe2:
Not sure how to revert my approval, but for the time being don't merge this PR, I found a bug that @maddyscientist is going to hunt down. It can be reproduced by compiling with quarter precision support (in my case, I also had MPI on, and turned off domain wall just for compile time reasons; I imagine neither of these make a difference) and running `./blas_test` without any arguments.

maddyscientist:
Just pushed a commit to address @weinbe2's comments:
* Extended `blas_test` to include testing of float2-ordered nSpin-4 fields when `GPU_MULTIGRID` is enabled.  This fixes a hole in the prior coverage, and ensures that none of the regressions fixed in this branch can occur again.
* Fix multiple float2-ordered field blas-related bugs
* Default internal field order for nSpin-2 fields should be float2 ordering

weinbe2:
Unless the other edits added coverage for this codepath in a way I didn't notice, this still needs the edit to line 222 of `color_spinor_param.h`:

` fieldOrder = (precision == QUDA_DOUBLE_PRECISION || nSpin == 1) ? QUDA_FLOAT2_FIELD_ORDER : QUDA_FLOAT4_FIELD_ORDER;`

Adding `||nSpin == 2`, or maybe by calling `setPrecision`?

maddyscientist:
@weinbe2 I've fixed color_spinor_field.h now: previously I'd only caught one of the constructors for `ColorSpinorParam`, but all good now.

@mathiaswagner I've added your patch to this branch to not include the cublas headers unless`CUBLAS_LIB` is enabled.  Amusingly, in doing so, I also found a bug (missing header in malloc_quda.h).

Conclusion: this pull is now complete, and can be merged when the build tests have completed.

weinbe2:
@maddyscientist FYI I'm doing a quick build right now (which I imagine will finish before the build tests) to do some tests.

weinbe2:
Running with `Nspin == 1`, (that is, `./blas_test --dslash-type asqtad`), getting the following two failures:

```
[ RUN      ] QUDA/BlasTest.verify/HeavyQuarkResidualNorm_single_default
/home/eweinberg/Data/quda_custom/pr894/quda/tests/blas_test.cu:1136: Failure
Expected: (deviation) <= (tol), actual: 2.25033 vs 1e-06
CPU and CUDA implementations do not agree
[  FAILED  ] QUDA/BlasTest.verify/HeavyQuarkResidualNorm_single_default, where GetParam() = (2, 29, 0) (528 ms)
[ RUN      ] QUDA/BlasTest.verify/xpyHeavyQuarkResidualNorm_single_default
/home/eweinberg/Data/quda_custom/pr894/quda/tests/blas_test.cu:1136: Failure
Expected: (deviation) <= (tol), actual: 2.25037 vs 1e-06
CPU and CUDA implementations do not agree
[  FAILED  ] QUDA/BlasTest.verify/xpyHeavyQuarkResidualNorm_single_default, where GetParam() = (2, 30, 0) (629 ms)
```

This is a run with `QUDA_PRECISION == 15`. I could reproduce the error by re-running the above command. To be explicit, the two blas routines do _not_ fail with quarter, half, or double.

weinbe2:
Doing one last build!

PR #896, opened 2019-09-10
Hotfix/develop cherry picking

maddyscientist:
Some fixes from #894 that should go into the 1.0.x release.


weinbe2:
We should pop in f19ef0a4224b8c0b9dab82a55dc89b3decef65e9 as well, I think? I'm guessing leaving out the quarter-precision related things was intentional.

weinbe2:
Approval currently pending the `blas_test` fail I noticed in #894

PR #898, opened 2019-09-11
Fixed old hard-coded reconstructs in staggered_invert_test.

weinbe2:
Removed a few hard codings of reconstruct values. Should be generally non-controversial.

### DIFF
--- a/tests/staggered_gauge_utils.cpp
+++ b/tests/staggered_gauge_utils.cpp
@@ -300,10 +300,10 @@ void computeStaggeredPlaquetteQDPOrder(void** qdp_link, double plaq[3], const Qu
   gauge_param.cpu_prec = gauge_param_in.cpu_prec;
 
   gauge_param.cuda_prec = gauge_param_in.cuda_prec;
-  gauge_param.reconstruct = gauge_param_in.reconstruct; //QUDA_RECONSTRUCT_NO;
+  gauge_param.reconstruct = gauge_param_in.reconstruct;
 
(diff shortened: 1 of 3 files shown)


weinbe2:
Don't merge this yet, accidentally introduced a bug where the long link recon gets used for the fat link during gauge reconstruct, which doesn't work very well...

tests/staggered_gauge_utils.cpp
```diff
@@ -300,10 +300,10 @@ void computeStaggeredPlaquetteQDPOrder(void** qdp_link, double plaq[3], const Qu
   gauge_param.cpu_prec = gauge_param_in.cpu_prec;
 
   gauge_param.cuda_prec = gauge_param_in.cuda_prec;
-  gauge_param.reconstruct = QUDA_RECONSTRUCT_NO;
+  gauge_param.reconstruct = gauge_param_in.reconstruct; //QUDA_RECONSTRUCT_NO;
```
maddyscientist: why leave the old buggy routine in the comment as opposed to just deleting this?  Else this will be confusing to anyone looking at this in the future.

weinbe2: Fixed, duly noted!

mathiaswagner:
@jenkins test this please

maddyscientist:
Merging.  Note that last build test shows as incomplete above (`ubuntu1804_gnu7_openmpi3_cmake3144_cuda101`), but looking at Jenkins, it did complete.  Looks like something is up with the Jenkins not updating the results on github.  Anyway, not the problem with this commit, so merging.

PR #901, opened 2019-09-13
Feature/alias sloppy

maddyscientist:
* Adds new method for allocating `ColorSpinorField` objects that alias to another's memory allocation, including in different precisions.  This allows us to reduce memory overhead when we require multiple concurrent temporaries in different precisions.
* Applied this optimization to GCR and CA-GCR.
* Some further cleanup and optimization of GCR to remove unneeded copy and zero calls

### DIFF
--- a/include/color_spinor_field.h
+++ b/include/color_spinor_field.h
@@ -540,6 +540,16 @@ namespace quda {
     static ColorSpinorField* Create(const ColorSpinorField &src, const ColorSpinorParam &param);
 
     /**
+       @brief Create a field that aliases this field's storage.  The
+       alias field can use a different precision than this field,
+       though it cannot be greater.  This functionality is useful for
+       the case where we have multiple temporaries in different
+       precisions, but do not need them simultaneously.  Use this functionality with caution.
+       @param[in] param Parameters for the alias field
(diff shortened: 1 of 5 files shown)


mathiaswagner:
Would that also allow to dynamically switch the precision of a field?

maddyscientist:
Not really, unless I misunderstand you.

alexstrel:
Would it be more convenient to pass to `CreateAlias` precision only, not the whole cs parameter?  It seems the alias field shares all info from the original field except precision (and the creation type). I think, in some situations the code could be a bit cleaner, e.g., when an alias is created in a different scope.

mathiaswagner:
I like the optimization and that we can reduce the memory footprint. Only thing I wonder about is the memory management. Having multiple fields using the same memory allocation scream smart pointer to me ...

alexstrel:
but same is true for the legacy reference field, isn't it?

mathiaswagner:
yeah, improving the memory management is probably beyond the scope of this. We have all that checking whether fields alias and then making sure to issue the right free/delete calls in the right order. Just a lot of code that we could get rid of now that we use C++14.

Issue #707, opened 2018-07-02
Incorrect logic for determining bi-directional coarsening in multigrid


weinbe2:
In feature/multigrid, the current logic for determining if you need to do bi-directional coarsening is if the previous level is preconditioned. This is true in every practical use case for Wilson-Clover, but it isn't strictly true: you need to do bi-directional coarsening if _any_ previous operator was preconditioned, because once that's the case you've broken the Hermitian conjugate symmetry between forward and backward links. This is relevant with, for example, staggered MG, where so far we've found _not_ preconditioning the lower levels leads to a better algorithm. 

The relevant boolean `bidirectional_links` is defined in the function `calculateY` at the bottom of the file `coarse_op.cuh`. 

Since this isn't an issue for any practical use case for Wilson-Clover, it's low priority. This is mostly a note for myself to take care of this in the future.

maddyscientist:
This will be closed by #902

weinbe2:
Closed by https://github.com/lattice/quda/pull/902

PR #902, opened 2019-09-16
Staggered+HISQ MG

### LINKED ISSUE #213: Support for nSpin = 1 in multigrid
Currently, multigrid assumes that the fine operator is a Wilson-type operator with 4 spin components.  Support for nSpin = 1 is needed for staggered multigrid.

weinbe2:
Many years in the making, this PR adds support for staggered and HISQ multigrid to mainline QUDA. This algorithm makes use of the method of Kahler-Dirac preconditioning in going from level 1 to level 2 (fine to "pseudo-fine"), along with a truncation of the long links in the HISQ case, and has specially optimized prolong, restrict, and coarsening routines which reflect the nature of this "coarsening" being simply a unitary transform as implemented in QUDA. This also required the addition of several checks of if this a staggered MG solve on the top level because some operators become pointless or don't make sense, for ex saving/loading near-null vectors for the top level. Beyond this custom top level behavior, on subsequent levels, the algorithm leverages standard coarsening routines used for Wilson-clover MG, though at higher `Nc = 64, 96`. 

So far as I can tell, I've put staggered compile guards around these extra these Nc instantiations so compile time should not go up for pure Wilson-clover MG builds (so long as you explicitly turn staggered off), but I'd appreciate it if folks double-checked that. 

The algorithm works ideally as a 5 level algorithm: fine, pseudo-fine, intermediate, coarse, SVD deflation thereof. The algorithm works very well on "sufficiently smooth" gauge fields, where the cut-off seems to be around 0.07 fm. The algorithm breaks down going between level 2 and level 3 ("pseudo-fine" and "intermediate"), so the problem there is more a statement of finding good near-null vectors (and perhaps exploring less aggressive aggregation: this is now more appealing because of memory oversubscription, I just haven't had the chance to explore it yet). 

I'll post a representative command for a single-GPU run in the comments below momentarily (I'm just on the wrong machine at the moment). 

Outstanding features that don't effect functional correctness and usefulness, in rough order of priority:
* Exposing the MG algorithm via the MILC interface
* Reducing memory overheads via a custom pseudo-fine operator which does not apply an explicit coarse hopping term but instead performs a fused gauge field plus explicit `Xinv` operation. This could cut the required number of nodes during linear solves by 30 to 40%. (For simplicity in further coarsening, it's useful to explicitly construct the coarse hopping term during set-up---this is where CUDA managed memory support comes in).
* Adding native support for the Schur-preconditioned staggered/HISQ operator on the fine level. Since the Schur-preconditioned linear system can be turned into an equivalent linear system on the full operator, this isn't a functional limitation. Properly including it requires adding support for right-preconditioned (residual-preserving) coarse operators, as opposed to the left-preconditioned (error-preserving) form.

One of the biggest pieces of help I could use with this PR is if folks could make sure clover MG and twisted mass MG weren't inadvertently broken, I haven't carefully tested this recently and don't know how to really probe the functional depths of the algorithm as used in practice anyway. In the interest of your sanity, make sure you compile with staggered operator support **off**---`Nc = 96` routines take forever to compile :) 

I think that covers thoughts off the top of my head... but ask every question you can think of!

Closes #213, #707


weinbe2:
As promised last week, here's a reference command:
```
ARGSTR="--dim 16 16 16 16 --prec double --prec-sloppy single --prec-precondition half --prec-null half --mg-smoother-halo-prec single --dslash-type asqtad --solve-type direct --compute-fat-long true --verbosity verbose --nsrc 20 "
ARGSTR=$ARGSTR"--load-gauge ../../l16t16b6p5 --mass 0.0 --tadpole-coeff 0.9 "
ARGSTR=$ARGSTR"--mg-levels 4 --mg-coarse-solve-type 0 direct "
ARGSTR=$ARGSTR"--mg-setup-inv 1 cgnr --mg-setup-maxiter 1 1000 --mg-setup-tol 1 1e-5 "
ARGSTR=$ARGSTR"--mg-setup-inv 2 cgnr --mg-setup-maxiter 1 1000 --mg-setup-tol 1 1e-5 "
ARGSTR=$ARGSTR"--mg-block-size 0 2 2 2 2 --mg-nvec 0 24 --mg-coarse-solve-type 1 direct-pc --mg-smoother-solve-type 0 direct    --mg-smoother 0 ca-gcr --mg-nu-pre 0 0 --mg-nu-post 0 8 --mg-smoother-tol 0 1e-10 --mg-coarse-solver-tol 1 5e-2 --mg-coarse-solver-maxiter 1 16   --mg-coarse-solver 1 gcr "
ARGSTR=$ARGSTR"--mg-block-size 1 2 2 2 2 --mg-nvec 1 64 --mg-n-block-ortho 1 2 --mg-coarse-solve-type 2 direct-pc --mg-smoother-solve-type 1 direct-pc --mg-smoother 1 ca-gcr --mg-nu-pre 1 0 --mg-nu-post 1 8 --mg-smoother-tol 1 1e-10 --mg-coarse-solver-tol 2 0.25 --mg-coarse-solver-maxiter 2 16   --mg-coarse-solver 2 gcr "
ARGSTR=$ARGSTR"--mg-block-size 2 2 2 2 2 --mg-nvec 2 96 --mg-n-block-ortho 2 2 --mg-coarse-solve-type 3 direct-pc --mg-smoother-solve-type 2 direct-pc --mg-smoother 2 ca-gcr --mg-nu-pre 2 0 --mg-nu-post 2 8 --mg-smoother-tol 2 1e-10 --mg-coarse-solver-tol 3 0.25 --mg-coarse-solver-maxiter 3 1024   --mg-coarse-solver 3 cgne "
./staggered_multigrid_invert_test $ARGSTR
```

weinbe2:
Updated reference command:
```
ARGSTR="--dim 16 16 16 16 --prec double --prec-sloppy single --prec-precondition half --prec-null half --mg-smoother-halo-prec single --dslash-type asqtad --solve-type direct --compute-fat-long true --verbosity verbose --nsrc 1 "
ARGSTR=$ARGSTR"--load-gauge ../../l16t16b6p5 --mass 0.0 --tadpole-coeff 0.9 "
ARGSTR=$ARGSTR"--mg-levels 4 --mg-coarse-solve-type 0 direct "
ARGSTR=$ARGSTR"--mg-setup-inv 1 cgnr --mg-setup-maxiter 1 1000 --mg-setup-tol 1 1e-5 "
ARGSTR=$ARGSTR"--mg-setup-inv 2 cgnr --mg-setup-maxiter 2 1000 --mg-setup-tol 2 1e-5 "
ARGSTR=$ARGSTR"--mg-block-size 0 2 2 2 2 --mg-nvec 0 24 --mg-coarse-solve-type 1 direct-pc --mg-smoother-solve-type 0 direct    --mg-smoother 0 ca-gcr --mg-nu-pre 0 0 --mg-nu-post 0 8 --mg-smoother-tol 0 1e-10 --mg-coarse-solver-tol 1 5e-2 --mg-coarse-solver-maxiter 1 16   --mg-coarse-solver 1 gcr "
ARGSTR=$ARGSTR"--mg-block-size 1 2 2 2 2 --mg-nvec 1 64 --mg-n-block-ortho 1 2 --mg-coarse-solve-type 2 direct-pc --mg-smoother-solve-type 1 direct-pc --mg-smoother 1 ca-gcr --mg-nu-pre 1 0 --mg-nu-post 1 2 --mg-smoother-tol 1 1e-10 --mg-coarse-solver-tol 2 0.25 --mg-coarse-solver-maxiter 2 16   --mg-coarse-solver 2 gcr "
ARGSTR=$ARGSTR"--mg-block-size 2 2 2 2 2 --mg-nvec 2 96 --mg-n-block-ortho 2 2 --mg-coarse-solve-type 3 direct-pc --mg-smoother-solve-type 2 direct-pc --mg-smoother 2 ca-gcr --mg-nu-pre 2 0 --mg-nu-post 2 2 --mg-smoother-tol 2 1e-10 --mg-coarse-solver-tol 3 2e-2 --mg-coarse-solver-maxiter 3 1024   --mg-coarse-solver 3 cgnr "
./staggered_multigrid_invert_test $ARGSTR
```

include/kernels/staggered_coarse_op_kernel.cuh
```diff
@@ -0,0 +1,316 @@
+#include <color_spinor_field_order.h>
```
maddyscientist: Remind me, why keep these kernels in a separate file vs `coarse_op_kernel.cuh`?

include/quda.h
```diff
@@ -623,6 +623,12 @@ extern "C" {
     /** Whether to run null vector oblique checks once set up is complete */
     QudaBoolean run_oblique_proj_check;
 
+    /** Whether to use eigenvectors for the nullspace */
+    QudaBoolean use_low_modes;
+
+    /** Deflate the coarsest grid  */
+    QudaBoolean deflate_coarsest;
```
maddyscientist: `use_low_modes` and `deflate_coarsest` don't look specific to staggered mg.  Why are these in this branch but not in develop?

include/transfer.h
```diff
@@ -307,7 +311,34 @@ namespace quda {
   void Restrict(ColorSpinorField &out, const ColorSpinorField &in, const ColorSpinorField &v, 
 		int Nvec, const int *fine_to_coarse, const int *coarse_to_fine, const int * const *spin_map,
 		int parity=QUDA_INVALID_PARITY);
-  
+
+
+#ifdef GPU_STAGGERED_DIRAC
```
maddyscientist: We prefer to define all functions in headers without macro guards, and leave the macro enablement to the implementation, with an appropriate error message if a function is called but not compiled.

lib/block_orthogonalize.cu
```diff
@@ -130,22 +130,22 @@ namespace quda {
...
-	}
+        if (V.FieldOrder() == QUDA_SPACE_SPIN_COLOR_FIELD_ORDER && B[0]->FieldOrder() == QUDA_SPACE_SPIN_COLOR_FIELD_ORDER) {
+          typedef FieldOrderCB<RegType,nSpin,nColor,nVec,QUDA_SPACE_SPIN_COLOR_FIELD_ORDER,vFloat,vFloat,DISABLE_GHOST> Rotator;
+          typedef FieldOrderCB<RegType,nSpin,nColor,1,QUDA_SPACE_SPIN_COLOR_FIELD_ORDER,bFloat,bFloat,DISABLE_GHOST> Vector;
+          CPU<Rotator,Vector>(B, std::make_index_sequence<nVec>());
+        } else {
+          errorQuda("Unsupported field order %d\n", V.FieldOrder());
+        }
+            } else {
+        #if __COMPUTE_CAPABILITY__ >= 300
```
maddyscientist: We can probably nuke this conditional compilation since we don't support Fermi any more.

lib/coarse_op.cu
```diff
@@ -124,6 +124,14 @@ namespace quda {
       calculateY<Float,vFloat,fineColor,fineSpin,24,coarseSpin>(Y, X, Yatomic, Xatomic, uv, av, T, g, c, kappa, mu, mu_factor, dirac, matpc);
     } else if (coarseColor == 32) {
       calculateY<Float,vFloat,fineColor,fineSpin,32,coarseSpin>(Y, X, Yatomic, Xatomic, uv, av, T, g, c, kappa, mu, mu_factor, dirac, matpc);
+#if 0 // only needed for Laplace MG. Not necessary for staggered MG Lanczos
```
maddyscientist: Random comment: at some point we should fill in the missing pieces for regular Laplace MG!

lib/coarse_op.cuh
```diff
@@ -8,7 +8,10 @@ namespace quda {
   // For coarsening un-preconditioned operators we use uni-directional
   // coarsening to reduce the set up code.  For debugging we can force
   // bi-directional coarsening.
-  static bool bidirectional_debug = false;
+  // ESW hack for now: need to add logic so bi-directional coarsening
+  // is triggered if ANY previous level is preconditioned, not just
+  // the current level.
+  static bool bidirectional_debug = true;
```
maddyscientist: This needs to be fixed, else it will double the cost of Wilson MG coarsening.

lib/coarsecoarse_op.cu
```diff
@@ -107,8 +107,16 @@ namespace quda {
 #endif
     } else if (coarseColor == 24) {
       calculateYcoarse<Float,vFloat,fineColor,fineSpin,24,coarseSpin>(Y, X, Yatomic, Xatomic, uv, T, g, clover, cloverInv, kappa, mu, mu_factor, dirac, matpc);
+#ifdef GPU_WILSON_DIRAC
```
maddyscientist: Do you need 24 for staggered as well, or can you extend this guard up to 24 too?

maddyscientist: Also, shouldn't these `GPU_WILSON_DIRAC` be `NSPIN4`.  E.g., we can compile twisted-mass without Wilson.

lib/cuda_color_spinor_field.cpp
```diff
@@ -295,6 +303,15 @@ namespace quda {
 #ifdef USE_TEXTURE_OBJECTS
   void cudaColorSpinorField::createTexObject() {
 
+    // We may set `v` and `norm` to `std::numeric_limits<unsigned long>::max()`
```
maddyscientist: What's the motivation for this?

lib/dirac_coarse.cpp
```diff
@@ -141,11 +146,19 @@ namespace quda {
     if (gpu_setup) dirac->createCoarseOp(*Y_d,*X_d,*transfer,kappa,mass,Mu(),MuFactor());
     else dirac->createCoarseOp(*Y_h,*X_h,*transfer,kappa,mass,Mu(),MuFactor());
 
+    printfQuda("ESW: About to run createYhat\n");
```
maddyscientist: Remove this print statements

lib/dirac_improved_staggered.cpp
```diff
@@ -122,6 +123,23 @@ namespace quda {
     // do nothing
   }
 
+  /* Creates the coarse grid dirac operator
```
maddyscientist: Make this doxgygen

maddyscientist: and move the comment to the header declaration

lib/dirac_staggered.cpp
```diff
@@ -126,10 +127,21 @@ namespace quda {
...
+  about the coarse grid blocking, as well as
+  having prolongate and restrict member functions
+  
+  Returns: Color matrices Y[0..2*dim] corresponding
+  to the coarse grid operator.  The first 2*dim
+  matrices correspond to the forward/backward
+  hopping terms on the coarse grid.  Y[2*dim] is
+  the color matrix that is diagonal on the coarse
+  grid
+  */
```
maddyscientist: Doxygen and move to header

lib/prolongator.cu
```diff
@@ -234,55 +234,86 @@ namespace quda {
...
       if (nVec == 4) {
-	Prolongate<Float,fineSpin,fineColor,coarseSpin,4>(out, in, v, fine_to_coarse, parity);
+        Prolongate<Float,fineSpin,fineColor,coarseSpin,4>(out, in, v, fine_to_coarse, parity);
       } else if (nVec == 6) { // Free field Wilson
-  Prolongate<Float,fineSpin,fineColor,coarseSpin,6>(out, in, v, fine_to_coarse, parity);
+        Prolongate<Float,fineSpin,fineColor,coarseSpin,6>(out, in, v, fine_to_coarse, parity);
       } else if (nVec == 24) {
-	Prolongate<Float,fineSpin,fineColor,coarseSpin,24>(out, in, v, fine_to_coarse, parity);
+        Prolongate<Float,fineSpin,fineColor,coarseSpin,24>(out, in, v, fine_to_coarse, parity);
+#ifdef GPU_WILSON_DIRAC
```
maddyscientist: Should be `NSPIN4` to ensure non-Wilson Dirac operator MG still works

lib/restrictor.cu
```diff
@@ -201,55 +202,86 @@ namespace quda {
...
       if (nVec == 4) {
-	Restrict<Float,fineSpin,fineColor,coarseSpin,4>(out, in, v, fine_to_coarse, coarse_to_fine, parity);
+        Restrict<Float,fineSpin,fineColor,coarseSpin,4>(out, in, v, fine_to_coarse, coarse_to_fine, parity);
       } else if (nVec == 6) { // free field Wilson
-  Restrict<Float,fineSpin,fineColor,coarseSpin,6>(out, in, v, fine_to_coarse, coarse_to_fine, parity);
+        Restrict<Float,fineSpin,fineColor,coarseSpin,6>(out, in, v, fine_to_coarse, coarse_to_fine, parity);
       } else if (nVec == 24) {
-	Restrict<Float,fineSpin,fineColor,coarseSpin,24>(out, in, v, fine_to_coarse, coarse_to_fine, parity);
+        Restrict<Float,fineSpin,fineColor,coarseSpin,24>(out, in, v, fine_to_coarse, coarse_to_fine, parity);
+#ifdef GPU_WILSON_DIRAC
```
maddyscientist: `NSPIN4` not `GPU_WILSON_DIRAC`

lib/spinor_noise.cu
```diff
@@ -163,8 +163,12 @@ namespace quda {
       spinorNoise<real,Ns,24>(src, randstates, type);
     } else if (src.Ncolor() == 32) {
       spinorNoise<real,Ns,32>(src, randstates, type);
+    } else if (src.Ncolor() == 64) {
+      spinorNoise<real,Ns,64>(src, randstates, type);
+    } else if (src.Ncolor() == 96) {
+      spinorNoise<real,Ns,96>(src, randstates, type);
     } else {
-      errorQuda("nColor = %d not implemented", src.Ncolor());
+      errorQuda("Ncolor = %d not implemented", src.Ncolor());
```
maddyscientist: Style.  Camel case is used typically for variables or templates that define a number of something, e.g., `n` amounts of something, e.g., `nSpin`, `nColor`, `nDim`.  Standard capitalization of these quantities saved for function name.  So in this print statement, I prefer camel case.

lib/util_quda.cpp
```diff
@@ -85,7 +85,7 @@ void pushVerbosity(QudaVerbosity verbosity)
   vstack.push(getVerbosity());
   setVerbosity(verbosity);
 
-  if (vstack.size() > 10) {
+  if (vstack.size() > 15) {
```
maddyscientist: Curious why this change was needed?

tests/multigrid_evolve_test.cpp
```diff
@@ -60,6 +60,31 @@ display_test_info()
   printfQuda("Outer solver paramers\n");
   printfQuda(" - pipeline = %d\n", pipeline);
 
+  if(low_mode_check || use_low_modes || deflate_coarsest) {
```
maddyscientist: Why are change to multigrid_evolve_test in this pull?

include/externals/cub/thread/thread_store.cuh
```diff
@@ -145,10 +145,10 @@ template <int MAX>
 struct IterateThreadStore<MAX, MAX>
 {
     template <CacheStoreModifier MODIFIER, typename T>
-    static __device__ __forceinline__ void Store(T * /*ptr*/, T * /*vals*/) {}
+    static __device__ __forceinline__ void Store(T *ptr, T *vals) {}
```
maddyscientist: The changes on lines 148 and 151 were accidental by me.  It's the change on 351 is what counts.  Can you revert the changes on lines 148 and 151 with your next commit @weinbe2

lib/coarse_op_preconditioned.cu
```diff
@@ -235,6 +237,10 @@ namespace quda {
     case 32: calculateYhat<storeFloat,Float,32>(Yhat, Xinv, Y, X); break;
     case 48: calculateYhat<storeFloat,Float,48>(Yhat, Xinv, Y, X); break;
     case 64: calculateYhat<storeFloat,Float,64>(Yhat, Xinv, Y, X); break;
+    case 96: calculateYhat<storeFloat,Float,96>(Yhat, Xinv, Y, X); break;
```
maddyscientist: How bad is compilation time for Yhat?

maddyscientist:
Had first code review and left comments.  Looks like we need to get some other pulls merged in before this one, e.g., #898 .

Definitely needs a another go through clang format, but wait on that until the end and other pulls are merged.

weinbe2:
Thanks @maddyscientist, I'll go through these soon.

include/quda.h
cpviolator: These are entirely vestigial. No longer in develop because we now use `--mg-eig <level> <true/false>`

include/externals/cub/thread/thread_store.cuh
weinbe2: Done locally!

include/kernels/staggered_coarse_op_kernel.cuh
weinbe2: Mostly historical at this point. The main blocker for a while has been that `coarse_op_kernel.cuh` is long/confusing enough as it is, and because the routines in this driver compile quickly enough the benefit of `jitify` support wasn't a good incentive. If you'd really rather I merged it into `coarse_op_kernel.cuh`, I will.

maddyscientist: No need.  Just asking.

include/quda.h
weinbe2: Good catch. It looks like at some point the copy of `multigrid_evolve_test` in this branch fell (far) out of sync with what's in `develop`, which is why this never threw a warning. I've both removed these declarations from `include/quda.h` and fixed `multigrid_evolve_test`.

include/transfer.h
weinbe2: Noted and fixed. `staggered_prolong_restrict.cu` already had the appropriate macro guards in it, so there were no changes needed there.

lib/block_orthogonalize.cu
weinbe2: Done.

lib/block_orthogonalize.cu
```diff
@@ -246,12 +246,14 @@ namespace quda {
       } else if (Nvec == 24) {
         BlockOrthogonalize<vFloat, bFloat, nSpin, spinBlockSize, nColor, 24>(V, B, fine_to_coarse, coarse_to_fine,
                                                                              geo_bs, n_block_ortho);
+#ifdef GPU_WILSON_DIRAC
       } else if (Nvec == 32) {
```
weinbe2: FYI---I've since replaced these `GPU_WILSON_DIRAC` and `GPU_STAGGERED_DIRAC`s with `NSPIN4` and `NSPIN1`, respectively.

lib/coarse_op.cu
weinbe2: Agreed! The main change we'd, in principle, want to make is getting rid of preserving chirality, since it's meaningless. (That said, if we kept staggered chirality around on a first pass, all of the pieces may already be there.)

lib/coarse_op.cuh
weinbe2: Good catch, fixed.

lib/coarse_op_preconditioned.cu
weinbe2: Not sure, tbh. Never bad enough for me to complain ;) But I'll add some macro guards since they won't hurt.

weinbe2: And I guess get rid of the unnecessary ones---we only need 12 (free field Wilson), 48 (Wilson and staggered), 64 (Wilson), 128 (staggered), and 192 (staggered).

lib/coarsecoarse_op.cu
weinbe2: We need 24 for coarsening the pseudo-fine/KD operator. (This will still be true after adding the optimized KD op, since we'll still explicitly build the coarse links exactly for coarsening, we'll just let managed memory sweep the details under the rug.) Duly noted on `NSPIN4` and `NSPIN1`, fixed.

lib/color_spinor_pack.cu
```diff
@@ -278,6 +300,18 @@ namespace quda {
       genericPackGhost<Float,ghostFloat,order,Ns,precision_spin_color_mapper<Float,ghostFloat,Ns,768>::nColor>(ghost, a, parity, nFace, dagger, destination);
     } else if (a.Ncolor() == 1024) {
       genericPackGhost<Float,ghostFloat,order,Ns,precision_spin_color_mapper<Float,ghostFloat,Ns,1024>::nColor>(ghost, a, parity, nFace, dagger, destination);
+#ifdef NSPIN1
+    } else if (a.Ncolor() == 1536) { // staggered KD 24 -> 64 nvec
+      genericPackGhost<Float,ghostFloat,order,Ns,precision_spin_color_mapper<Float,ghostFloat,Ns,1536>::nColor>(ghost, a, parity, nFace, dagger, destination);
```
weinbe2: FYI, added some compile guards, removed some instantiations that are unnecessary for Wilson or staggered.

lib/cuda_color_spinor_field.cpp
weinbe2: This was a clever little hack we employed once we added the specialized routines for staggered prolong/restrict. We couldn't easily strip out the `B`/`V` fields in that case because they were used in some places as metadata containers used to query the number of degrees of freedom, etc, so we created a special flavor of `ColorSpinorField` that never actually allocated anything.

This created an issue with the texture backend where it would try to bind to data that didn't exist, so we chose `std::numeric_limits<long long unsigned int>>::max()` as a value to query for before trying to bind a texture, since the value of a memory address effectively will never equal that.

lib/dirac_coarse.cpp
weinbe2: I elected to add verbosity guards and make the debug statements more descriptive (and remove the `ESW: `) instead. I found them useful during development and debugging.

weinbe2: I also used this as a reminder to get rid of/edit out all `ESW` comments

lib/dirac_improved_staggered.cpp
weinbe2: These comments also reside in `dirac_wilson.cpp`, etc, effectively verbatim (i.e., without doxygen). If you'd like, I can remove them from both places.

In all cases there were doxygen comments in the header declaration already.

lib/dirac_staggered.cpp
weinbe2: See above

lib/multigrid.cpp
```diff
@@ -492,7 +499,7 @@ namespace quda
       param_coarse_solver->pipeline = 8;
 
       param_coarse_solver->maxiter = param.mg_global.coarse_solver_maxiter[param.level+1];
-      param_coarse_solver->Nkrylov = param_coarse_solver->maxiter < 20 ? param_coarse_solver->maxiter : 20;
+      param_coarse_solver->Nkrylov = param_coarse_solver->maxiter < param_coarse_solver->Nkrylov ? param_coarse_solver->maxiter : param_coarse_solver->Nkrylov;
```
weinbe2: @maddyscientist I feel like this change caused controversy at some point---do you remember why? I think it was just because it changed some default behavior that was being assumed in Chroma, but I'm not sure.

lib/prolongator.cu
weinbe2: Done.

lib/restrictor.cu
weinbe2: Done.

lib/spinor_noise.cu
weinbe2: Fixed here, and elsewhere in the file where it's relevant. Not exactly sure why I had that in the first place.

lib/transfer.cpp
```diff
@@ -123,6 +136,16 @@ namespace quda {
     param.fieldOrder = location == QUDA_CUDA_FIELD_LOCATION ? QUDA_FLOAT2_FIELD_ORDER : QUDA_SPACE_SPIN_COLOR_FIELD_ORDER;
     param.setPrecision(location == QUDA_CUDA_FIELD_LOCATION ? null_precision : B[0]->Precision());
 
+    if (is_staggered) {
```
weinbe2: @maddyscientist re: assigning pointers to `std::numeric_limits<long long unsigned int>::max()`

lib/util_quda.cpp
weinbe2: I never did a super-explicit trace and counting, but doing a four-level solve blew out a size of 10, while 15 seemed to be enough. The `Level 1`, etc, prints during the setup and solve always made consistent sense, so it wasn't actually because of a `popVerbosity/OutputPrefix` mismatch.

weinbe2: I don't deny that it's surprising 10 wasn't sufficient.

tests/multigrid_evolve_test.cpp
weinbe2: Something just went out of whack at some point, I can't explain it. In lieu of thinking about it I grabbed the latest version of `test/multigrid_evolve_test` and added the `is_staggered = QUDA_BOOLEAN_NO` check.

include/multigrid.h
```diff
@@ -99,6 +99,9 @@ namespace quda {
     /** Filename for where to load/store the null space */
     char filename[100];
 
+    /** Whether or not this is a staggered solve or not */
+    QudaBoolean is_staggered;
```
maddyscientist: There is no reason not to simply make this a `bool`: `QudaBoolean` exists only for the C interface.

lib/color_spinor_pack.cu
```diff
@@ -210,24 +210,42 @@ namespace quda {
...
   template<int nColor_> struct precision_spin_color_mapper<double,double,4,nColor_> { static constexpr int nColor = 3; };
+#endif
 #endif
 
   template <typename Float, typename ghostFloat, QudaFieldOrder order, int Ns>
   inline void genericPackGhost(void **ghost, const ColorSpinorField &a, QudaParity parity,
 			       int nFace, int dagger, MemoryLocation *destination) {
-    
+
+#ifndef GPU_STAGGERED_DIRAC
```
maddyscientist: Any reason why this guard isn't using `NSPIN1`?

lib/cuda_color_spinor_field.cpp
maddyscientist: Better to use `uint64_t` than `long long unsigned int` here.
https://en.cppreference.com/w/cpp/types/integer

lib/dirac_coarse.cpp
```diff
@@ -56,11 +56,14 @@ namespace quda {
   void DiracCoarse::createY(bool gpu, bool mapped) const
   {
     int ndim = transfer->Vectors().Ndim();
+    if (ndim == 5 && transfer->Vectors().Nspin() != 4)
```
maddyscientist: When we re-add multi-RHS operator support we will be ripping out the old 5th dimensional trick.  With this in mind, and given this pull be merged then, can you add a comment to each line like this that we will need to update when we do so.  E.g., add `// FIXME MRHS NDIM hack` that we can easily grep.

lib/dirac_improved_staggered.cpp
maddyscientist: In that case, make sure that any details from here and dirac_wilson.cpp are percolated to their respective comments in the header, and delete the comments here.  No need to duplication of comments, and the convention is to have details like this in the header.

lib/dirac_staggered.cpp
maddyscientist: Again, make sure all details are in header and delete from the .cpp file.  Can put in a in line or two of comments if warranted, but details like this should be in header.

lib/interface_quda.cpp
```diff
@@ -2479,6 +2479,16 @@ multigrid_solver::multigrid_solver(QudaMultigridParam &mg_param, TimeProfile &pr
   csParam.fieldOrder = mg_param.setup_location[0] == QUDA_CUDA_FIELD_LOCATION ? QUDA_FLOAT2_FIELD_ORDER : QUDA_SPACE_SPIN_COLOR_FIELD_ORDER;
   csParam.mem_type = mg_param.setup_minimize_memory == QUDA_BOOLEAN_YES ? QUDA_MEMORY_MAPPED : QUDA_MEMORY_DEVICE;
   B.resize(mg_param.n_vec[0]);
+
+  if (mg_param.is_staggered == QUDA_BOOLEAN_YES) { 
+    // Create the ColorSpinorField as a "container" for metadata.
+    csParam.create = QUDA_REFERENCE_FIELD_CREATE;
+
+    // These never get accessed, `nullptr` on its own leads to an error in texture binding
+    csParam.v = (void*)std::numeric_limits<long long unsigned int>::max();
```
maddyscientist: as above, change this to `uint64_t`

lib/transfer.cpp
maddyscientist: change to `uint64_t`

tests/multigrid_invert_test.cpp
```diff
@@ -377,6 +377,9 @@ void setMultigridParam(QudaMultigridParam &mg_param)
   mg_param.run_low_mode_check = low_mode_check ? QUDA_BOOLEAN_YES : QUDA_BOOLEAN_NO;
   mg_param.run_oblique_proj_check = oblique_proj_check ? QUDA_BOOLEAN_YES : QUDA_BOOLEAN_NO;
 
+  // Is NOT a staggered solve
+  mg_param.is_staggered = QUDA_BOOLEAN_NO;
```
maddyscientist: Having to set this is an interface change.  I also just realized that you haven't set `is_staggered` in check_params.h.  Probably best to default to `QUDA_BOOLEAN_NO` so that all codes that use QUDA MG for Wilson/clover/twisted-mass fermions don't suddenly break.

tests/test_util.cpp
```diff
@@ -111,7 +111,7 @@ void initComms(int argc, char **argv, int *const commDims)
   initRand();
```
maddyscientist: Lots of lines changed here unnecessarily.  Any reason for this?  I realize there are some formatting fixes, but all these changes seem excessive given we like to preserve history.

lib/multigrid.cpp
maddyscientist: The "controversy" was over this line
```c++
param_coarse_solver->Nkrylov = param.mg_global.coarse_solver_ca_basis_size[param.level+1];
```
since it changed the prior behaviour when using the CA-GCR coarse solver.  I got over that though, and that's no problem now 😜

include/externals/cub/thread/thread_store.cuh
```diff
@@ -348,7 +348,7 @@ __device__ __forceinline__ void ThreadStoreVolatilePtr(
     for (int i = 0; i < SHUFFLE_MULTIPLE; ++i)
         reinterpret_cast<ShuffleWord*>(words)[i] = reinterpret_cast<ShuffleWord*>(&val)[i];
 
-    IterateThreadStore<0, VOLATILE_MULTIPLE>::template Dereference(
+    IterateThreadStore<0, VOLATILE_MULTIPLE>::template Dereference<volatile VolatileWord*,VolatileWord>(
```
maddyscientist: Just noting that this is something that this patch is something that I need to ensure is included in official CUB.....

lib/dslash_coarse.cu
```diff
@@ -4,6 +4,7 @@
 #include <worker.h>
```
maddyscientist: Lots of formatting changes for no apparent reason in this file.

maddyscientist:
Second review done. Looks mostly good and shouldn't be too hard to fix this up.  @weinbe2 once you have finished making the changes, the last step would be for you to write up a wiki page with instructions for how to run QUDA MG for a given staggered lattice.  Then I or someone else can work from those instructions and check we can reproduce sanity.  Beyond the usual battery of tests, with additional testing for Wilson MG, I think this pull should be complete.

Sound like a plan?

weinbe2:
@maddyscientist sounds good! I also need to merge in the latest version of develop, that's more of a reminder for myself than anything else.

lib/transfer.cpp
weinbe2: Did a sweep with `grep`, replaced it everywhere

lib/interface_quda.cpp
weinbe2: Done

lib/color_spinor_pack.cu
weinbe2: Nope :) Fixed

include/multigrid.h
weinbe2: (Probably) fixed, need to go to a machine where I can compile.

lib/dirac_coarse.cpp
weinbe2: Done.

tests/multigrid_invert_test.cpp
weinbe2: Done, I think correctly.

lib/dslash_coarse.cu
weinbe2: Had an offline discussion that not fixing this one is fine.

tests/test_util.cpp
weinbe2: Looks like the _only_ changes were formatting, so I just copied the file from develop directly. Should hopefully cure this.

lib/cuda_color_spinor_field.cpp
weinbe2: Done

lib/dirac_improved_staggered.cpp
weinbe2: Will fix soon

lib/dirac_staggered.cpp
weinbe2: Will fix soon

lib/dirac_improved_staggered.cpp
weinbe2: Done

lib/dirac_staggered.cpp
weinbe2: done

weinbe2:
@maddyscientist this is ready for a "final" review and specifically hardcore Wilson-type testing

weinbe2:
@cpviolator whenever you have the chance, can you please run the most vicious battery of Wilson-type MG tests you can conceive of against this branch? In my mind that'd at least be running your standard tests of Twisted MG to make sure everything's ship-shape, along with anything else you can think of. 

My only specific requests are to compile _without_ staggered enabled, at least as a first pass---I want to make sure I didn't accidentally wrap something essential in an `#ifdef NSPIN1` or `#ifdef [whatever staggered is]`. Also, when you have a compile without staggered enabled, try to run with Nc = 64 and 96---this shouldn't work, but I want to make sure. (You'd know if they got compiled in accidentally anyway, `dslash_coarse.cu` takes absolutely forever to compile now...)

Of course, don't let this stop you from also testing a build with staggered too :)

cpviolator:
@weinbe2 Sure thing. I'll zip through all the physical lattices I have on summit with staggered off, then recompile with on and do the same. Just be aware that I found #934 a few days ago so that might occur library wide too.

weinbe2:
Thanks @cpviolator , as well as for the FYI. I've been running HISQ MG with 18/13/9 without issue for a while, FWIW, though at some point 18/9/[9 or 13] was broken, and I can't say I've followed up since.

cpviolator:
Aside from #934 note seeing anything amiss with wilson types on summit. `nVec=64/96` does indeed error out. Compiling with staggered. See you in the new year!

weinbe2:
That's awesome to hear @cpviolator, thanks!!

cpviolator:
Compiled with staggered off and on. No issues with Wilson or twisted clover physical lattices. Indeed, Nc=64/96 will error out without staggered enabled.

cpviolator:
No comments at all that will hold up this epic pull!

One point of note: while I was testing with staggered I fell foul of missing CMake parameter switches `QUDA_LINK_HISQ`, `QUDA_LINK_ASQTAD`. Can these be hidden behind activation of staggered fermions, or automatically switched to `ON` if staggered fermions are compiled?

maddyscientist:
Good call @cpviolator.  I would happy just removing these cmake flags and having hisq link stuff built conditional purely on whether staggered fermions are enabled.  Thoughts @weinbe2 @mathiaswagner

mathiaswagner:
Building HISQ link whenever staggered is enabled sounds good to me.

weinbe2:
@maddyscientist should I make that part of this pull, or do we want to have a separate pull where we can discuss flattening HISQ and ASQTAD support as well?

As far as I know there's nothing you need for ASQTAD links that you don't also need for HISQ links (HISQ is ASQTAD fattening -> unitarization -> ASQTAD fattening+long links, optionally plus non-relativistic correction.

weinbe2:
Also @maddyscientist, when you have the chance can you do a quick run of `multigrid_evolve_test`? I doubt I broke anything there, especially because nothing broke during @cpviolator 's testing, but may as well be thorough.

mathiaswagner:
@weinbe2 yes, you can make that a part of your PR. Might be worth to grep for `define` in the code and reduce it to only the HISQ one. If not exposed separately from HISQ in the build system there is no need for separate defines.

maddyscientist:
One alternative is that I put this into my open pull into 1.0.0. Might be save a few more folks from falling into the trap of enabling staggered but not hisq link.

mathiaswagner:
Good idea. I like it.

include/dirac_quda.h
```diff
@@ -365,6 +377,13 @@ namespace quda {
     /**
      * @brief Create the coarse clover operator
      *
+     * @details Takes the multigrid transfer class, which knows
+     *          about the coarse grid blocking, as well as
+     *          having prolongate and restrict member functions,
+     *          and returns color matrices Y[0..2*dim-1] corresponding
+     *          to the coarse grid hopping terms and X corresponding to
+     *          the coarse grid "clover" term.
```
maddyscientist: Text book doxygen 👍

PR #903, opened 2019-09-16
Add support for overlapping Schwarz (MSPCG) and EOFA for Mobius DWF

hummingtree:
This PR adds two features for Möbius DWF:
### 1. Overlapping schwarz (MSPCG) [arXiv:1811.08488]

See the [wiki page](https://github.com/lattice/quda/wiki/The-Multi-Splitting-Preconditioned-Conjugate-Gradient-(MSPCG),-an-application-of-the-additive-Schwarz-Method).

- [x] ~... and a later developed method of preconditioning with a DWF preconditioner with a smaller `Ls`, together with a trained interpolating/restricting matrix to make it working.~
    1. Specifically, a local MdagM operator is added for (constant, real) Möbius domain wall fermion.
    2. The operator uses tensor cores on Volta and Turing with the `wmma` routines.
    3. For Volta, the PTX instruction `mma.sync.m8n8k4` is used when possible (CUDA >= 10.1).

The local `MdagM` operator for (contant, real) MDWF can be tested with
```
tests/dslash_test \
  --matpc even-even \
  --dim 12 12 12 16 \
  --Lsdim 16 \
  --gridsize 1 1 1 1 \
  --dslash-type mobius \
  --b5 2.5 --c5 1.5 \
  --mass 0.01 \
  --prec half \
  --recon 12 \
  --test MatPCDagMatPCLocal
```

The MSPCG inverter can be tested with

```
tests/invert_test \
  --matpc even-even \
  --dim 12 12 12 16 \
  --Lsdim 12 \
  --gridsize 1 1 1 1 \
  --dslash-type mobius \
  --b5 2.5 --c5 1.5 \
  --inv-type pcg \
  --precon-type cg \
  --precon-schwarz-type additive \
  --tol 1e-10 \
  --tol-precondition 1e-6 \
  --niter 860 \
  --max-precondition 6 \
  --mass 0.01 \
  --prec double \
  --prec-sloppy half \
  --prec-precondition half 
```

As a reference for the interface, the solve is invoked when we solve using either PCG or GCR as the outer solver (with PCG being superior), using CG or MR as the inner solver (with CG being superior), as well as explicitly requesting additive schwarz: 
```
inv_param.inv_type = QUDA_PCG_INVERTER;
inv_param.precon_type = QUDA_CG_INVERTER;
inv_param.schwarz_type = QUDA_ADDITIVE_SCHWARZ;
```
 
Currently only half or quarter precisions and the symmetric even-odd preconditioning is supported
```
gauge_param.cuda_prec_precondition = QUDA_HALF_PRECISION;
inv_param.cuda_prec_precondition = QUDA_HALF_PRECISION;
inv_param.matpc_type = QUDA_MATPC_EVEN_EVEN or QUDA_MATPC_ODD_ODD;
```
One needs to use a normal operator solve in order to use preconditioned CG:
```
inv_param.solution_type = QUDA_MATPCDAG_MATPC_SOLUTION;
inv_param.solve_type = QUDA_NORMOP_PC_SOLVE;
```

### 2. Exact one flavor algorithm (EOFA) [arXiv:1706.05843]

See the [wiki page](https://github.com/lattice/quda/wiki/The-Exact-One-Flavor-Algorithm-(EOFA)).

It can be tested with

```
tests/dslash_test \
  --matpc even-even \
  --dim 12 12 12 16 \
  --Lsdim 12 \
  --gridsize 1 1 1 1 \
  --dslash-type mobius-eofa \
  --b5 2.5 --c5 1.5 \
  --tol 1e-10 \
  --prec double \
  --test MatPCDagMatPC(for example)

```

The (known) remaining parts that have not been finished:
- [x] CPU test code for the local MDWF including the snake terms.
- [x] Integration of the MSPCG solver into the general solver framework.
- [x] ~Fix the preconditioned CG in #960.~
- [x] Separate the mixed MSPCG solver into operators and the preconditioned CG solver.


include/quda.h
```diff
@@ -107,8 +107,21 @@ extern "C" {
     double m5;    /**< Domain wall height */
     int Ls;       /**< Extent of the 5th dimension (for domain wall) */
 
-    double_complex b_5[QUDA_MAX_DWF_LS]; /**< Mobius coefficients - only real part used if regular Mobius */
-    double_complex c_5[QUDA_MAX_DWF_LS]; /**< Mobius coefficients - only real part used if regular Mobius */
+    double _Complex b_5[QUDA_MAX_DWF_LS]; /**< Mobius coefficients - only real part used if regular Mobius */
```
maddyscientist: This should be reverted to `double_complex` - this is a work around since nvrtc is a pure C++ compiler and cannot use C99 complex variables.

host_reference/domain_wall_dslash_reference.cpp
```diff
@@ -975,6 +1235,216 @@ void mdw_matpc(void *out, void **gauge, void *in, double _Complex *kappa_b, doub
...
+  }
+  cpuGaugeField *padded_cpu = new cpuGaugeField(gauge_field_param);
+
+  copyExtendedGauge(*padded_cpu, cpu, QUDA_CPU_FIELD_LOCATION);
+  padded_cpu->exchangeExtendedGhost(R, true); // Do comm to fill halo = true
+
+  return padded_cpu;
+}
+
+inline int index_4d_cb_from_coordinate_4d(const int coordinate[4], const int dim[4])
```
weinbe2: I won't hold up the PR on this b/c these host routines are a mess (in general, not yours specifically), but these routines are likely redundant of the routines around here: https://github.com/lattice/quda/blob/develop/utils/host_utils.h#L128

host_reference/domain_wall_dslash_reference.cpp
```diff
@@ -975,6 +1235,216 @@ void mdw_matpc(void *out, void **gauge, void *in, double _Complex *kappa_b, doub
...
+    }
+  } else {
+    errorQuda("Unsupported matpc_type=%d dagger=%d", matpc_type, dagger);
+  }
+
+  free(tmp);
+}
+
+// helper for creating extended gauge fields
+static cpuGaugeField *create_extended_gauge(void **gauge, QudaGaugeParam &gauge_param, const int *R)
```
weinbe2: Won't hold the PR up on this, but we should put this routine in a file in `utils/`. @hummingtree if you could file a bug (well, feature request) for this I'd be grateful.

include/dirac_quda.h
```diff
@@ -742,6 +765,59 @@ namespace quda {
...
+    virtual void M(ColorSpinorField &out, const ColorSpinorField &in) const;
+    virtual void MdagM(ColorSpinorField &out, const ColorSpinorField &in) const;
+
+    virtual void prepare(ColorSpinorField *&src, ColorSpinorField *&sol, ColorSpinorField &x, ColorSpinorField &b,
+                         const QudaSolutionType) const;
+    virtual void reconstruct(ColorSpinorField &x, const ColorSpinorField &b, const QudaSolutionType) const;
+  };
+
+  // 4d Even-odd preconditioned Mobius domain wall with EOFA
+  class DiracMobiusPCEofa : public DiracMobiusEofa
```
weinbe2: Naming convention: this should be `DiracMobiusEofaPC`, unless there's a specific reason we're interchanging the order (which we should note in the comment)

include/dirac_quda.h
```diff
@@ -1568,8 +1644,43 @@ namespace quda {
...
+    DiracMdagMLocal(const Dirac &d) : DiracMatrix(d) {}
+    DiracMdagMLocal(const Dirac *d) : DiracMatrix(d) {}
+
+    void operator()(ColorSpinorField &out, const ColorSpinorField &in) const { dirac->MdagMLocal(out, in); }
+
+    void operator()(ColorSpinorField &out, const ColorSpinorField &in, ColorSpinorField &tmp) const
+    {
+      dirac->tmp1 = &tmp;
+      dirac->MdagMLocal(out, in);
+      if (shift != 0.0) blas::axpy(shift, const_cast<ColorSpinorField &>(in), out);
```
weinbe2, line 1661: Any reason there isn't a fused `MdagMLocal`+`axpy`?

include/dslash_quda.h
```diff
@@ -475,11 +475,29 @@ namespace quda {
...
+    DSLASH5_MOBIUS_PRE,
+    DSLASH5_MOBIUS,
+    M5_INV_DWF,
+    M5_INV_MOBIUS,
+    M5_INV_ZMOBIUS,
+    M5_EOFA,
+    M5INV_EOFA
+  };
+
+  enum MdwfFusedDslashType { // the details are too complicated to describe here.
```
weinbe2: This comment needs to be improved. This can be done by expanding the details despite it being long, or by linking to a wiki page. The wiki page can be a placeholder for now.

include/enum_quda.h
```diff
@@ -309,6 +310,8 @@ extern "C" {
     QUDA_DOMAIN_WALL_4DPC_DIRAC,
     QUDA_MOBIUS_DOMAIN_WALL_DIRAC,
     QUDA_MOBIUS_DOMAIN_WALLPC_DIRAC,
+    QUDA_MOBIUS_DOMAIN_WALL_EOFA_DIRAC,
```
weinbe2, line 314: The FORTRAN enum header needs to be updated as well.

include/kernels/dslash_mobius_eofa.cuh
```diff
@@ -0,0 +1,271 @@
...
+      const int volume_4d_cb; // 4-d checkerboarded volume
+      const int_fastdiv Ls;   // length of 5th dimension
+
+      const real m_f; // fermion mass parameter
+      const real m_5; // Wilson mass shift
+
+      const bool dagger; // dagger
+      const bool xpay;   // whether we are doing xpay or not
+
+      real b = 0.; // real constant Mobius coefficient
```
weinbe2: These should be initialized in the constructor, not here.

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -0,0 +1,475 @@
...
+#include <inline_ptx.h>
+#include <math_helper.cuh>
+#include <shared_memory_cache_helper.cuh>
+
+#if (__CUDACC_VER_MAJOR__ >= 9 && __COMPUTE_CAPABILITY__ >= 700)
+#include <mma.h>
+
+// The `mma.sync` PTX is only available with or after CUDA 10.1
+#if ((__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 1) || (__CUDACC_VER_MAJOR__ > 10))                         \
+  && (__COMPUTE_CAPABILITY__ < 750)
```
weinbe2: Replace this with `== 700` (I believe) as opposed to `< 750`. Alternatively, if this also works well on A100, add a `__COMPUTE_CAPABILITY == 800` on top of the original check.

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -0,0 +1,475 @@
...
+    // if we rescale, then the actual matrix is alpha*m5inv+beta.
+    // Otherwise a = 1., b = 0.;
+    const float b = arg.beta;
+
+    int offset_k = threadIdx.y * 4;
+    int x = threadIdx.x;
+
+    while (x < Ls) {
+      int offset_m = x * 2;
+      int exp = x - threadIdx.y;
```
weinbe2: Rename `exp`, that's a standard math function name.

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -0,0 +1,475 @@
...
+      if (mp && mm) {
+        if (dagger) {
+          factorR = mp[x * Ls + threadIdx.y];
+          factorL = mm[x * Ls + threadIdx.y];
+        } else {
+          factorR = mp[threadIdx.y * Ls + x];
+          factorL = mm[threadIdx.y * Ls + x];
+        }
+      } else {
+        int exp;
```
weinbe2: Rename `exp`, it's a math function name.

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -0,0 +1,475 @@
...
+  __device__ inline void wmma_gemm(T *a_frag, half *sm_a, half *sm_b, half *sm_c)
+  {
+    constexpr int WMMA_M = 16;
+    constexpr int WMMA_N = 16;
+    constexpr int WMMA_K = 16;
+
+    constexpr int tm_dim = M / WMMA_M;
+    constexpr int tn_dim = N / WMMA_N;
+
+    constexpr int total_warp = BlockDimX * Ls / 32;
```
weinbe2: Probably an overkill comment, but we may want to replace this hard coded `32` with some sort of `warp_size` variable.

lib/dirac_mobius.cpp
```diff
@@ -17,46 +24,27 @@ namespace quda {
     }
 
     if (getVerbosity() > QUDA_VERBOSE) {
-      if (zMobius)
+      if (zMobius) {
         printfQuda("%s: Detected variable or complex cofficients: using zMobius\n", __func__);
-      else
+        errorQuda("zMobius has NOT been fully tested in QUDA.\n");
```
weinbe2: An `errorQuda` should not be included in a check against `QUDA_VERBOSE`, it should trigger regardless.

lib/dirac_mobius.cpp
```diff
@@ -9,6 +9,13 @@ namespace quda {
     memcpy(b_5, param.b_5, sizeof(Complex) * param.Ls);
     memcpy(c_5, param.c_5, sizeof(Complex) * param.Ls);
 
+    double b = b_5[0].real();
+    double c = c_5[0].real();
+    mobius_kappa_b = 0.5 / (b * (m5 + 4.) + 1.);
```
weinbe2, line 14: Let's add a check for variable coefficients and add an `errorQuda` for the case that there are variable coefficients. I'll accept just comparing the first two coefficients. I assume that would also require a check for kappa vs mass normalization?

lib/dirac_mobius.cpp
```diff
@@ -81,26 +69,26 @@ namespace quda {
...
+                                const ColorSpinorField &x, const double &k) const
   {
     checkDWF(in, out);
     checkParitySpinor(in, out);
     checkSpinorAlias(in, out);
 
     ApplyDomainWall4D(out, in, *gauge, k, m5, b_5, c_5, x, parity, dagger, commDim, profile);
 
-    flops += (1320LL+48LL)*(long long)in.Volume();
+    flops += (1320LL + 48LL) * (long long)in.Volume();
```
weinbe2: Why is this an explicit sum? I understand if it's useful to distinguish the two contributions, but a quick comment on it would be instructive.

lib/dirac_mobius.cpp
```diff
@@ -125,17 +113,17 @@ namespace quda {
...
+    flops += (96LL) * (long long)in.Volume() + 96LL * bulk + 120LL * wall;
   }
 
   void DiracMobius::M(ColorSpinorField &out, const ColorSpinorField &in) const
   {
     checkFullSpinor(out, in);
 
     // FIXME broken for variable coefficients
-    double kappa_b = 0.5 / (b_5[0].real() * (4.0 + m5) + 1.0);
+    double mobius_kappa_b = 0.5 / (b_5[0].real() * (4.0 + m5) + 1.0);
```
weinbe2, line 128: Re:my request for checking for variable coefficients in the constructor, add a reference to that check here.

lib/dirac_mobius.cpp
```diff
@@ -186,17 +174,26 @@ namespace quda {
...
+    extended_gauge = createExtendedGauge(*gauge, R, profile, true);
+  }
 
-  DiracMobiusPC::DiracMobiusPC(const DiracParam &param) : DiracMobius(param) {  }
-
-  DiracMobiusPC::DiracMobiusPC(const DiracMobiusPC &dirac) : DiracMobius(dirac) {  }
+  DiracMobiusPC::DiracMobiusPC(const DiracMobiusPC &dirac) : DiracMobius(dirac)
+  {
+    const int R[] = {2, 2, 2, 2};
+    extended_gauge = createExtendedGauge(*gauge, R, profile, true);
```
weinbe2: Do we always need to create an extended gauge? Can we check for this on the first invocation of `MdagMlocal` and only create it then? We also need some way to check if the resident gauge has updated, though that can wait for a PR I need to work on for HISQ MG.

lib/dirac_mobius.cpp
```diff
@@ -372,4 +363,438 @@ namespace quda {
...
+    if (zMobius) { errorQuda("DiracMobiusPC::MdagMLocal doesn't currently support zMobius.\n"); }
+
+    checkDWF(in, out);
+    // checkParitySpinor(in, out);
+    checkSpinorAlias(in, out);
+
+    ColorSpinorParam csParam(out);
+    csParam.create = QUDA_NULL_FIELD_CREATE;
+
+    ColorSpinorField *unextended_tmp1 = ColorSpinorField::Create(csParam);
```
weinbe2, line 391: We should create these fields once and persist it thereafter. Technically the pool allocator should keep this from being a problem, but it's still a better practice.

lib/dslash5_domain_wall.cu
```diff
@@ -11,7 +10,6 @@ namespace quda
 
   /*
     FIXME
-    - fix flops counters
     - check dagger operators are correct - there might need to be a
```
weinbe2: Has this been checked?

lib/extended_color_spinor_utilities.cu
```diff
@@ -486,6 +486,16 @@ namespace quda {
...
         errorQuda("Unsupported Precision %d", src.Precision());
       }
+    } else if (dst.Precision() == QUDA_QUARTER_PRECISION){
+      if(src.Precision() == QUDA_DOUBLE_PRECISION){
+        CopyExtendedColorSpinor(dst, src, parity, location, static_cast<char*>(Dst), static_cast<double*>(Src), static_cast<float*>(dstNorm), 0);
+      }else if(src.Precision() == QUDA_SINGLE_PRECISION){
+        CopyExtendedColorSpinor(dst, src, parity, location, static_cast<char*>(Dst), static_cast<float*>(Src), static_cast<float*>(dstNorm), 0);
+      }else if(src.Precision() == QUDA_HALF_PRECISION){
+        CopyExtendedColorSpinor(dst, src, parity, location, static_cast<char*>(Dst), static_cast<short*>(Src), static_cast<float*>(dstNorm), static_cast<float*>(srcNorm));
+      }else{
```
weinbe2, line 496: Is there a reason copying quarter -> quarter results in an `errorQuda`?

lib/dslash5_domain_wall.cu
maddyscientist: For this file, it is correct.

lib/inv_cg_quda.cpp
```diff
@@ -220,6 +220,8 @@ namespace quda {
 
   void CG::operator()(ColorSpinorField &x, ColorSpinorField &b, ColorSpinorField *p_init, double r2_old_init)
   {
+    if (param.is_preconditioner && param.global_reduction == false) commGlobalReductionSet(false);
```
weinbe2, line 223: For true robustness, should this be a push/pop type operation for nested CG inversions?

lib/inv_gcr_quda.cpp
```diff
@@ -38,6 +38,7 @@ namespace quda {
 
     inner.inv_type_precondition = QUDA_INVALID_INVERTER;
     inner.is_preconditioner = true; // tell inner solver it is a preconditioner
+    inner.pipeline = true;
```
weinbe2, line 41: Why was this change made?

lib/inv_pcg_quda.cpp
```diff
@@ -134,13 +155,14 @@ namespace quda
 
     if (K) {
       csParam.create = QUDA_COPY_FIELD_CREATE;
-      csParam.setPrecision(param.precision_precondition);
+      csParam.setPrecision(Kparam.precision);
+#ifdef FLOAT8
+      csParam.fieldOrder = QUDA_FLOAT8_FIELD_ORDER;
```
weinbe2, line 162: Is this safe for all possible preconditioners?

lib/mdw_fused_dslash.cu
```diff
@@ -0,0 +1,757 @@
...
+        coordinate[d]++;
+      } // nDim
+    }
+
+    /**
+    -> Everything should be understood in a 4d checkboarding sense.
+    */
+    template <class T>
+    __device__ inline void coordinate_from_shrinked_index(int coordinate[4], int shrinked_index, const T shrinked_dim[4],
+                                                          const int shift[4], int parity) // s is the 5d stuff,
```
weinbe2: The `// s is the 5d stuff` should go into the Doxygen iirc

lib/mdw_fused_dslash.cu
```diff
@@ -0,0 +1,757 @@
...
+            in.VolumeCB() > out.VolumeCB() ? in.X(3) : out.X(3)},
+        shrinked_dim{dim[0] - 2 * shift[0], dim[1] - 2 * shift[1], dim[2] - 2 * shift[2], dim[3] - 2 * shift[3]},
+        volume_4d_cb_shift(shrinked_dim[0] * shrinked_dim[1] * shrinked_dim[2] * shrinked_dim[3] / 2),
+        type(type_)
+      {
+        if (in.Nspin() != 4) { errorQuda("nSpin = %d NOT supported.\n", in.Nspin()); }
+
+        if (nParity == 2) { errorQuda("nParity = 2 NOT supported, yet.\n"); }
+
+        // if (!in.isNative() || !out.isNative())
```
weinbe2: Remove this comment

lib/mdw_fused_dslash.cu
```diff
@@ -0,0 +1,757 @@
...
+        //   errorQuda("Unsupported field order out=%d in=%d\n", out.FieldOrder(), in.FieldOrder());
+
+        b = b_5[0].real();
```
weinbe2, line 130: Add a check to make sure these are real, constant (again, I'm content with just comparing the first and second elements)

lib/mdw_fused_dslash.cu
```diff
@@ -0,0 +1,757 @@
...
+      float *smem_scale = shared_memory_data;
+
+      half2 *sm_b = reinterpret_cast<half2 *>(smem_scale + 32);
+      half *sm_c = reinterpret_cast<half *>(sm_b);
+
+      half *sm_a = reload ? sm_c + M * N_sm : sm_c;
+      // This is for type == 1 ONLY.
+      half *sm_a_black = sm_a + M * M_sm;
+
+      if (type_ == 0) {
```
weinbe2: Are the different `type_` described anywhere?

lib/mdw_fused_dslash.cu
```diff
@@ -0,0 +1,757 @@
...
+      int gridStep() const { return deviceProp.multiProcessorCount; }
+      unsigned int maxGridSize() const { return (arg.volume_4d_cb_shift + blockMin() - 1) / blockMin(); }
+      unsigned int minGridSize() const { return deviceProp.multiProcessorCount; }
+
+      unsigned int sharedBytesPerBlock(const TuneParam &param) const
+      {
+        const int a_size = (param.block.y * 4) * (param.block.y * 4 + sm_m_pad_size(arg.Ls));
+        const int b_size = (param.block.y * 4) * (param.block.x * 6 + sm_n_pad_size());
+        // (Ls*4) by (Ls*4), (Ls*4) by (volume_4d*6 + 16)
+        if (param.aux.x == 1) { // aux.x == 1 --> reload == true
```
weinbe2, line 561: Can the different `aux` params be documented concisely somewhere?

weinbe2: Nevermind, saw it below

lib/mdw_fused_dslash.cu
```diff
@@ -0,0 +1,757 @@
...
+    // out = Dslash5 * in
+    template <typename storage_type, int nColor, QudaReconstructType recon> struct FusedApply {
+
+      inline FusedApply(ColorSpinorField &out, const ColorSpinorField &in, const GaugeField &U, ColorSpinorField &y,
+                        const ColorSpinorField &x, double m_f, double m_5, const Complex *b_5, const Complex *c_5,
+                        bool dagger, int parity, int shift[4], int halo_shift[4], MdwfFusedDslashType type)
+      {
+        // switch for Ls
+        switch (in.X(4)) {
+        case 4: {
```
weinbe2, line 708: I imagine the factors of 4 are due to the tensor core sizes, can a quick comment on this get added?

utils/command_line_params.cpp
```diff
@@ -785,6 +808,18 @@ void add_multigrid_option_group(std::shared_ptr<QUDAApp> quda_app)
                          "The verbosity to use on each level of the multigrid (default summarize)");
 }
 
+void add_eofa_option_group(std::shared_ptr<QUDAApp> quda_app)
+{
+  auto opgroup = quda_app->add_option_group("EOFA", "Options controlling EOFA parameteres");
+
+  opgroup->add_option("--eofa-pm", eofa_pm,
+                      "Set to evalute plus (non-zero values) or minus (zero) EOFA operator (default 1)");
```
weinbe2: Can this be updated to verbatim request the string `plus` or `minus`?

lib/dirac_mobius.cpp
hummingtree: The check for variable coefficients is done later in this constructor. I don't think the `kappa` and/or mass normalization is used here?

lib/dirac_mobius.cpp
hummingtree: So something like
```
static ColorSpinorField *unextended_tmp1 = ColorSpinorField::Create(csParam);
```
?
I am little bit uncomfortable about having these pointer staying there util the end of the app.

maddyscientist: This was benchmarked, and found not to be a problem.  

In general, beyond the pool allocation, allocating fields has an overhead when a field has communication.  By definition these fields have no communication, and the overhead of this reallocation has been tested to be zero.  There are persistent memory pointers in the dirac objects already, e.g., `tmp1` and `tmp2`, but there use needs to be cleaned up and that design is far from perfect.  Hence why this present design is sufficient for now, until we have an improved pool allocator for temporary fields like this.

lib/inv_cg_quda.cpp
maddyscientist: Yes.  The design for that goes beyond the scope of this pull.  It would be covered by #986.

lib/extended_color_spinor_utilities.cu
hummingtree: This part of the code is not used by this PR and is commented out entirely. It shows diff because of historical reasons.

@maddyscientist does it make sense to simply remove this code?

lib/inv_cg_quda.cpp
hummingtree: @maddyscientist

lib/inv_gcr_quda.cpp
hummingtree: @maddyscientist

maddyscientist: To accelerate the preconditioner.  Robustness is not a problem for a preconditioner since it will only run for a handful of iterations.

lib/inv_pcg_quda.cpp
maddyscientist: No and that's intentional.  As noted above, `FLOAT8` support is a known wart since it is a non-native type which will require some thought if and how to deploy en masse in QUDA.  There's no way to determine if the preconditioner is safe to use with `FLOAT8` mode, but I will add a warning at this point to let the user know they are using an experimental feature.

hummingtree: @maddyscientist

lib/dirac_mobius.cpp
hummingtree: @weinbe2 So I get around this with checking at the beginning of `MdagMLocal` and allocate `extended_gauge_field` if it doesn't exist yet. The problem with that is `MdagMLocal` is a `const` method and we are changing `extended_gauge_field` in its implementation. I have put `extended_gauge_field` `mutable` but let me know if you think that's too dirty a solution.

include/dirac_quda.h
hummingtree: @weinbe2 Do you mean a fused kernel for `MdagMLocal` + `axpy`? The `if (shift != 0.0) blas::axpy(...)` pattern is used in the other methods (`MdagM`/`M`) as well: I am just replicating here. In principle a fused kernel is certainly better but I would push that to the following boundary-condition-cleaning PR that's on my list.

maddyscientist: This is a non issue.  I introduced this as an initial hack for some variant of Hasenbusch preconditioning, for plugging into BQCD IIRC.  No need for a fused kernel for this.

host_reference/domain_wall_dslash_reference.cpp
hummingtree: I actually didn't find the one I wanted in the `utils/host_utils.h` ...

lib/extended_color_spinor_utilities.cu
maddyscientist: A proper treatment of extended color-spinor fields is on my TODO list.  If this file is taking up actual compilation time and/or is a non-trivial binary size, then I would suggest commenting out these instantiations and just leaving an error message regardless of precision.

host_reference/domain_wall_dslash_reference.cpp
weinbe2: Can you move it there?

include/dslash_quda.h
```diff
@@ -475,11 +475,35 @@ namespace quda {
...
      @param[in] profile The TimeProfile used for profiling the dslash
   */
+
   void ApplyDomainWall4D(ColorSpinorField &out, const ColorSpinorField &in, const GaugeField &U, double a, double m_5,
                          const Complex *b_5, const Complex *c_5, const ColorSpinorField &x, int parity, bool dagger,
                          const int *comm_override, TimeProfile &profile);
 
-  enum Dslash5Type { DSLASH5_DWF, DSLASH5_MOBIUS_PRE, DSLASH5_MOBIUS, M5_INV_DWF, M5_INV_MOBIUS, M5_INV_ZMOBIUS };
+  enum Dslash5Type {
+    DSLASH5_DWF,
```
weinbe2, line 484: Since these are internal to QUDA, they should be replaced with an `enum class` I believe. @mathiaswagner , what's your take?

include/dslash_quda.h
```diff
@@ -475,11 +475,35 @@ namespace quda {
...
+  };
+
+  enum MdwfFusedDslashType {
+    /**
+      Applying the following five kernels in the order of 4-0-1-2-3 is equivalent to applying
+      the full even-odd preconditioned symmetric MdagM operator:
+      op = (1 - M5inv * D4 * D5pre * M5inv * D4 * D5pre)^dag
+            * (1 - M5inv * D4 * D5pre * M5inv * D4 * D5pre)
+    */
+    dslash4_dslash5pre_dslash5inv = 0,
```
weinbe2: QUDA convention, `enum` (or `enum class`, see above) should be all capitals. Comment should go outside the `enum`.

include/enum_quda_fortran.h
```diff
@@ -280,6 +281,8 @@
 #define QUDA_DOMAIN_WALL_4DPC_DIRAC 7
 #define QUDA_MOBIUS_DOMAIN_WALL_DIRAC 8
 #define QUDA_MOBIUS_DOMAIN_WALLPC_DIRAC 9
+#define QUDA_MOBIUS_DOMAIN_WALL_EOFA_DIRAC 23
```
weinbe2: This is a PITA `enum` thing, but based on the ordering in `enum_quda.h`, this should be 10, `QUDA_MOBIUS_DOMAIN_WALLPC_EOFA_DIRAC` becomes 11, `QUDA_STAGGERED_DIRAC` becomes 12...

This is another motivator to get this file auto-generated from `enum_quda.h`.

include/enum_quda_fortran.h
```diff
@@ -92,6 +92,7 @@
 #define QUDA_DOMAIN_WALL_DSLASH 2
 #define QUDA_DOMAIN_WALL_4D_DSLASH 3
 #define QUDA_MOBIUS_DWF_DSLASH 4
+#define QUDA_MOBIUS_DWF_EOFA_DSLASH 6
```
weinbe2: See comment on `enum` below

include/mma_tensor_op/mma_m16n16k16_sm70.cuh
```diff
@@ -0,0 +1,216 @@
...
+      thread_offset_c = idx_strided * stride + (idx_contiguous + 2);
+      C[thread_offset_c] = __floats2half2_rn(reg[4], reg[5]);
+
+      thread_offset_c = (idx_strided + 2) * stride + (idx_contiguous + 2);
+      C[thread_offset_c] = __floats2half2_rn(reg[6], reg[7]);
+    }
+  };
+
+  template <int BlockDimX, int Ls, int M, int N, int M_PAD, int N_PAD, bool reload, class T>
+  __device__ inline void mma_sync_gemm(T op_a[], half *sm_a, half *sm_b, half *sm_c, const WarpRegisterMapping &wrm)
```
weinbe2, line 121: Do we need a `__CUDA_ARCH__` check within this routine? i.e., will it barf when compiled against `sm60`?

lib/dirac_mobius.cpp
weinbe2: @maddyscientist for insight

lib/dirac_mobius.cpp
weinbe2: If it's benchmarked sounds good. I'd like a comment added on that, and a `FIXME`

include/mma_tensor_op/mma_m16n16k16_sm70.cuh
maddyscientist: Nope.  This is fine since the file is included conditionally on `__CUDA_ARCH__`, Jenkins compilation tests pass, and I've tested this by hand already.  All is fine.

include/enum_quda_fortran.h
hummingtree: I have zero knowledge of fortran ...

hummingtree: @weinbe2 Let me know if this has been fixed now.

host_reference/domain_wall_dslash_reference.cpp
hummingtree: I will add this to `utils/`.

include/gauge_field.h
```diff
@@ -750,6 +751,28 @@ namespace quda {
...
+                                      bool redundant_comms = false, QudaReconstructType recon = QUDA_RECONSTRUCT_INVALID);
+
+  /**
+     This function is used for creating an exteneded (cpu) gauge field from the input,
+     and copying the gauge field into the extended gauge field.  Defined in lib/gauge_field.cpp.
+     @param in The input field from which we are extending
+     @param R By how many do we want to extend the gauge field in each direction
+     @return the pointer to the extended gauge field
+  */
+  cpuGaugeField *create_extended_gauge(void **gauge, QudaGaugeParam &gauge_param, const int *R);
```
weinbe2: Why is there a different naming convention for this routine?

maddyscientist:
Confirming this branch is now good to merge.  Whenever the build tests complete merge can be done.

tests/invert_test.cpp
```diff
@@ -104,6 +104,8 @@ int main(int argc, char **argv)
   auto app = make_app();
   add_eigen_option_group(app);
   add_deflation_option_group(app);
+  // add_multigrid_option_group(app);
```
mathiaswagner: guess this can go away

mathiaswagner:
Just asking as the commit history is so long I might have overlooked this: Did this get the clang-format treatment?

hummingtree:
> Just asking as the commit history is so long I might have overlooked this: Did this get the clang-format treatment?

I just did a clang-format-diff (and I hope I did it correctly). The majority of the code should be in good shape for clang-format anyway.

PR #904, opened 2019-09-19
Add experimental NVSHMEM support

mathiaswagner:
This PR adds initial nvshmem support to QUDA. It requires NVSHMEM which can be requested at https://developer.nvidia.com/nvshmem.

It is a rewrite of exploratory work on top of #899 and has not been fully tested (correctness/performance) and hence is considered to be experimental. 

A future PR will change that status.


mathiaswagner:
This has been outdated and pushed out while nvshmem is in EA. I'll reopen a PR again soon.

PR #905, opened 2019-09-20
Hotfix/misc

maddyscientist:
Some quick fixes
* Add covdef_test to ctest list
* Fix compiler warnings on gcc-8
* Small cleanup of experimental fast reduce code


mathiaswagner:
Looks good. Why does the new reduce code go into release? If this is not fixing a bug it should technically go to develop.

maddyscientist:
The reduce code was already in release, this was just a clean up and making it more robust.

PR #906, opened 2019-09-24
Feature/thinning

maddyscientist:
Pull request to reduce some of the instantiated template bloat
* Add more support for `QUDA_PRECISION` cmake parameter, e.g., don't build a given precision unless enabled
* Remove Wuppertal smearing kernel - this is functionally equivalent to Laplace kernel with additional scale factor - extend Laplace kernel to cover both cases
* Some clean up of max_gauge and max_clover files.

For example, before this pull request, if we do a pure Wilson single-precision build of QUDA with only reconstruct-18 enabled, the library is reduced by around ~25%.


weinbe2:
I'm running this through `ctest -V` with a fun precision combination, `10`, because it might spring some interesting issues because it skips an intermediate precision. Assuming that goes through I'll give it the approval.

PR #909, opened 2019-09-26
Further thinning and fixes

maddyscientist:
This pull request focuses of further reductions in build time through avoiding instantiation of unnecessary templates
* When building non-degenerate twisted-mass fermions, do not build singlet or regular Wilson fermions
* Added `NSPIN1`, `NSPIN2`, `NSPIN4` macros for cleaner control of which templates are instantiated
* More kernels respect both these `NSPIN` macros, as well as the `QUDA_PRECISION` and `QUDA_RECONSTRUCT` macros
* Cleanup of gauge_plaq and gauge_qcharge using variadic templates to instantiate templates
* Fixed bug with extended fields using recon-9 (wrong template was being instantiated)
* Removed legacy copy_quda kernels since their functionality is covered by the generic copy colorspinor kernels
* Adds new advanced CMake option `QUDA_REDUCE_SINGLE_WARP` for only compiling one warp per CTA for the reduction kernels for faster builds

With these changes in place, build time nvsocal2 for building a single-precision recon-18 with only non-degenerate twisted mass fermions is reduced by 25%


maddyscientist:
I've just pushed latest set of fixes and improvements.  This should be done now.  This also includes the outstanding changes in the release/1.0.x branch not already merged into develop, notably the staggered boundary condition fix.

include/kernels/dslash_staggered.cuh
```diff
@@ -41,8 +41,11 @@ namespace quda
     const GL L; /** the long gauge field */
 
     const real a; /** xpay scale factor */
-    const real tboundary;
+    const real tboundary; /** temporal boundary condition */
+    const bool is_first_time_slice; /** are we on the first (global) time slice */
```
weinbe2: @maddyscientist is there a reason you've added these variables if they're never used? (in contrast to in `index_helper.cuh`, where you did make use of them)

include/quda_internal.h
```diff
@@ -18,6 +18,21 @@
...
+  || defined(GPU_TWISTED_MASS_DIRAC) || defined(GPU_TWISTED_CLOVER_DIRAC) || defined(GPU_NDEG_TWISTED_MASS_DIRAC)      \
+  || defined(GPU_CLOVER_HASENBUSCH_TWIST) || defined(GPU_COVDEV)
+#define NSPIN4
+#endif
+
+#if defined(GPU_MULTIGRID)
+#define NSPIN2
+#endif
+
+#if defined(GPU_STAGGERED_DIRAC)
```
weinbe2: Should there be a define related to the laplace operator too? In the chance someone is linking against QUDA to use it, you'd need `NSPIN1` blas routines...

lib/blas_quda.cu
```diff
@@ -234,23 +234,23 @@ namespace quda {
...
+#if defined(NSPIN4)
             const int M = 6;
             nativeBlas<float4, short4, short4, M, Functor, writeX, writeY, writeZ, writeW, writeV>(
                 a, b, c, x, y, z, w, v, x.Volume());
 #else
             errorQuda("blas has not been built for Nspin=%d fields", x.Nspin());
 #endif
           } else if (x.Nspin() == 4 && x.FieldOrder() == QUDA_FLOAT2_FIELD_ORDER) { // wilson
-#if defined(GPU_MULTIGRID)
+#if defined(GPU_MULTIGRID) // FIXME eventually we should get rid of this and use float4 ordering
```
weinbe2: Can this be `#if defined(NSPIN2)` instead? Or is `GPU_MULTIGRID` a stronger requirement here?

weinbe2: Nevermind---didn't pay close enough attention to the context.

maddyscientist: This is intentional (it's not `NSPIN2`), it's the funky float2 ordering that MG fields use in single precision.  And I want to remove these eventually and have MG use float4 always for `nSpin=4`.

include/quda_internal.h
maddyscientist: We have both `nSpin=1` and `nSpin=4` Laplace operators, so these are conditionally compiled depending on whether staggered and Wilson are enabled, respectively.

include/kernels/dslash_staggered.cuh
maddyscientist: Not sure I follow.  These variables are used in index_helpher.cuh, and this is where they are defined.

weinbe2: Oh, yep, mental disconnect. In my mind they were also defined in `include/index_helper.cuh`.

lib/laplace.cu
```diff
@@ -158,6 +157,7 @@ namespace quda
 #endif
       } else if (in.Nspin() == 4) {
 #ifdef GPU_WILSON_DIRAC
```
weinbe2: Can this be replaced with `NSPIN4`? (Likewise for staggered and `NSPIN1` above)

maddyscientist: It was intentional I didn't do this, and just have Laplace enabled by Wilson / staggered, as opposed to enabling it any generic nspin-4 or nspin-1 operator is enabled.  Alternative is to make a Cmake variable for this specifically.  Not sure I have strong feelings either way.

tests/CMakeLists.txt
```diff
@@ -223,6 +223,7 @@ set(QUDA_CTEST_LAUNCH ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${MPIEXEC_MA
 if(QUDA_DIRAC_WILSON
    OR QUDA_DIRAC_CLOVER
```
weinbe2: + `QUDA_DIRAC_CLOVER_HASENBUSCH`? Or does that already imply `QUDA_DIRAC_CLOVER`?

maddyscientist: Indeed, it already implies both regular clover and twisted clover.

PR #900, opened 2019-09-12
Feature/persistent deflation space

cpviolator:
One can now set an MG parameter flag `preserve_deflation` and when MG updates, it will transfer the deflation space from the bottom level to the updated MG object. This will be useful when computing with multiple non-degenerate fermions. The routine will recompute the eigenvalues as needed in the transfer, and thence the SVD.

A test command is given:

`./multigrid_evolve_test --solve-type direct-pc --sdim 8 --tdim 16 --mg-eig 1 true --mg-eig-nEv 1 24 --mg-eig-nKr 1 48 --mg-nvec 1 24 --verbosity verbose --mg-verbosity 0 verbose --mg-verbosity 1 verbose --mg-eig-preserve-deflation true --mg-levels 2 --mg-block-size 0 2 2 2 2 --mass -0.1 --dslash-type twisted-clover --mu 0.1`

PR #910, opened 2019-09-27
Feature/solver deflation

cpviolator:
This PR branches allows for deflation of the fine operator. Currently supported solvers are CG and GCR types. All of these types now have a `matPrecon` operator, and the eigensolver is run in the precision dictated by `--prec-precondition` and deflation is requested by `--inv-deflate`. An example command is given.

```
./invert_test --sdim 8 --tdim 8 --prec double --prec-precondition single --prec-sloppy half --niter 10000 --eig-amax 50 --eig-amin 2.2 --eig-poly-deg 20 --eig-nEv 32 --eig-nConv 32 --eig-nKr 96 --eig-tol 1e-7 --tol 1e-6 --inv-deflate true --verbosity verbose
```


cpviolator:
Some thoughts on this.

1. Evan and I share concern that using `matPrecon` to compute the eigendecomposition can be dangerous, but having another, say, `matEigen` in the solver may be cumbersome. Do we know any use cases where one would want to use one of the existing preconditioners in QUDA in conjunction with deflation?

2. Which solvers should be included? At the moment, I've added CG and GCR, bit BiCGStab should be included for sure.

maddyscientist:
Is there any relation between this pull and #900?

cpviolator:
Yeah, I based it off that PR. I think this was an error in retrospect...

maddyscientist:
@cpviolator now that #900 has been merged can you go ahead and merge develop in to this?  We can then review this and get it merged.

cpviolator:
The eigensolver, when created from inside an inverter, will always be created using `matPrecon`.

The outstanding issue I see here is the insertion of
```
inv_param.eig_param = nullptr;
```
in the invert parameter structure to prevent an eigensolver instance being instantiated when running a solver.

maddyscientist:
On first pass this looks mostly good I think.  I'll actually get this compiled up and tested properly though.

maddyscientist:
I've done some testing, and the deflated solver is working mostly fine.  I've pushed some fixes:
* Ensure correct boundary condition is used with multi-GPU since preconditioner Dirac operator is used
* Construct the preconditioner operator components from the precise components to ensure that no unnecessary precision truncation occurs

There's outstanding work left to be done here though
* Add restart ability to the deflated solver to reproject out the eigenvalues when a given residual drop occurs.  This will ensure continued fast convergence of the solver.  This should occur for example in the reliable updates, but triggered by a new `delta` parameter, e.g., `deflate_delta` or something like that
* Presently the deflation space is not preserved between solves.  This means the deflation is reconstructed each time we do a new solve.  For correct operator we should adopt a similar strategy as is used by MG, where we return a pointer to the constructed deflation space that is preserved between solves.  The `QudaInvertParam::preconditioner` member could perhaps be used for this.

Beyond the above changes needed, this looks like a nice addition to QUDA.

cpviolator:
@maddyscientist Thanks for your comments and additions.

@alexstrel So it looks like this solver deflation and your EigCG deflation are starting to converge even more in their code path similarities. @maddyscientist would like the Eigensolver to be used as a preconditioner in very much the same way you use the deflation space as a preconditioner to preserve the space between calls to `invertQuda`. So, finally (after much discussion) I'll amalgamate as much code as I can between the two routines.

alexstrel:
@cpviolator yes, it is the last point for the unification,  I've made changes.

cpviolator:
@alexstrel Awesome. I can get back to this on Friday or late Thursday. Are your commits EigCG/develop branch pushed?

alexstrel:
No problem, it's still in feature/deflated-solver-revisit. I can wait for the commit. (But it may take longer time for me to prepare for PR)

maddyscientist:
One other comment: I will likely write up my finding for tuning the parameters for the deflated solvers in a wiki page.  This should provide a nice user guide for this.

Great work @cpviolator putting this pull together.

lib/inv_ca_cg.cpp
```diff
@@ -760,7 +795,7 @@ namespace quda {
       param.secs += profile.Last(QUDA_PROFILE_COMPUTE);
 
       // store flops and reset counters
-      double gflops = (blas::flops + mat.flops() + matSloppy.flops())*1e-9;
+      double gflops = (blas::flops + mat.flops() + matSloppy.flops() + matPrecon.flops()) * 1e-9;
```
weinbe2: Remind me, can `matPrecon` and `matSloppy` alias each other? If so, is this math wrong?

maddyscientist: no they cannot

maddyscientist: and even if they did, no error, since the act of fetching the flops also zeros the counter

weinbe2:
@mathiaswagner not sure why the last Jenkins build failed:
```
[...]
[ 58%] Building CUDA object lib/CMakeFiles/quda.dir/copy_color_spinor_mg_ss.cu.o
[ 59%] Building CUDA object lib/CMakeFiles/quda.dir/copy_color_spinor_mg_sh.cu.o
[ 59%] Building CUDA object lib/CMakeFiles/quda.dir/copy_color_spinor_mg_sq.cu.o
[ 60%] Building CUDA object lib/CMakeFiles/quda.dir/copy_color_spinor_mg_hs.cu.o
sh: 1: cannot create /tmp/tmpxft_00000e1b_00000000-0: Permission denied
lib/CMakeFiles/quda.dir/build.make:1037: recipe for target 'lib/CMakeFiles/quda.dir/copy_color_spinor_mg_sq.cu.o' failed
make[2]: *** [lib/CMakeFiles/quda.dir/copy_color_spinor_mg_sq.cu.o] Error 2
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:1027: recipe for target 'lib/CMakeFiles/quda.dir/all' failed
make[1]: *** [lib/CMakeFiles/quda.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
Build step 'CMake Build' marked build as failure
```
Anything really stopping us from committing this merge? @maddyscientist

mathiaswagner:
This pull breaks quda for MILC as `bool` is not a valid type in `quda.h` for a C compiler ...

PR #911, opened 2019-09-27
avoid QMP/QIO configure in every make invocation

mathiaswagner:
When QUDA_DOWNLOAD_USQCD is enabled avoid running configure multiple times by pointing QUDA_QMPHOME / QUDA_QIOHOME / QUDA_LIMEHOME to the result of the previous build and remove the target dependency on QMP/QIO from QUDA


mathiaswagner:
About to board my plane. Quick check looked ok but it might need more in the wild testing. Hence only a draft for now.

maddyscientist:
Excellent addition 😄

mathiaswagner:
Was to quick yesterday. A bit more testing revealed that it did not work reliably. It does work now but requires to manually `cmake .` once QIO/QMP have been build (unless anything else triggers a rerun of cmake already).

mathiaswagner:
This should now work as intended.

weinbe2:
Noticed this was still pending, so adding a quick ping to @maddyscientist and @cpviolator . `QUDA_DOWNLOAD_USQCD` has some quirks on my end that I think are just related to my machine setup as opposed to something fundamental, but it does block me from giving this an approval myself.

cpviolator:
Works perfectly for me using `ccmake` on NVSOCAL2.

cpviolator:
@maddyscientist Happy with this?

PR #912, opened 2019-10-01
Use an enum for test_type in dslash_test

mathiaswagner:
Can now specify the test type using a string

`--test ENUM:value in {Dslash->0,MatPC->1,Mat->2,MatPCDagMatPC->3,MatDagMat->4,M5->5,M5inv->6,Dslash4pre->7}`

Also add some helpers to use the `CLI::TransformPairs<T>` to also convert from `T` to `string` for potential use in all the `get_` functions.


weinbe2:
This visually looks good to me, but @maddyscientist knows best what needs to be tested for `dslash_[c]test`. @mathiaswagner does this break staggered? One way or the other, do you have the cycles to also add staggered versions? I probably can a bit later today, but I may end up blocked until Friday.

include/enum_quda.h
```diff
@@ -526,6 +526,8 @@ extern "C" {
     QUDA_EXTLIB_INVALID = QUDA_INVALID_ENUM
   } QudaExtLibType;
 
+  enum class dslash_test_type { Dslash = 0, MatPC, Mat, MatPCDagMatPC, MatDagMat, M5, M5inv, Dslash4pre };
```
maddyscientist: This is C++ code - enum_quda.h must be a pure C header since it is part of the interface

maddyscientist:
Apart from the inclusion of C++ code in the C header, this looks good.  Will approve once this is fixed.

include/enum_quda.h
mathiaswagner: Good point. Reason it is there (apart from me forgetting about not putting c++ in the interface here) is that there are test only functions in the interface like `void dslashQuda_mdwf(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity parity, dslash_test_type test_type)`.
Probably best to move them into a helper test thing.

mathiaswagner:
@weinbe2 Staggered does not care, though we should use an enum for staggered as well.

maddyscientist:
Could the repeated boilerplate in dslash_test and dslash_ctest be moved to the new helper file?

maddyscientist:
Also, will you make the enum changes for staggered as well in this pull?

maddyscientist:
Ok, this looks all good to me.  The code is now much more readable using enums instead of arbitrary numbers.  Final comment: can you deleted the commented out CLI transform lines, since these are now superfluous and only serve to confuse.

PR #914, opened 2019-10-01
Saving full and half parity fields via QIO, adding XML descriptors

weinbe2:
Update: good to discuss merging

This is meant as a launchpad to discuss more complete support for loading/saving `ColorSpinorField`s using QIO. In the current `develop` branch the `write_spinor_field` routine will save either a single or full parity field without qualms, which in the single parity case breaks lexicographical ordering and lattice volume assumptions.

As a quick hack to give useable eigenvectors, the routine to save eigenvectors detects if the field is a single-parity `nColor == 3` field, and if so expands it into a full parity field with zeroes on the other parity. (For now, it assumes the single parity field is even, begging a discussion about adding metadata to a single parity field of a "suggested" parity---there are good technical reasons why a single-parity `ColorSpinorField` shouldn't error-enforce being even or odd in general.)

Per a suggestion from @detar, I also put in a tentative XML descriptor for fields saved by QUDA. This should _not_ be considered final and the exact format of it could break at any time.

Right now these changes are for save only, making this branch especially dangerous in workflows with both saving and loading, since the save will ALWAYS expand a single-parity `nColor == 3` field into a full parity field on save, but won't do anything to detect this case on load (though the metadata is there in the XML).


weinbe2:
Note: this branch assumes #910 has already been merged in.

The latest changes fix the hacky-ness of the initial QIO fix, and is ultimately intended to avoid a divergence in the features @detar needs for deflation to work as part of his workflow (and for the general good of the community too).

This branch adds a new metadata tag, `QudaParity suggested_parity;` to `ColorSpinorField` (and an analogous component to `ColorSpinorParam`) which is made to indicate the parity of a single-parity field in algorithms where it matters. For a full-parity field, `suggested_parity` should just be set to `QUDA_INVALID_PARITY`. This addition in no way changes the behavior of any existing algorithm: QUDA assumes you can use a single parity field interchangeably for even and odd parity usage, and that's unchanged.

The primary place this is used for now is saving and loading eigenvectors. Since a single parity field isn't a well-defined QIO type, they need to be expanded to full parity objects before being saved, and the correct parity needs to be extracted when full parity objects are loaded. This additional metadata tag serves to specify if it's even or odd. In the case of eigenvalue deflation, whether a `ColorSpinorField` is even or odd can be deduced from the operator preconditioning type, which is what is done in `eigensolve_quda.cpp`. 

For the time being, this is only done for `nColor == 3`, since for the forseeable future I don't see other software loading the single-parity eigenvectors that are generated for deflating coarse operators. 

As a bonus, this PR adds some XML information to saved color spinors and gauge fields, per a suggestion from Carleton, and does some additional checks on loading.

weinbe2:
As a note, the changes for loading and saving only went into `eigensolve_quda.cpp` and not into the save and load routines in `deflation.cpp`, so those are effectively depreciated. IIRC @cpviolator mentioned eventually the deflation file would disappear so I don't consider this an issue, but please correct me if I'm wrong; it shouldn't be hard to swap it in.

@maddyscientist and @cpviolator this is good for review and, pending that, is ready to merge once #910 is in.

include/color_spinor_field.h
```diff
@@ -96,6 +99,11 @@ namespace quda {
 
     QudaPCType pc_type; // used to select preconditioning method in DWF
 
+    QudaParity suggested_parity; // used to specify whether a single parity field is even/odd
```
maddyscientist: This should use doxygen markup.

include/color_spinor_field.h
```diff
@@ -338,6 +352,11 @@ namespace quda {
...
```
maddyscientist: Doxygen markup

lib/eigensolve_quda.cpp
```diff
@@ -409,18 +462,58 @@ namespace quda
...
+          else if (spinor_parity == QUDA_ODD_PARITY)
+            blas::copy(tmp[i]->Odd(), *tmp_intermediate);
+          else
+            errorQuda("When saving single parity vectors, the suggested parity must be set.");
+        }
+        delete tmp_intermediate;
       }
     } else {
-      for (int i = 0; i < Nvec; i++) { tmp.push_back(eig_vecs[i]); }
+      // FIXME: add conversion to full site subset from single parity fields.
```
maddyscientist: what does this FIXME refer to?

include/color_spinor_field.h
```diff
@@ -16,6 +16,9 @@ namespace quda {
 
   struct FullClover;
 
+  // Declaration in advance
```
maddyscientist: What's the point of this pre-declaration?  Why not have the definition here too?

maddyscientist:
Seems like a mostly un-controversial pull request 😄.  @weinbe2 have you verified that MG null space and coarse-grid vector I/O is working as it should be with these changes.  Worth specifically testing since Nc != 3 is a special case. Apart from that, mostly trivial changes.

include/color_spinor_field.h
weinbe2: Done, using this as a reference: https://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/doxygen-howto.html

include/color_spinor_field.h
weinbe2: Done

lib/eigensolve_quda.cpp
weinbe2: Outdated, my original hack didn't support fixing single-parity CPU fields. Removed.

weinbe2: Nevermind, I never actually finished fixing that! Good catch. Addressing it now.

include/color_spinor_field.h
weinbe2: No good reason, moving it.

weinbe2:
Doing a fresh build and test of MG now!

weinbe2:
All fixes in. Verified MG in the following ways:
* Verified saving and loading of near-null vectors by checking that the convergence behavior on the original run and the loading run was identical: matched down to however many decimal places.
* Verified saving and loading of singular vectors/eigenvectors for the coarsest level via the criteria above for both the direct and the preconditioned operator.
* Verified that, as expected, the file size for the preconditioned operator was one half the size of the full operator. This behavior is consistent with the expanding single -> full parity fields on the save only being enabled for `Nc == 3`
* Verified that loading vectors on the coarsest level barfs if you pass single/full parity fields to a preconditioned/direct solve, respectively. 
** This is not a relevant test for near-nulls because you need the full parity field. This doesn't change even if we add the optimization of throwing out one parity of the prolongator/restrictor during the solve because we need the full parity fields to perform the coarse link construct on subsequent runs.

PR #921, opened 2019-10-08
Hotfix/cast fix

maddyscientist:
This is a bug fix pull request:
* Changed function signatures of gauge-accessor `load`/`save` functions to use complex-valued arrays instead of real-valued arrays: this avoids casting to a real array that causes undefined behaviour and lead to bad code being generated by the compiler.
* Switched all uses of `load`/`save` to use `operator=` overloading to remove any pointer casting from kernel code to accommodate the above change.
* Minor fix to CMake build (don't build tests if building with `QUDA_QDPJIT` enabled, rather than if `QUDA_INTERFACE_QDPJIT` is enabled)

### DIFF
--- a/include/gauge_field_order.h
+++ b/include/gauge_field_order.h
@@ -73,7 +73,7 @@ namespace quda {
        */
       template<typename M>
       __device__ __host__ inline void operator=(const M &a) {
-	gauge.save(a.data, x_cb, dim, parity);
+        gauge.save(a.data, x_cb, dim, parity);
       }
     };
 
@@ -139,7 +139,7 @@ namespace quda {
        */
       template<typename M>
       __device__ __host__ inline void operator=(const M &a) {
-	gauge.saveGhost(a.data, ghost_idx, dim, parity);
+        gauge.saveGhost(a.data, ghost_idx, dim, parity);
       }
     };
 
@@ -1137,39 +1137,36 @@ namespace quda {
         {
         }
 
-        __device__ __host__ inline void Pack(real out[N], const complex in[N/2], int idx) const
+        __device__ __host__ inline void Pack(real out[N], const complex in[N / 2], int idx) const
         {
           if (isFixed<Float>::value) {
 #pragma unroll
-            for (int i = 0; i < N/2; i++) {
-              out[2*i+0] = scale_inv * in[i].real();
-              out[2*i+1] = scale_inv * in[i].imag();
+            for (int i = 0; i < N / 2; i++) {
+              out[2 * i + 0] = scale_inv * in[i].real();
+              out[2 * i + 1] = scale_inv * in[i].imag();
             }
           } else {
 #pragma unroll
-            for (int i = 0; i < N/2; i++) {
-              out[2*i+0] = in[i].real();
-              out[2*i+1] = in[i].imag();
+            for (int i = 0; i < N / 2; i++) {
+              out[2 * i + 0] = in[i].real();
+              out[2 * i + 1] = in[i].imag();
             }
           }
         }
 
         template <typename I>
-        __device__ __host__ inline void Unpack(complex out[N/2], const real in[N], int idx, int dir, real phase, const I *X, const int *R) const
+        __device__ __host__ inline void Unpack(complex out[N / 2], const real in[N], int idx, int dir, real phase,
+                                               const I *X, const int *R) const
         {
           if (isFixed<Float>::value) {
 #pragma unroll
-            for (int i = 0; i < N/2; i++) {
-              out[i] = scale * complex(in[2*i+0], in[2*i+1]);
-            }
+            for (int i = 0; i < N / 2; i++) { out[i] = scale * complex(in[2 * i + 0], in[2 * i + 1]); }
           } else {
 #pragma unroll
-            for (int i = 0; i < N/2; i++) {
-              out[i] = complex(in[2*i+0], in[2*i+1]);
-            }
+            for (int i = 0; i < N / 2; i++) { out[i] = complex(in[2 * i + 0], in[2 * i + 1]); }
           }
         }
-        __device__ __host__ inline real getPhase(const complex in[N/2]) const { return 0; }
+        __device__ __host__ inline real getPhase(const complex in[N / 2]) const { return 0; }
       };
 
       /**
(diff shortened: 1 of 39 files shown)


include/kernels/clover_sigma_outer_product.cuh
```diff
@@ -58,10 +58,9 @@ namespace quda
 
     result -= conj(result);
 
-    Matrix<Complex, 3> temp;
-    arg.oprod.load(reinterpret_cast<real *>(temp.data), idx, (mu - 1) * mu / 2 + nu, parity);
+    Matrix<Complex, 3> temp = arg.oprod((mu - 1) * mu / 2 + nu, idx, parity);
```
weinbe2: Can this `3` become an `nColor`, or is there a reason it shouldn't be because it's an outer product?

maddyscientist: trying minimize code changes that are beyond scope of this pull

lib/extract_gauge_ghost_extended.cu
```diff
@@ -82,21 +83,22 @@ namespace quda {
 
     int oddness = (parity+arg.localParity[dim])&1;
     
+    Matrix<complex<typename mapper<Float>::type>, Ncolor(length)> u;
```
weinbe2: -> create a `constexpr int nColor = Ncolor(length);` above as you've done elsewhere

maddyscientist: Since this is a one line use is there any point?

lib/ks_force_quda.cu
```diff
@@ -63,40 +62,21 @@ namespace quda {
 #endif
 #endif
 
-      Matrix<complex<Float>,3> O;
-      Matrix<complex<Float>,3> G;
-      Matrix<complex<Float>,3> M;
-
+      Matrix<complex<Float>,3> O, G, M;
```
weinbe2: `nColor`? (similar to clover comment)

maddyscientist: Again these are changes beyond scope

lib/staggered_oprod.cu
```diff
@@ -182,15 +182,14 @@ namespace quda {
   template<typename real, typename Output, typename InputA, typename InputB>
   __global__ void interiorOprodKernel(StaggeredOprodArg<real, Output, InputA, InputB> arg)
     {
+      using complex = complex<real>;
+      using matrix = Matrix<complex,3>;
```
weinbe2: `nColor` comment as before. iirc this is the first time I've seen a `using matrix = Matrix<complex,3>;` in the `*_oprod.cu` files, let's propagate that elsewhere while we're making this global change (unless there's a reason to not)

lib/extract_gauge_ghost_extended.cu
weinbe2: Duly noted, but I've noticed declarations like this have become a style convention elsewhere in the code, so I think it'd be a good change to make just for consistency.

I also think this `constexpr ...` convention is very useful for code readability.

weinbe2: That said, it's not a critical change.

weinbe2:
As a due-diligence FYI, I did the best I could to go through and make `idx <-> mu`-style swaps went through correctly. Example to clarify what I mean:
```
- arg.force.save((Float*)(result_tmp.data), i, dir, parity);
+ arg.force(dir, i, parity) = result_tmp;
```

include/gauge_field_order.h
```diff
@@ -1375,72 +1373,62 @@ namespace quda {
...
-        __device__ __host__ inline void Pack(RegType out[12], const RegType in[18], int idx) const
+        __device__ __host__ inline void Pack(real out[12], const complex in[9], int idx) const
         {
           reconstruct_12.Pack(out, in, idx);
         }
 
         template <typename I>
-        __device__ __host__ inline void Unpack(RegType out[18], const RegType in[12], int idx, int dir,
-            const RegType phase, const I *X, const int *R) const
+        __device__ __host__ inline void Unpack(complex out[18], const real in[12], int idx, int dir, real phase,
```
mathiaswagner: Should this be `complex out[9]` ?

include/gauge_field_order.h
```diff
@@ -1487,100 +1475,102 @@ namespace quda {
...
-        const I *X, const int *R, const Complex scale = Complex(static_cast<RegType>(1.0), static_cast<RegType>(1.0))) const
-    {
-      Complex u = dir < 3 ? anisotropy :
-                            timeBoundary<ghostExchange_>(idx, X, R, tBoundary, scale, firstTimeSliceBound,
-                                lastTimeSliceBound, isFirstTimeSlice, isLastTimeSlice, ghostExchange);
-      Unpack(out, in, idx, dir, phase, X, R, scale, u);
-    }
+        template <typename I>
+        __device__ __host__ inline void
+        Unpack(complex out[18], const real in[8], int idx, int dir, real phase, const I *X, const int *R,
```
mathiaswagner: out[9] ?

mathiaswagner:
@Jenkins build this please

include/gauge_field_order.h
maddyscientist: Thanks.  This nicely highlights one issue: that there is no array-size type checking using naked C++ pointers.  One reason why I'd like to move us to using `std::array` eventually.

tests/plaq_test.cpp
```diff
@@ -79,6 +79,7 @@ void setGaugeParam(QudaGaugeParam &gauge_param)
   pad_size = MAX(pad_size, t_face_size);
   gauge_param.ga_pad = pad_size;
 #endif
+  gauge_param.ga_pad = 0;
```
weinbe2: Should this be inside an `#else/#endif`? Otherwise the `gauge_param.ga_pad = pad_size;` in the multigpu block is pointless.

include/gauge_field_order.h
mathiaswagner: All in on `std::array` - just beyond this PR

kostrzewa:
Just fyi: there seems to have been an unreported regression with twisted-clover *CG* (reliable residual norm increases after many iterations) in `develop` which this seems to have fixed, so that's great.

maddyscientist:
@kostrzewa can this issue be reproduced with the internal tests?  Was it isolated to one architecture over another?  Or one compiler other another?

The main work in this branch concerned removal of direct casting of stack arrays from complex to real, and vice versa.  We found found cases where certain  compilers could do this incorrectly, and it was determined this usage was probably undefined behavior.

kostrzewa:
I would need to try to reproduce this using internal tests. What I did the last couple of days chasing the issue on Kepler was testing several QUDA versions (going back to the very old version from November 2018 that I mentioned to you via e-mail) on both Pascal and Kepler. One thing which I noticed today, using the `feature/preserve_deflation` branch (from before it was merged into develop) testing CG versus MG on P100 for a baseline to compare tests using the `cast_fix` and current `develop` branches, is that CG was having some trouble:

```
# QUDA: WARNING: CG: new reliable residual norm 3.488228e-07 is greater than previous reliable residual norm 3.254169e-07 (total #inc 1)
# QUDA: WARNING: CG: new reliable residual norm 3.476218e-07 is greater than previous reliable residual norm 3.410334e-07 (total #inc 2)
# QUDA: WARNING: CG: new reliable residual norm 3.413971e-07 is greater than previous reliable residual norm 3.399292e-07 (total #inc 3)
# QUDA: WARNING: CG: new reliable residual norm 3.372967e-07 is greater than previous reliable residual norm 3.365700e-07 (total #inc 4)
# QUDA: WARNING: CG: new reliable residual norm 3.296888e-07 is greater than previous reliable residual norm 3.275644e-07 (total #inc 5)
# QUDA: WARNING: CG: new reliable residual norm 3.260449e-07 is greater than previous reliable residual norm 3.252879e-07 (total #inc 6)
# QUDA: WARNING: CG: new reliable residual norm 3.235423e-07 is greater than previous reliable residual norm 3.231298e-07 (total #inc 7)
# QUDA: CG: Convergence at 9935 iterations, L2 relative residual: iterated = 3.149286e-10, true = 3.149286e-10 (requested = 3.162278e-10)
```

On the `cast_fix` branch, instead, this does not happen anymore and I also tried the `develop` branch from a few days ago which also had this issue, hence my, somewhat poorly supported conclusion.

I'll try to see what I can do with the internal tests.

kostrzewa:
This was using gcc 6.3.0

PR #924, opened 2019-11-05
Feature/lu batched rotate

cpviolator:
This PR allows for a significant reduction in memory overhead when using the Eigensolver. Instead of creating up to `nEv` extra eigenvectors for basis rotation, one can limit this number using the flag `--eig-batched-rotate <int>`. If this number is greater than 0, or less that the number required at a particular restart, the rotation with be done via an LU decomposition.

N.B. The batched `caxpy_U` does not seem to work as expected. I have defaulted those operations to `caxpy` in lieu of resolving the problem

A test command is given:

```
./eigensolve_test --sdim 8 --tdim 8 --eig-nEv 128 --eig-nKr 256 --verbosity verbose --eig-amin 1.5 --prec double --eig-tol 1e-9 --eig-poly-deg 25 --eig-batched-rotate 10
```


lib/eigensolve_quda.cpp
```diff
@@ -214,6 +216,95 @@ namespace quda
...
+        }
+      }
+    }
+    // Sanity check
+    for (int i = 0; i < size; i++) {
+      if (pivots[i] > 0) errorQuda("Error at pivot %d", i);
+    }
+  }
+
+  void EigenSolver::blockRotate(std::vector<ColorSpinorField *> &kSpace, double *array, int rank, int is, int ie,
```
weinbe2: For readability/clarity, let's change `int type` into a file-local enum indicating full, lower, upper.

cpviolator: 39e1a1ab7b8a533573944594f54db6c537305236

maddyscientist:
@weinbe2 @mathiaswagner any comments on this pull before we merge?

mathiaswagner:
Is the fortran interface affected here? Have not yet checked but also don't want to forget.

maddyscientist:
No effect directly on the Fortran interface in this pull.  We should do need to enable deflation using Fortran, since the `QudaEigParam`does not have a Fortran equivalent (neither does `QudaMultigridParam` for that matter).

Anyway, beyond the scope of this pull.

lib/eigensolve_quda.cpp
```diff
@@ -214,6 +214,94 @@ namespace quda
...
+                                int js, int je, blockType b_type)
+  {
+    // is = i_start
+    // ie = i_end
+    // js = j_start
+    // je = j_end
```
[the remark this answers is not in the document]
weinbe2: ^ Agreed

lib/eigensolve_quda.cpp
```diff
@@ -214,6 +214,94 @@ namespace quda
...
+      }
+    }
+    // Sanity check
+    for (int i = 0; i < size; i++) {
+      if (pivots[i] > 0) errorQuda("Error at pivot %d", i);
+    }
+  }
+
+  void EigenSolver::blockRotate(std::vector<ColorSpinorField *> &kSpace, double *array, int rank, int is, int ie,
+                                int js, int je, blockType b_type)
```
weinbe2: I wouldn't hold this PR up on this point, but if it's easy, I liked the `range = std::pair<size_t,size_t>` @maddyscientist used in `lib/multi_blas_quda.cu`, in large part for readability.

lib/multi_blas_quda.cu
```diff
@@ -701,17 +702,18 @@ namespace quda {
...
       } else {
         // if at the bottom of recursion,
-        // return if on lower left for upper triangular,
-        // return if on upper right for lower triangular.
         if (is_valid_NXZ(x.size(), false, x[0]->Precision() < QUDA_SINGLE_PRECISION)) {
-          if (upper == 1 && j_idx < i_idx) { return; }
-          if (upper == -1 && j_idx > i_idx) { return; }
+          // since tile range is [first,second), e.g., [first,second-1], we need >= here
+          // if upper triangular and upper-right tile corner is below diagonal return
+          if (upper == 1 && range_y.first >= range_x.second) { return; }
```
weinbe2: Took me a while to picture this in my head, but now that I understand it, I agree with this logic (insofar as I don't think there's a corner case testing hasn't caught yet).

weinbe2:
@maddyscientist doing my due-diligence local testing, then this will have my approval.

tests/multigrid_invert_test.cpp
```diff
@@ -130,6 +130,7 @@ void setEigParam(QudaEigParam &mg_eig_param, int level)
   mg_eig_param.nEv = mg_eig_nEv[level];
   mg_eig_param.nKr = mg_eig_nKr[level];
   mg_eig_param.nConv = nvec[level];
+  mg_eig_param.batched_rotate = mg_eig_batched_rotate[level];
```
weinbe2: This should be added to `multigrid_evolve_test.cpp` for completeness.

tests/test_params.cpp
```diff
@@ -371,6 +376,9 @@ std::shared_ptr<QUDAApp> make_app(std::string app_description, std::string app_n
   quda_app->add_option("--msrc", Msrc,
                        "Used for testing non-square block blas routines where nsrc defines the other dimension");
   quda_app->add_option("--mu", mu, "Twisted-Mass chiral twist of Dirac operator (default 0.1)");
+  quda_app->add_option("--m5", m5, "Mass of shift of five-dimensional Dirac operators (default -1.5)");
```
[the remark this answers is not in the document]
weinbe2: @mathiaswagner not for the scope of this pull, but how easy is it to add variable amounts of additional params, for ex for flavors of dwf with variable coefficients?

lib/eigensolve_quda.cpp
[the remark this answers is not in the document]
cpviolator: https://github.com/lattice/quda/commit/77359bde3409ee68f2c88f5b2339a7e4eb8fcca4

tests/multigrid_invert_test.cpp
cpviolator: https://github.com/lattice/quda/commit/4bb34ceb48acf7061377d2ef5bce24f33ed11fa2

lib/multi_blas_quda.cu
maddyscientist: I have some ideas how to clean this up significantly, but I'll save that for a new branch.  Namely to create a `tile` object that bakes this logic into it.  From which we can query its properties, such as `isLower`, `isUpper`, etc.  Only need to implement it once and use it in this and extend it into multi-reduce, etc.

include/quda_cuda_api.h
```diff
@@ -43,11 +43,21 @@ namespace quda {
 #define qudaMemcpy2DAsync(dst, dpitch, src, spitch, width, height, kind, stream) \
   ::quda::qudaMemcpy2DAsync_(dst, dpitch, src, spitch, width, height, kind, stream, __func__, quda::file_name(__FILE__), __STRINGIFY__(__LINE__));
 
+#define STRINGIFY__(x) #x
+#define __STRINGIFY__(x) STRINGIFY__(x)
+#define qudaMemset(ptr, value, count)                                                                                  \
+  ::quda::qudaMemset_(ptr, value, count, __func__, quda::file_name(__FILE__), __STRINGIFY__(__LINE__));
+
+#define STRINGIFY__(x) #x
```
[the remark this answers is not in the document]
maddyscientist: Not following.  What do you mean by twice here?

lib/eigensolve_quda.cpp
maddyscientist: I'm not following this comment @weinbe2.  What does the `range` have to do with the `blockRotate` function here.  Not sure how this would apply here.

include/quda_cuda_api.h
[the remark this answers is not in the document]
mathiaswagner: STRINGIFY is defined twice. once for qudaMemeset and once for qudaMemsetasync.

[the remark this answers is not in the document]
maddyscientist: Doh.  Long standing issue I never noticed before.  Will push fix now.

lib/eigensolve_quda.cpp
```diff
@@ -970,44 +1073,173 @@ namespace quda
...
+      FullPivLU<MatrixXd> matLU(mat);
+      // RitzLU now contains the LU decomposition
+
+      MatrixXd matUpper = MatrixXd::Zero(dim, iter_keep);
+      matUpper = matLU.matrixLU().triangularView<Eigen::Upper>();
+      MatrixXd matLower = MatrixXd::Identity(dim, dim);
+      matLower.block(0, 0, dim, iter_keep).triangularView<Eigen::StrictlyLower>() = matLU.matrixLU();
+
+      // Multi-BLAS friendly arrays to store the L/U matrices
+      double *host_L = (double *)safe_malloc((dim * iter_keep) * sizeof(double));
```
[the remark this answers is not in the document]
cpviolator: https://github.com/lattice/quda/commit/7680832f41af0faad8beab689b26820c24060386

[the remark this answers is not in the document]
maddyscientist: Nice cleanup @cpviolator!

cpviolator:
One final comment on this. Now that the full matrix rotation is done with a single MB call, it will make tuning a bit more lengthy. Every new instance of `iter_keep` for a given production run will trigger a new tuning run. This might kill performance of the eigensolver and would make estimating the run times for production code difficult. One could combat this with a tuning function that iterates over `1-N` for `iter_keep` using random data. What are the collective thoughts on adding such functions?

maddyscientist:
I don't think this is needed Dean.  The tuning overheads with different parameters doesn't seem any more onerous than other algorithms, e.g., multigrid when using CA solvers.  On the larger lattices I was testing, the multi-blas tuning overhead isn't bad compared to other thing.

weinbe2:
@cpviolator If it takes less time than tuning the HISQ MG coarse link construct (~1 hour), it doesn't bother me ;)

lib/eigensolve_quda.cpp
weinbe2: @maddyscientist he's passing in a start and an end, a `std::pair == range` is a nice way to make that compact.

weinbe2:
As of commit `d30e2aa` this branch passed a `ctest -V` run. I don't see any reason to re-run that with the couple of commits since then.

lib/eigensolve_quda.cpp
maddyscientist: will add this now

include/eigensolve_quda.h
```diff
@@ -5,6 +5,9 @@
 #include <dirac_quda.h>
 #include <color_spinor_field.h>
 
+// Local enum for the LU axpy block type
+enum blockType { PENCIL, LOWER_TRI, UPPER_TRI };
```
weinbe2: This should go within the `quda` namespace

maddyscientist: I'll fix in my next push

weinbe2:
Just finished verifying this on the coarsest level of an MG solve, which included passing in some relative prime values for `--mg-nvec`, `--mg-eig-nEv`, `--mg-eig-nKr` and `--mg-eig-batched-rotate` to test checking the remainder cases, and everything passed with flying colors. 

I know there are a few outstanding small pushes from @maddyscientist, but pending those plus a fresh Jenkins build this gets my approval. I'm very impressed with the work done in this pull.

maddyscientist:
Ok, all needed changes and cleanup should now be pushed.

weinbe2:
Did a quick test with a pull with the latest additions, all looks good.

PR #925, opened 2019-11-07
Implement the function BatchInvertMatrix on CPU without CUBLAS

ybyang:
This pull request is trying to implement the function BatchInvertMatrix on CPU, and then it can be done without CUBLAS. Unfortunately the result based on CPU is different from the CUBLAS one in the multigrid_invert_test and the reason is unknown.


maddyscientist:
@ybyang  I'll have a go at cleaning up you pull before we merge.  Thank you for your contribution 😄

cpviolator:
This PR is now up to date with the latest develop, compiles, and successfully MG inverts with a CPU target using the following command:
```
./invert_test --sdim 8 --tdim 16 --inv-multigrid true --mass -0.42 --tol 1e-10 --anisotropy 2.38 --solve-type direct-pc --prec double --prec-sloppy single --prec-precondition half --prec-null half --ngcrkrylov 50 --niter 5000 --verify false --mg-levels 3 --mg-block-size 0 2 2 2 4 --mg-block-size 1 2 2 2 2 --mg-setup-inv 0 cg --mg-setup-inv 1 cg --mg-nvec 0 24 --mg-nvec 1 32 --mg-nvec 2 32 --mg-smoother 0 ca-gcr --mg-smoother 1 ca-gcr --mg-coarse-solver 2 ca-gcr --mg-coarse-solver-maxiter 2 100 --mg-coarse-solver-tol 2 0.001 --mg-nu-pre 0 0 --mg-nu-pre 1 0 --mg-nu-post 1 8 --mg-nu-post 1 8 --mg-eig 2 true --mg-eig-nEv 2 64 --mg-eig-nKr 2 128 --mg-eig-use-poly-acc 2 false --mg-low-mode-check false
```

lib/targets/cuda/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # generate an object library for all target specific files 
-add_library(quda_cuda_target OBJECT malloc.cpp tune.cpp)
+add_library(quda_cuda_target OBJECT malloc.cpp tune.cpp blas_cublas.cu)
```
mathiaswagner: I admit this is something we need to look into but I had hoped not to have .cu files here. We might need to work out how to handle that, in particular making sure all flags are passed properly.

mathiaswagner:
Is the `CPU` target the right thing to do this? I had assumed a `CPU` target means we run everything on the CPU.

mathiaswagner:
Who is stepping up to build a CI for the CPU target?

cpviolator:
@mathiaswagner I'm in the middle of changing that. I will nuke the `cpu` target and create a `targets/generic` directory, in which the Eigen lapack will be placed. If the target implementation does not have its own BatchedInvert Lapack (like ROCm) QUDA will fall back on the generic implementation that uses Eigen and OMP.

CMakeLists.txt
```diff
@@ -57,9 +57,9 @@ if(BUILD_TYPE_VALID LESS 0)
 endif()
 
 # Target type
-set(VALID_TARGET_TYPES CUDA HIP)
+set(VALID_TARGET_TYPES CUDA HIP CPU)
```
maddyscientist: Need to remove this CPU target

include/blas_lapack.h
```diff
@@ -0,0 +1,53 @@
...
+       @param[in] n Dimension each matrix
+       @param[in] batch Problem batch size
+       @param[in] precision Precision of the input/output data
+       @param[in] Location of the input/output data
+       @return Number of flops done in this computation
+    */
+    long long BatchInvertMatrix(void *Ainv, void *A, const int n, const uint64_t batch, QudaPrecision precision,
+                                QudaFieldLocation location);
+
+    //long long BatchInvertMatrixGENERIC(void *Ainv, void* A, const int n, const uint64_t batch, QudaPrecision prec,
```
maddyscientist: Delete this commented out code

lib/CMakeLists.txt
```diff
@@ -206,11 +206,17 @@ if(${QUDA_TARGET_TYPE} STREQUAL "CUDA")
   add_subdirectory(targets/cuda)
 endif()
 if(${QUDA_TARGET_TYPE} STREQUAL "HIP")
-  #set(QUDA_TARGET_DIR "hip")
+  #add_subdirectory(targets/hip)
+endif()
+if(${QUDA_TARGET_TYPE} STREQUAL "CPU")
```
maddyscientist: Remove CPU target here

CMakeLists.txt
```diff
@@ -216,7 +216,9 @@ option(QUDA_TEX "enable texture reads?" ON)
 option(QUDA_FLOAT8 "enable float-8 ordered fixed-point fields?" OFF)
 option(QUDA_NVML "use NVML to report CUDA graphics driver version" OFF)
 option(QUDA_NUMA_NVML "experimental use of NVML to set numa affinity" OFF)
-option(QUDA_VERBOSE_BUILD "display kernel register useage" OFF)
+option(QUDA_VERBOSE_BUILD "display kernel register usage" OFF)
+option(QUDA_CUDA_LAPACK "use cuBLAS lapack" ON)
```
maddyscientist: Just wondering if the option should be `QUDA_BLAS_LAPACK` with the options being `NATIVE` or `GENERIC`.  On CUDA, `NATIVE` would correspond to cuBLAS.  This would allow one to use the generic variant on CUDA if desired, e.g., for benchmarking.

lib/targets/cuda/CMakeLists.txt
maddyscientist: @mathiaswagner this file at present is a .cu file since it uses a mini kernel to set some pointers.  If this is avoided, then this file can be cpp.

maddyscientist:
@cpviolator Had a quick review.  This still needs the CPU target to be removed.  One thought for how to choose which BLAS_LAPACK to use: how about on CUDA we always compile all contents of `targets/cuda` and `targets/generic`, but we have a CMake variable that sets which BLAS_LAPACK we are using by using a particular namespace.  E.g.,
```c++
#if BLAS_LAPACK == BLAS_LAPACK_CUDA
  using blas_lapack = cuda::blas_lapack;
#elif BLAS_LAPACK == BLAS_LAPACK_GENERIC
  using blas_lapack = generic::blas_lapack;
#endif
```

cpviolator:
@maddyscientist An issue here is that when we use the generic blas, a D2H memcpy, and subsequent H2D memcpy must be performed before and after the call to the generic blas function. I can put the memcpy inside the generic blas function (which would get rid of this awkward preprocessor flag (https://github.com/lattice/quda/blob/feature/cpu_invertmatrix/lib/coarse_op_preconditioned.cu#L162) but then it would have a target dependency through the calls to `qudaMemcpy`, which will have to be abstracted.

Also, I don't know how to combine preprocessor defines and namespace defines, and I'm not sure of what the benefit is. If we are going to direct the blas calls at compile time, why compile both? If we will compile both, we can make it a runtime choice and an `mg_param` flag. The latter choice will necessitate the abstraction of `qudaMemcpy`.

cpviolator:
@maddyscientist This PR will now build the generic LAPACK by default, and we have a new param `--native-lapack <true/false> | inv_param.native_lapack` to control the code path used by all functions requiring LAPACK.

@ybyang I found a bug in the CPU batch invert where the Eigen objects were being populated in row major order. Eigen is, by default, column major. This likely accounts for the discrepancy you observed in the initial PR (and is further evidence of the wilson MG robustness as a solver!) With this in place you should see good algorithmic performance of the coarseY inversions.

cpviolator:
From the Jenkins console output, it looks like the builds are failing because the `lib/target/generic/blas_eigen_lapck.cpp` file is being compiled before Eigen is downloaded and expanded. Will look at repositioning the compilation of `target` object files.

mathiaswagner:
There might be an `add_dependency` call missing for the target directory.

cpviolator:
@mathiaswagner I'm trying a lot of combinations to add the EIGEN dependency to `quda_generic_target` and/or  `quda_target` in both `lib/CMakeLists.txt` and `lib/targets/generic/CMakeLists.txt` with no luck.

Errors like:
```
CMake Error at lib/targets/generic/CMakeLists.txt:13 (add_dependencies):
  The dependency target "EIGEN" of target "quda_generic_target" does not
  exist.
```
arise.

cpviolator:
```
CMake Error at lib/CMakeLists.txt:535 (add_dependencies):
  Cannot add target-level dependencies to alias target "quda_generic".
```

lib/targets/cuda/CMakeLists.txt
maddyscientist: I've addressed this: I got rid of the kernel, and now set the pointers on the host, and moved the file to cpp now.

maddyscientist:
Adding the addendum to make it clear: we can now set whether to use the target-native BLAS-LAPACK library (which on CUDA means cuBLAS) or to use a target-independent generic BLAS-LAPACK on the host (which means Eigen).  The default is to use the native target, and it can be switched with `--native-blas-lapack false` on the command line, or similarly with `QudaInvertParam::native_blas_lapack`.

weinbe2:
This looks good from a visual inspection standpoint. I'm going to take this for a due-diligence build+run before giving this a final approve+merge (which should be done before the build tests finish, anyway).

I assume this PR will block #1022 for the time being.

PR #931, opened 2019-12-02
Hotfix/cub

maddyscientist:
This pull request updates CUB to v1.8.  This is required for CUDA 11 compatibility.


mathiaswagner:
Code is obviously not for review here. Do we need excessive testing or do we trust cub?

maddyscientist:
We trust.  This is the same cub that will be included in the next release of the CUDA toolkit.

Issue #932, opened 2019-12-07
TRLM: upper bound of operator should be optionally found using Raleigh quotient


maddyscientist:
At present the user must estimate the upper and lower bounds of the spectrum to include in the polynomial acceleration.  We could make this process a bit easier for the user, such that if the upper bound is found automatically using a Raleigh quotient method in the same way that the CA-CG with Chebyshev basis does at present if the user doesn't specify the bound.

cpviolator:
Closed with #933

PR #933, opened 2019-12-07
Feature/raleigh quotient

cpviolator:
Addresses #932 

This PR enables the use of power iteration to estimate the spectral radius of the operator passed to the eigensolver. If a specific value of `--eig-amax` or `eig_param.amax` is not set, then a value of zero will be passed to the solver. If Chebyshev acceleration is also requested, the solver will compute the Raleigh quotient and use this to make a safe estimate of the largest eigenvalue. If a non zero value of `amax` is passed, that will be used instead.

A test command is given here:

*Estimation*
`./eigensolve_test --sdim 8 --tdim 16`
_Note that the default value of amax passed to the solver is zero_

*Explicit value*
`./eigensolve_test --sdim 8 --tdim 16 --eig-amax 4.0`


lib/eigensolve_quda.cpp
```diff
@@ -191,6 +191,42 @@ namespace quda
     delete tmp2;
   }
 
+  double EigenSolver::estimateChebyOpMax(const DiracMatrix &mat, ColorSpinorField &out, ColorSpinorField &in)
+  {
+
+    if (in.Location() == QUDA_CPU_FIELD_LOCATION) {
+      in.Source(QUDA_RANDOM_SOURCE);
+    } else {
+      RNG *rng = new RNG(in, 1234);
```
weinbe2: While it sort of doesn't matter, I think we should use a different "magic number" whenever we create an RNG. I know we use 1234 elsewhere.

Food for thought @maddyscientist @cpviolator , I'm definitely not going to hold up the PR on this.

lib/qio_field.cpp
```diff
@@ -109,7 +109,7 @@ int read_field(QIO_Reader *infile, int count, void *field_in[], QudaPrecision cp
     errorQuda("QIO_get_typesize %d does not match expected datasize %d", in_typesize, file_prec * len);
 
   // Print the XML string.
-  if (QIO_string_length(xml_record_in) > 0) printfQuda("QIO string: %s\n", QIO_string_ptr(xml_record_in));
+  // if (QIO_string_length(xml_record_in) > 0) printfQuda("QIO string: %s\n", QIO_string_ptr(xml_record_in));
```
weinbe2: This was addressed in #937

lib/eigensolve_quda.cpp
```diff
@@ -191,6 +191,42 @@ namespace quda
...
+        blas::ax(1.0 / norm, *in_ptr);
+      }
+      matVec(mat, *out_ptr, *in_ptr);
+      std::swap(out_ptr, in_ptr);
+    }
+
+    // Compute spectral radius estimate
+    double result = blas::reDotProduct(*out_ptr, *in_ptr);
+    // Zero out workspace
+    blas::zero(out);
```
weinbe2: Do we need these two calls to `blas::zero`, i.e., does a subsequent part of the algorithm assume one or both of these are zero? Otherwise we can save the cost of the call.

weinbe2:
Thanks for adding this @cpviolator , this is a great small addition. I'm going to do a quick due-diligence compile test on my end, but pending that and my couple of suggestions, this is good.

lib/qio_field.cpp
cpviolator: Merged develop.

lib/eigensolve_quda.cpp
cpviolator: The Krylov space passed to the eigensolver from the C interface is zeroed out. I think it's asking for trouble to populate two vectors with essentially junk in a minor function like this, then handing them back to the eigenslolver. Given that zeroing out two vectors, one time, will not have any significant impact on performance, I'd prefer it if we maintained the zeroed out state of the initial Krylov space.

weinbe2: I see it the other way: if the Krylov space vectors don't need to be zeroed out, we shouldn't be zeroing that out either.

I'd let this go if they weren't trivial changes, but since they are trivial, I'd like to see them implemented.

maddyscientist: I agree with @weinbe2 here.  If there is subsequent zeroing that needs to be done, this should be handled outside of there.  A given function shouldn't be zeroing data proactively just in case for some future use.

cpviolator: Ok, took out the zeroing. I had to move the function and use different vectors because as it stood it overwrote potential initial guess data. 

While we're here, I noticed that there is no way for the user to pass in an initial guess via the C interface. Is this intentional? If the device Krylov space referenced the host Krylov space at creation, this would allow for an initial guess. I can hot fix that if there are no objections.

weinbe2:
@cpviolator sorry for the delay: specifically testing this now.

Issue #936, opened 2019-12-09
QIO_string_ptr causes a segfault when loading some Chroma gauge fields


weinbe2:
With some gauge fields generated by Chroma, calling `QDP_string_ptr` on the XML from the gauge field _record_ (as opposed to the overall file) leads to a segfault. This, in particular, happens on line ~112 of `lib/qio_field.cpp`.

For now there's a WAR in place for gauge fields since I've never hit this issue with spinor fields (and MILC has also successfully read spinor fields generated by QUDA), which suggests this is a "narrow" issue.

So far as I can tell this is an issue with the way these gauge fields were written by QIO in the first place (via QDP++), but I haven't nailed it down yet---I haven't hit this issue with any files written by QUDA, and there's been success loading files written by QUDA into MILC. Anyway, further debugging abounds.

weinbe2:
@bjoo I can take point on this, but I wanted to tag you for awareness.

PR #937, opened 2019-12-09
WAR for segfaults when loading some Chroma-generated gauge fields

weinbe2:
Progress making this a formal fix tracked on github via #936 

Should be a non-controversial WAR. `len == 18` corresponds to a gauge field (9 elements * 2 for complex). I've never seen issues with spinor fields, and MILC can verifiably load spinor fields generated by QUDA, so I didn't want to do a blanket comment out of the line.

@cpviolator @alexstrel if one of you can verify, let's get this merged in once it passes build tests.


weinbe2:
FYI @alexstrel confirmed this WAR works on a single GPU

PR #938, opened 2019-12-13
Hotfix/release note clang

maddyscientist:
* First pass at updating the NEWS and README files for the 1.0 release
* Bump the internal version to 1.0.0
* Various fixes to allow QUDA to compiler on clang-10 with the CUDA 9+ headers (enable use of shfl_sync for cub, generics, trove, etc.)
* Remove some legacy texture reference code in gauge_fix_fft.cu and dslash_textures.h


NEWS
```diff
@@ -1,3 +1,82 @@
...
+  precision.  QUDA_RECONSTRUCT is a 3-bit number corresponding to
+  which reconstruct types are enabled, with 1 = reconstruct-8/9, 2 =
+  reconstruct-12/13 and 4 = reconstruct-18, the default is 7 which
+  enables all reconstruct types.
+
+- Completely rewritten all dslash kernels using the accessor
+  framework.  This dramatically reduces code complexity and improve
+  performance.
+
+- New physics functionalithy added: gauge Laplace kernel, Gaussian
```
weinbe2: functionalithy -> functionality

NEWS
```diff
@@ -1,3 +1,82 @@
...
+  quark smearing, topological charge density.
+
+- QUDA can now be built to either utilize texture-memory reads or to
+  use direct memory accessing (cmake option QUDA_TEX).  The default is
```
weinbe2: Isn't this backwards? Default is on for Pascal and earlier, off for Volta and Turing?

README.md
```diff
@@ -106,19 +113,24 @@ details).  MAGMA is available from
...
+QUDA_DOWNLOAD_ARPACK, and QUDA_DOWNLOAD_USQCD.
+
 ### Application Interfaces
 
 By default only the QDP and MILC interfaces are enabled.  For
-interfacing support with QDPJIT, BQCD or CPS; this should be enabled at
-by setting the corresponding `QUDA_INTERFACE_<application>` variable e.g., `QUDA_INTERFACE_BQCD=ON`.
-To keep compilation time to a minimum it is recommended to only enable
-those interfaces that are used by a given application.  
+interfacing support with QDPJIT, BQCD or CPS; this should be enabled
```
weinbe2: And TIFR?

README.md
```diff
@@ -241,13 +246,16 @@ http://lattice.github.com/quda .
...
+*  Eloy Romero (William and Mary)
 *  Claudio Rebbi (Boston University) 
 *  Guochun Shi (NCSA)
 *  Hauke Sandmeyer (Bielefeld)
 *  Mario Schröck (INFN)
 *  Alexei Strelchenko (Fermi National Accelerator Laboratory)
+*  Jiqun Tu (Columbia)
 *  Alejandro Vaquero (Utah University)
 *  Mathias Wagner (NVIDIA)
+*  Andre Walker-Loud
```
weinbe2: (LBL)?

NEWS
maddyscientist: Indeed, this is backwards.  And moreover, I didn't apply the change suggested in the notes that default is to have texture off for Volta/Turing, since getting it to work properly requires two calls to cmake (first sets arch, second sets tex).  So I just left the default as is for all arch.  Eventually texture will go away which make this a moot point.

Fixed.

include/kernels/color_spinor_pack.cuh
```diff
@@ -173,26 +173,26 @@ namespace quda {
...
+    const int dim0 = dim_dir / 2; // this is the dim offset for each thread =0 for dim_threads=1)
     const int dir = dim_dir % 2;
     const int parity = (arg.nParity == 2) ? (parity_dim_dir / (2*dim_threads) ) : arg.parity;
     const int spinor_parity = (arg.nParity == 2) ? parity : 0;
     const int spin_block = (spin_color_block / (Nc / Mc)) * Ms;
     const int color_block = (spin_color_block % (Nc / Mc)) * Mc;
 
 #pragma unroll
-    for (int dim=dim0; dim<4; dim+=dim_threads) {
+    for (int dim=0; dim<4; dim+=dim_threads) {
```
weinbe2: Was this a bug before? Or one of those "right for the wrong reasons" fixes?

maddyscientist: Not a correctness bug before, but was a performance bug.  Compiler could not unroll this loop: it was only that clang-cuda actually reported the lack of loop unrolling where as nvcc was silent on the matter.  As a result we may see small performance uplift for this routine.

include/kernels/dslash_twisted_mass_preconditioned.cuh
```diff
@@ -58,7 +58,7 @@ namespace quda
     typedef Matrix<complex<real>, nColor> Link;
     const int their_spinor_parity = nParity == 2 ? 1 - parity : 0;
 
-#pragma unroll
+#pragma unroll nDim
     for (int d = 0; d < nDim; d++) { // loop over dimension
```
weinbe2: Is there a reason this is a loop over `nDim` and not over `4` like for the other fermion types? (Not asking this to be changed if it doesn't really matter, but it's a noteworthy difference.)

lib/gauge_fix_fft.cu
```diff
@@ -19,17 +19,12 @@ namespace quda {
 
 #ifdef GPU_GAUGE_ALG
```
weinbe2: What type of correctness checking was done for gauge fixing? Just the standard test executable?

include/kernels/dslash_twisted_mass_preconditioned.cuh
maddyscientist: It's a case that it was `nDim` elsewhere, but I ended up hardcoding to 4 where needed.  E.g., in the dslash_wilson.cuh kernel is used for 5-d and 4-d fermion kernels, so I hardcoded to 4.  For this operator, the number of grid dimensions is equal to stencil dimension, so might as well leave the code generic.

Older code is likely hard-coded to 4 where it could be more generic.

lib/unitarize_links_quda.cu
```diff
@@ -335,12 +335,12 @@ namespace{
       for(int dir=0; dir<4; ++dir){
 	if(field.Precision() == QUDA_SINGLE_PRECISION){
 	  copyArrayToLink(&link, ((float*)(field.Gauge_p()) + (i*4 + dir)*18)); // order of arguments?
-	}else if(field.Precision() == QUDA_DOUBLE_PRECISION){     
+	}else if(field.Precision() == QUDA_DOUBLE_PRECISION){
 	  copyArrayToLink(&link, ((double*)(field.Gauge_p()) + (i*4 + dir)*18)); // order of arguments?
```
weinbe2: Is it worth taking the time to turn this into an `operator=` workflow as opposed to a `load/store`-esque workflow like we've done elsewhere?

lib/gauge_fix_fft.cu
maddyscientist: Yes, which checks most of the code paths in question here but certainly not all.  The gauge fixing code is one of those parts of the code that is overly due for a rewrite (using accessors) and deployment into ctest to improve coverage.  Obviously beyond nature of this pull which is just reformatting `<< <` into `<<<` and deleting the old texture reference code path.

lib/unitarize_links_quda.cu
maddyscientist: This kernel has already been significantly rewritten in develop branch.  More cleanup certainly possible, but not in this pull.

README.md
```diff
@@ -18,29 +18,32 @@ fermion actions:
...
 ## Software Compatibility:
 
-The library has been tested under Linux (CentOS 6 and Ubuntu 16.04)
-using releases 7.5, 8.0 and 9.0, 9.1 and 9.2 of the CUDA toolkit.
-CUDA 7.0 is not supported, though they may continue to work fine.
-Eariler versions of the CUDA toolkit will not work.  QUDA has been
-tested in conjuction with both x86-64 and IBM POWER8/POWER9 CPUs.  The
-library also works on 64-bit Intel-based Macs.
+The library has been tested under Linux (CentOS 7 and Ubuntu 18.04)
+using releases 7.5 through 10.2 of the CUDA toolkit.  Eariler versions
```
mathiaswagner: might be worth running a spell checker on README and NEWS.

include/atomic.cuh
```diff
@@ -48,8 +50,10 @@ static inline __device__ double atomicAdd(double* address, double val)
 */
 static inline __device__ double2 atomicAdd(double2 *addr, double2 val){
   double2 old = *addr;
+#if defined(__CUDA_ARCH__) || CUDA_VERSION >= 8000
```
mathiaswagner: Why is that function just doing nothing on CUDA 7.5? Is that safe?

maddyscientist: It doesn't do nothing on CUDA 7.5, since it is also activated if `__CUDA_ARCH__` is defined.  The issue is here that clang-cuda will not allow us to provide our own definition of `atomicAdd` since it sees the definition in the header which was provided for Pascal and above on CUDA 8.0 and above.  This is a restriction of how clang and nvcc resolve functions marked as `__device__` / `__host__`.

The end result here is that on clang-cuda we cannot provide support on Kepler architectures (which require a homegrown version of `atomicAdd` using `atomicCAS`) without using a different name for `atomicAdd` or applying some hacks to the CUDA headers.  Since Kepler support is on the way out for perhaps QUDA 2.0, I don't see this as a major issue.

mathiaswagner: Makes sense now. Thanks. 
Maybe worth adding a comment so this doesn’t confuse anyone in the future.

maddyscientist:
Comments addressed, and I have added the changes discussed in #902 regarding enablement of link fattening kernels whenever staggered fermions are requested.  MILC build instructions https://github.com/lattice/quda/wiki/MILC-with-QUDA updated accordingly.

maddyscientist:
This is now ready to be merged.  All tests passing.

mathiaswagner:
@weinbe2 is this good for you as well? I';d hit merge then.

Issue #797, opened 2019-05-01
Wilson Flow


cpviolator:
I'm homogenising the gauge smearing and fixing routines. The addition of Wilson Flow routines to QUDA would be a nice addition.

AlexVaq:
I had (somewhere) code for that (not production ready but fully working). Once I’m a bit more free I can try to find it, but it might be lost with my IRLM...

Alex

cpviolator:
No worries if you can't find it. The core of the algorithm would be contained to a `computeWFStep` function call, which would perform a single Wilson Flow step. The rest is all boiler plated template instantiation.

But it would be good to refactor working code rather than starting from scratch :)

cpviolator:
Will close with #1556

PR #942, opened 2019-12-22
Feature/wilson flow

cpviolator:
This PR adds Wilson Flow to QUDA. Addresses #797


cpviolator:
Addresses #797

weinbe2:
@dean I'm sorry I haven't had the chance to give this a serious look yet, I've been busy the past couple of days. Which configuration are you using? Once I have a chance I can run a Wilson flow in another program so we can figure out the "ground truth" to base debugging on.

cpviolator:
@weinbe2 I am using `/scratch/weinbe2/config/l32t64b6p4_v3` and I want to run it through MILC as a check. I'm going to need help building MILC for sure, I have no idea. If you know how to acquire some trustworthy Wilson Flow data from that lattice it would be amazing.

I'm just about to fix the anisotropic plaquette calculation so we can use the L=16,64 wilson too.

cpviolator:
@weinbe2 Fairly obvious clue, which I still can't find the source of. Run these three commands:

`./su3_test --sdim 16 --tdim 64 --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --test 3 --prec double`

`mpirun -np 2 ./su3_test --sdim 16 --tdim 32 --gridsize 1 1 1 2 --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --test 3 --prec double`

`mpirun -np 4 ./su3_test --sdim 16 --tdim 16 --gridsize 1 1 1 4 --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --test 3 --prec double`

And you'll see that the topological charge at each step is roughly the same for both commands, but the energy is a factor of two and four different. Clearly an MPI thing. But why the gauge energy is affected and not the topological charge is a mystery.

cpviolator:
@weinbe2 @maddyscientist Fixed the MPI related issue: I used the wrong volume normalisation `Fmunu.Volume()`.

Specific things that I think need to be checked:
1. The variation of the computed energy and topological charge in the Wilson flow output when using different MPI splitting is, in my opinion, too large. One can see that in the APE and stout smearing routines, the variation of the output is tiny. Wilson Flow is nothing more than fancy stout smearing, so this should not be occurring.
2. Is this the correct volume normalisation? https://github.com/lattice/quda/blob/feature/wilson_flow/lib/gauge_energy.cu#L71
3. in both `gauge_energy.cu` and `gauge_qcharge.cu` I use the `Fmunu` as the first argument to the instantiate template, not the gauge field itself as is expected. Is this OK?

erinaldi:
@cpviolator the Wilson Flow "smearing" is different from APE and stout because it acts in all 4 dimensions. Could this be a reason for the MPI difference? What if you only split the volume in (x,y,z) and not t?

cpviolator:
@erinaldi I have used all sorts of MPI splittings in only x, only y, only z, only t, and all the mixtures. There is always an O(1.0e-3) variation. With stout and APE, that variation is at the Machine Precision level. I definitely need a second opinion on this.

cpviolator:
@erinaldi As a further cross reference, we can use the Over-Improved stout routines which smear in 4D, and the variation there is again at MP.

1112 split:
```
Computed topological charge gauge precise is -1.7945808710091891e+00 Done in 0.022511 secs
Computed topological charge gauge precise from density function is -1.7945808710091891e+00
GPU value -1.794581e+00 and host density sum -1.794581e+00. Q charge deviation: -2.737810e-13
Plaquette after 0 OvrImpSTOUT steps: 6.305487e-01 6.305831e-01 6.305143e-01
Q charge at step 005 = -4.4713373484012449e+00
Q charge at step 010 = -2.7192971136502440e+00
Q charge at step 015 = -2.0971305931007485e+00
Q charge at step 020 = -2.6745686155136554e+00
Q charge at step 025 = -5.5589585248045577e+00
Q charge at step 030 = -4.3266058929103757e+00
Q charge at step 035 = -2.3924769021051020e+00
Q charge at step 040 = +3.2428606514855933e+00
Q charge at step 045 = +1.6661119504195128e+00
Q charge at step 050 = +5.8359304937376892e+00
Plaquette after 50 OvrImpSTOUT steps: 6.243327e-01 6.245489e-01 6.241165e-01
Total time for Over Improved STOUT = 5.41714 secs
```

and a 1121 split:
```
Computed topological charge gauge precise is -1.7945808710091888e+00 Done in 0.021589 secs
Computed topological charge gauge precise from density function is -1.7945808710091888e+00
GPU value -1.794581e+00 and host density sum -1.794581e+00. Q charge deviation: -3.952394e-14
Plaquette after 0 OvrImpSTOUT steps: 6.305487e-01 6.305831e-01 6.305143e-01
Q charge at step 005 = -4.4713373484012440e+00
Q charge at step 010 = -2.7192971136502444e+00
Q charge at step 015 = -2.0971305931007485e+00
Q charge at step 020 = -2.6745686155136541e+00
Q charge at step 025 = -5.5589585248045585e+00
Q charge at step 030 = -4.3266058929103730e+00
Q charge at step 035 = -2.3924769021051011e+00
Q charge at step 040 = +3.2428606514855929e+00
Q charge at step 045 = +1.6661119504195128e+00
Q charge at step 050 = +5.8359304937376866e+00
Plaquette after 50 OvrImpSTOUT steps: 6.243327e-01 6.245489e-01 6.241165e-01
Total time for Over Improved STOUT = 6.59829 secs
```

cpviolator:
And just so everyone can see, here is the kind of variation I see with Wilson flow. Notice that the topological charge now sees a lot of variation too.

The Energy Density line is t = 2*t^2 * <E> and <E>

1121:
```
Computed topological charge gauge precise is -1.7945808710091891e+00 Done in 0.021649 secs
Computed topological charge gauge precise from density function is -1.7945808710091891e+00
GPU value -1.794581e+00 and host density sum -1.794581e+00. Q charge deviation: -2.737810e-13
Plaquette after 0 WFlow steps: 6.305487e-01 6.305831e-01 6.305143e-01
Q charge at step 005 = -4.0011192939715228e+00
Energy Density at step 5.000000e-02 = +1.5767714653668344e-02 +3.1535429307336686e+00
Q charge at step 010 = -5.1761587468912493e+00
Energy Density at step 1.000000e-01 = +5.0467189356880000e-02 +2.5233594678440001e+00
Q charge at step 015 = -5.6096479403624429e+00
Energy Density at step 1.500000e-01 = +8.8466938286632504e-02 +1.9659319619251667e+00
Q charge at step 020 = -5.6486579253348390e+00
Energy Density at step 2.000000e-01 = +1.2196430607044763e-01 +1.5245538258805953e+00
Q charge at step 025 = -5.5674074530552868e+00
Energy Density at step 2.500000e-01 = +1.4903622790458484e-01 +1.1922898232366788e+00
Q charge at step 030 = -5.4996622715496493e+00
Energy Density at step 3.000000e-01 = +1.7035161672990964e-01 +9.4639787072172032e-01
```

1121:
```
Computed topological charge gauge precise is -1.7945808710091888e+00 Done in 0.020631 secs
Computed topological charge gauge precise from density function is -1.7945808710091888e+00
GPU value -1.794581e+00 and host density sum -1.794581e+00. Q charge deviation: -3.952394e-14
Plaquette after 0 WFlow steps: 6.305487e-01 6.305831e-01 6.305143e-01
Q charge at step 005 = -4.0042828643414623e+00
Energy Density at step 5.000000e-02 = +1.5768195389420658e-02 +3.1536390778841312e+00
Q charge at step 010 = -5.1820180927416910e+00
Energy Density at step 1.000000e-01 = +5.0470463322638034e-02 +2.5235231661319015e+00
Q charge at step 015 = -5.6162034625018356e+00
Energy Density at step 1.500000e-01 = +8.8475922475910893e-02 +1.9661316105757978e+00
Q charge at step 020 = -5.6537977808574427e+00
Energy Density at step 2.000000e-01 = +1.2198109578886278e-01 +1.5247636973607848e+00
Q charge at step 025 = -5.5703558868016749e+00
Energy Density at step 2.500000e-01 = +1.4906168474832859e-01 +1.1924934779866287e+00
Q charge at step 030 = -5.5007388749321615e+00
Energy Density at step 3.000000e-01 = +1.7038565114136053e-01 +9.4658695078533628e-01
```

erinaldi:
Ok, I see. So after you changed the normalization in Q, the variation under MPI splits is larger than before and similar to the energy density. Definitely sounds like a problem. Thanks for sharing the output.

cpviolator:
@erinaldi I've addressed the issue with data variance with 0772930.

@weinbe2 @maddyscientist I'm happy with the algorithm and the data, so now it's ready for code review. The issue was the three consecutive instantiations in `WilsonFlowStep`.

include/kernels/gauge_wilson_flow.cuh
```diff
@@ -0,0 +1,192 @@
...
+      for (int dir = 0; dir < 4; ++dir) {
+        border[dir] = in.R()[dir];
+        X[dir] = in.X()[dir] - border[dir] * 2;
+        threads *= X[dir];
+      }
+      threads /= 2;
+    }
+  };
+
+  // Wilson Flow as defined in https://arxiv.org/abs/1006.4518v3
```
maddyscientist: The three kernels here, while all different, clearly have much boiler plate in comment, e.g., all the indexing.  How about instead a single templated kernel, with a switch statement inside doing either `W1`, `W2` or `Wt` paths?  Keep separate tune keys for the different kernel types of course.

lib/gauge_wilson_flow.cu
```diff
@@ -0,0 +1,198 @@
...
+#include <tune_quda.h>
+#include <gauge_field.h>
+
+#include <jitify_helper.cuh>
+#include <kernels/gauge_wilson_flow.cuh>
+#include <instantiate.h>
+
+namespace quda {
+
+  template <typename Float, int nColor, QudaReconstructType recon> class GaugeWFlowW1 : TunableVectorYZ
```
maddyscientist: Similarly to the above comment on using a kernel.  Should be using a single Tunable instance here for the three kernel variants.

include/kernels/gauge_wilson_flow.cuh
cpviolator: @maddyscientist Just looking for clarification because my C++ isn't that great. There will be a single `Tunable` instance, such as:
```C++
template <typename Float, int nColor, QudaReconstructType recon> class GaugeWFlow : TunableVectorYZ
```
which has members:
```C++
GaugeWFlowW1(GaugeField &out, GaugeField &temp, GaugeField &in, double epsilon);
GaugeWFlowW2(GaugeField &out, GaugeField &temp, GaugeField &in, double epsilon);
GaugeWFlowVt(GaugeField &out, GaugeField &temp, GaugeField &in, double epsilon);
```
so that the tuning strings are separated, and a case/swicth on
```C++
computeWFlowStep[W1,W2,Vt]<<<tp.grid, tp.block, tp.shared_bytes>>>(arg);
```
Is that the gist of it?

maddyscientist: I'm thinking something like coarse_op_preconditioned.cu:
* Although there's only one kernel, there is a templated option which controls whether the kernel does a norm or the actual calculation.  I'm thinking here we simply have a template enum parameter.  Here's a cartoon.
```c++
template <FlowType type_>
struct Arg {
  static constexpr FlowType type = type_;
};

template <Arg arg>
__global__ void foo(Arg arg)
{
  // first do common indexing etc.

  // now specialize according to type...
  switch(arg.type) {
  case W1type: // do W1 compute
  case W2type: // do W2 compute
  case WTtype: // do WT compute
  }
}
```
* Same `Tunable` class is used for both, and there's an external knob to control which compute-type is being done.  And there is an explicit string added to the `TuneKey` so to break the degeneracy of the tuning.

cpviolator: OK. I didn’t realise that one could update the halo regions from inside a `.cu` file. That gets rid of the calls to the gauge exchange in the interface. With that it’s simple to put the nStep loop in the `.cu`, exchange ghost zones if using MPI, and templetise a WFlowStep call on step type. I’ll update the smearing routines to follow this style in a separate pull. Then I’ll push my updates to over-relaxation gauge fixing.

erinaldi:
@maddyscientist the primary target would be MILC. They have the `wilson_flow` app folder where they do WF on the CPU. Checking against that would be best (both Wilson and Symanzik flow). It would also be great to have a standalone QUDA app for this, just in case we have a GPU ready and we don’t want to compile the whole of MILC. How does it sound?

cpviolator:
OK, some good news, some bad news, and a puzzle.

Good News: The Wilson and Symanzik flow energy data are in excellent agreement with MILC.
Bad News: The topological charge calculation differs from MILC's by a few percent.
Puzzle: The discrepancy in the topological charge goes away as flow time increases....

Some data from an L=72, T=96 lattice.

Symanzik flow MILC:
```
#LABEL time Et Es charge
WFLOW 0 1.00452 1.00431 -12.4383
WFLOW 0.03 0.910357 0.910152 -12.3468
WFLOW 0.06 0.791053 0.790867 -11.6439
WFLOW 0.09 0.672429 0.672269 -8.81809
...
WFLOW 4.92 0.00507238 0.00508143 11.4874
WFLOW 4.95 0.00503267 0.00504173 11.4883
WFLOW 4.98 0.00499353 0.0050026 11.4893
Number of steps = 166
Time to complete flow = 3.831483e+03 seconds
```

Symanzik flow QUDA:
```
0.00e+00 +1.0045222754242917e+00 +1.0043149041108701e+00 -1.2594814221561426e+01
3.00e-02 +9.1035735404933038e-01 +9.1015264921426875e-01 -1.2464389533881425e+01
6.00e-02 +7.9105329210727826e-01 +7.9086791413079394e-01 -1.1707962690681248e+01
9.00e-02 +6.7242897039676142e-01 +6.7226892401941951e-01 -8.8445007220016496e+00
...
4.92e+00 +5.0723843877207358e-03 +5.0814308494653293e-03 +1.1487354517026457e+01
4.95e+00 +5.0326666437905060e-03 +5.0417255773647337e-03 +1.1488324712414881e+01
4.98e+00 +4.9935273659577166e-03 +5.0025992434286704e-03 +1.1489312671369468e+01
Total time for Wilson Flow = 244.17 secs
```
Notice that the energy (cols 2 and 3) are always in good agreement, the Q charge in early flow times does not agree, but it does at later times.

Should we bug hunt the Q charge issue before review?

erinaldi:
> OK, some good news, some bad news, and a puzzle.
> Notice that the energy (cols 2 and 3) are always in good agreement, the Q charge in early flow times does not agree, but it does at later times.
> 
> Should we bug hunt the Q charge issue before review?

I think all gluonic observables have to agree at each step. 
In particular at t=0 there should be no ambiguity coming from the flowing kernel, hence I don't see why Q can be different is the definition is the same. 
Is it using a different definition than MILC? Maybe some rounding errors?
The fact that it agrees at later time might be a coincidence or pointing to the fact that smoother gauge links (less fluctuations at the lattice spacing level) produce a smaller discrepancy when calculating Q.

cpviolator:
I agree. This discrepancy must be addressed. However, the Q charge code has
been in Quda for some time so it’s really separate issue to this PR.

I’ll open up a separate issue on the Q charge tomorrow.

weinbe2:
@cpviolator @erinaldi because they differ by so little, my guess is they take advantage of different amounts of symmetry per-site. Epsilon_ijkl F_ij F_kl has 24 non-zero terms, but using symmetry you can write it as just the sum of 3 (which is what I did back in FUEL). That'd be consistent with the deviation getting smaller with increased flow time: more flow -> smoother, more isotropic fields -> less sensitivity to using more/less symmetry.

There's a _tiny_ chance it could be different order calculations (QUDA re-using what's computed for `symmE`, MILC using the improved definition with 2x2 and rectangle terms), but in practice I've found that leads to _far_ larger deviations (O(1)) at small flow time, so I bet it's what I noted above.

tests/heatbath_test.cpp
```diff
@@ -249,12 +251,30 @@ int main(int argc, char **argv)
...
+
+    computeFmunu(Fmunu, *gauge);
+    double charge = quda::computeQCharge(Fmunu);
+    // double charge = qChargeQuda();
+    printfQuda("Initial gauge field plaquette = %.16e topological charge = %.16e\n", plaq.x, charge);
 
     // Reunitarization setup
     setReunitarizationConsts();
-    plaquette(*gaugeEx);
+    // CallUnitarizeLinks(gaugeEx);
```
maddyscientist: What's the motivation for adding these commented out calls?

cpviolator: MILC performs a reunitarisation as standard before the Wilson Flow routine is called. I thought the discrepancies were coming from that, but it transpired not to be the case as even further data divergence was seen, even in the energy. I suspected I was not using it properly. It would perhaps be a good idea to allow the user to call reunitarisation for any QUDA test, orthogonal to this PR.

lib/gauge_qcharge.cu
```diff
@@ -68,16 +70,16 @@ public:
...
         QChargeArg<Float, nColor, recon, false> arg(Fmunu, (Float*)qDensity);
         QChargeCompute<decltype(arg)> qChargeCompute(arg, Fmunu);
         qChargeCompute.apply(0);
         qudaDeviceSynchronize();
-
+	
         checkCudaError();
         comm_allreduce((double *)arg.result_h);
-        charge = arg.result_h[0];
+        charge = arg.result_h[0] / (-4*M_PI*M_PI);
```
maddyscientist: @cpviolator I guess this is the source of the minus sign discrepancy with tmLQCD?

cpviolator: And the minus sign agreement with MILC.

lib/gauge_field_strength_tensor.cu
```diff
@@ -19,8 +20,9 @@ namespace quda
 public:
     Fmunu(const GaugeField &u, GaugeField &f) :
       TunableVectorYZ(2, 6),
+      //TunableVector(),
```
maddyscientist: What's this line here for (and 48 and 66)?

lib/interface_quda.cpp
```diff
@@ -5741,15 +5799,14 @@ double qChargeQuda()
...
-  // Do we keep the smeared extended field on memory, or the unsmeared one?
 
   profileQCharge.TPSTART(QUDA_PROFILE_INIT);
   // create the Fmunu field
 
   GaugeFieldParam tensorParam(gaugePrecise->X(), gauge->Precision(), QUDA_RECONSTRUCT_NO, 0, QUDA_TENSOR_GEOMETRY);
   tensorParam.siteSubset = QUDA_FULL_SITE_SUBSET;
   tensorParam.order = QUDA_FLOAT2_GAUGE_ORDER;
-  tensorParam.ghostExchange = QUDA_GHOST_EXCHANGE_NO;
+  tensorParam.ghostExchange = QUDA_GHOST_EXCHANGE_PAD;
```
maddyscientist: This is a bug that has been introduced here.  This should remain at `QUDA_GHOST_EXCHANGE_NO`.

FYI: `QUDA_GHOST_EXCHANGE_PAD` is a specific "inline ghost" format used by the stencil operators, but for all pure-gauge algorithms we should only be using either `QUDA_GHOST_EXCHANGE_NO` or `QUDA_GHOST_EXCHANGE_EXTENDED`

maddyscientist:
Approving this commit.  I have made some changes in the last few commits:
* Cleaned up the `gaugeObservableQuda` interface I add, to ensure that the parameter struct is never uninitialized
* Added optional SU(N) projection prior to making any measurements
* Fused the gauge-field energy and topological charge kernels

cpviolator:
@weinbe2 @walkloud Any comments concerns before this gets merged?

walkloud:
no concerns from me.
My comment is “yay!"

weinbe2:
Looks good @cpviolator !

PR #945, opened 2019-12-30
Feature/cuda backend

cpviolator:


cpviolator:
Just a draft at the moment. I'm nearing the point where I think I've identified all the CUDA api and datatypes. @maddyscientist I recall you telling me that there were some Memcpy calls that you left as cudaMemcpy and did not wrap around qudaMemcpy. Can we identify those and deal with them?

cpviolator:
@maddyscientist This one comes to mind: https://github.com/lattice/quda/blob/develop/lib/cuda_color_spinor_field.cpp#L601

maddyscientist:
@cpviolator I'll look into the `cudaMemcpy` stuff once you have merged develop to resolve the conflicts.

One thing funny about the conflicts shown by github are that there are CUB conflicts.  This should not be happening.  There should be no changes to CUB by this branch, since CUB is a straight external drop in.

alexstrel:
@cpviolator a few minor issues have to be resolved: 
1. remove an error code check, starting from the line
https://github.com/lattice/quda/blob/feature/cuda-backend/lib/comm_common.cpp#L488 
since ```qudaMemset_``` already performs the needed check. 
2. same issue for the line
https://github.com/lattice/quda/blob/feature/cuda-backend/lib/comm_common.cpp#L598
3. ```qudaMemcpy2D_``` definition is missing, but required e.g. by the line
https://github.com/lattice/quda/blob/feature/cuda-backend/lib/comm_common.cpp#L560

include/quda_cuda_api.h
```diff
@@ -1,68 +1,68 @@
 #pragma once
 
 #ifndef __CUDACC_RTC__
-#include <cuda.h>
-#include <cuda_runtime.h>
+#include <quda_backend_api.h>
```
maddyscientist, line 4: Perhaps this file should be renamed, since this header is now backend agnostic?

include/qudarand_kernel.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
```
maddyscientist, line 1: I think this file should be renamed to `quda_rand_kernel.h` to make it more readable.

lib/coarse_op_preconditioned.cu
```diff
@@ -82,8 +82,8 @@ namespace quda {
 #endif
         if (compute_max_only) {
           if (!activeTuning()) { // only do copy once tuning is done
-            qudaMemcpyAsync(arg.max_h, arg.max_d, sizeof(Float), cudaMemcpyDeviceToHost, stream);
-            qudaStreamSynchronize(const_cast<cudaStream_t&>(stream));
+            qudaMemcpyAsync(arg.max_h, arg.max_d, sizeof(Float), qudaMemcpyDeviceToHost, stream);
+            qudaStreamSynchronizeDriver(const_cast<qudaStream_t&>(stream));
```
maddyscientist, line 86: Why does have have `qudaStreamSynchronizeDriver` as opposed to `qudaStreamSynchronize`

lib/tune.cpp
```diff
@@ -755,33 +755,35 @@ namespace quda {
...
 	    tunable.apply(0);  // calls tuneLaunch() again, which simply returns the currently active param
 	  }
-	  cudaEventRecord(end, 0);
-	  cudaEventSynchronize(end);
-	  cudaEventElapsedTime(&elapsed_time, start, end);
-	  cudaDeviceSynchronize();
-	  error = cudaGetLastError();
+	  qudaEventRecord(end, 0);
+	  qudaEventSynchronizeDriver(end);
+	  //cudaEventSynchronize(end);
```
maddyscientist, line 764: these commented out lines should be deleted

include/deflation.h
```diff
@@ -62,7 +62,7 @@ namespace quda {
         //Check that RV is a composite field:
         if(RV->IsComposite() == false) errorQuda("\nRitz vectors must be contained in a composite field.\n");
 
-        cudaHostRegister(matProj,ld*tot_dim*sizeof(Complex),cudaHostRegisterDefault);
+        cudaHostRegister(matProj,ld*tot_dim*sizeof(Complex),qudaHostRegisterDefault);
```
maddyscientist, line 65: Shouldn't this be `qudaHostRegister` instead of `cudaHostRegister`?

maddyscientist:
I've done an initial review of this.  Mostly looks fine, and nothing really really surprising:
* This will need develop merged back in, as it looks like there are a lot of conflicts with present develop
* None of the contents of `include/externals` should be modified: these are external libraries that we include in QUDA.  So all changes to the contents of `include/externals` should be reverted.
   * For jitify this shouldn't matter, since jitify is a CUDA-only thing anyway
   * For cub, there is a HIP-variant of CUB that would be used instead
* I've made some over comments in the code that will be quick fixes

Once the above are fixed I will do a second pass review, but there's nothing really problematic that I see.

Quite the global edit @cpviolator!  There are of course significant improvements we can make to this, but I'm happy if this is in an incremental effort.  Getting this merged in would already be a solid step forward.

cpviolator:
Work moved to #982 Will keep open until the CUDA backend is full merged into develop.

PR #952, opened 2020-01-10
Hotfix/gauge memset

maddyscientist:
Bugfix pull
* Fix `cpuGaugeField::zero()` when using qdp-ordered fields
* Silence ASAN warning about invalid memory read when using bi-directional gauge fields on the CPU

### DIFF
--- a/lib/cpu_gauge_field.cpp
+++ b/lib/cpu_gauge_field.cpp
@@ -173,7 +173,7 @@ namespace quda {
     if (link_direction != QUDA_LINK_BACKWARDS)
       errorQuda("link_direction = %d not supported", link_direction);
 
-    void *recv[QUDA_MAX_DIM];
+    void *recv[2*QUDA_MAX_DIM];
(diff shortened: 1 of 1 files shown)


lib/cpu_gauge_field.cpp
```diff
@@ -173,7 +173,7 @@ namespace quda {
     if (link_direction != QUDA_LINK_BACKWARDS)
       errorQuda("link_direction = %d not supported", link_direction);
 
-    void *recv[QUDA_MAX_DIM];
+    void *recv[2*QUDA_MAX_DIM];
```
mathiaswagner: just by looking at the use of recv: What motivated the change to 2*QUDA_MAX_DIM. This is not obvious to me.

maddyscientist: This comes from the constructor for the QDP fine-grain accessor (gauge_field_order.h).  It assumes the ghost array is of length `2*nDim`.  The upper elements are never used, but this silences the ASAN error.

PR #954, opened 2020-01-10
Hotfix/cpu mg setup

maddyscientist:
Hotfix pull
* Fix bug when running MG setup on CPU (fixes bug introduced in #902)
* Reduced compilation time of coarsecoarse_op.cu through ensuring we only instantiate the needed template compilations (approximate halving of compilation overhead)

For correct running of the MG setup on CPU this will require #952 to be merged in first.


mathiaswagner:
Just that we don't miss it: As #952 goes to release we need another PR that merges release into develop before merging.

weinbe2:
@maddyscientist you accidentally lost coarsening Nc = 24 -> Nc = 64, 96 when staggered is enabled. I'm popping in the fix now myself, no need to worry about it.

weinbe2:
FWIW, fixing it didn't increase the compile time by anything I was really aware of.

PR #958, opened 2020-01-22
improve linking against cufft and curand

mathiaswagner:
This should fix having to run cmake twice for cufft and curand.

@kostrzewa do you mind giving it a try?


kostrzewa:
will do

weinbe2:
@mathiaswagner I believe we should be linking `curand` no matter what because it gets used for generating random spinor fields. I can get to this a little later if you don't have the cycles to quickly pop it in now.

kostrzewa (approved):
works for me on PizDaint

mathiaswagner:
Looking at the code more I think we don't even need to link against curand at all. We only use the device interface.

 @maddyscientist @weinbe2 Do you agree? If yes I'd just remove that dependency. I have test a build locally and it seems fine.

weinbe2:
@mathiaswagner if it's building and running fine, sounds good, I don't know if @maddyscientist has other plans in mind.

maddyscientist:
I hadn’t realized this, but I think this is correct that we only use the device interface.  I think this is something that @nmrcardoso probably added and we percolated without thinking too hard.

maddyscientist (approved):

PR #960, opened 2020-01-27
Fix the timer conflicts and a potential double delete in PCG

hummingtree:
1. The preconditioned CG uses the same `TimeProfile` for both the outer solver and the inner preconditioned solver, which leads to timer conflicts. This PR creates a separate `TimeProfile` and pass it to the preconditioned solver.
2. Change the `delete` condition for `x_sloppy` and `r_sloppy` the same as their corresponding `new` condition to avoid potential segmentation faults.

### DIFF
--- a/lib/inv_pcg_quda.cpp
+++ b/lib/inv_pcg_quda.cpp
@@ -13,8 +13,6 @@
 namespace quda
 {
 
-  static TimeProfile profileInvertPrecon("invertPreconQuda");
-
   using namespace blas;
 
   // set the required parameters for the inner solver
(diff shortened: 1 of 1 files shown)


maddyscientist:
One question: GCR (inv_gcr_quda.cpp) uses the same `TimeProfile` instance for the preconditioner and the outer solver.  Is there something in CG that makes this not possible?

hummingtree:
> One question: GCR (inv_gcr_quda.cpp) uses the same `TimeProfile` instance for the preconditioner and the outer solver. Is there something in CG that makes this not possible?

I just tried GCR with CG as the preconditioning solver and saw the same timer conflict.

maddyscientist:
😄  Can you tell me what command you used for this so I can reproduce?

hummingtree:
Hmm ... I was trying in my branch and say the timer conflict. And I can't reproduce the problem in develop for either PCG or GCR ...

maddyscientist:
It's possible this is a bug then that has since been fixed since you branched.

maddyscientist:
Does the bug disappear if you merge in develop into your branch?

hummingtree:
I believe so since CG (not PCG) has changed a lot since last time I merged. Let me do the merge (which will take a while.)

hummingtree:
I have merged the lastest `develop` into my `feature/overlapping-schwarz` and still gets timer conflicts. I haven't been able to find any clue in `inv_pcg_quda.cpp` and `inv_cg_quda.cpp`.

maddyscientist (approved):

maddyscientist:
This pull is all good and fixes a bug that lead to segmentation faults.  We can merge this in once the latest builds have completed. 

Thanks @hummingtree!

PR #961, opened 2020-01-27
Explicit UVM prefetch API

weinbe2:
This PR adds a new virtual `prefetch(QudaFieldLocation, cudaStream_t)` function to the `LatticeField` class, to be overwritten by its children class, which manually calls `cudaMemPrefetchAsync` under the hood on all device fields allocated via `pool_device_malloc`. The `QudaFieldLocation` is used to specify fetching to the host or device, and the `cudaStream_t` is an argument with default value `0` specifying the CUDA stream to use, thinking forward to overlapping memory transfer and compute for deflation. This function has also been added to the `Dirac` class, overriden as appropriate for its children, to prefetch its component fields.

The `prefetch` routine is a no-op for CPU fields. It's also a no-op for CUDA fields unless both managed memory is enabled, as well as a new environment variable `QUDA_ENABLE_MANAGED_PREFETCH`. The latter environment variable is motivated by the fact that `cudaMemPrefetchAsync` suffers an O(launch latency) overhead even if the memory is already resident in the destination.

So far the `prefetch` mechanism is only used in MG setup: it evicts the `B` fields to the host once they are done being used for the coarse construct and verify, and it explicitly migrates all gauge links and clover terms (top level and coarse) to their appropriate device to the device, avoiding swapping during the first MG solve.

This PR required a refactor of the `verify` routine to be called on a per-level basis immediately after constructing a new coarse operator. This was because verifying `|(1 - P P^dagger) B| = 0` would otherwise fetch the `B` fields back to the GPU. There were other possible refactors to avoid this issue, but I thought this was helpful because it forces the check of the coarse op before building subsequent coarse levels, which is a waste of time if the coarse op is messed up somehow.

Verified for Wilson and staggered, GPU and CPU setup.

I'll add `QUDA_ENABLE_MANAGED_PREFETCH` to the wiki once this PR is approved.


include/multigrid.h
```diff
@@ -334,12 +334,16 @@ namespace quda {
...
+       starting from the top down.
+       @details This method verifies the correctness of the MG method.  It checks:
        1. Null-space vectors are exactly preserved: v_k = P R v_k
        2. Any coarse vector is exactly preserved on the fine grid: eta_c = R P eta_c
        3. The emulated coarse Dirac operator matches the native one: D_c = R D P
+       4. The preconditioned operator was correctly formulated: \hat{D}_c - X^{-1} D_c
+       5. The normal operator is indeed normal: im(<x|D^\dag D|x>) < epsilon
      */
-    void verify();
+    void verify(bool recrusively = false);
```
maddyscientist: describe the `recursively` parameter in the doxygen

lib/clover_field.cpp
```diff
@@ -373,6 +373,23 @@ namespace quda {
     checkCudaError();
   }
 
+  void cudaCloverField::prefetch(QudaFieldLocation mem_space, cudaStream_t stream) const
```
maddyscientist, line 376: For clover I'm thinking we should add a specialization of `prefetch` that allows us to specify whether to fetch the clover field or its inverse as opposed to only acting on both.

lib/cuda_color_spinor_field.cpp
```diff
@@ -574,6 +574,21 @@ namespace quda {
...
+
+    // conditionals based on destructor
+    if (is_prefetch_enabled() && alloc && mem_type == QUDA_MEMORY_DEVICE) {
+      int dev_id = 0;
+      if (mem_space == QUDA_CUDA_FIELD_LOCATION) dev_id = comm_gpuid();
+      else if (mem_space == QUDA_CPU_FIELD_LOCATION) dev_id = cudaCpuDeviceId;
+      else errorQuda("Invalid QudaFieldLocation.");
+
+      cudaMemPrefetchAsync(v, bytes, dev_id, stream);
+      if (precision == QUDA_HALF_PRECISION || precision == QUDA_QUARTER_PRECISION) cudaMemPrefetchAsync(norm, norm_bytes, dev_id, stream);
```
maddyscientist: Can you change this to `(precision == QUDA_HALF_PRECISION || precision == QUDA_QUARTER_PRECISION) && norm_bytes > 0`.  This will allow for future optimization when doing fixed-point fields without a norm field (e.g., V fields).

lib/dirac_coarse.cpp
```diff
@@ -373,6 +373,14 @@ namespace quda {
     }
   }
 
+  void DiracCoarse::prefetch(QudaFieldLocation mem_space, cudaStream_t stream) const {
+    Dirac::prefetch(mem_space, stream);
+    if (Y_d) Y_d->prefetch(mem_space, stream);
```
maddyscientist, line 379: Perhaps we should only fetch Yhat and not Y here?  Since Yhat is only used in the solver?

lib/dirac_clover.cpp
```diff
@@ -92,6 +92,11 @@ namespace quda {
     CoarseOp(Y, X, T, *gauge, &clover, kappa, a, mu_factor, QUDA_CLOVER_DIRAC, QUDA_MATPC_INVALID);
   }
 
+  void DiracClover::prefetch(QudaFieldLocation mem_space, cudaStream_t stream) const {
+    Dirac::prefetch(mem_space, stream);
+    clover.prefetch(mem_space, stream);
```
maddyscientist: Perhaps should specialize for preconditioned operator.  `DiracClover` should only fetch clover field, and `DiracCloverPC` operator should grab the direct and inverse field according to the type of preconditioning, e.g., both parities of inverse for symmetric preconditioned operator or one parity of each for asymmetric operator.

lib/dirac_coarse.cpp
maddyscientist: My comment on Clover also applies here I guess.  Only prefetch what we want according to `DiracCoarse` vs `DiracCoarsePC`.

stevengottlieb:
@maddyscientist recursively is misspelled in the + line of code.

Steve

maddyscientist:
Review done.  Looks mostly good, though see comments to be addressed.  Nice work 😄

weinbe2:
@maddyscientist I've addressed your comments. In particular, I added a special version of `prefetch` for `Dirac[Twisted]Clover[PC]` which only fetches the appropriate parities of the clover field and its inverse depending on the preconditioning form.... as best as I can tell, please double check my logic.

include/clover_field.h
```diff
@@ -6,6 +6,14 @@
 
 namespace quda {
 
+  // Prefetch type
```
maddyscientist, line 9: For an internal enum I wouldn't bother with the `Quda` prefix

mathiaswagner: For an internal enum I would suggest to always use enum classes. No need to stick to plain old C. https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum

mathiaswagner (changes requested):
This looks good (apart from my enum class comment)

weinbe2:
As requested, I replaced `enum` with `enum class` in `clover_field.h` and propagated the required changes downstream. I also remembered I had an internal `enum` in `staggered_prolong_restrict.cu`, so I made the same change there. 

Also applied a `clang-format` and did a due diligence build to make sure it goes through. Since this is the home stretch, I preemptively added `QUDA_MANAGED_MEMORY_PREFETCH` to the wiki.

maddyscientist (approved):

lib/clover_field.cpp
```diff
@@ -373,6 +373,67 @@ namespace quda {
...
+  {
+    prefetch(mem_space, stream, CloverPrefetchType::BOTH_CLOVER_PREFETCH_TYPE);
```
mathiaswagner, line 378: what about just naming the element `BOTH`. `Having CloverPrefetchType twice seems a bit wordy.

mathiaswagner (approved):
I am not picky about my duplicate CloverPrefetchType redundancy comment. This is nice to have not blocking.

weinbe2:
Totally assumed this got merged. Any objections to me hitting merge @maddyscientist @mathiaswagner >

maddyscientist:
Only objection is that I hit merge first 😜

PR #962, opened 2020-02-11
Feature/float8

maddyscientist:
This PR adds support for a new data order, `QUDA_FLOAT8_FIELD_ORDER`, for the fermion fields and `QUDA_FLOAT8_GAUGE_FIELD_ORDER` for gauge fields
* Verified from generated assembly code that 128-bit load/store transactions are used for both the `colorspinor::FloatNOrder` accessor and for the blas kernels.
* Since this is an experimental option, must be opted into with cmake advanced option `QUDA_FLOAT8=ON`
* When enabled, half and quarter precision blas and copy kernels will be instantiated
* When enabled, half and quarter precision with either recon-8 or recon-9 gauge fields will use float-8 ordering as their native data order
* Robustness additions to accessor framework to allow float-8 ordered fields to work even when the (spin-projected) halo is not compatible with float-8 ordering - use float-4 ordering for the halo, and force use of halo-packing kernel in the temporal direction
* While I have not added this to this PR, I have verified that Wilson dslash, including multi-GPU, with float-8 ordering just works
* float-8 ordering is only enabled when `QUDA_TEX=OFF`.  Compile time check added to ensure that this is the case
* Fix some bugs introduced with #942


lib/reduce_quda.cu
```diff
@@ -440,6 +440,15 @@ namespace quda {
...
             errorQuda("blas has not been built for Nspin=%d fields", x.Nspin());
+#endif
+          } else if (x.Nspin() == 4 && x.FieldOrder() == QUDA_FLOAT8_FIELD_ORDER) { // wilson
+#if defined(NSPIN4) && defined(FLOAT8)
+            const int M = 3; // determines how much work per thread to do
+            value
+                = nativeReduce<doubleN, ReduceType, float8, short8, short8, M, Reducer, writeX, writeY, writeZ, writeW, writeV>(
+                    a, b, x, y, z, w, v, y.Volume());
+#else
+            errorQuda("blas has not been built for Nspin=%d fields", x.Nspin());
```
mathiaswagner: should we include the FieldOrder in this error message?

include/float_vector.h
```diff
@@ -45,7 +52,7 @@ namespace quda {
     return make_double4(x.x + y.x, x.y + y.y, x.z + y.z, x.w + y.w);
   }
 
-  __host__ __device__ inline float4 operator*(const float a, const float4 x) {
+  __host__ __device__ inline float4 operator*(const float &a, const float4 &x) {
```
mathiaswagner: what's the motivation for making `a` a reference here? Shouldn't passing by value be perfectly fine here?

include/float_vector.h
```diff
@@ -77,14 +84,21 @@ namespace quda {
...
+
+  __host__ __device__ inline float2 operator+(const float2 &x, const float2 y) {
     float2 z;
     z.x = x.x + y.x;
     z.y = x.y + y.y;
     return z;
   }
 
-  __host__ __device__ inline float4 operator+(const float4 x, const float4 y) {
+  __host__ __device__ inline float4 operator+(const float4 &x, const float4 y) {
```
mathiaswagner: why is x by-reference and y by-value?

mathiaswagner (approved):
Some minor comments but everything else looks good. Pending some actual testing.

maddyscientist:
This is now ready for merging.  I have addressed @mathiaswagner's comments.  Moreover, I have reverted a couple of the optimizations as these were found to regress performance on Volta.  With the pull request as is, I will investigating that for another day in another pull.

maddyscientist:
I should note that one specific rework I did was in b4cf12e566a12011c82eff055050399760e9d395, which created a `Coord` object for easier reuse of indexing computations and simplifies the stencil kernels.  This allows for an optimization removing the need to recompute the full lattice index in the neighbour index computation.

include/dslash_helper.cuh
```diff
@@ -85,28 +84,28 @@ namespace quda
      @return checkerboard space-time index
   */
   template <QudaPCType pc_type, KernelType kernel_type, typename Arg, int nface_ = 1>
-  __host__ __device__ inline int getCoords(int coord[], const Arg &arg, int &idx, int parity, int &dim)
+  __host__ __device__ inline auto getCoords(const Arg &arg, int &idx, int s, int parity, int &dim)
```
mathiaswagner, line 87: love auto here

mathiaswagner (approved):
Looks good to me

PR #969, opened 2020-03-05
Aggressive saving of tunecache

cpviolator:
Tuning the eigensolver can be tricky. If a set of algorithmic parameters does not result in convergence, the whole hardware tuning cache is lost. This hot fix saves the tunecache at various stages of the solve, including the very beginning so that hardware tuning from the inverter is also saved.

Shall we do the same for the inverters?


maddyscientist (approved):
Happy to merge this as is.  However, I view this as a short-term hack.

The real long-term solution is for rank 0 to periodically save the tune cache after some amount of time has passed.  Likely not hard to implement, but can be for another day.

PR #970, opened 2020-03-14
Expose createDirac to the user via dirac_quda.h

cpviolator:
This hot fix simply adds forward declarations to the two `createDirac` functions in `interface_quda.cpp` so that users may create solvers outside of `invertQuda()`

### DIFF
--- a/include/dirac_quda.h
+++ b/include/dirac_quda.h
@@ -1686,27 +1686,6 @@ public:
     }
   };
 
-  /**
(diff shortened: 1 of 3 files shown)


maddyscientist:
Perhaps we should just move the `createDirac` functions into dirac.cpp?

cpviolator:
I like that idea. I'll look around for more opportunities to move functions away from the interface.

PR #971, opened 2020-03-14
Hotfix/expose create dirac

cpviolator:
This hot fix simply exposes the createDirac helper functions to the user via dirac.h. This is sufficient for users to construct solvers and operators.

### DIFF
--- a/include/dirac_quda.h
+++ b/include/dirac_quda.h
@@ -15,7 +15,7 @@ namespace quda {
   // Forward declare: MG Transfer Class
   class Transfer;
 
-  // Forward declar: Dirac Op Base Class
+  // Forward declare: Dirac Op Base Class
   class Dirac;
 
   // Params for Dirac operator
@@ -92,7 +92,6 @@ namespace quda {
         printfQuda(
             "b_5[%d] = %e %e \t c_5[%d] = %e %e\n", i, b_5[i].real(), b_5[i].imag(), i, c_5[i].real(), c_5[i].imag());
     }
-
   };
 
   // This is a free function:
@@ -1701,7 +1700,7 @@ public:
    * @param[in/out] d        User prec
    * @param[in/out] dSloppy  Sloppy prec
    * @param[in/out] dPre     Preconditioner prec
-   * @param[in] dRef         Reference prec (EigCG and deflation)
+   * @param[in] dRef         Refine prec (EigCG and deflation)
    * @param[in] param        Invert param container
    * @param[in] pc_solve     Whether or not to perform an even/odd preconditioned solve
    */
--- a/include/dirac_quda.h
+++ b/include/dirac_quda.h
@@ -1695,7 +1695,7 @@ public:
    * @param[in] pc_solve     Whether or not to perform an even/odd preconditioned solve
    */
   void createDirac(Dirac *&d, Dirac *&dSloppy, Dirac *&dPre, QudaInvertParam &param, const bool pc_solve);
-  
+
   /**
    * Create the Dirac operator
    * @param[in/out] d        User prec
@@ -1706,7 +1706,7 @@ public:
    * @param[in] pc_solve     Whether or not to perform an even/odd preconditioned solve
    */
   void createDirac(Dirac *&d, Dirac *&dSloppy, Dirac *&dPre, Dirac *&dRef, QudaInvertParam &param, const bool pc_solve);
-  
+
 } // namespace quda
 
 #endif // _DIRAC_QUDA_H
--- a/include/dirac_quda.h
+++ b/include/dirac_quda.h
@@ -1686,6 +1686,27 @@ public:
     }
   };
 
+  /**
+   * Create the Dirac operator
+   * @param[in/out] d        User prec
+   * @param[in/out] dSloppy  Sloppy prec
+   * @param[in/out] dPre     Preconditioner prec
+   * @param[in] param        Invert param container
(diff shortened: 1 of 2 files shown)


mathiaswagner:
I have mixed feelings about exposing that. Mostly because exposing it in its current form means people will start using it and then we need to keep it.
While it was in the interface_quda.h I think it was considered an internal function.

One thing I would definitely like to see is to have a create dirac with unique_ptr, see 
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-ptr

Before we switched to C++14 rewriting the CreateDirac to use make_unique was not trivial and while we need to change it in a few place where CreateDirac is used I suggest to expose a modern version of it.

Oh, as things might alias (Dsloppy = D) we might need to think about whether a unique_ptr is the right thing here ...

cpviolator:
I agree with @mathiaswagner here. This is very much a quick and dirty fix to allow users to create solver objects in application code. What it is clearly pointing towards is a complete rethink on how users interface with the QUDA library. We are moving away from the territory where user data came only in the form of CPU side fermion vectors, and more towards users writing code that makes direct use of GPU side fermion vectors and operators.

For the time being, I can write application code with the headers in `dirac_quda.h`, but while that it going on we can all think about the best way to shake up `interface_quda.cpp` Currently, the `setDiracParam` family of functions rely on global pointers (`gaugePrecise`, `cloverPrecise`, etc) declared in the preamble of `interface_quda.cpp` so moving the functions causes severe failures.

As for the unique_ptr issue, I think all this stems from having N different Dirac objects for N different ways we wish to use the operator. At the moment we have things like mat, matSloppy, matPrecon, matRefine, and this list could grow indefinitely as algorithms become more complex. In the end, however, we only ever have K different gauges in memory, where K is the number of different precisions needed. How about we change the names of the pointers in the preamble to something like:
```C++
GaugeField *gaugeDouble = nullptr;
GaugeField *gaugeSingle = nullptr;
GaugeField *gaugeHalf = nullptr;
GaugeField *gaugeQuarter = nullptr;
```

We then only instantiate the K different gauges that have been requested at run time, as usual, and add some new members to the `diracParam`, for example:
```C++
switch(inv_param.clover_cuda_prec) {
case: QUDA_DOUBLE_PRECISION : diracParam.clover = cloverDouble;
case: QUDA_SINGLE_PRECISION : diracParam.clover = cloverSingle;
case: QUDA_HALF_PRECISION : diracParam.clover = cloverHalf;
case: QUDA_QUARTER_PRECISION : diracParam.clover = cloverQuarter;
}

switch(inv_param.clover_cuda_prec_sloppy) {
case: QUDA_DOUBLE_PRECISION : diracParam.cloverSloppy = cloverDouble;
case: QUDA_SINGLE_PRECISION : diracParam.cloverSloppy = cloverSingle;
case: QUDA_HALF_PRECISION : diracParam.cloverSloppy = cloverHalf;
case: QUDA_QUARTER_PRECISION : diracParam.cloverSloppy = cloverQuarter;
}
```
etc.

Of course, this means a lot of new members in the parameter struct, but that is the price we pay for not aliasing with multiple instances of the same dirac op.

We should then define these gauges and clover arrays in some new `global_interface_quda.h` file, which is then included in a more structured set of .cpp and header files such as `gauge_interface_quda.cpp` `colver_interface_quda.cpp`, etc to define construction of data, and of course `dirac_interface_quda.cpp` to define the construction of operators. Then there could be  `inversion_interface_quda.cpp` for inversion, `eigensolve_interface_quda.cpp` for eigensolvers, etc. This immediately makes the interface more manageable and readable, and with the above changes to the way we handle `dirac` objects, much less risk of memory leaks and bugs.

mathiaswagner:
@cpviolator As we might not have a quick fix fir smart pointers we might just merge this. What I want to avoid is making it a public quda API function so that we can easily change it.

Would that do for you for now?
(Strictly speaking dirac_quda.h is a quda internal header so if we change it and someone used it we would not care). 
There has been an ask coming over and over again to have a lower level cpp interface for QUDA but this has never gotten enough traction.

cpviolator:
@mathiaswagner The only reason for this PR is so CalLat can construct physics routines that use the QUDA internal objects, while also keeping those routines separate from the main library. In the short term we are writing routines that we can weave into an existing workflow via this branch, based off the improved test suite pull: https://github.com/lattice/quda/tree/experimental/external_progs

cpviolator:
The above branch needs this simple exposure to work. If that were not there, we have have to manually edit every new QUDA pull so the exposed dirac can work with our branch. I'm happy with any suggestion or method that will allow us to maintain an external programs repository that doesn't require us make edits to develop ad develop evolves.

In the medium to long term, a very thin public API that allows users to create objects on the GPU directly, and not go through the canonical method of assigning CPU memory and transferring to the GPU, would be very helpful. Failing that, I can continue modify the existing test suite to the point where a rudimentary, but well organised, CPU level application library can be constructed.

mathiaswagner (approved):

mathiaswagner:
Ok. This is new territory. If these programs are using QUDA they should live in or close to the QUDA repository to avoid any issues with QUDA and these programs running out of sync.
So as this is the case for the mentioned branch I think we could break an internal API without doing too much damage (like would be the case if things live in a fork that just runs out of sync, see some of the MILC versions floating around). 

So I'd merge that with a heads up that at some point we might change the createDirac to use smart pointers. Sounds good?

cpviolator:
That works for me! Thank you @mathiaswagner

PR #974, opened 2020-03-19
Feature/improved test suite

cpviolator:
This PR rearranges the test routines significantly. We now have a `utils` directory which is the contains a low level library of host side and device side routines. On top of that we build `host_reference` code and  `tests` directory. Many helper functions are now defined and can be shared among test scripts.

In addition, the multigrid_invert routines have been eliminated. In their place, one passes a command line flag `inv-multigrid true` and the test scripts will initialise the MG preconditioner.

Still much work to be done across the test suite. This PR is to canvass opinions on the dependency hierarchy and invert/eigensolve scripts.


cpviolator:
Pinging this PR.

weinbe2:
@cpviolator thanks for the reminder. I'm aware and grateful for your improvements, but I'll be slow to get to it because
- I want to give it a very thorough test (hit every corner case I can think of) since it is a big re-engineering, though again very worthwhile
- Since it is a usability improvement as opposed to exposing new features, addressing correctness, etc, it's a bit further down on the priority list by construction

cpviolator:
@weinbe2 Can you please just give me the script that will do all the the very thorough tests? It's obviously better for me to clean up those corner cases than for you.

Also, no worries that this will take you some time to get to. I can always merge in develop, and the testing suite is rarely updated. All this is for the benefit of CalLat creating a set of programs for physics calculations.

mathiaswagner:
@cpviolator as you mentioned, this is a significant rearrangement. As soon as I find cycles I'll try to get on it.

host_reference/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This directory contains host side reference code that cross checks the GPU kerels in QUDA
+with host side calculated routines. Currenly supported reference routines for matrix 
+operators are:
+
+	 1. Wilson dslash
+	 2. Clover dslash
+	 3. Staggered dslash
+	 4. Domain Wall dslash
+	 5. Covariant derivative
```
maddyscientist: Missing twisted mass, non-degenerate twisted mass, twisted clover

maddyscientist: and also should state 4-d and 5-d Shamir DWF as well as Möbius

host_reference/contract_reference.h
```diff
@@ -1,9 +1,9 @@
 #pragma once
 
-#include <blas_reference.h>
+#include <host_utils.h>
+//#include <blas_reference.h>
```
maddyscientist: If it's not being used just delete rather than comment out.

host_reference/covdev_reference.h
```diff
@@ -1,13 +1,11 @@
...
 #ifndef _COVDEV_QUDA_DSLASH_REF_H
 #define _COVDEV_QUDA_DSLASH_REF_H
-#include <blas_reference.h>
 #include <quda_internal.h>
 #include "color_spinor_field.h"
 
-extern int Z[4];
-extern int Vh;
-extern int V;
+// extern int Z[4];
```
maddyscientist: again: if not being used, delete

tests/covdev_test.cpp
```diff
@@ -25,10 +25,10 @@ using namespace quda;
 
 #define MAX(a,b) ((a)>(b)?(a):(b))
 
-QudaPrecision cpu_prec = QUDA_DOUBLE_PRECISION;
-QudaPrecision cuda_prec;
+// QudaPrecision cpu_prec = QUDA_DOUBLE_PRECISION;
```
maddyscientist: delete if not used

tests/gauge_alg_test.cpp
```diff
@@ -32,12 +26,7 @@ int *num_failures_dev;
 
 #define MAX(a,b) ((a)>(b)?(a):(b))
 #define DABS(a) ((a)<(0.)?(-(a)):(a))
-
-QudaPrecision &cpu_prec = prec;
-QudaPrecision &cuda_prec = prec;
-QudaPrecision &cuda_prec_sloppy = prec_sloppy;
-
-
+/*
```
maddyscientist: Any reason not to delete this?

utils/command_line_params.cpp
```diff
@@ -55,8 +55,9 @@ quda::mgarray<char[256]> mg_vec_infile;
 quda::mgarray<char[256]> mg_vec_outfile;
 QudaInverterType inv_type;
 bool inv_deflate = false;
+bool inv_multigrid = false;
 QudaInverterType precon_type = QUDA_INVALID_INVERTER;
-int multishift = 0;
+int multishift = 1;
```
maddyscientist, line 60: Why has this been set to 1 from the default of 0?

tests/staggered_dslash_ctest.cpp
```diff
@@ -72,8 +63,7 @@ bool global_skip = true; // hack to skip tests
 
 QudaParity parity = QUDA_EVEN_PARITY;
 
-static int n_naiks = 1; // Number of naiks. If eps_naik is 0.0, we only need to construct one naik.
-
+// static int n_naiks = 1; // Number of naiks. If eps_naik is 0.0, we only need to construct one naik.
```
maddyscientist: delete

tests/plaq_test.cpp
```diff
@@ -23,9 +23,9 @@
 
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 
-QudaPrecision cpu_prec = QUDA_DOUBLE_PRECISION;
-QudaPrecision &cuda_prec = prec;
-QudaPrecision &cuda_prec_sloppy = prec_sloppy;
+// QudaPrecision cpu_prec = QUDA_DOUBLE_PRECISION;
```
maddyscientist: delete

tests/invert_test.cpp
```diff
@@ -3,36 +3,74 @@
 #include <time.h>
 #include <math.h>
 #include <string.h>
-#include <limits>
 
+// QUDA headers
+#include <quda.h>
+#include <quda_internal.h>
```
maddyscientist: Why is `quda_internal.h` included here?

lib/interface_quda.cpp
```diff
@@ -3618,10 +3623,14 @@ void invertMultiShiftQuda(void **_hp_x, void *_hp_b, QudaInvertParam *param)
 
   // grow resident solutions to be big enough
   for (int i=solutionResident.size(); i < param->num_offset; i++) {
+    printfQuda("Adding vector %d to solutionsResident\n", i);
     solutionResident.push_back(new cudaColorSpinorField(cudaParam));
   }
   for (int i=0; i < param->num_offset; i++) x[i] = solutionResident[i];
 
+  // DMH Heisenbug...
```
maddyscientist: Heisenbug still there?

lib/interface_quda.cpp
```diff
@@ -3618,10 +3623,14 @@ void invertMultiShiftQuda(void **_hp_x, void *_hp_b, QudaInvertParam *param)
 
   // grow resident solutions to be big enough
   for (int i=solutionResident.size(); i < param->num_offset; i++) {
+    printfQuda("Adding vector %d to solutionsResident\n", i);
```
maddyscientist: Unguarded `printfQuda` without any verbosity check.  I guess this line should either be deleted or guarded by a `QUDA_DEBUG_VERBOSE` check.

lib/interface_quda.cpp
```diff
@@ -52,7 +52,8 @@
 #define MAX(a,b) ((a)>(b)? (a):(b))
 #define TDIFF(a,b) (b.tv_sec - a.tv_sec + 0.000001*(b.tv_usec - a.tv_usec))
 
-#define spinorSiteSize 24 // real numbers per spinor
+// Is this even used?
```
maddyscientist: Not use, and should be deleted.  This is a remnant I missed from QUDA 0.1 😄

lib/interface_quda.cpp
```diff
@@ -695,7 +696,7 @@ void initQuda(int dev)
 static cudaGaugeField* createExtendedGauge(cudaGaugeField &in, const int *R, TimeProfile &profile,
 					   bool redundant_comms=false, QudaReconstructType recon=QUDA_RECONSTRUCT_INVALID)
 {
-  profile.TPSTART(QUDA_PROFILE_INIT);
+  // profile.TPSTART(QUDA_PROFILE_INIT);
```
maddyscientist: Why comment these out?  Was this fixing an issue with timers somewhere?

host_reference/dslash_reference.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+    int vol = inv_param.solution_type == QUDA_MAT_SOLUTION ? V : Vh;
+    mxpy(spinorIn, spinorCheck, vol * spinor_site_size * inv_param.Ls, inv_param.cpu_prec);
+    double nrm2 = norm_2(spinorCheck, vol * spinor_site_size * inv_param.Ls, inv_param.cpu_prec);
+    double src2 = norm_2(spinorIn, vol * spinor_site_size * inv_param.Ls, inv_param.cpu_prec);
+    double l2r = sqrt(nrm2 / src2);
+
+    printfQuda("Residuals: (L2 relative) tol %g, QUDA = %g, host = %g; (heavy-quark) tol %g, QUDA = %g\n",
+               inv_param.tol, inv_param.true_res, l2r, inv_param.tol_hq, inv_param.true_res_hq);
+  }
```
maddyscientist, line 141: This needs the fix from #979 applied to it to check for an unsupported `solution_type`.

maddyscientist:
I've done a visual review.  Overall this is great, and a wonderful cleanup of the tests.  Lots of good additions here.

I've left a few comments, though these are mostly cosmetic comments, but a couple of important issues that need to be addressed (cf need to apply the fix in #979).

host_reference/README.md
cpviolator: bfc720c91

host_reference/contract_reference.h
cpviolator: 9accde037

host_reference/covdev_reference.h
cpviolator: 9accde037

tests/covdev_test.cpp
cpviolator: 251c560af

tests/gauge_alg_test.cpp
cpviolator: 251c560af

utils/command_line_params.cpp
cpviolator: I used `multishift` as a parameter describing the number of masses. If set to one, the invert_test goes through the single mass code path, `invertQuda()`. If greater than one, multiple masses are created (as a function of the base mass) and `invertMultiShiftQuda()` is called. I suppose `multishift` in its current context means `masses`.

tests/staggered_dslash_ctest.cpp
cpviolator: 251c560af

tests/plaq_test.cpp
cpviolator: 251c560af

tests/invert_test.cpp
cpviolator: 251c560af

lib/interface_quda.cpp
cpviolator: da1904c5e

lib/interface_quda.cpp
cpviolator: da1904c5e

lib/interface_quda.cpp
cpviolator: Will check with some thorough testing.

lib/interface_quda.cpp
cpviolator: Deleted

lib/interface_quda.cpp
```diff
@@ -693,7 +693,7 @@ void initQuda(int dev)
 static cudaGaugeField* createExtendedGauge(cudaGaugeField &in, const int *R, TimeProfile &profile,
 					   bool redundant_comms=false, QudaReconstructType recon=QUDA_RECONSTRUCT_INVALID)
 {
-  // profile.TPSTART(QUDA_PROFILE_INIT);
+  profile.TPSTART(QUDA_PROFILE_INIT);
```
maddyscientist: Something looks wrong here.  You've added back the `TPSTART` but not the `TPSTOP`.  Intentional or bug?

cpviolator: cb055928c

utils/command_line_params.cpp
maddyscientist: Ah ok, that makes sense.  Can you add a comment explaining this?

maddyscientist (approved):

mathiaswagner:
I did not review yet as there was still a lot of activity. Is this calming down now and a good time to review?

maddyscientist:
I have gone through this pull request quite thoroughly and iterated back and forth a lot on this with @cpviolator.  All tests are now passing, and this is a great clean up of the tests. 

@weinbe2 and @mathiaswagner a cursory review would be welcome on your part, but this pull is ready I believe.

Noting that in doing the various testing in this pull, I have fixed some orthogonal long-standing bugs and done further cleanup.  Bugs fixed include:
* Covariant derivative fused kernel evaluated incorrectly (specifically if only a subset of the partitioned dimensions were enabled then the fused halo update kernel logic in general broke).  Fix was to remove the legacy `DslashArg::ghostDim` array which was a holdover from old dslash kernels.
* `--dagger` command line option was broken
* dslash_test downloaded the gauge field and clover field multiple times

One outstanding issue that I found in this PR is that the gauge_alg_test is flaky with autotuning enabled.  This is a pre-existing condition that was present prior to this PR and will be tracked elsewhere.

Awesome work @cpviolator.

host_reference/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+set(QUDA_REFERENCE_COMMON covdev_reference.cpp dslash_test_helpers.cpp hisq_force_reference.cpp 
+  dslash_reference.cpp wilson_dslash_reference.cpp clover_reference.cpp domain_wall_dslash_reference.cpp 
+  gauge_force_reference.cpp hisq_force_reference2.cpp staggered_dslash_reference.cpp)
+cuda_add_library(quda_reference STATIC ${QUDA_REFERENCE_COMMON})
+if(QUDA_QMP AND QUDA_DOWNLOAD_USQCD AND NOT QUDA_QMPHOME)
+  add_dependencies(quda_reference QMP)
+endif()
+
+if(QUDA_QDPJIT)
+  target_link_libraries(quda
```
mathiaswagner: Curious, why is this here?

host_reference/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+include_directories(../utils)
+
+# enable utils build a common library for all utils
+set(QUDA_REFERENCE_COMMON covdev_reference.cpp dslash_test_helpers.cpp hisq_force_reference.cpp 
+  dslash_reference.cpp wilson_dslash_reference.cpp clover_reference.cpp domain_wall_dslash_reference.cpp 
+  gauge_force_reference.cpp hisq_force_reference2.cpp staggered_dslash_reference.cpp)
+cuda_add_library(quda_reference STATIC ${QUDA_REFERENCE_COMMON})
```
mathiaswagner: can we use `add_library` here please? We are trying to get rid of the `cuda_` prefixed things as they are deprecated and actually currently a wrapper. And this should not have any cuda code in . Also I don't think we need QUDA_REFERENCE_COMMON here but can just pass the source directly to add_library.

host_reference/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+endif()
+
+if(QUDA_QIO AND QUDA_DOWNLOAD_USQCD AND NOT QIO_FOUND)
+  add_dependencies(quda_reference QIO)
+endif()
+if(QUDA_QMP AND QUDA_DOWNLOAD_USQCD AND NOT QMP_FOUND)
+  add_dependencies(quda_reference QMP)
+endif()
+
+macro(QUDA_CHECKBUILDREFERENCE mytarget qudabuildreference)
```
mathiaswagner: Do we need that?

utils/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+# enable utils build a common library for all utils
+set(QUDA_UTILS_COMMON misc.cpp face_gauge.cpp host_utils.cpp host_blas.cpp command_line_params.cpp llfat_utils.cpp staggered_gauge_utils.cpp set_params.cpp)
+cuda_add_library(quda_utils STATIC ${QUDA_UTILS_COMMON})
```
mathiaswagner: same comments for host_reference. No need for a variable here, use `add_library` directly.

utils/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+if(QUDA_QMP AND QUDA_DOWNLOAD_USQCD AND NOT QUDA_QMPHOME)
+  add_dependencies(quda_utils QMP)
+endif()
+
+if(QUDA_QDPJIT)
```
mathiaswagner: Is this needed?

utils/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  target_link_libraries(quda
+    INTERFACE ${QDP_LDFLAGS} ${QDP_LIB} ${QDP_LIBS} ${QIO_LIB} ${LIME_LIB} ${QUDA_QMP_LDFLAGS}
+    ${QMP_LIB} ${MPI_CXX_LIBRARIES})
+endif()
+
+if(QUDA_QIO AND QUDA_DOWNLOAD_USQCD AND NOT QIO_FOUND)
```
mathiaswagner: As this shows up in many places, can we make sure to only add it where it is actually used. So only if it actually is a dependency. (Must admit I have not verified this yet)

tests/CMakeLists.txt
```diff
@@ -1,9 +1,7 @@
-include_directories(. googletest/include googletest)
-
-# enable_language(Fortran)
+include_directories(. googletest/include googletest ../utils ../host_reference)
 
 # enable tests build a common library for all test utilities
-set(QUDA_TEST_COMMON googletest/src/gtest-all.cc test_util.cpp test_params.cpp misc.cpp face_gauge.cpp)
+set(QUDA_TEST_COMMON googletest/src/gtest-all.cc)
```
mathiaswagner: if all that is leftover here is googletest maybe we can clean that up. But we can clean that up later and just file an issue.

host_reference/CMakeLists.txt
cpviolator: This is because I copied the tests/CMakeLists.txt file and left it in. In the tests, I see that the building of tests is disabled when QDPJIT is enabled, so I don't know why it's the tests. Why is that bit of code needed in tests/CMakeLists.txt?

host_reference/CMakeLists.txt
cpviolator: We do not, I'll remove it.

host_reference/hisq_force_reference.cpp
```diff
@@ -4,10 +4,10 @@
 #include <string.h>
 #include <type_traits>
 
-#include "quda.h"
-#include "test_util.h"
-#include "misc.h"
-#include "hisq_force_reference.h"
+#include <quda.h>
```
mathiaswagner, line 7: Do we actually have a convention when we use <> and when "" for includes? @maddyscientist ?

lib/gauge_field_strength_tensor.cu
```diff
@@ -1,6 +1,7 @@
+#include <typeinfo>
```
mathiaswagner: Do we need typeinfo in every .cu file now? If yes maybe that should be done through a header.

cpviolator: This came from one of Kate's recent PRs.

utils/misc.cpp
```diff
@@ -468,68 +468,66 @@ link_sanity_check(void* link, int len, int precision, int dir, QudaGaugeParam* g
...
+
+        //even
+        for (i=0;i < len/2 ;i++){
+            rc = link_sanity_check_internal_12(mylink + gauge_site_size*i, dir, i, gaugeParam, 0);
+            if (rc != 0){
+                printf("ERROR: even link sanity check 12 failed, i=%d\n",i);
+                exit(1);
+            }
+
+            // rc = link_sanity_check_internal_8(mylink + gauge_site_size*i, dir, i, gaugeParam, 0);
```
mathiaswagner: can we delete that?

utils/llfat_utils.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
+
+template <typename real> struct su3_matrix {
+  std::complex<real> e[3][3];
+};
+template <typename real> struct su3_vector {
+  std::complex<real> e[3];
+};
+
+#ifdef __cplusplus
```
mathiaswagner: do we ever include that in a C code?

utils/misc.cpp
cpviolator: I couldn't see it being used anywhere in the tests, and nothing breaks when it's commented out. But they look like useful debugging routines.

mathiaswagner: Indeed. But then we should use `#ifdef` I guess.

utils/llfat_utils.h
cpviolator: In the original tests, they were wrapped around the extern "C" https://github.com/lattice/quda/blob/cfac928a1fa97f407b993b08901b90e8b5ec0d0c/tests/llfat_reference.h

cpviolator: Maybe for the MILC interface?

lib/gauge_field_strength_tensor.cu
mathiaswagner: Yes, I saw the PR that introduced it in some places. Just wanted to bring it up here. Nothing blocking but would wait for @maddyscientist to comment.

tests/dslash_test.cpp
```diff
@@ -43,10 +36,14 @@ cudaColorSpinorField *cudaSpinor, *cudaSpinorOut, *tmp1=0, *tmp2=0;
 
 void *hostGauge[4], *hostClover, *hostCloverInv;
 
-Dirac *dirac = NULL;
+Dirac *dirac = nullptr;
 
 QudaDagType not_dagger;
 
+// For loading the gauge fields
+int argc_copy;
```
mathiaswagner: why do we need the copies?

tests/eigensolve_test.cpp
```diff
@@ -75,303 +58,86 @@ void display_test_info()
...
     app->parse(argc, argv);
   } catch (const CLI::ParseError &e) {
     return app->exit(e);
   }
 
-  if (prec_sloppy == QUDA_INVALID_PRECISION) prec_sloppy = prec;
-  if (link_recon_sloppy == QUDA_RECONSTRUCT_INVALID) link_recon_sloppy = link_recon;
-  if (link_recon_precondition == QUDA_RECONSTRUCT_INVALID) link_recon_precondition = link_recon_sloppy;
+  // Set some default values for precisions if none are passed through the command line
+  setQudaDefaultPrecs();
```
mathiaswagner: should we make that explicit? Or have the command line defaults be compatible with what is set here?

tests/eigensolve_test.cpp
```diff
@@ -75,303 +58,86 @@ void display_test_info()
...
     dw_setDims(gauge_param.X, eig_inv_param.Ls);
   } else {
     setDims(gauge_param.X);
   }
-
-  // set spinor site size
-  int sss = 24;
-  if (dslash_type == QUDA_LAPLACE_DSLASH) sss = 6;
+  // Set spinor site size
+  int sss = (dslash_type == QUDA_LAPLACE_DSLASH ? 6 : 24);
```
mathiaswagner: looks like LAPLACE is not listed as compatible above

tests/pack_test.cpp
```diff
@@ -100,8 +98,9 @@ void end() {
 
 void packTest() {
 
-  printf("Sending fields to GPU...\n"); fflush(stdout);
-  
+  printfQuda("Sending fields to GPU...\n");
+  fflush(stdout);
```
mathiaswagner: I know that has been here for a long time but so we need the flush?

tests/unitarize_link_test.cpp
```diff
@@ -36,7 +36,7 @@ static double svd_rel_error  = 1e-4;
 static double svd_abs_error  = 1e-4;
 static double max_allowed_error = 1e-11;
 
-static QudaPrecision cpu_prec = QUDA_DOUBLE_PRECISION;
+// static QudaPrecision cpu_prec = QUDA_DOUBLE_PRECISION;
```
mathiaswagner: no need to keep that

tests/staggered_eigensolve_test.cpp
```diff
@@ -57,372 +33,178 @@ void display_test_info()
...
-  eig_param.max_restarts = eig_max_restarts;
-  eig_param.cuda_prec_ritz = prec;
-
-  eig_param.use_norm_op = eig_use_normop ? QUDA_BOOLEAN_TRUE : QUDA_BOOLEAN_FALSE;
-  eig_param.use_dagger = eig_use_dagger ? QUDA_BOOLEAN_TRUE : QUDA_BOOLEAN_FALSE;
-  eig_param.compute_svd = eig_compute_svd ? QUDA_BOOLEAN_TRUE : QUDA_BOOLEAN_FALSE;
-  if (eig_compute_svd) {
-    eig_param.use_dagger = QUDA_BOOLEAN_FALSE;
-    eig_param.use_norm_op = QUDA_BOOLEAN_TRUE;
+  if (test_type != 0 && test_type != 3 && test_type != 4) {
```
mathiaswagner: can we catch that using a specific option for test like we do for the dslashes? Checking for numbers here seems brittle.

tests/staggered_eigensolve_test.cpp
```diff
@@ -57,372 +33,178 @@ void display_test_info()
...
+    host_evecs[i] = (void *)malloc(V * my_spinor_site_size * eig_inv_param.cpu_prec);
   }
+  double _Complex *host_evals = (double _Complex *)malloc(eig_param.nEv * sizeof(double _Complex));
 
+  double time = 0.0;
+
+  // QUDA eigensolver test
+  //----------------------------------------------------------------------------
   switch (test_type) {
   case 0: // full parity solution
```
mathiaswagner, line 181: this is already caught above

utils/host_utils.cpp
```diff
@@ -39,11 +47,677 @@ int V_ex, Vh_ex;
...
+  param.nDim = 4;
+  param.siteSubset = QUDA_FULL_SITE_SUBSET;
+  param.siteOrder = QUDA_EVEN_ODD_SITE_ORDER;
+  param.location = QUDA_CPU_FIELD_LOCATION; // DMH FIXME so one can construct device noise
+  for (int d = 0; d < 4; d++) param.x[d] = x[d];
+  quda::cpuColorSpinorField spinor_in(param);
+  quda::spinorNoise(spinor_in, rng, QUDA_NOISE_UNIFORM);
+}
+
+void constructStaggeredTestSpinorParam(quda::ColorSpinorParam *cs_param, const QudaInvertParam *inv_param,
```
mathiaswagner: should we move the staggered specific function in their own file?

tests/staggered_invert_test.cpp
```diff
@@ -1,692 +1,406 @@
...
-  eig_param.spectrum = eig_spectrum;
-  if ((eig_type == QUDA_EIG_TR_LANCZOS || eig_type == QUDA_EIG_IR_LANCZOS)
-      && !(eig_spectrum == QUDA_SPECTRUM_LR_EIG || eig_spectrum == QUDA_SPECTRUM_SR_EIG)) {
-    errorQuda("Only real spectrum type (LR or SR) can be passed to Lanczos type solver");
+  setQudaDefaultMgTestParams();
+  // Parse command line options
+  auto app = make_app();
+  add_eigen_option_group(app);
+  add_deflation_option_group(app);
+  add_multigrid_option_group(app);
```
mathiaswagner, line 109: should this depend on whether QUDA is built with MG?

tests/invertmsrc_test.cpp
```diff
@@ -371,130 +363,131 @@ int main(int argc, char **argv)
...
-//  }
-//    #if 0
-//   else
-  for (int i=0; i <inv_param.num_src; i++){
+  printfQuda("\nDone: %i iter / %g secs = %g Gflops, total time = %g secs\n", inv_param.iter, inv_param.secs,
+             inv_param.gflops / inv_param.secs, time0);
+
+  for (int i = 0; i < inv_param.num_src; i++) { invertQuda(spinorOutMulti[i], spinorIn[i], &inv_param); }
+
+  //  if (true) {
```
mathiaswagner: is this dead code?

tests/dslash_ctest.cpp
```diff
@@ -256,87 +169,53 @@ void init(int precision, QudaReconstructType link_recon) {
...
+  spinorRef = new cpuColorSpinorField(csParam);
+  spinorTmp = new cpuColorSpinorField(csParam);
 
-  // set verbosity prior to loadGaugeQuda
-  setVerbosity(verbosity);
-  inv_param.verbosity = verbosity;
+  csParam.x[0] = gauge_param.X[0];
 
-  // printfQuda("Sending gauge field to GPU\n");
+  constructHostGaugeField(hostGauge, gauge_param, argc_copy, argv_copy);
```
mathiaswagner, line 199: might have missed it but where is copy initialized?

tests/dslash_ctest.cpp
```diff
@@ -396,8 +275,7 @@ void init(int precision, QudaReconstructType link_recon) {
 void end() {
   printfQuda("%s\n", __func__);
   if (!transfer) {
-    if(dirac != NULL)
-    {
+    if (dirac != NULL) {
```
mathiaswagner, line 278: guess we so need smart pointers here ...

tests/dslash_ctest.cpp
mathiaswagner: found it, but what is the point of duplicating the pointer?

mathiaswagner (approved):
Some questions but I like it.  Of course it reminds me what else should be done to the test code but definitely a big step in the right direction.

lib/gauge_field_strength_tensor.cu
maddyscientist: This was a fix for GCC-6 that was broken (missed because not in Jenkins).  The need for `typeinfo` comes from using RTTI for identifying the name for `Tunable`.  I can just put this include in `tune_quda.h` and we can remove all the unnecessary `typeinfo` inclusions.

host_reference/README.md
```diff
@@ -0,0 +1,32 @@
...
+This directory contains host side reference code that cross checks the GPU kerels in QUDA
+with host side calculated routines. Currenly supported reference routines for matrix 
+operators are:
+
+	 1. Wilson dslash
+	 2. Clover dslash
+	 3. Twisted mass
+	 4. Twisted clover
+	 5. Non-degenerate twisted mass/clover	
+	 6. Staggered dslash
```
weinbe2, line 14: I assume this is both staggered and HISQ, please update!

include/dslash_helper.cuh
```diff
@@ -240,7 +240,6 @@ namespace quda
     const int_fastdiv dim[5]; // full lattice dimensions
     const int volumeCB;       // checkerboarded volume
     int commDim[4];           // whether a given dimension is partitioned or not (potentially overridden for Schwarz)
```
weinbe2, line 242: @maddyscientist was Schwarz tested?

maddyscientist: yes

include/invert_quda.h
```diff
@@ -657,42 +657,67 @@ namespace quda {
...
+         TimeProfile &profile);
+    virtual ~CGNR();
 
-  class CG3NE : public Solver {
+    void operator()(ColorSpinorField &out, ColorSpinorField &in);
+
+    virtual bool hermitian() { return false; } /** CGNR is for any system */
+  };
+
+  class CG3 : public Solver
```
weinbe2, line 701: Orthogonal question---is there any reason we still have CG3? Is anyone actually using it?

maddyscientist: @sbacchio ?

lib/inv_pcg_quda.cpp
```diff
@@ -30,139 +31,133 @@ namespace quda {
...
-    if(K){
+    if (K) {
       csParam.create = QUDA_COPY_FIELD_CREATE;
       csParam.setPrecision(param.precision_precondition);
-      rPre = new cudaColorSpinorField(rSloppy,csParam);
-      // Create minvrPre 
+      rPre = new cudaColorSpinorField(rSloppy, csParam);
+      // Create minvrPre
       minvrPre = new cudaColorSpinorField(*rPre);
       commGlobalReductionSet(false);
```
weinbe2: @maddyscientist  should this `commGlobalReductionSet` be enabled for every possible preconditioner? Seems suspicious...

utils/misc.cpp
weinbe2: Aye, I second using `#if 0` if anything for this

weinbe2:
I'm very happy with these modifications. I made a few comments in the pull, but I imagine nothing too painful.

I'm trusting @maddyscientist on testing and @mathiaswagner on cmake verification.

I'll hit approve on it since my suggestions are largely cosmetic, but I hope they'll be addressed or put off in a new issue if they aren't fixed for now.

lib/inv_pcg_quda.cpp
maddyscientist: This is years old code....yes it's suspicious and should be fixed.  Not here though.....  I'll file an issue now.

weinbe2 (approved):

host_reference/README.md
maddyscientist: fixed

lib/gauge_field_strength_tensor.cu
maddyscientist: fixed

host_reference/hisq_force_reference.cpp
maddyscientist: No convention that I've ever used: I always just used `< >` since it looked it better to my eye, but then you told me the standard convention was to use `" "` within a library and `< >` for externals.

mathiaswagner: never mind. Sure we can clean that if we ever want to.

utils/llfat_utils.h
maddyscientist: Fixed

utils/misc.cpp
maddyscientist: Fixed (deleted since it's dead code).

tests/eigensolve_test.cpp
cpviolator: `setQudaDefaultPrecs()` is just a small helper function that sets the precisions. I guess I should rename it to `setPrecisions()` and clarify the comment. The comment and function appear in a few places.

cpviolator: https://github.com/lattice/quda/commit/32bad4291b25cc819820f86420689048465192a0

tests/eigensolve_test.cpp
cpviolator: Yep, laplace is in the staggered file only, fixed with 1fdd1152c

tests/pack_test.cpp
cpviolator: I don't know. I strongly suspect a `fflush` call is not needed.

tests/unitarize_link_test.cpp
cpviolator: 30dea12cc

tests/dslash_ctest.cpp
cpviolator: Minor hangover from dslash_ctest.cpp. In that file, one needs to copy the args to global pointers so that they are visible to functions uses by Google Test. I took a quick look and, short of hacking the google test suite, this is the best option (in dslash_cstest.cpp) In this file, they are not needed. 8e1caf1ad

tests/dslash_ctest.cpp
cpviolator: Which is preferred here, NULL or nullptr... both instances occur in this file.

maddyscientist: `nullptr`

cpviolator: Library wide?

maddyscientist: In general yes, but no need to go crazy on this pull 😜

cpviolator: #987 will deal with this in due time.

tests/pack_test.cpp
maddyscientist: Fixed

tests/staggered_eigensolve_test.cpp
cpviolator: https://github.com/lattice/quda/commit/77a672f8699d8bbc4a0a9c7614bbbee008263f91

tests/staggered_eigensolve_test.cpp
cpviolator: https://github.com/lattice/quda/commit/77a672f8699d8bbc4a0a9c7614bbbee008263f91

utils/host_utils.cpp
cpviolator: b8e9ba5ee

tests/staggered_invert_test.cpp
cpviolator: Well, the `add_multigrid_option_group(app);` line will have to remain, albeit around preprocessor guards. I merged the `tests/staggered_invert_test.cpp` and `tests/staggered_multigrid_invert_test.cpp` and placed the relevant mg stuff around a run time evaluated boolean. The worst that can happen here is that a user invokes the help command, MG options are displayed, but MG is not built... which they will quickly discover if they run with `--inv-multigrid true`

weinbe2: I like your idea of adding preprocessor guards @cpviolator , either here or perhaps better within the `add_multigrid_option_group` definition itself. It's a simple fix and could save the end user confusion.

tests/invertmsrc_test.cpp
cpviolator: https://github.com/lattice/quda/commit/29dc0b5b91836824b59a9b96b4555fda2be45285

tests/staggered_invert_test.cpp
cpviolator: If we block out all the mg option group code from command_line_params.cpp, then I think it would cause even more confusion to the user.

If the user fails to build the multigrid code (which is defaulted not to build) and they try to pass **correct** mg options via the command line, then the user will be presented with an error message informing them that their mg options are unrecognised. This would be a huge red herring, given that every other unbuilt option in QUDA would still be recognised by the CLI parser.

On the other hand, as the code is presently, if the user passes valid mg options and mg is not built, they will be given sensible error messages at runtime.

weinbe2: So is the solution adding a guard within `[staggered_]invert_test` instead? That's fine, I trust your judgement.

mathiaswagner: Ok, I guess we can leave the options in. I think we already have a ton of options that are only caught at runtime later. @cpviolator can you file an issue so maybe we can look at that when someone has a good idea how to catch that when processing command line options...

tests/staggered_invert_test.cpp
```diff
@@ -170,6 +163,11 @@ int main(int argc, char **argv)
   QudaEigParam mg_eig_param[mg_levels];
 
   if (inv_multigrid) {
+
+    // Set some default values for MG solve types
+    setQudaMgSolveTypes();
+    reliable_delta = 1e-4;
```
mathiaswagner: Do we want that to be hard-coded? Looks like it should be an option @weinbe2 ?

mathiaswagner:
@cpviolator as invertmsrc was likely broken before your pull and we need to work on that a lot I would not consider it being broken a blocker.

tests/staggered_invert_test.cpp
maddyscientist: Fixed

maddyscientist:
Ok, all comments have been addressed.  We can merge this PR once the final set of build tests have completed.  4000 lines of code nuked, nice work @cpviolator 😎

cpviolator:
Careful with that relatable delta change when comparing the staggered MG
output with develop:

https://github.com/lattice/quda/blob/56c5dd93292bc7fff53d5fad2a97015042591395/tests/staggered_multigrid_invert_test.cpp#L433

cpviolator:
*reliable.

maddyscientist:
Sure.  That line should never have been there though I would assert.

PR #979, opened 2020-04-01
Hotfix/various

### LINKED ISSUE #973: sp_pad causing algorithmic anomalies
At the present state, the `QudaInvertParam` member `sp_pad` is set to zero library wide. However, when this value is set to non-zero, it will cause the CG algorithm to behave differently.

To see this, run `./staggered_invert_test` with default parameters. Then notice that the `X[]` variable used to fill `sp_param` is actually never set. Remedy this by replacing

```C++
int tmpint = MAX(X[1] * X[2] * X[3], X[0] * X[2] * X[3]);
tmpint = MAX(tmpint, X[0] * X[1] * X[3]);
tmpint = MAX(tmpint, X[0] * X[1] * X[2]);
inv_param.sp_pad = tmpint;
```
with
```C++
int tmpint = MAX(ydim * zdim * tdim, xdim * zdim * tdim);
tmpint = MAX(tmpint, xdim * ydim * tdim);
tmpint = MAX(tmpint, xdim * ydim * zdim);
inv_param.sp_pad = tmpint;
```
which fills `sp_param`. You will then see that the CG solver converges with a different iteration count

maddyscientist:
Bugfix pull
* The `QudaInvertParam::sp_pad` parameter is now ignored, and all `ColorSpinorField` instances will use zero padding.  This closes #973, which wasn't really a bug, rather a statement that different pad values can lead to different tunings which will slightly change the convergence due to floating point rounding.
* Ensure that unsupported `solution_type` values are caught in `invert_test` and don't just silently fail.

### DIFF
--- a/include/color_spinor_field.h
+++ b/include/color_spinor_field.h
@@ -225,7 +225,7 @@ namespace quda {
     // normally used to create cuda param from a cpu param
     ColorSpinorParam(ColorSpinorParam &cpuParam, QudaInvertParam &inv_param,
                      QudaFieldLocation location = QUDA_CUDA_FIELD_LOCATION) :
-      LatticeFieldParam(cpuParam.nDim, cpuParam.x, inv_param.sp_pad, inv_param.cuda_prec),
+      LatticeFieldParam(cpuParam.nDim, cpuParam.x, 0, inv_param.cuda_prec),
       location(location),
       nColor(cpuParam.nColor),
       nSpin(cpuParam.nSpin),
--- a/tests/invert_test.cpp
+++ b/tests/invert_test.cpp
@@ -589,7 +589,7 @@ int main(int argc, char **argv)
         }
       }
 
-    } else if(inv_param.solution_type == QUDA_MATPC_SOLUTION) {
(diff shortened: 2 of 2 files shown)


weinbe2:
Nothing controversial here, but also arguably nothing urgent. For that reason I'm using this PR as an excuse to run a full `ctest` battery. I'll merge this once that's done.

weinbe2:
Looks good. Merging.

weinbe2 (approved):

cpviolator (commented):
Is the cl_pad still used? There in an instance in interface_quda.cpp where cl_pad is read, so old application code that sets cl_pad may still be affected.

https://github.com/lattice/quda/blob/0e4f3587290e2d443453adaf8ab041fb3b530ac4/lib/interface_quda.cpp#L1046

maddyscientist:
There's a reason why it's still there.  It leaves scope for a possible future optimization with respect to strides memory access patterns in the clover dslash.

Note that since the clover field is never involved in reductions, there is no numerical variance when `cl_pad` is modified.

PR #983, opened 2020-04-17
Add configure time check for simultaneous TEX and FLOAT8

cpviolator:
Adding an addition to the CMakeLists.txt compile time checks. If QUDA_FLOAT8 and QUDA_TEX are simultaneously enabled, CMake will throw an error.


mathiaswagner (approved):

mathiaswagner:
This is one of the little annoyances that I wanted to fix in cmake for some time but whenever it bothered me I did not actually just do it ... so thanks a lot @cpviolator !

PR #984, opened 2020-04-21
Feature/block trlm

cpviolator:
This PR add block solver functionality to the TRLM, based on this paper https://arxiv.org/pdf/1902.02064.pdf. There are some crucial differences to this implementation over the one in the paper:

1. This implementation infers the residuals of the eigenvectors, rather than explicitly measuring the change in the eigenvalues during convergence. It uses the vector norms of the last block of vectors, in analogy to the way the non-blocked TRLM uses the final beta value.

2. Rather than keeping the arrow position fixed (as in the paper) this implementation has a variable arrow position to save on computation. Blocks of vectors, not individual ones, are deemed to be kept, locked, and converged.

3. In order to maintain feature 2., this implementation restricts the user to request values of nEv and nKr that are multiples of the block size.

At the moment it is a separate class because of one code path variation. For the non-blocked algorithm, the Ritz matrix is purely real, and therefore the rotation/compression routines use `Eigen::MatrixXd` and `blas::axpy_(U/L)`. The blocked code path generates a complex Ritz matrix, and so uses `Eigen::MatrixXcd` and `blas::caxpy_(U/L)`. Furthermore, constructing the arrow matrix itself is much more involved in the blocked regime than it is in the non blocked case, so there is also a separate routine for that.

Block sizes of 1 give identical algorithmic behaviour (to MP). However, for a given problem it takes more restarts to converge as the block size increases, in stark contrast to the claims in the paper that fewer restart iterations are required for convergence. This perceived performance loss is likely due to a combination of two things:

1. The loss of orthogonality during the lanczos factorisation. As the block size is increased, more vectors are operated on by the Chebyshev operator simultaneously, and the loss of orthonormality is enhanced.
2. This implementation uses the residual norm of the eigenvector as metric for convergence, which is a tighter constraint than the change in eigenvalue as used in the paper.

What remains to be seen is whether or not a combination of block TRLM and a multi right-hand-side operator will exhibit a faster wall clock convergence over non-blocked TRLM. 

```
./eigensolve_test --sdim 8 --tdim 8 --mass -0.1 --eig-tol 1e-10 --solve-type direct-pc --eig-spectrum SR --verbosity verbose --eig-use-poly-acc true --eig-max-restarts 100 --eig-use-normop true --prec-precondition double --prec double --prec-sloppy double --eig-nEv 48 --eig-nKr 128 --eig-batched-rotate 5 --eig-type blktrlm --eig-block-size 4
```


cpviolator:
Update:

After actually reading the paper, It seems I misunderstood their accounting procedure for iterations. When they use a block size of N, they count the N mat vecs as a single lanczos step. They also have their own version of multi RHS, so they also observe wall clock speed-up.

This implementation exhibits the expected algorithmic speed up (using their accounting) so I'm confident that an efficient multi RHS will give us the expected wall clock speed up.

Using the command:
```
mpirun -np 2 ./eigensolve_test --sdim 16 --tdim 32 --gridsize 1 1 1 2 --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --mass -0.42 --eig-tol 1e-10 --anisotropy 2.38 --solve-type direct-pc --eig-poly-deg 100 --eig-amin 7e-4 --eig-use-poly-acc true --eig-spectrum SR --verbosity verbose --eig-use-normop true --prec-precondition single --prec double --prec-sloppy half --eig-nEv 48 --eig-nKr 192 --eig-batched-rotate 0 --eig-type blktrlm --eig-block-size ${BLOCK} --eig-max-restarts 100
```

| Block Size | Blocked Lanczos Steps | Actual Lanczos Steps | Wall Clock 
| ---- | ---- | ---- | ---- |
| 1 | 192 | 192 | 19.691888|
| 2 | 126 | 252 | 25.566567|
| 3 | 91 | 273 |27.402543 |
| 4 | 72 | 288 |28.373102 |
| 6 | 48 | 288 |28.781603 |
| 8 | 36 | 288 | 28.105595|
| 12 | 32 | 384 |38.734093 |
| 16 | 28 | 448 | 44.646723|
| 24 | 23 | 552 | 57.735302|

By this data, we would need to see a 3x speed up for 24 multi RHS to break even.

lib/eig_block_trlm.cpp
```diff
@@ -0,0 +1,771 @@
...
+      break;
+    case QUDA_HALF_PRECISION:
+      epsilon = 2e-3;
+      if (getVerbosity() >= QUDA_SUMMARIZE) printfQuda("Running Eigensolver in half precision\n");
+      break;
+    case QUDA_QUARTER_PRECISION:
+      epsilon = 5e-2;
+      if (getVerbosity() >= QUDA_SUMMARIZE) printfQuda("Running Eigensolver in quarter precision\n");
+      break;
+    default: errorQuda("Invalid precision %d", prec);
```
maddyscientist: Can we reuse the code already in regular Lanczos here?  Perhaps create a `setEpsilon` helper in the parent class?

lib/eig_block_trlm.cpp
```diff
@@ -0,0 +1,771 @@
...
+        idx = b * block_size + c;
+        jth_block[idx] = -1.0 * block_alpha[arrow_offset + idx];
+      }
+    }
+
+    // r = r - a_j * v_j
+    blas::caxpy(jth_block, vecs_ptr, r);
+
+    // Orthogonalise R[0:block_size] against the Krylov space V[0:j + block_size]
+    for (int k = 0; k < 1; k++) blockOrthogonalize(v, r, j + block_size);
```
maddyscientist, line 294: Is this loop here for potential future exploration with multiple passes of orthogonalization?

lib/eig_block_trlm.cpp
```diff
@@ -0,0 +1,771 @@
...
+
+    // QR decomposition via modified Gram Scmidt
```
maddyscientist: Schmidt

lib/eig_block_trlm.cpp
```diff
@@ -0,0 +1,771 @@
...
+          kSpace.push_back(ColorSpinorField::Create(csParamClone));
+        }
+      }
+
+      profile.TPSTART(QUDA_PROFILE_EIGEN);
+      MatrixXcd mat = MatrixXcd::Zero(dim, iter_keep);
+      for (int j = 0; j < iter_keep; j++)
+        for (int i = 0; i < dim; i++) mat(i, j) = block_ritz_mat[j * dim + i];
+
+      FullPivLU<MatrixXcd> matLU(mat);
```
maddyscientist, line 541: Is full pivoting required here?  I believe the most stable solvers in Eigen are the SVD based ones.  Have you done any exploration of the best to use?

include/externals/jitify.hpp
```diff
@@ -118,9 +118,9 @@
 
 // For use by get_current_executable_path().
 #ifdef __linux__
-#include <linux/limits.h>  // For PATH_MAX
+#include <linux/limits.h> // For PATH_MAX
 
-#include <cstdlib>  // For realpath
+#include <cstdlib> // For realpath
```
maddyscientist: All changes to Jitify need to be reverted.  Jitify is an external inclusion and should never be clang-formatted (it has its own custom clang-format file which is distinct from QUDA's).

maddyscientist:
I've done a visual review of this PR and mostly looks good.  I'll proceed with hands-on testing once  the above issues are taken care of.

lib/eig_block_trlm.cpp
cpviolator: I went ahead and made a bunch of helpers in the parent class. I think we could do away with having two separate classes and integrate the block solver aspects into the TRLM class.

lib/eig_block_trlm.cpp
cpviolator: I think, during the time I was exploring Implicitly Restarted Lanczos, I had to crank this up to maintain orthonormality. TRLM is very stable though, so this can go.

lib/eig_block_trlm.cpp
cpviolator: I can't see anything in the Eigen documentation about doing LU via SVD. Full pivoting is done because we need not only the P permutation matrix, but also the Q permutation matrix. Partial pivoting provides P only:

https://eigen.tuxfamily.org/dox/group__LU__Module.html

include/externals/jitify.hpp
cpviolator: I have no idea how that happened. Will revert.

lib/eig_block_trlm.cpp
maddyscientist: Let's keep them separate classes for now and just move a few helpers into the parent.  In the future where we hopefully have Arnoldi, block Arnoldi, IRLM, etc. (a girl can dream? 😉) the right level of abstraction might be different then that is now.

lib/eig_block_trlm.cpp
maddyscientist: Gotcha, my misunderstanding.

lib/eig_block_trlm.cpp
cpviolator: On the Arnoldi, I think I know where I went wrong. The algorithm (as written in Arpack) will check to see how orthogonal the next Arnoldi vector is with respect to the current factorisation. If it is not sufficiently orthogonal, it is replaced with a random vector.

When I was developing, I was using a very small operator and the above never occurred in Arpack. Recently, I ran old faithful (L=16, T=64) through Arpack and the above happened several times. I will dig up my old Arnoldi code (which did work for small lattices) and implement this step.

maddyscientist: 🤞 🙏 🙇‍♀️ 👍 💯

lib/eig_block_trlm.cpp
maddyscientist: You could always expose the number of orthogonalization sweeps as a parameter.  We've done in this multigrid.  Might be nice to leave this in there and just default the value to 1.

cpviolator:
Script to use the solver:

```bash
#!/bin/bash

PREC=double
PRECON=single
SLOPPY=half

export QUDA_RESOURCE_PATH=.

DEG=100

# single solve
COMMAND="mpirun -np 2 ./eigensolve_test --sdim 16 --tdim 32 --gridsize 1 1 1 2  --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --mass -0.42 \
		--eig-tol 1e-10 --anisotropy 2.38 --solve-type direct-pc --eig-poly-deg ${DEG} --eig-amin 7e-4 --eig-use-poly-acc true --eig-spectrum SR --verbosity verbose \
		--eig-use-normop true --prec-precondition ${PRECON} --prec ${PREC} --prec-sloppy ${SLOPPY} --eig-nEv 48 --eig-nKr 192 --eig-batched-rotate 5 \
                --eig-type trlm --eig-block-size 10 --eig-max-restarts 100 --eig-require-convergence false "

echo $COMMAND
$COMMAND 

# block solve
for BLOCK in 1 2 3 4 6 8 12 16 24 ; do
    COMMAND="mpirun -np 2 ./eigensolve_test --sdim 16 --tdim 32 --gridsize 1 1 1 2  --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --mass -0.42 \
		--eig-tol 1e-10 --anisotropy 2.38 --solve-type direct-pc --eig-poly-deg ${DEG} --eig-amin 7e-4 --eig-use-poly-acc true --eig-spectrum SR --verbosity verbose \
		--eig-use-normop true --prec-precondition ${PRECON} --prec ${PREC} --prec-sloppy ${SLOPPY} --eig-nEv 48 --eig-nKr 192 --eig-batched-rotate 5 --eig-type blktrlm \
		--eig-block-size ${BLOCK} --eig-max-restarts 100 --eig-require-convergence false "
    
    echo $COMMAND
    $COMMAND 
    wait
done
```

maddyscientist (approved):

maddyscientist:
This is a fantastic addition, thank you @cpviolator.  I've done some testing, and this looks very promising.  The benefit will be very apparent for the coarse-operator deflation when have multi-RHS support enabled.

This is of course a perfect motivator to get the multi-RHS dslash framework done.  Beyond that, as noted already [here](https://github.com/lattice/quda/wiki/Thick-Restarted-eigensolver#block-thick-restarted-eigensolver), there are a number of further optimizations we can make to this to improve this but that will be for another day.  Thus approving this pull request as is.

maddyscientist:
Ok, just after I approved the pull, I happened to realize that there is a very obvious optimization can be done in the `orthoCheck` routine to avoid a quadratic number of inner product kernels being called.  This can be trivially replaced with a single `hDotProduct` kernel call to bring this back to O(1).  I'll push this fix.

cpviolator:
This works just as it did before, all good.

PR #990, opened 2020-05-05
Feature/dwf io

maddyscientist:
This PR is focussed on improving the I/O support in QUDA for quark fields:
* Added support for 5-d field I/O, e.g., domain-wall and Möbius
* Fixed single-parity I/O support: the previous functions in layout_hyper implicitly assumed that all fields were dual parity and as such, I/O broke when changing the node geometry between saving and loading. 
* Restored single-parity I/O default when loading / storing sets of single-parity fields.  
* Inflation to a full field can be enabled with the `set_io_parity_inflation_quda` function - this is set in the MILC interface to preserve the prior behaviour being preserved which is necessary for MILC interoperability  
* Removal of redundant code
* The I/O routines have been moved from the Eigensolver class into their own class `VectorIO` located in vector_io.cpp

This PR has already been tested on Summit to verify that Möbius eigenvector I/O is working as expected (thanks @cpviolator).  In addition, I have tested both 4-d and 5-d (4-d and 5-d preconditioned) eigen-vector I/O across a range of different geometries and see perfect consistency in the eigenvalues calculated.


weinbe2:
Inflation to full parity needs be exposed as a command line argument

The use case of using the standalone `test` executables to generate and save eigenvectors was why I introduced support for inflating fields in the first place. This was necessary for one of @detar 's workflows. Fixing the issues with saving single parity fields is orthogonal to this use case.

weinbe2:
Reference from the original PR: https://github.com/lattice/quda/pull/914#issuecomment-552017709

cpviolator (approved):
Performed a load of jobs with different geometry splits on summit and all worked perfectly. Thank you!

maddyscientist:
I have addressed @weinbe2's comment
* The option to inflate single-parity fields has been exposed through the `QudaEigParam::io_parity_inflate` option which can be used to enable I/O parity inflation
* I have added the option `--eig-io-parity-inflate` to the command line
```
 --eig-io-parity-inflate BOOL
         Whether to inflate single-parity eigenvectors onto dual parity full fields for file I/O (default = false)
```
* The inflation option is enabled when QUDA's eignsolvers are called using the MILC interface
* I have removed the `set_io_parity_inflation_quda` function since this is now redundant.

Thus the default behaviour is now as it was prior to #914, and to restore behaviour since #914 one needs to add the `--eig-io-parity-inflate true` flag when running QUDA's tests (@detar as an FYI).

weinbe2:
I've performed a visual code review and I'm very happy with this PR. I'll focus my remaining checks on testing this with the WIP MILC HISQ MG interface (where the linker errors originally reared their head). Assuming I don't hit any problems there I'll hit merge.

weinbe2 (approved):
All's good on the MILC front. Hitting merge now.

PR #992, opened 2020-05-13
Fix staggered_invert_test with naive staggered fermions

weinbe2:
This quick hotfix re-arranges `staggered_invert_test` to compute the plaquette before formally loading the fat and long gauge links. This is necessary because the routines to compute the plaquette for the staggered links re-upload the gauge fields. 

We would've picked up on this one if anyone had recently run `./staggered_invert_test --dslash-type staggered`.


maddyscientist (approved):
Good catch Evan.

PR #993, opened 2020-05-15
Feature/ampere

maddyscientist:
* Add sm_80 support to building QUDA
* `Tunable::maxSharedMemoryPerBlock` and `Tunable::maxBlocksPerSM` now query the device rather than use hardcoding
* Work around for building QUDA with CUDA 11 until we remove CUB from QUDA 
* Updates the default architecture to Volta (sm_70)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,8 +95,8 @@ message(STATUS "")
 
 set(QUDA_GPU_ARCH
     ${QUDA_DEFAULT_GPU_ARCH}
-    CACHE STRING "set the GPU architecture (sm_20, sm_21, sm_30, sm_35, sm_37, sm_50, sm_52, sm_60, sm_70, sm_75 sm_80)")
-set_property(CACHE QUDA_GPU_ARCH PROPERTY STRINGS sm_20 sm_21 sm_30 sm_35 sm_37 sm_50 sm_52 sm_60 sm_70 sm_75 sm_80)
+    CACHE STRING "set the GPU architecture (sm_35, sm_37, sm_60, sm_70, sm_80)")
+set_property(CACHE QUDA_GPU_ARCH PROPERTY STRINGS sm_35 sm_37 sm_60 sm_70 sm_80)
 # build options
 set(QUDA_DIRAC_WILSON ON CACHE BOOL "build Wilson Dirac operators")
 set(QUDA_DIRAC_CLOVER ON CACHE BOOL "build clover Dirac operators")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,7 +86,7 @@ message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 if(DEFINED ENV{QUDA_GPU_ARCH})
   set(QUDA_DEFAULT_GPU_ARCH $ENV{QUDA_GPU_ARCH})
 else()
-  set(QUDA_DEFAULT_GPU_ARCH sm_35)
+  set(QUDA_DEFAULT_GPU_ARCH sm_70)
 endif()
 if(NOT QUDA_GPU_ARCH)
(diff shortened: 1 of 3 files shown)


include/tune_quda.h
```diff
@@ -152,21 +152,25 @@ namespace quda {
...
+      if (!tuneGridDim())
 	param.grid = dim3((minThreads()+param.block.x-1)/param.block.x, 1, 1);
 
       return ret;
     }
 
     /**
-     * @brief For some reason this can't be queried from the device
-     * properties, so here we set set this.  Based on Table 14 of the
-     * CUDA Programming Guide 10.0 (Technical Specifications per
```
weinbe2, line 164: We should leave this note in as a reference for the `#else` path

include/tune_quda.h
```diff
@@ -204,14 +209,16 @@ namespace quda {
     }
 
     /**
-     * @brief This can't be correctly queried in CUDA for all
-     * architectures so here we set set this.  Based on Table 14 of
-     * the CUDA Programming Guide 10.0 (Technical Specifications per
-     * Compute Capability).
+     * @brief Returns the maximum dynamics shared memory per block.
```
weinbe2: Same comment as above

maddyscientist: That was intentional, but I've restored the reference in a new comment in the else path

include/tune_quda.h
maddyscientist: That was intentional, but I've restored the reference in a new comment in the else path

include/tune_quda.h
```diff
@@ -204,14 +210,17 @@ namespace quda {
...
      * @return The maximum dynamic shared memory to CUDA thread block
      */
     unsigned int maxDynamicSharedBytesPerBlock() const
     {
+#if CUDA_VERSION >= 9000
+      static int max_shared_bytes = 0;
+      if (!max_shared_bytes) cudaDeviceGetAttribute(&max_shared_bytes, cudaDevAttrMaxSharedMemoryPerBlockOptin, comm_gpuid());
+      return max_shared_bytes;
+#else
+      // these variables are taken from Table 14 of the CUDA 10.2 prgramming guide
```
mathiaswagner, line 223: do we need to have Volta hardcoded if we query it for CUDA 9+ ?

CMakeLists.txt
```diff
@@ -95,8 +95,8 @@ message(STATUS "")
 
 set(QUDA_GPU_ARCH
     ${QUDA_DEFAULT_GPU_ARCH}
-    CACHE STRING "set the GPU architecture (sm_20, sm_21, sm_30, sm_35, sm_37, sm_50, sm_52, sm_60, sm_70, sm_75)")
-set_property(CACHE QUDA_GPU_ARCH PROPERTY STRINGS sm_20 sm_21 sm_30 sm_35 sm_37 sm_50 sm_52 sm_60 sm_70 sm_75)
+    CACHE STRING "set the GPU architecture (sm_20, sm_21, sm_30, sm_35, sm_37, sm_50, sm_52, sm_60, sm_70, sm_75 sm_80)")
+set_property(CACHE QUDA_GPU_ARCH PROPERTY STRINGS sm_20 sm_21 sm_30 sm_35 sm_37 sm_50 sm_52 sm_60 sm_70 sm_75 sm_80)
```
mathiaswagner: We should remove less common hardware. Suggest is to keep the HPC arch Kepler, Pascal, Volta, Ampere.

mathiaswagner (changes requested):
looks good apart from minor points. I still want to fire up a manual compile test.

CMakeLists.txt
maddyscientist: Done

include/tune_quda.h
maddyscientist: Removed Volta

mathiaswagner (approved):

PR #995, opened 2020-05-16
CMake cleanup

mathiaswagner:
This is was intended to be a small cleanup, only fixing issues with cmake 3.17 and taking first steps into using `FindCUDAToolkit`.

It has now evolved into a significant overhaul of cmake. 

I tried to 
* remove all global / directory scope `include_directories` / `add_definitions` and make them dependencies of QUDA
* clean up the use of CMAKE_CUDA Flags and use `target_compile_defintions`

TODO:

- [x] ~create imported targets for USQCD libs~ (will look into this later, maybe we can switch to cmake versions of this), see #997
- [x] create imported target for MAGMA and add compile definitions to it
- [x] do a final cleaning pass and make sure things are somewhat ordered 
- [x] check which compile definitions can be made PRIVATE

Currently this PR is a still draft only but good to have build tests running for a wider range of cmake versions.


mathiaswagner:
@maddyscientist @weinbe2 @hummingtree @bjoo @cpviolator 

Since a change in the build system is quite intrusive and all things might break that are not covered with my testing/jenkins I appreciate if you can (as time permits) fire up a build and let me know if things still work for you.

This is as time permits and I will all ask you for a more formal review later when this is beyond draft status.

mathiaswagner:
@alexstrel can you please test if MAGMA still works?

hummingtree:
I have tested with `cmake` `3.16.2` and `3.17.2` and they work fine for me. Thanks @mathiaswagner for the work.

cmake/FindCUDAWrapper.cmake
```diff
@@ -8,51 +8,51 @@
...
+# # wrapper for cuda_add_library
+# # Issues:
+# #
+# function(cuda_add_executable)
+#   add_executable(${ARGV})
+#   target_include_directories(${ARGV0} PUBLIC
+#                              ${CUDA_INCLUDE_DIRS})
+#   target_link_libraries(${ARGV0}  ${CUDA_LIBRARIES})
+#   set_target_properties(${ARGV0} PROPERTIES LINKER_LANGUAGE CUDA)
+# endfunction()
```
maddyscientist: The above is all commented out.  Any reason not to delete?

tests/CMakeLists.txt
```diff
@@ -1,24 +1,22 @@
-#TODO: we might be able to clean that up a bit and have the include directories propagate with the libs
-include_directories(. googletest/include googletest ../utils ../host_reference)
+# TODO: we might be able to clean that up a bit and have the include directories propagate with the libs
+# include_directories(.  ../utils ../host_reference)
```
maddyscientist: Delete this?

maddyscientist (approved):
This is building great for me.  Beyond a couple of a trivial suggestions regarding deleting some commented out lines, this is looking good.

tests/CMakeLists.txt
mathiaswagner: All deleted. Had forgotten to remove that.

maddyscientist:
Looks like you have a final TODO list item: ` check which compile definitions can be made PRIVATE`.  Is that something yet to be done prior to merging?

mathiaswagner:
Just forgot to check that off. I think most a PRIVATE. There is certainly room for further improvement e.g. if we make sure that some headers do not depend on defines and we could improve (re)build times when defines that only affect a single or few source files are only applied to these (see what id done for backwards ).
But I think that is more a code cleanup and header reorganization which is beyond this PR.

CMakeLists.txt
```diff
@@ -228,65 +265,26 @@ mark_as_advanced(QUDA_INTERFACE_NVTX)
...
-  if(NOT EIGEN_FOUND)
-    message(
-      FATAL_ERROR
-        "QUDA requires Eigen (http://eigen.tuxfamily.org). Please either set EIGEN_INCLUDE_DIRS to path to eigen3 include directory, e.g. /usr/local/include/eigen3 or set QUDA_DOWNLOAD_EIGEN to ON to enable automatic download of the necessary components."
-      )
-  endif()
+if(QUDA_MPI AND QUDA_QMP)
+  message(
+    WARNING
+      "Specifying QUDA_QMP and QUDA_MPI might result in undefined behavior. If you intend to use QMP set QUDA_MPI=OFF.")
```
weinbe2, line 284: Should we just make this an error? Is there ever a reason why someone would want both QMP and MPI defined?

mathiaswagner: Good idea. In fact if you enable both compilation will fail so better to error out.

weinbe2 (approved):
Pending the checks with Chroma and CPS this looks good to me. I tested this with MILC and there weren't any surprises.

mathiaswagner:
@hummingtree can you try CPS? Apart from that I would let the builds finish and then merge this.

PR #999, opened 2020-05-23
Feature/quda targets

cpviolator:
This PR is a more broad attempt to incorporate different targets in QUDA. It accomplishes 4 things:

1. Make CMake changes to expose different target options
2. Move `malloc.cpp` and `tune.cpp` into targets/CUDA
3. Abstract cudaStreamSynchronize to qudaStreamSynchronize
4. Define a general data type `qudaStream_t` in the library and define it as `cudaStream_t` at compile time.

The general idea to incrementally add more of these types of changes.


maddyscientist (approved):
Thanks for this short and sharp pull @cpviolator.  Nothing controversial and all looks good.  I've changed one thing, and that was that I prefer we always have lower-case directories, e.g., `targets/cuda` instead of `targets/CUDA`.

Adding @weinbe2 and @mathiaswagner to also sign off.

mathiaswagner:
Before doing a full review: We should get #995 merged first and then update the changes to CMake here accordingly.

CMakeLists.txt
```diff
@@ -665,6 +682,16 @@ endif()
 # set which precisions to enable
 add_definitions(-DQUDA_RECONSTRUCT=${QUDA_RECONSTRUCT})
 
+# Set target preprocessor flags
+if(${QUDA_TARGET_TYPE} STREQUAL "CUDA")
+  set(QUDA_TARGET_DIR "cuda")
+  add_definitions(-DQUDA_TARGET_QUDA)
```
mathiaswagner: This needs to be updated after #995 is merged.

alexstrel:
@maddyscientist @cpviolator  would it be preferable to move `comm_common.cpp` to `targets/cuda`? It relies on cuda-specific stuff that might be better to isolate.

CMakeLists.txt
```diff
@@ -665,6 +682,16 @@ endif()
...
+endif()
+if(${QUDA_TARGET_TYPE} STREQUAL "HIP")
+  set(QUDA_TARGET_DIR "hip")
+  add_definitions(-DQUDA_TARGET_HIP)
```
mathiaswagner: I would only want to have that definition on specific source files. Most of QUDA should not care about this and not having it available reduces potential abuse.

include/quda_cuda_api.h
```diff
@@ -5,6 +5,8 @@
 #include <cuda_runtime.h>
 #include <quda_cuda_api.h>
 
+#define qudaStream_t cudaStream_t
```
mathiaswagner: does this need to be a define. Can we get away with a `using` ?

lib/CMakeLists.txt
```diff
@@ -9,7 +9,9 @@ set (QUDA_OBJS
   eigensolve_quda.cpp quda_arpack_interface.cpp
   multigrid.cpp transfer.cpp block_orthogonalize.cu inv_bicgstab_quda.cpp
   prolongator.cu restrictor.cu staggered_prolong_restrict.cu
-  gauge_phase.cu timer.cpp malloc.cpp
+  gauge_phase.cu timer.cpp
+  targets/${QUDA_TARGET_DIR}/malloc.cpp
```
mathiaswagner: should also be updated with `target_sources`

mathiaswagner (dismissed):
mostly cmake changes after merging #995.

include/quda_cuda_api.h
maddyscientist: This was for expediency for this PR only.  I'm not sure if `using` works, but I can try.

cpviolator: @maddyscientist It compiles with `using`, I'll do a run time check in a minute.

cpviolator: And it runs.

cpviolator:
@alexstrel for sure, but this PR is just for a few simple things, and all subsequent PRs will be incremental. This is to avoid any undetected runtime errors creeping in, as happened in the last attempt.

lib/CMakeLists.txt
cpviolator: Is the `PUBLIC` keyword correct here: https://github.com/lattice/quda/commit/087daceb0cd9a69db9a7660d7484d36167869e21

CMakeLists.txt
cpviolator: https://github.com/lattice/quda/commit/087daceb0cd9a69db9a7660d7484d36167869e21

cpviolator:
Picking up this error after merging cmake317:

```
[ 32%] Building CUDA object lib/CMakeFiles/quda.dir/dslash_coarse.cu.o
nvcc fatal   : redefinition of argument 'ftz'
make[2]: *** [lib/CMakeFiles/quda.dir/build.make:63: lib/CMakeFiles/quda.dir/dslash_coarse.cu.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1040: lib/CMakeFiles/quda.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

mathiaswagner:
Thanks @cpviolator. I'll give the cmake another review pass with the latest changes in but everything else looks good.

mathiaswagner:
I moved the target files into a separate CMakeLists and added it through an object library. Idea is to be able to move target specific compile flags there and keep them contained in there as much as possible.

Sure there is more room for optimization as we go done that route. But I think this is my preferred idea right now.

PR #1000, opened 2020-05-26
fixing a error in the construction of the MatDagger for Mobius fermions

walkloud:
Balint found an error in the construction of the Dagger operator in `DiracMobius::M` when building full MAT operators (not preconditioned) - we added a switch to properly construct the dagger operator.  We verified the fixed operator agrees with that from Chroma in the application of the full (both checkerboard parities) Möbius Matrix.

### DIFF
--- a/host_reference/domain_wall_dslash_reference.cpp
+++ b/host_reference/domain_wall_dslash_reference.cpp
@@ -808,18 +808,30 @@ void mdw_mat(void *out, void **gauge, void *in, double _Complex *kappa_b, double
   void *outEven = out;
   void *outOdd = (char *)out + V5h * spinor_site_size * precision;
 
-  mdw_dslash_4_pre(tmp, gauge, inEven, 0, dagger, precision, gauge_param, mferm, b5, c5, true);
-  dslash_4_4d(outOdd, gauge, tmp, 1, dagger, precision, gauge_param, mferm);
-  mdw_dslash_5(tmp, gauge, inOdd, 1, dagger, precision, gauge_param, mferm, kappa5, true);
+  if (!dagger) {
+    mdw_dslash_4_pre(tmp, gauge, inEven, 0, dagger, precision, gauge_param, mferm, b5, c5, true);
+    dslash_4_4d(outOdd, gauge, tmp, 1, dagger, precision, gauge_param, mferm);
+    mdw_dslash_5(tmp, gauge, inOdd, 1, dagger, precision, gauge_param, mferm, kappa5, true);
+  } else {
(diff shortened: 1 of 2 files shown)


mathiaswagner:

bjoo:
👍

weinbe2:
The better way to do this is to define a separate `Mdag`:

`void Dirac::Mdag(ColorSpinorField &out, const ColorSpinorField &in) const`

The parent implementation in lib/dirac.cpp switches dagger <-> non-dagger, calls `M`, and reverts, which is why this fix technically works.

weinbe2:
But excuse my manners---thank you for catching this 🙂

walkloud:
I'll let you guys fix it how you want to - meaning - it is still a bit above me to "do the better way" so I'll let someone who knows get it right

bjoo:
Yes, originally I looked for a Mdag() but didn't find it, so I figured it was implemented by
a default which just called M() with arg.dagger set appropriately. This works for regular 4D DWF where the op is just an addition of two ops and there is no need to reverse order of applications when applying the dagger.

So since I thought Mdag was in a default this if-else statement seemed minimally invasive and quick to test. Perhaps having a separate Mdag would be cleaner and trivial to add...

Best
B
Sent from Blue<http://www.bluemail.me/r?b=15860>
On May 26, 2020, at 7:33 PM, Evan Weinberg <notifications@github.com<mailto:notifications@github.com>> wrote:

The better way to do this is to define a separate Mdag:

void Dirac::Mdag(ColorSpinorField &out, const ColorSpinorField &in) const

The parent implementation in lib/dirac.cpp switches dagger <-> non-dagger, calls M, and reverts, which is why this fix technically works.

maddyscientist:
I should add that the style as André has written is consistent with other Dirac operators, where the dagger status is queried in the `M` operator and the appropriate action taken, as opposed to deriving and specializing the `Mdag` operator.  So I might suggest @weinbe2 that the present PR is fine as is, and if it was to be updated with `Mdag` specialization then the same style should be applied to all Dirac classes.  Just my $0.02 😜

mathiaswagner:
@Jenkins test this please.

weinbe2:
@maddyscientist duly noted! Let's get it merged in and I'll make those changes separately.

maddyscientist:
I have merged in the develop branch which will allow all the build tests to finish.  We'll merge this in once they complete.  Thanks @walkloud!

maddyscientist (approved):

maddyscientist:
@Jenkins test this please

PR #1001, opened 2020-05-26
Allow for relaxed orthonormalisation constraints for low prec solves, minor improvements to stdout

cpviolator:
Quick PR to relax the orhonormalisation constraints on the eigensolver if low prec solves are requested. Also improved some stdout.


maddyscientist:
Looks good, but needs a clang-format application.

maddyscientist (approved):

PR #1002, opened 2020-05-29
MILC interface for HISQ MG

weinbe2:
This PR adds a MILC interface for calling HISQ multigrid from within MILC. This PR is complimentary to https://github.com/weinbe2/milc_qcd/tree/feature/staggMG , for which I'll file a separate PR once the interface changes I added in this PR are approved (FYI @detar). Note that there are the appropriate `#ifdef` guards are in MILC such that this version can be safely linked against older versions of QUDA.

The main design choice of this PR is to define an interface where MG-specific parameters are passed in via an input file. This decouples the ever-expanding universe of MG parameters from changing the signature of the MILC interface functions themselves. For the time being I've included support for a "minimal" set of MG parameters which I've personally most relevant for tuning HISQ MG. These will be documented in the wiki momentarily. Adding support for more parameters is largely trivial.

This PR adds support for a "thin" multigrid update where only fine level Dirac matrices are updated with the latest gauge fields/clover fields/fat and long links as appropriate, as well as the mass, without rebuilding the entire coarse operator. Edit: this PR adds support for a thin update via a new field to `QudaMultigridParam`, `thin_update_only`, with a default value of `QUDA_BOOLEAN_FALSE` as to not break existing behavior.

This PR also adds rebuilding the clover field to `multigrid_evolve_test` (initially added as a proxy for debugging issues within thin updates).

This PR includes the changes in feature/access_rank. 

TBD work, some of which may or may not be considered blocking:

- [x] Approval of the interface itself (reason why this is a draft PR for now)
- [x] Creation of a wiki page on the interface, required changes to MILC compile defines, changes to MILC input files, the format and parameters of MG input files
- [x] Formal perf tests of enabling/disabling `QUDA_ENABLE_P2P_MAX_ACCESS_RANK` re:feature/access_rank
- [ ] Merge of weinbe2:feature/staggMG into mainline MILC (not strictly necessary, no interfaces unrelated to MG were changed)

The merge of this PR enables/begs some future changes for which I'll file issues/feature requests (#1003, #1004, #1005).


include/quda_milc_interface.h
```diff
@@ -310,6 +310,62 @@ extern "C" {
...
+                    const void *const milc_longlink, void *mg_pack_ptr, int mg_rebuild_type, void *source,
+                    void *solution, double *const final_residual, double *const final_fermilab_residual, int *num_iters);
+
+  /**
+   * Clean up a staggered/HISQ multigrid object, freeing all internal
+   * fields and otherwise allocated memory.
+   *
+   * @param mg_pack_ptr Void pointer mapping to the multigrid structure returned by qudaSetupMultigrid
+   */
+  void qudaCleanupMultigrid(void *mg_pack_ptr);
```
maddyscientist: Not sure I like this name "cleanup".  I'd suggest changing to `qudaMultigridDestroy`, and above setup to `qudaMultigridCreate`.

lib/multigrid.cpp
```diff
@@ -88,8 +88,9 @@ namespace quda
             }
           }
           if (param.mg_global.num_setup_iter[param.level] > 0) {
-            if (strcmp(param.mg_global.vec_infile[param.level], "")
-                != 0) { // only load if infile is defined and not computing
+            if (param.mg_global.vec_load[param.level] == QUDA_BOOLEAN_TRUE
```
maddyscientist, line 91: Was this a bug fix?

lib/milc_interface.cpp
```diff
@@ -1184,6 +1191,895 @@ void qudaEigCGInvert(int external_precision, int quda_precision, double mass, Qu
...
+
+  QudaParity local_parity = inv_args.evenodd; // ignored, just needed to set some defaults
+  const double reliable_delta = 1e-4;
+
+  setInvertParams(localDim, host_precision, device_precision, device_precision_sloppy, mass, target_residual,
+                  target_fermilab_residual, inv_args.max_iter, reliable_delta, local_parity, verbosity,
+                  QUDA_GCR_INVERTER, &invertParam);
+
+  // Override a few things for WAR
+#if 0
```
maddyscientist: What are these lines about?  What's the WAR needed for?

include/quda.h
```diff
@@ -1013,6 +1013,16 @@ extern "C" {
...
+  /**
+   * @brief Performs a thin update to the multigrid preconditioner for the new gauge / clover field.
+   * This update only recreates the Dirac classes and does not rebuild the entire coarse setup.
+   * Assumes gauge parameters haven't changed.
+   * @param mg_instance Pointer to instance of multigrid_solver
+   * @param param Contains all metadata regarding host and device
+   * storage and solver parameters
+   */
+  void thinUpdateMultigridQuda(void *mg_instance, QudaMultigridParam *param);
+
```
maddyscientist: I'm not sure I understand why we need a new interface for this functionality.  Why can't this be parameterized and deployed in the already existing `updateMultigridQuda` interface function?  This approach is much more intrusive and would require more extensive changes in case, for example was wanted to deploy a "thin" update in other applications.

maddyscientist:
Had my first review at this.  Beyond my inline comments
* There's a lot of code in milc_interface.cpp devoted to loading in parameters.  I thought the CLI11 had this functionality already.  Is there a reason why this can't be deployed here?

tests/multigrid_evolve_test.cpp
```diff
@@ -395,6 +407,19 @@ int main(int argc, char **argv)
         if (inv_multigrid) mg_param.invert_param->mu = inv_param.mu;
       }
 
+      // as needed to bake in mu
+      if (dslash_type == QUDA_TWISTED_CLOVER_DSLASH) {
```
maddyscientist: Why is the clover field only being recomputed for twisted-clover here and not regular clover?  Since this loop will be modifying the mass and not just the twist, I think we need to recompute the clover field.

include/quda_milc_interface.h
weinbe2: Done.

lib/multigrid.cpp
weinbe2: Technically, though there was no code path where it was a problem in practice: `*.vec_load[*]` was actually never checked, and it was ultimately redundant of `*.vec_infile[*]` being non-null.

Previous behavior was by default `*.vec_load[*]` was unset, and was only set to `QUDA_BOOLEAN_TRUE` if `*.vec_infile[*]`was set to something non-null. This PR changes the behavior to initialize `*.vec_load[*]` to `QUDA_BOOLEAN_FALSE` and only set it to true as appropriate.

Another option is to just remove `*.vec_load[*]`, again because it's redundant of `*.vec_infile[*]` being non-null.

lib/milc_interface.cpp
weinbe2: That can be cleaned up. This was from the earlier days of interface checking where I just wanted to make sure I was setting up full parity solves correctly, independent of what the solver was.

include/quda.h
weinbe2: Two part reply:

1. The core of the issue here is identical to what we discussed on Slack for #1018, extending an existing interface with new functionality, where @mathiaswagner introduced the tentative solution of adding `qudaGaugeForcePhased` alongside `qudaGaugeForce`. The issues we discussed were:
* Not wanting to modify the signature of `qudaGaugeForce` to avoid breaking existing interfaces
* C not supporting default arguments, so we couldn't just add an int/enum/etc corresponding to phased vs unphased (in that context), or thin vs full here.
* Not wanting to modify a data structure passed in to avoid breaking existing interfaces (the issue there more being that structures aren't guaranteed to be zero-initialized, so you risk ending up with undefined behavior).

As I said, that solution was tentative, so there's room for further discussion that would apply in both places. If we're willing to break long-standing interfaces (`qudaGaugeForce` in #1018, `updateMultigridQuda` here), then things become a lot easier, but I don't like that approach.

2. In the sense that this is MG-specific, I'm certainly open to modifying this signature to be more future-proof.

weinbe2:
> Had my first review at this. Beyond my inline comments
> 
>     * There's a lot of code in milc_interface.cpp devoted to loading in parameters.  I thought the CLI11 had this functionality already.  Is there a reason why this can't be deployed here?

I created #1004 almost a month ago to track this. tl;dr what I've implemented was a quick, "just works" solution because I didn't want to hold up opening this PR on thinking about a parser.

tests/multigrid_evolve_test.cpp
weinbe2: Good catch, fixed.

include/quda.h
maddyscientist: I still don't understand this.  Why can't any objectives be implemented through parameters in `QudaMultigridParam`?  The issue is different from #1018 since this is a QUDA interface, not a MILC-QUDA interface.

maddyscientist: We can simply ensure that the default behaviour matches the present `updateMultigridQuda`.  Perhaps I am missing some subtlety? (or not so subtlety 😜 )

weinbe2: Ah, fair point. I can add it to `QudaMultigridParam`.

lib/comm_common.cpp
```diff
@@ -164,8 +165,10 @@ static bool intranode_enabled[2][4] = { {false,false,false,false},
 static bool peer2peer_present = false;
 
 /** by default enable both copy engines and load/store access */
-static int enable_peer_to_peer = 3; 
+static int enable_peer_to_peer = 3;
 
+/** sets whether we cap which peers can use peer-to-peer */
```
mathiaswagner, line 170: why is this mulled in here? Guess we need to handle #1024 first then?

include/quda.h
weinbe2: Schematically done, though it may not compile yet ;)

lib/comm_common.cpp
weinbe2: ^ @mathiaswagner yes

maddyscientist:
> > Had my first review at this. Beyond my inline comments
> > ```
> > * There's a lot of code in milc_interface.cpp devoted to loading in parameters.  I thought the CLI11 had this functionality already.  Is there a reason why this can't be deployed here?
> > ```
> 
> I created #1004 almost a month ago to track this. tl;dr what I've implemented was a quick, "just works" solution because I didn't want to hold up opening this PR on thinking about a parser.

Ah, good point.  Ok, ignore my comment here then.  Pre- ☕ reviews are suboptimal.

Would you consider then moving out this file i/o to a separate file to keep the milc_interface.cpp code a bit cleaner?

weinbe2:
> > > Had my first review at this. Beyond my inline comments
> > > ```
> > > * There's a lot of code in milc_interface.cpp devoted to loading in parameters.  I thought the CLI11 had this functionality already.  Is there a reason why this can't be deployed here?
> > > ```
> > 
> > 
> > I created #1004 almost a month ago to track this. tl;dr what I've implemented was a quick, "just works" solution because I didn't want to hold up opening this PR on thinking about a parser.
> 
> Ah, good point. Ok, ignore my comment here then. Pre- ☕ reviews are suboptimal.
> 
> Would you consider then moving out this file i/o to a separate file to keep the milc_interface.cpp code a bit cleaner?

Sure---do you mind if make that a subsequent PR? I'll do that cleanup sooner than #1004.

weinbe2:
Made all of the various requested changes, verified everything compiles and behaves as expected (and correctly). This should be good mod further review, build checks going through, and @mathiaswagner 's comment on reconciling #1024.

weinbe2:
I've verified these changes don't effect Chroma's MG-accelerated HMC.

maddyscientist (approved):
Important step forward for QUDA + MILC.  Thanks @weinbe2!

Issue #164, opened 2014-10-16
Add outer product to internal quda test suite


mathiaswagner:

maddyscientist:
Assigning to @weinbe2 to take a look at this.

PR #1007, opened 2020-06-03
Feature/oprod

maddyscientist:
This pull request is mainly clean up pull, but also will give some small boosts in MILC performance. 
* Cleanup of the outer-product routines to use `colorspinor` accessors instead of the legacy textures accessors.  Still much optimization to be done for these routines, but this can wait until multi-RHS support is present in QUDA.
* Removal of legacy functions that are no longer used.
* Fix for latent race condition in `cudaColorSpinorField::exchangeGhost`: function does not return until any outstanding P2P sending is complete.
* Staggered outer product kernel now uses the regular `exchangeGhost` function for improved scaling.
* `GaugeField::fat_link_max` is now not computed until it is needed, as opposed to doing it whenever a fat-link host field is created.  This means we can do this evaluation on the device as opposed to the host, the end result is more than a halving in time spent in `loadGaugeQuda` when running using MILC.

This cleanup and optimization has been verified against MILC RHMC.

### DIFF
--- a/include/color_spinor.h
+++ b/include/color_spinor.h
@@ -992,7 +992,6 @@ namespace quda {
   __device__ __host__ inline Matrix<complex<Float>, Nc> outerProdSpinTrace(const ColorSpinor<Float, Nc, Ns> &a,
                                                                            const ColorSpinor<Float, Nc, Ns> &b)
   {
-
     Matrix<complex<Float>, Nc> out;
 
     // outer product over color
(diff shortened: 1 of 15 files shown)


weinbe2:
Has this been tested against Chroma HMC? I see the clover outer product routines were also updated.

maddyscientist:
No, because Chroma doesn't use QUDA for anything other than the solver.

weinbe2:
What correctness testing has been/should be done for the clover changes otherwise?

maddyscientist:
We don't have any tests off-hand that use this any more.  This code was used by the old TIFR MILC clover RHMC code, but that's not used in production any more.  For now, these routines are kept around for if / when we have HMC natively in QUDA.

weinbe2:
I'm not pleased by the idea that we have this code exposed in the interface if we don't have active testing/verification, but I guess it's not unprecedented re:the issues in gauge fixing code, if I remember correctly.

I won't hold up the PR for it, but please file an issue/feature request for it (if there isn't one already) so we have it recorded somewhere.

maddyscientist:
Long outstanding issue: https://github.com/lattice/quda/issues/164

mathiaswagner:
Do we need to clang-format this? Just thought of it and might forget later.

include/color_spinor.h
```diff
@@ -1020,6 +1019,35 @@ namespace quda {
...
+#pragma unroll
+    for (int i = 0; i < Nc; i++) {
+#pragma unroll
+      for (int j = 0; j < Nc; j++) {
+        // trace over spin (manual unroll for perf)
+        out(j, i).real(a(0, j).real() * b(0, i).real());
+        out(j, i).real(out(j, i).real() + a(0, j).imag() * b(0, i).imag());
+        out(j, i).imag(a(0, j).imag() * b(0, i).real());
+        out(j, i).imag(out(j, i).imag() - a(0, j).real() * b(0, i).imag());
+        // out(j,i) = a(0,j) * conj(b(0,i));
```
mathiaswagner, line 1045: can this be deleted?

maddyscientist:
> Do we need to clang-format this? Just thought of it and might forget later.

I already did.

weinbe2 (approved):
Gave this a visual inspection and did a due-diligence build+link with MILC, all looks good.

mathiaswagner (changes requested):
Looks good. MILC test also looked fine.

PR #1009, opened 2020-06-04
Hacked Mods for Clang based builds

bjoo:
Hi @mathiaswagner . This pull is to specifically address using Clang to build the device code as well as the host code. I have used it only as a Devel build on Summit so far. Currently I have had to use many PATH oriented flags on Summit with my Clang-10.1 build:

```       
        -DCMAKE_CXX_COMPILER="clang++"\
        -DCMAKE_CXX_FLAGS=" -std=c++14  -cxx-isystem ${GCC_HOME}/include -cxx-isystem $GCC_HOME/include/c++/7.4.0 -cxx-isystem $GCC_HOME/include/c++/7.4.0/powerpc64le-none-linux-gnu" \
        -DCMAKE_C_COMPILER="clang" \
        -DCMAKE_EXE_LINKER_FLAGS="-Wl,-rpath,${GCC_HOME}/lib64 -L${GCC_HOME}/lib64 -Wl,-rpath,${CUDA_TOOLKIT_ROOT_DIR}/lib64 -L${CUDA_TOOLKIT_ROOT_DIR}/lib64 -Wl,-rpath,${CUDA_TOOLKIT_ROOT_DIR}/targets/ppc64le-linux/lib -lcudart -lcublas -lstdc++" \
        -DCMAKE_CUDA_COMPILER=clang++ \
        -DCMAKE_CUDA_FLAGS=" --cuda-path=${CUDA_TOOLKIT_ROOT_DIR} --cuda-gpu-arch=sm_70  -cxx-isystem $GCC_HOME/include/c++/7.4.0 -cxx-isystem $GCC_HOME/include/c++/7.4.0/powerpc64le-none-linux-gnu " \
        -DCMAKE_CUDA_HOST_COMPILER="clang++" \
        -DCMAKE_CUDA_HOST_LINK_LAUNCHER=clang++  \
        -DQUDA_BUILD_SHAREDLIB=ON \
        -DCMAKE_CUDA_ARCHITECTURES="70" \
```

This is mostly all related to picking up the C++ header files from the more modern toolchain (7.4.0) which is not in the standard location on Summit, and linking the appropriate version of `libstdc++` and `CUBLAS` (including `libcublasLt.so`) .  I would hope, that on a more standard system ( CUDA in /usr/local/ and Clang in /usr/bin etc, none of this complication would be needed.)

The primary change is to enable a flag which for now is called QUDA_NVCC_BUILD
based on the CUDA_COMPILER_ID matching. Based on this we may need to disable `-Xcompiler` and also when passing `-Wno-feature,-Wno-feature2` to instead pass a space separated list rather than a comma separated list (as needed by Xcompiler) we pass a space separated list (regular compiler args). 

With these mods I could compile QUDA fully on Summit using a self built Clang (modulo the path stuff), but this may need more tidying/systematizing. I am happy to iterate with you on that. Also, I have not tried any other build types than 'Devel' at this point.


mathiaswagner:
I think https://github.com/Kitware/CMake/commit/80d37167fed1178872d28cbcbf57c7a3660bf244 has some good examples on how to use generator expressions to determine whether NVCC is used. I think we should use this approach here as well -- as much as possible.

mathiaswagner:
@Jenkins build this please

mathiaswagner:
replaced by #1025.

PR #1011, opened 2020-06-09
Delete the solver.... then update the parameters.

cpviolator:
Restores persistent deflation space behaviour.

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -2994,13 +2994,13 @@ void invertQuda(void *hp_x, void *hp_b, QudaInvertParam *param)
       DiracMdagMLocal mPreLocal(diracPre);
       Solver *solve = Solver::create(solverParam, m, mSloppy, mPreLocal, profileInvert);
       (*solve)(*out, *in);
-      solverParam.updateInvertParam(*param);
       delete solve;
+      solverParam.updateInvertParam(*param);
     } else {
       Solver *solve = Solver::create(solverParam, m, mSloppy, mPre, profileInvert);
       (*solve)(*out, *in);
-      solverParam.updateInvertParam(*param);
       delete solve;
+      solverParam.updateInvertParam(*param);
     }
   } else { // norm_error_solve
     DiracMMdag m(dirac), mSloppy(diracSloppy), mPre(diracPre);


mathiaswagner:
Changes look good. Just curious: Why is this needed?

And can we add a comment here? I think we still want to switch to using smart pointers (preferably `unique_ptr`) sometime and then we loose some control about when `delete` is called (agree, we can still force it). But I am worried the issue might get reintroduced then, if we just delete the `delete`.

cpviolator:
@mathiaswagner The transfer of the deflation space is performed in the update function:

```C++
/**                                                                                                                                                                                                     
       Update the QudaInvertParam with the data from this                                                                                                                                                   
       @param param the QudaInvertParam to be updated                                                                                                                                                       
     */
    void updateInvertParam(QudaInvertParam &param, int offset=-1) {
      param.true_res = true_res;
      param.true_res_hq = true_res_hq;
      param.iter += iter;
      reduceDouble(gflops);
      param.gflops += gflops;
      param.secs += secs;
      if (offset >= 0) {
	param.true_res_offset[offset] = true_res_offset[offset];
        param.iter_res_offset[offset] = iter_res_offset[offset];
        param.true_res_hq_offset[offset] = true_res_hq_offset[offset];
      } else {
        for (int i=0; i<num_offset; i++) {
          param.true_res_offset[i] = true_res_offset[i];
          param.iter_res_offset[i] = iter_res_offset[i];
          param.true_res_hq_offset[i] = true_res_hq_offset[i];
        }
      }
      //for incremental eigCG:                                                                                                                                                                              
      param.rhs_idx = rhs_idx;

      param.ca_lambda_min = ca_lambda_min;
      param.ca_lambda_max = ca_lambda_max;

      if (deflate) *static_cast<QudaEigParam *>(param.eig_param) = eig_param;
    }
```

Apparently, if it's done in the wrong order, certain booleans are destroyed and the subsequent solve has no memory of the previous deflation space construction.

mathiaswagner:
Thanks @cpviolator. Looking at the code I think there might be an underlying issue. I'd certainly call that behavior unexpected as deleting the solver after the param struct has been updated should not modify the param structure again.

maddyscientist:
Ok, I see what is going on.  The issue is that the solver deletion calls `Solver::destroyDeflationSpace()`, which sets the critical parameters for preserving the deflation
```c++
      if (param.eig_param.preserve_deflation) {
        if (getVerbosity() >= QUDA_VERBOSE) printfQuda("Preserving deflation space of size %lu\n", evecs.size());

        if (param.eig_param.preserve_deflation_space) {
          deflation_space *space = reinterpret_cast<deflation_space *>(param.eig_param.preserve_deflation_space);
          // first ensure that any existing space is freed                                                                                                                                                                                                    
          for (auto &vec : space->evecs)
            if (vec) delete vec;
          space->evecs.resize(0);
          delete space;
        }

        deflation_space *space = new deflation_space;

        // if evecs size = 2x evals size then we are doing an SVD deflation                                                                                                                                                                                   
        space->svd = (evecs.size() == 2 * evals.size()) ? true : false;

        space->evecs.reserve(evecs.size());
        for (auto &vec : evecs) space->evecs.push_back(vec);

        space->evals.reserve(evals.size());
        for (auto &val : evals) space->evals.push_back(val);

        param.eig_param.preserve_deflation_space = space;
```
If we try to transfer before the solver is deleted, we do so before these parameters are set, and we instead just fail as Dean observed.

maddyscientist (approved):

Issue #1010, opened 2020-06-08
QUDA throws linker errors when built against MPI with the C++ interface


weinbe2:
As it stands, QUDA throws an error at link time if MPI has been built with the (depreciated) C++ interface (`--enable-mpi-cxx`). The root of the issue appears to be that, in this case, some file is being compiled with `mpicc` instead of `mpicxx`, even though CMake _should_ be smart enough to know where to use one vs. the other. 

One hacky WAR is to manually set `MPI_C_LIB_NAMES = MPI_CXX_LIB_NAMES` so `mpicc` will grab the appropriate C++ library. What this would cause in practice is:

MPI build without C++ bindings: `MPI_C_LIB_NAMES == MPI_CXX_LIB_NAMES == mpi` (standard behavior, sane)
MPI build with C++ bindings: `MPI_C_LIB_NAMES == MPI_CXX_LIB_NAMES == mpi_cxx;mpi` (hacky, could have unintended consequences)

While this works, it's gross, which is why we aren't implementing it in practice. The fact that the C++ interface to MPI is depreciated is another reason to not bother.

This issue and WAR is here for the reference of others, but unless an urgent need arises this is a WNF bug.

weinbe2:
Closing because WNF. This has also been noted near the top of the wiki page on building CUDA with CMake: https://github.com/lattice/quda/wiki/QUDA-Build-With-CMake

weinbe2:
Addressed by #1012

PR #1012, opened 2020-06-09
Potential fix for MPI CXX link issue

mathiaswagner:
Use `MPI:CXX` instead of `MPI::C` in cmake.
This seems to link fine for MPI built without C++ bindings.

@weinbe2 @hummingtree  can you check this fixes your issues (#1010) ?

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -364,7 +364,7 @@ if(QUDA_MPI OR QUDA_QMP)
 endif()
 
 if(QUDA_MPI)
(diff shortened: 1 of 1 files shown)


weinbe2:
Testing now

weinbe2:
Verified this works for builds of OpenMPI 4.0.* that both do and don't have the C++ interface.

As due diligence I also made sure both versions link to MILC (not that I was expecting surprises).

weinbe2 (approved):

PR #1013, opened 2020-06-09
Support using managed memory for MILC allocations

mathiaswagner:
To go with https://github.com/lattice/milc_qcd/tree/feature/managed_multix

This reduces some memory transfer overheads in MILC.

### DIFF
--- a/include/timer.h
+++ b/include/timer.h
@@ -135,6 +135,7 @@ namespace quda {
     QUDA_PROFILE_MEMCPY_D2H_ASYNC,   /**< device to host async copy */
     QUDA_PROFILE_MEMCPY2D_D2H_ASYNC, /**< device to host 2-d memcpy async copy*/
     QUDA_PROFILE_MEMCPY_H2D_ASYNC,   /**< host to device async copy */
+    QUDA_PROFILE_MEMCPY_DEFAULT_ASYNC,   /**< default async copy */
 
     QUDA_PROFILE_COMMS_START, /**< initiating communication */
     QUDA_PROFILE_COMMS_QUERY, /**< querying communication */
@@ -146,8 +147,6 @@ namespace quda {
   };
 
 #ifdef INTERFACE_NVTX
-
-
 
 #define PUSH_RANGE(name,cid) { \
     int color_id = cid; \
--- a/lib/cuda_gauge_field.cpp
+++ b/lib/cuda_gauge_field.cpp
@@ -248,7 +248,7 @@ namespace quda {
 	recvStart(dim, dir); // prepost the receive
 	if (!comm_peer2peer_enabled(dir,dim) && !comm_gdr_enabled()) {
 	  cudaMemcpyAsync(my_face_dim_dir_h[bufferIndex][dim][dir], my_face_dim_dir_d[bufferIndex][dim][dir],
-			  ghost_face_bytes[dim], cudaMemcpyDefault, streams[2*dim+dir]);
+			  ghost_face_bytes[dim], cudaMemcpyDeviceToHost, streams[2*dim+dir]);
 	}
       }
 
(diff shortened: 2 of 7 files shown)


lib/quda_cuda_api.cpp
```diff
@@ -189,6 +189,9 @@ namespace quda {
       case cudaMemcpyDeviceToDevice:
         PROFILE(cuMemcpyDtoDAsync((CUdeviceptr)dst, (CUdeviceptr)src, count, stream), QUDA_PROFILE_MEMCPY_D2D_ASYNC);
         break;
+      case cudaMemcpyDefault:
+        PROFILE(cuMemcpyAsync((CUdeviceptr)dst, (CUdeviceptr)src, count, stream), QUDA_PROFILE_MEMCPY_D2D_ASYNC);
```
weinbe2: Is `QUDA_PROFILE_MEMCPY_D2D_ASYNC` appropriate for default? I thought default was supposed to deduce the correct type of copy?

lib/cuda_gauge_field.cpp
```diff
@@ -681,7 +681,7 @@ namespace quda {
 	  if (error != cudaSuccess) errorQuda("Failed to get device pointer for MILC site / BQCD array");
 
 	  if (src.GhostExchange() == QUDA_GHOST_EXCHANGE_NO) {
-	    copyGenericGauge(*this, src, QUDA_CUDA_FIELD_LOCATION, gauge, src_d);
+	    copyGenericGauge(*this, src, QUDA_CUDA_FIELD_LOCATION, gauge);
```
maddyscientist: Is this a bug that got introduced here: `src_d` is not being passed to `copyGenericGauge`?  This will lead to direct use of the CPU pointer on the GPU which can fail, e.g., if allocated in CPU memory outside of the address space of the GPU.

mathiaswagner: good catch. I think this should not be in. It was an experiment to put all pinned allocations in managed, including FN links and then hostregister fails.

maddyscientist: Understood.  So I guess fix this for now, and we can revisit when in the future if/when we consider fermion links?

mathiaswagner: yes

lib/quda_cuda_api.cpp
mathiaswagner: Yes. It only affects which profile counter captures the copy. And with MemcpyDefault we don't really know ...

lib/cuda_gauge_field.cpp
mathiaswagner: Thinking about it more we could probably revert all changes to this file if we don't use managed for gauge fields for now.

maddyscientist: I think there's no harm keeping these changes , they enable future optimization and don't harm anything.

lib/quda_cuda_api.cpp
maddyscientist: I've fixed this with  e9e5737

lib/cuda_gauge_field.cpp
maddyscientist: Fixed with 3d34837

mathiaswagner:
Thanks for the cleanup @maddyscientist. Looks good to me now.

maddyscientist:
I'll do some tests with MILC with https://github.com/milc-qcd/milc_qcd/pull/35 and then sign off.

include/timer.h
```diff
@@ -135,6 +135,7 @@ namespace quda {
     QUDA_PROFILE_MEMCPY_D2H_ASYNC,   /**< device to host async copy */
     QUDA_PROFILE_MEMCPY2D_D2H_ASYNC, /**< device to host 2-d memcpy async copy*/
     QUDA_PROFILE_MEMCPY_H2D_ASYNC,   /**< host to device async copy */
+    QUDA_PROFILE_MEMCPY_DEFAULT_ASYNC,   /**< default async copy */
```
weinbe2, line 138: This needs a corresponding update in `lib/timer.cpp`: https://github.com/lattice/quda/blob/develop/lib/timer.cpp#L38

weinbe2: Would replacing `std::string TimeProfile::pname[]` with `std::string TimeProfile::pname[QUDA_PROFILE_TOTAL]` be enough to get a compile-time assert on the length of the initializer list? Or is that not the case?

maddyscientist: Pushed the fix.  I don't know, that's an interesting suggestion.  There's a bunch places through QUDA that should probably use `std:string` instead of `char*`.  To be honest this `PROFILE` needs a good cleanup.

weinbe2 (approved):
Nothing controversial here.

PR #1016, opened 2020-06-11
Feature/deflation vecs control

cpviolator:
Uncontroversial PR. I added a new parameter `deflation_vecs` to the `eig_param` structure. It allows the user to control how many eigenvectors to use in the deflation routines. Useful for studying what is optimal for a given workflow (rather than having to recompute different values of `nConv` for each study.)

Wiki updated, example command therein: https://github.com/lattice/quda/wiki/Thick-Restarted-eigensolver#initial-guess-deflation


cpviolator:
Ok, this collected some controversy. I changed the CamelCase variables in the eigensolver to underscore, and allowed the user to compute eigenvectors in double precision, but save in single.

lib/eigensolve_quda.cpp
```diff
@@ -395,13 +416,13 @@ namespace quda
       for (int j = 0; j < size; j++) {
         auto cnorm = H[i*size + j];
         if (j != i) {
-          if (abs(cnorm) > epsilon) {
+          if (abs(cnorm) > 5.0 * epsilon) {
```
maddyscientist, line 420: Why this arbitrary factor of 5.0?

cpviolator: Just using `epsilon=2e-16` would sometimes fail. I chose 5 to be safe after a few random iterations.

include/eigensolve_quda.h
```diff
@@ -21,12 +21,13 @@ namespace quda
...
-    int nEv;          /** Size of initial factorisation */
-    int nKr;          /** Size of Krylov space after extension */
-    int nConv;        /** Number of converged eigenvalues requested */
-    double tol;       /** Tolerance on eigenvalues */
-    bool reverse;     /** True if using polynomial acceleration */
-    char spectrum[3]; /** Part of the spectrum to be computed */
+    int n_ev;            /** Size of initial factorisation */
+    int n_kr;            /** Size of Krylov space after extension */
+    int n_conv;          /** Number of converged eigenvalues requested */
+    int deflation_vecs; /** Number of converged eigenvalues to use in deflation */
```
maddyscientist: Given the convention of `n_ev`, `n_kr`, etc., perhaps a more consistent name for `deflation_vec` would be `n_ev_deflate`?

maddyscientist:
This looks good.  Just a few things, then we can merge this:
* Probably rename `deflation_vecs` to `n_ev_deflate` for consistency with the other eigen parameter names
* Have you updated the wiki's various eigen solver instructions for consistency change for `nEv` -> `n_ev` etc.?
* Needs a clang format

maddyscientist (approved):
Thanks, as always for this addition @cpviolator. Will merge after build tests complete.

cpviolator:
Not seeing the build error when using `cmake version 3.16.1`...

PR #1018, opened 2020-06-16
Handle MILC staggerd phase removal in QUDA

mathiaswagner:
This is the first steps towards better gauge field residency in MILC. 
It adds functions to the MILC interface that take an argument `phase_in` that signals QUDA that MILC has not removed the staggered phases and QUDA can handle it.
This removes the need for MILC to modify the gauge field in the HMC evolution.

This does not break compatibility with older MILC version but modifications to MILC are required to exploit the benefits. Those can be found in https://github.com/lattice/milc_qcd/tree/feature/quda_rephase

I'll file a PR for that later.

PR #1019, opened 2020-06-18
Add gauge field residency for MILC

mathiaswagner:
This builds on and depends on #1018.

This adds some gauge field residency for update U, gauge force and reunitarize to MILC. The fermion force still breaks this up.
Also, at the SU3 projections / unitarization at the end of a trajectory breaks the residency.


lib/interface_quda.cpp
```diff
@@ -5043,6 +5043,7 @@ void updateGaugeFieldQuda(void* gauge,
    if (param->use_resident_gauge) {
      if (!gaugePrecise) errorQuda("No resident gauge field to use");
      cudaGauge = gaugePrecise;
+     gaugePrecise = nullptr;
```
weinbe2, line 5046: What's the purpose of this addition? More specifically, why wasn't it there in the first place?

mathiaswagner: Later 5070 bad things might happen if gaugePrecise and cudaGauge alias (I think)

weinbe2 (approved):

PR #1021, opened 2020-06-23
Feature/no tex

maddyscientist:
This pull request's primary purpose is to remove textures from QUDA, however, in doing so it also completely rewrites the blas kernels in order remove any performance regressions from not having textures.
* Complete removal of textures from QUDA, this leads to optimal performance out of the box on Pascal, Volta and Ampere
* Complete rewrite of `convert.h` and now more robust through using `enable_if`.  Alternative fast `I2F` included in QUDA, but not enabled by default.
* Complete rewrite of the blas, reduce, multi-blas and multi-reduce kernels to be much cleaner and more composable.
* With removal of textures, we no longer need to instantiate various precision / ordering combinations, e.g., all single-precision blas can now use the `float4` variant.
* nSpin=4 MG null-space fields now use float4 ordering on the fine grid
* Removal of old CPU generic blas and reduce kernels, with CPU now directly targeted in same path as GPU. 
* Enabling `FLOAT8` ordering now enables `FLOAT8` for all half and quarter precision nSpin=4 fields.  E.g., FLOAT8 is now the library-wide native order when it is set.  This cleans up the hacks for FLOAT8 present from #903. 
* Removal of various dead code across the library
* Fix jitify failure introduced with #999 
* Various miscellaneous cleanup and perf fixes


include/color_spinor_field_order.h
```diff
@@ -1601,12 +1613,18 @@ namespace quda {
     typedef colorspinor::FloatNOrder<float, 1, Nc, 2, false, huge_alloc> type;
   };
 
+#ifdef FLOAT8
+#define N8 8
+#else
+#define N8 4
```
mathiaswagner, line 1669: wondering if we should undefine N8 again

lib/inv_bicgstabl_quda.cpp
```diff
@@ -125,16 +125,15 @@ namespace quda {
   void BiCGstabL::updateXRend(Complex* gamma, Complex* gamma_prime, Complex* gamma_prime_prime,
                             std::vector<ColorSpinorField*> r, ColorSpinorField& x, int nKrylov)
   {
-    /*
-    blas::caxpy(gamma[1], *r[0], x_sloppy);
+#if 0
```
mathiaswagner, line 124: Is this intended here? wondering about the #if / #else path here were else just does the delete

weinbe2:
Out of curiosity, what's the performance delta for Kepler?

include/kernels/reduce_core.cuh
```diff
@@ -642,66 +624,27 @@ namespace quda
...
-      {
-        tmp = x;
-        x = b.x * (x - a.x * z) + b.y * y;
-        y = tmp;
-        norm2_<ReduceType>(sum, x);
+#pragma unroll
+        for (int i = 0; i < x.size(); i++) {
+          auto tmpx = x[i];
+          auto tmpy = y[i];
+          x[i] = b * (x[i] + a * y[i]) + ((real)1.0 - b) * z[i];
```
weinbe2, line 638: Do you have a feeling for why the bugfix of replacing `b` with `1-b` (based on the wasn't caught before?

maddyscientist: Previously, `b` was a two-component number with `b.y = 1 - b.x`.  This replacement with a real-valued `b` is equivalent and in my option cleaner.

maddyscientist:
> Out of curiosity, what's the performance delta for Kepler?

Double and single precision are unchanged.  For half precision Wilson and hisq dslash kernels, the regressions are around 20% for the interior dslash with the halo kernels unaffected.  Interestingly, I found for Wilson I can narrow the regression back to around 10% by switching to `FLOAT8` (which is because we have less indexing operations as a result, and it amortizes somewhat the overhead).  I fully expect that for staggered, we'll get back to parity or perhaps even improve upon the texture performance when I switch to 128-bit load/store for the staggered fermion field, since this will cut down the number of load/store instructions by a factor of 4, together with a commensurate reduction in indexing ops.

Regardless, likely any dslash regression of this order is approaching the noise for multi-node calculations that don't have good GDR enabled (e.g., the K80 cluster at BNL).

include/kernels/blas_core.cuh
```diff
@@ -9,153 +9,160 @@ namespace quda
...
+      axpbyz_(const real &a, const real &b, const real &c) : a(a), b(b) { ; }
+      template <typename T> __device__ __host__ void operator()(T &x, T &y, T &z, T &w, T &v)
       {
-        v = a.x * x + b.x * y;
+#pragma unroll
+        for (int i = 0; i < x.size(); i++) v[i] = a * x[i] + b * y[i];
       }                                  // use v not z to ensure same precision as y
-      static int streams() { return 3; } //! total number of input and output streams
-      static int flops() { return 3; }   //! flops per element
+      int streams() const { return 3; } //! total number of input and output streams
```
mathiaswagner: guess we could even go `constexpr` here ...

include/kernels/reduce_core.cuh
```diff
@@ -11,287 +10,253 @@ namespace quda
...
-      sum += (ReduceType)a.w * (ReduceType)b.w;
-    }
-
-    template <typename ReduceType> __device__ __host__ void dot_(ReduceType &sum, const float8 &a, const float8 &b)
+    template <typename reduce_t, typename T>
+    __device__ __host__ void dot_(reduce_t &sum, const typename VectorType<T, 2>::type &a, const typename VectorType<T, 2>::type &b)
     {
-      dot_(sum, a.x, b.x);
-      dot_(sum, a.y, b.y);
+      sum += (reduce_t)a.x * (reduce_t)b.x;
```
mathiaswagner: style question: should we use C++-style instead of C-style casts?

mathiaswagner (approved):
This looks good. Guess we suffer now from not having proper regression testing in place.

include/color_spinor_field_order.h
```diff
@@ -562,6 +585,36 @@ namespace quda {
...
+       * parameter n is only used for indexed into the packed
+       * null-space vectors.
+       * @param x 1-d checkerboard site index
+       * @param s spin index
+       * @param c color index
+       * @param v vector number
+       */
+      template <int nSpinBlock>
+      __device__ __host__ inline void load(complex<Float> out[nSpinBlock * nColor * nVec], int parity, int x_cb,
+                                           int chi) const
```
hummingtree, line 649: It would be great if the documentation can be updated here.

hummingtree (approved):
The changes, the number of lines of code reduction, look great to me.

lib/inv_bicgstabl_quda.cpp
maddyscientist: This was intended.  I just fixed the broken debugging code that was there before.  Of course we can just delete it but I didn't think it did any harm by being there.

include/kernels/reduce_core.cuh
maddyscientist: Since this is a simple cast from one scalar type to another, the C-style cast and C++ static_cast are identical.  I went with C since it looked more readable.  If there's a good reason to change it, I'm happy to do so.

include/color_spinor_field_order.h
maddyscientist: Good point, I'll do that.

include/color_spinor_field_order.h
maddyscientist: Good catch, will do.

lib/reduce_quda.cu
```diff
@@ -264,487 +311,104 @@ namespace quda {
...
-
-      blas::bytes += reduce.bytes();
-      blas::flops += reduce.flops();
-      checkCudaError();
-      return value;
-    }
-
-    /*
-      Wilson
-      double double2 M = 1/12
```
weinbe2, line 332: These types of `M` values scattered across multiple files confuse the life out of me. Maybe this gets added in a later PR, but if not, can there be some documentation explaining what they mean?

weinbe2:
Were the FLOAT8 changes verified for correctness against @hummingtree 's PCG/etc work? More of a due-diligence check than anything else.

weinbe2 (approved):
This PR is insane in the best of ways. A commit-by-commit visual inspection was very enlightening. I put a few comments in the PR for documentation and some corner-case testing, and I'll do some due-diligence builds on my end, but this gets my good-faith early approve.

maddyscientist:
> Were the FLOAT8 changes verified for correctness against @hummingtree 's PCG/etc work? More of a due-diligence check than anything else.

Yes, I've tested the block preconditioner with and without `FLOAT8` in this branch.  All works as expected.

lib/reduce_quda.cu
weinbe2: Everything was cleared up in a separate convo with @maddyscientist , all is good.

maddyscientist:
I have address all the above PR comments, thanks for the reviews folks.  I also realized there was an issue with `FLOAT8` ordering for MG, where the necessary templates weren't instantiated (implemented!) for float-8 ordering with fine-grain accessors.  That's fixed now.

lib/eigensolve_quda.cpp
```diff
@@ -668,7 +668,7 @@ namespace quda
     // 1. Take block inner product: L_i^dag * vec = A_i
     std::vector<ColorSpinorField *> left_vecs;
     left_vecs.reserve(n_defl);
-    for (int i = n_defl; i < 2 * n_defl; i++) left_vecs.push_back(evecs[i]);
+    for (int i = eig_param->n_conv; i < eig_param->n_conv + n_defl; i++) left_vecs.push_back(evecs[i]);
```
mathiaswagner, line 671: Why is this in the no_tex branch?

maddyscientist: It was a quick fix that was needed after #1016 was merged.  Since this PR was almost ready this was a quick fix to add here.

mathiaswagner: I’d generally prefer a quick hotfix pr for these things. It is orthogonal to the no Tex work.

maddyscientist: I figured there were enough unrelated minor fixes in this PR already that another wouldn't hurt.  Regardless, point taken.

cpviolator (approved):

PR #1022, opened 2020-06-25
Feature/cublas interface

cpviolator:
This PR adds an interface to the cuBLAS GEMM suite of functions. All data types (S,D,C,Z) are supported, as well as their batched variants. Host checking is done with Eigen.

Command line parameters are given here:

```
--cublas-data-type ENUM:value in {C,Z,S,D}=2
                              Whether to use single(S), double(D), and/or complex(C/Z) data types (default C)
  --cublas-data-order ENUM:value in {row,col}=1
                              Whether data is in row major or column major order (default row)
  --cublas-trans-a ENUM:value in {N,T,C}=0
                              Whether to leave the A GEMM matrix as is (N), to transpose (T) or transpose conjugate (C) (default N) 
  --cublas-trans-b ENUM:value in {N,T,C}=0
                              Whether to leave the B GEMM matrix as is (N), to transpose (T) or transpose conjugate (C) (default N) 
  --cublas-alpha FLOAT=[1,0] x 2
                              Set the complex value of alpha for GEMM (default {1.0,0.0}
  --cublas-beta FLOAT=[1,0] x 2
                              Set the complex value of beta for GEMM (default {1.0,0.0}
  --cublas-mnk INT=[128,128,128] x 3
                              Set the dimensions of the A, B, and C matrices GEMM (default 64 64 64)
  --cublas-leading-dims INT=[64,64,64] x 3
                              Set the leading dimensions A, B, and C matrices GEMM (default 128 128 128) 
  --cublas-offsets INT=[0,0,0] x 3
                              Set the offsets for matrices A, B, and C (default 0 0 0)
  --cublas-batch INT=16       Set the number of batches for GEMM (default 16)
```

And and example command:

```
./cublas_interface_test --cublas-mnk 64 64 64 --cublas-leading-dims 128 128 128 --cublas-offsets 16 16 16 --cublas-data-type Z --cublas-data-order row --cublas-batch 20 --cublas-alpha 1.0 2.0 --cublas-beta -3.0 1.5 --cublas-trans-a T --cublas-trans-b C
```

The example command will compute a batched ZGEMM. It will create data arrays with leading dims of 128, but matrix sizes of 64x64 (this tests the leading dim logic.) The offset is 16, so each matrix will be defined starting from the 16th element of the array (this tests the starting position logic.) There are 20 batches, and arbitrary alpha and beta coefficients are given. We declare the the data is in ROW major, forcing the routine to test its logic in converting from row major (C++ standard) to column major (CUDA/FORTRAN standard). Lastly, we use transposed OP types for both A and B to test the transposition logic.


cpviolator:
This PR will fail Jenkins tests with the following error:

```C++
/scratch/CPviolator/work/QUDA_EXT_PROGS/quda/lib/check_params.h:23:32: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
 #define P(x, val) if (param->x == val && param->x != 0) errorQuda("Parameter " #x " undefined")
/scratch/CPviolator/work/QUDA_EXT_PROGS/quda/lib/check_params.h:914:3: note: in expansion of macro ‘P’
   P(n, INVALID_INT);
```
This is because I needed to declare some parameters as `uint64_t` and the `check_param.h` defines do not know how to handle anything other than `int` or `double`.

cpviolator:
Build failures are from `creal` and `cimag` used in `lib/targets/cuda/blas_lapack_cublas.cpp` Fine when using g++-7, seems to go ~away~ awry when using g++ >= 8.

maddyscientist:
These errors arise from the use of C complex.h functions in C++ code.  Can you replace the use of `creal` / `cimag` with their equivalents in `std::complex`?  Note that it is specified in the C++ standard that C++ and C complex types are directly compatible, e.g., it is legal to `reinterpret_cast` from one to the other.

cpviolator:
New example command:
```
blas_interface_test --blas-mnk 64 64 64 --blas-leading-dims 128 128 128 --blas-offsets 16 16 16 --blas-data-type Z --blas-data-order row --blas-batch 20 --blas-alpha 1.0 2.0 --blas-beta -3.0 1.5 --blas-trans-a T --blas-trans-b C
```

The last update merged in develop and renamed all the BLAS objects and functions from `XXXcublasXXX` to `XXXblasXXX` so that one can make generic targets.

cpviolator:
New example command:
```
./blas_interface_test --blas-mnk 64 64 64 --blas-leading-dims 128 128 128 --blas-offsets 16 16 16 --blas-data-type Z --blas-data-order row --blas-batch 20 --blas-alpha 1.0 2.0 --blas-beta -3.0 1.5 --blas-trans-a T --blas-trans-b C  --blas-strides 1 2 1 --native-blas-lapack false
```

This command includes the option to perform the BLAS test on the host(generic) or device. Please test using a variety of data types, strides, transpose operations, etc.

include/blas_lapack.h
```diff
@@ -60,6 +60,36 @@ namespace quda
       long long BatchInvertMatrix(void *Ainv, void *A, const int n, const uint64_t batch, QudaPrecision precision,
                                   QudaFieldLocation location);
 
+      /**
+         Strided Batch GEMM. This function performs N GEMM type operations in a
```
maddyscientist: Missing `@brief` here

include/enum_quda.h
```diff
@@ -1,5 +1,4 @@
-#ifndef _ENUM_QUDA_H
-#define _ENUM_QUDA_H
+#pragma once
```
maddyscientist, line 1: These additions need also to be mirrored in `enum_quda_fortran.h`

lib/check_params.h
```diff
@@ -918,6 +918,62 @@ void printQudaGaugeObservableParam(QudaGaugeObservableParam *param)
...
+  P(a_offset, INVALID_INT);
+  P(b_offset, INVALID_INT);
+  P(c_offset, INVALID_INT);
+  P(strideA, INVALID_INT);
+  P(strideB, INVALID_INT);
+  P(strideC, INVALID_INT);
+  P(batch_count, INVALID_INT);
+  P(data_type, QUDA_BLAS_DATATYPE_INVALID);
+  P(data_order, QUDA_BLAS_DATAORDER_INVALID);
+#endif
```
maddyscientist, line 970: Is the `INIT_PARAM` specialization needed here?  If all of the above parameters need to be set, and it looks like most of the initialized values aren't  valid for computation, e.g., `m = n = k = 0`, then just going with the `INVALID_INT` initialization is likely better here I think.

lib/interface_quda.cpp
```diff
@@ -510,6 +512,68 @@ void initQudaDevice(int dev)
   profileInit.TPSTOP(QUDA_PROFILE_TOTAL);
 }
 
+void printDeviceProp()
```
maddyscientist: This whole function should be moved to the `target/cuda/device.cpp` file, and called with something like  `device::print_prop()`.

tests/CMakeLists.txt
```diff
@@ -162,6 +162,13 @@ if(QUDA_MULTIGRID)
 
 endif()
 
+if(QUDA_BUILD_NATIVE_LAPACK)
+  add_executable(blas_interface_test blas_interface_test.cpp)
+  target_link_libraries(blas_interface_test ${TEST_LIBS})
+  quda_checkbuildtest(blas_interface_test QUDA_BUILD_ALL_TESTS)
+  install(TARGETS blas_interface_test ${QUDA_EXCLUDE_FROM_INSTALL} DESTINATION ${CMAKE_INSTALL_BINDIR})
+endif()
+
```
maddyscientist, line 171: Can you add a `blas_interface_test` to the ctest list?

lib/interface_quda.cpp
```diff
@@ -5960,3 +6026,143 @@ void gaugeObservablesQuda(QudaGaugeObservableParam *param)
   gaugeObservables(*gauge, *param, profileGaugeObs);
   profileGaugeObs.TPSTOP(QUDA_PROFILE_TOTAL);
 }
+
+void blasGEMMQuda(void *arrayA, void *arrayB, void *arrayC, bool use_native, QudaBLASParam *blas_param)
```
maddyscientist: At some point we'll need to refactor interface_quda.cpp since it's getting rather unwieldy.  We could start that now: perhaps move this function to its own file, blas_gemm_quda.cpp?  Perhaps leaving this stub here and pass the `profileBLAS` as a parameter to the function in that file?

maddyscientist:
This looks great @cpviolator.  Solid addition to the QUDA repertoire.  I've left some comments for things that should be fixed, but these should be quick, and once that's done, I see no reason not to get this merged.

lib/interface_quda.cpp
weinbe2: I support splitting up the `interface_quda.cpp` file, but if we are let's stick them in a `lib/interface/` subdirectory or something of the sort---start clawing back at the file sprawl 🙂

maddyscientist: Excellent idea!

include/blas_lapack.h
```diff
@@ -60,6 +60,36 @@ namespace quda
       long long BatchInvertMatrix(void *Ainv, void *A, const int n, const uint64_t batch, QudaPrecision precision,
                                   QudaFieldLocation location);
 
+      /**
+         @breif Strided Batch GEMM. This function performs N GEMM type operations in a
```
maddyscientist: spelling of brief

lib/interface_quda.cpp
cpviolator: I'm afraid I'm going to need some help on this. I moved the `blasGEMMQuda` function to its own file `lib/interfaces/blas_interface.cpp`, constructed a `CMakeLists.txt`, added the sub directory, etc... Rather than use an `extern` declaration for the BLAS `TimeProfile` I created a simple method to return a reference to it in `interface_quda.cpp`:
```C++
//!< Profiler for GEMM and other BLAS                                                                                                                                             
static TimeProfile profileBLAS("blasQuda");
TimeProfile* getProfileBLAS() {
  return &profileBLAS;
}
```

which naturally must have a forward declaration in `quda.h`. But `quda.h` has no knowledge of what a `TimeProfile` object is. I don't know how to proceed.

Also, the call to `checkBLASParam(blas_param);` now causes an error, because the function is defined in `check_params.h`  which is no longer visible to the file in `lib/interfaces/blas_interface.cpp`, and the compilation of `interface_quda.cpp` causes an unused-function warning because there is no instance of `checkBLASParam()` in the file anymore.

Are we sure that this is in the scope of this PR, or am I missing an obvious solution?

cpviolator: Nevermind... All fixed, thank you @maddyscientist .

weinbe2:
I'm doing a review right now, and I'm very impressed. I'm going to use your comments if I ever need to figure out cublas, @cpviolator !

A few issues have come up:
* The clang Jenkins builds are throwing errors. This is the cause, easy to fix:
```
../tests/blas_interface_test.cpp:22:15: error: unused variable 'NcontractType' [-Werror,-Wunused-const-variable]
constexpr int NcontractType = 2;
```
* I'm hitting an error when I run `ctest -V`:
```
test 1
    Start 1: blas_interface_test

1: Test command: /home/eweinberg/Data/quda_custom/2020-11-05CublasInterface/build/tests/blas_interface_test "--blas-mnk" "64" "64" "64" "--blas-leading-dims" "128" "128" "128" "--blas-offsets" "16" "16" "16" "--blas-data-type" "Z" "--blas-data-order" "row" "--blas-batch" "20" "--blas-alpha" "1.0" "2.0" "--blas-beta" "-3.0" "1.5" "--blas-trans-a" "T" "--blas-trans-b" "C" "--gtest_output=xml:blas_interface_test.xml"
1: Test timeout computed to be: 1500
1: The following argument was not expected: --gtest_output=xml:blas_interface_test.xml
1: Run with --help for more information.
1/1 Test #1: blas_interface_test ..............***Failed    0.28 sec
```
This is a single GPU build, all dslash types disabled, otherwise default parameters.

weinbe2:
I see that even though I'm specifying a data type (here, using your command, `--blas-data-type Z`), gtests for all four types (S/D/C/Z) are getting run. Is this intentional? This isn't a huge problem, it's not like these tests take long, but it's a bit confusing.

weinbe2:
Error check: leading dims needs to be larger than mnk.

```
./blas_interface_test --blas-mnk 64 64 64 --blas-leading-dims 32 32 32
```

Throws an error w/in cuBLAS, from an abstraction standpoint I think it'd be better if this got caught in an `errorQuda`.

weinbe2:
Error when running with `--native-blas-lapack false`, looks like a QUDA timer isn't being closed out correctly.

```
> ./blas_interface_test --blas-mnk 64 64 64 --native-blas-lapack false
[...]
batch 13: (C_host - C_gpu) Frobenius norm = 1.658820e+02. Relative deviation = 4.049854e-02
batch 14: (C_host - C_gpu) Frobenius norm = 1.806672e+02. Relative deviation = 4.410820e-02
batch 15: (C_host - C_gpu) Frobenius norm = 1.511400e+02. Relative deviation = 3.689941e-02
[       OK ] QUDA/BLASTest.verify/realSingle (17 ms)
[ RUN      ] QUDA/BLASTest.verify/realDouble
ERROR: Cannot start an already running timer (/home/eweinberg/Data/quda_custom/2020-11-05CublasInterface/quda/lib/interface/blas_interface.cpp:15 in blasGEMMQuda())
ERROR: Aborting (/home/eweinberg/Data/quda_custom/2020-11-05CublasInterface/quda/lib/../include/timer.h:66 in Start())
```

Specifying `--blas-data-type Z`, etc, doesn't effect that, but it shouldn't happen regardless

weinbe2:
Other checks needed:

* `--blas-mnk -2 -2 -2` --- catch negative values. This throws an error w/in cuBLAS
* `--blas-leading-dims -2 -2 -2` --- maybe less necessary b/c it'd be caught by a leading dims < mnk error
* ...I guess blas offsets being negative is fine, despite my best efforts to get it to throw errors
* `--blas-batch -2` --- catch negative values. This one is especially important b/c -2 gets wrapped around to a very large unsigned value, so you get an attempt to make a very large allocation

It also appears there's some flaw in the verification check logic:
* `--blas-strides 1 -8 1`, whether or not it's valid, caused some weird issues

```
> ./blas_interface_test --blas-mnk 64 64 64 --blas-leading-dims 96 96 96 --blas-offsets 27 3 16 --blas-data-type Z --blas-data-order col --blas-batch 4 --blas-alpha 1.0 0.0 --blas-beta -3.0 1.5 --blas-trans-a T --blas-trans-b C  --blas-strides 1 -8 1
[...]
[ RUN      ] QUDA/BLASTest.verify/complexSingle
Computing Eigen matrix opertaion a * A_{64,64} * B_{64,64} + b * C_{64,64} = C_{64,64}
batch 0: (C_host - C_gpu) Frobenius norm = 3.357449e-04. Relative deviation = 8.196898e-08
batch 1: (C_host - C_gpu) Frobenius norm = 2.609108e+07. Relative deviation = 6.369892e+03
batch 2: (C_host - C_gpu) Frobenius norm = 2.366428e+02. Relative deviation = 5.777412e-02
batch 3: (C_host - C_gpu) Frobenius norm = nan. Relative deviation = nan
[       OK ] QUDA/BLASTest.verify/complexSingle (13 ms)
[...]
```

Despite one relative deviation >> 1 and a `nan` relative deviation, the test went through as "passed", which shouldn't happen even if the input flags are inherently flawed.

cpviolator:
@weinbe2 Nice pick up on errors! I went strict with checks for nonsense values of `m,n,k`, `strides`, `offsets` and `batches` which seems to clear up the majority of the errors. Also, I recognised that the generic code path will always be computed in Row major, so no need for the ColMaj helpers.

As for the `ctest -V`, I'm not sure. I compiled with wilson and I get failures along that path too.

weinbe2:
Thanks @cpviolator , I appreciate the update.

I'm still concerned about the verification tests. I understand the inputs I provided were non-sense, but that doesn't change the fact that the deviation checks should have indicated that the tests failed.

I can look into `ctest -V`.

weinbe2:
It looks like more of the Google test infrastructure needs to go into `blas_interface_test.cpp`. While it's much more complicated than what's needed here, `dslash_ctest.cpp` is representative of what the tests may need to look like, namely the norm check and `ASSERT_LE` in `TEST_P([...],verify)`.

On a separate note, though, I confirmed that all of your updates to the range checks covered the cases I uncovered!

maddyscientist (approved):
Great work @cpviolator.  This is a great PR.

Noting that Dean and I have discussed this PR at length offline, with all issues being resolved.  this PR also adds `contract_test` to the the `ctest` list.

PR #1027, opened 2020-06-30
Add the correct boundary condition to the Mobius preconditioner

hummingtree:
This PR adds the correct boundary condition to the `MdagMLocal` operator for Mobius domain wall fermion.

Previously no matter what the operator always have a zero Dirchlet boundary condition, for which each direction needs to be padded to catch the snake terms. Now with this PR, if a direction is not partitioned, i.e. `comm_dim_partitioned(d) == 0`, the direction `d` will not be padded and the stencils will simply extract data from the other side of that direction.

This PR also fixes a bug where the reduction in the overall scaling process (in order to fit the range of fp16 for the tensor cores) tries access illegal shared memory addresses. Now this reduction is done by CUB.

This way if the x direction is not partitioned, we no longer have the non-coalesced memory accesses previously experienced, which improves the performance of the preconditioner.

As an example on a Titan V, `Ls=16`, with a local volume of `12x12x12x16`, the timing of the preconditioner with different boundary conditions are,

comm | time [s]
:-: | :-:
comm0000 | 0.000598
comm0001 | 0.000629
comm0010 | 0.000658
comm0011 | 0.000695
comm0100 | 0.000708
comm0101 | 0.000734
comm0110 | 0.000799
comm0111 | 0.000821
comm1000 | 0.000723
comm1001 | 0.000749
comm1010 | 0.000818
comm1011 | 0.000837
comm1100 | 0.000844
comm1101 | 0.000863
comm1110 | 0.000880
comm1111 | 0.000971

time_vs_comm_both

### DIFF
--- a/include/mdw_dslash5_tensor_core.cuh
+++ b/include/mdw_dslash5_tensor_core.cuh
@@ -8,8 +8,7 @@
 #include <math_helper.cuh>
 #include <shared_memory_cache_helper.cuh>
 
-#define THRUST_IGNORE_CUB_VERSION_CHECK
-#include <cub/cub.cuh>
+#include <cub_helper.cuh>
 
 #if (__CUDACC_VER_MAJOR__ >= 9 && __COMPUTE_CAPABILITY__ >= 700)
 #include <mma.h>
--- a/include/mdw_dslash5_tensor_core.cuh
+++ b/include/mdw_dslash5_tensor_core.cuh
@@ -8,6 +8,9 @@
 #include <math_helper.cuh>
 #include <shared_memory_cache_helper.cuh>
 
+#define THRUST_IGNORE_CUB_VERSION_CHECK
+#include <cub/cub.cuh>
+
 #if (__CUDACC_VER_MAJOR__ >= 9 && __COMPUTE_CAPABILITY__ >= 700)
 #include <mma.h>
 
@@ -265,7 +268,8 @@ namespace quda
 
   constexpr float target_scale = 2e3;
 
-  template <class Vector> __device__ inline void block_wise_reduce_vector(const Vector &v, float *smem_scale)
+  template <int block_x, int block_y, class Vector>
+  __device__ inline void block_wise_reduce_vector(const Vector &v, float *smem_scale)
   {
 
     __syncthreads();
@@ -283,23 +287,19 @@ namespace quda
         __float_max_abs_floats__(warp_max, v(spin, color).imag());
       }
     }
-    // Find the maximum absolute value in a warp across different lanes
-    warp_wise_reduce_float(warp_max);
-    // Now lane 0 of each warp holds the maximum value
-    if (lane_id == 0) { smem_scale[warp_id] = warp_max; }
-    __syncthreads();
-    if (warp_id == 0) {
-      warp_max = (lane_id < ((blockDim.x * blockDim.y) >> 5)) ? smem_scale[lane_id] : 0.0f;
-      warp_wise_reduce_float(warp_max);
-      if (lane_id == 0) { smem_scale[0] = warp_max / target_scale; }
-    }
+
+    typedef cub::BlockReduce<float, block_x, cub::BLOCK_REDUCE_WARP_REDUCTIONS, block_y, 1> BlockReduce;
+    __shared__ typename BlockReduce::TempStorage temp_storage;
+    float aggregate = BlockReduce(temp_storage).Reduce(warp_max, cub::Max());
+
+    if (warp_id == 0 && lane_id == 0) { smem_scale[0] = aggregate / target_scale; }
     __syncthreads();
   }
 
   // Actually does more than the function name suggests.
   // will find the maximum absolute value among the vector, scale that, and store
   // to sm_b
-  template <int N_sm_d2, bool accumulate, bool store = true, class Vector>
+  template <int block_x, int block_y, int N_sm_d2, bool accumulate, bool store = true, class Vector>
   __device__ inline void load_matrix_b_vector(Vector &v, half2 *sm_b, float *smem_scale, float m_scale = 1.0f)
   {
     if (accumulate) {
@@ -315,7 +315,7 @@ namespace quda
       }
     }
     if (store) {
-      block_wise_reduce_vector(v, smem_scale);
+      block_wise_reduce_vector<block_x, block_y>(v, smem_scale);
       // Now smem_scale[0] contains the maximum value
       float current_scale = smem_scale[0];
 #pragma unroll
--- a/lib/mdw_fused_dslash.cu
+++ b/lib/mdw_fused_dslash.cu
@@ -30,16 +30,6 @@ namespace quda
 
 #if (__CUDACC_VER_MAJOR__ >= 9 && __COMPUTE_CAPABILITY__ >= 700)
 
-    template <typename F> inline void setMaxDynamicSharedBytesPerBlock(F *func)
-    {
-      qudaFuncSetAttribute((const void *)func, cudaFuncAttributePreferredSharedMemoryCarveout,
-                           (int)cudaSharedmemCarveoutMaxShared);
-      cudaFuncAttributes attr;
-      cudaFuncGetAttributes(&attr, (const void *)func);
-      qudaFuncSetAttribute((const void *)func, cudaFuncAttributeMaxDynamicSharedMemorySize,
-                           deviceProp.sharedMemPerBlockOptin - attr.sharedSizeBytes);
-    }
-
     /**
       @brief Parameter structure for applying the Dslash
     */
@@ -638,7 +628,7 @@ namespace quda
 
       template <typename T> inline void launch(T *f, const TuneParam &tp, Arg &arg, const qudaStream_t &stream)
       {
-        quda::mobius_tensor_core::setMaxDynamicSharedBytesPerBlock(f);
+        setMaxDynamicSharedBytesPerBlock(f);
         void *args[] = {&arg};
         qudaLaunchKernel((const void *)f, tp.grid, tp.block, args, tp.shared_bytes, stream);
       }
--- a/lib/mdw_fused_dslash.cu
+++ b/lib/mdw_fused_dslash.cu
@@ -284,7 +284,7 @@ namespace quda
       coordinate[3] = aux[3];
 
       // Find the full coordinate in the shrinked volume.
-      coordinate[0] += (parity + coordinate[3] + coordinate[2] + coordinate[1]) & 1;
+      coordinate[0] += (shift[0] + shift[1] + shift[2] + shift[3] + parity + coordinate[3] + coordinate[2] + coordinate[1]) & 1;
 
 // Now go back to the extended volume.
 #pragma unroll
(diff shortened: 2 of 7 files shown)


maddyscientist:
@cpviolator / @walkloud can you test this branch with the MSPCG preconditioner on Lassen? Notably, this should improve the performance of the preconditioner, both in terms of GFLOPS and in terms of the quality of the preconditioner.  Thx

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -8,6 +8,9 @@
 #include <math_helper.cuh>
 #include <shared_memory_cache_helper.cuh>
 
+#define THRUST_IGNORE_CUB_VERSION_CHECK
```
maddyscientist: I think we should avoid direct inclusion of `cub/cub.cuh` in QUDA code, instead use `cub_helper.cuh`.

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -283,23 +287,19 @@ namespace quda
...
-      warp_max = (lane_id < ((blockDim.x * blockDim.y) >> 5)) ? smem_scale[lane_id] : 0.0f;
-      warp_wise_reduce_float(warp_max);
-      if (lane_id == 0) { smem_scale[0] = warp_max / target_scale; }
-    }
+
+    typedef cub::BlockReduce<float, block_x, cub::BLOCK_REDUCE_WARP_REDUCTIONS, block_y, 1> BlockReduce;
+    __shared__ typename BlockReduce::TempStorage temp_storage;
+    float aggregate = BlockReduce(temp_storage).Reduce(warp_max, cub::Max());
+
+    if (warp_id == 0 && lane_id == 0) { smem_scale[0] = aggregate / target_scale; }
```
maddyscientist, line 294: Curious: any perf difference from using CUB for the reduction here?

lib/mdw_fused_dslash.cu
```diff
@@ -597,38 +620,25 @@ namespace quda
...
+        sprintf(config, "shift%d%d%d%d,halo%d%d%d%d,comm%d%d%d%d", arg.shift[0], arg.shift[1], arg.shift[2],
+                arg.shift[3], arg.halo_shift[0], arg.halo_shift[1], arg.halo_shift[2], arg.halo_shift[3], arg.comm[0],
+                arg.comm[1], arg.comm[2], arg.comm[3]);
+        strcat(aux, config);
       }
 
       template <typename T> inline void launch(T *f, const TuneParam &tp, Arg &arg, const qudaStream_t &stream)
       {
-        setMaxDynamicSharedBytesPerBlock(f);
+        quda::mobius_tensor_core::setMaxDynamicSharedBytesPerBlock(f);
```
maddyscientist: Why the need for the custom setMaxDynamicSharedBytesPerBlock here?

maddyscientist:
Nice PR, thanks @hummingtree.  I've left some comments above, just some trivial fixes and questions.  Hopefully @cpviolator and or @walkloud can also check this through Chroma MDWF solves.

include/mdw_dslash5_tensor_core.cuh
hummingtree: There is no visible difference in performance.

maddyscientist: CUB is faster I assume?

lib/mdw_fused_dslash.cu
hummingtree: It's because in order to use CUB to do the reduction the kernel will use static shared memory, and for that reason when setting function attributes, we need to this:
```
qudaFuncSetAttribute((const void *)func, cudaFuncAttributeMaxDynamicSharedMemorySize,
                           deviceProp.sharedMemPerBlockOptin - attr.sharedSizeBytes);
```
I should probably change the `setMaxDynamicSharedBytesPerBlock` [here](https://github.com/lattice/quda/blob/f5d1db135824bf6df98cf261fecc6810f4cf41fd/include/tune_quda.h#L202).

maddyscientist: Ah, that makes sense.  Do you know if these additional queries have any appreciable overhead?  (I would hope not, but good to check)

include/mdw_dslash5_tensor_core.cuh
hummingtree: Sorry previously I forget to type the "no" ... there is *no* visible difference

lib/mdw_fused_dslash.cu
hummingtree: I checked on nsys and I am not seeing visible overheads (< 2 us). Although it might not be a good idea to set the function attributes every time the kernel is called, the overhead is negeligible.

walkloud:
Performance comparison with physical pion mass solve on 64c96x12 volume.
Normally solved on 4 Summit nodes.  This test is on 16 Summit nodes.  CG solve args to `invert_test`
```
ARGS="--dslash-type mobius $dim_grid \
      --Lsdim 12 --m5 -1.1 --b5 1.5 --c5 0.5 --mass 0.00152 \
      --load-gauge ${production}/${ens}/cfgs_flow/${ens_long}.${CONF}_wflow1.0.lime \
      --verbosity verbose \
      --solution-type mat --matpc even-even --solve-type normop-pc \
      --prec double --niter 14000 --prec-sloppy half --reliable-delta 0.1 \
      --inv-type cg "
```
resulting performance
```
CG: 5451 iterations, <r,r> = 2.242103e-06, |r|/|b| = 9.839016e-08
CG: Reliable updates = 9
CG: Convergence at 5451 iterations, L2 relative residual: iterated = 9.839016e-08, true = 9.839016e-08 (requested = 1.000000e-07)
Solution = 4.30787e+08
Reconstructed: CUDA solution = 3.0144e+08, CPU copy = 3.0144e+08
Done: 5451 iter / 419.571 secs = 25973.1 Gflops
```
Pre pull request MSPCG solve - args
```
ARGS="--dslash-type mobius $dim_grid \
      --Lsdim 12 --m5 -1.1 --b5 1.5 --c5 0.5 --mass 0.00152 \
      --load-gauge ${production}/${ens}/cfgs_flow/${ens_long}.${CONF}_wflow1.0.lime \
      --verbosity verbose \
      --solution-type mat --matpc even-even --solve-type normop-pc \
      --prec double --niter 14000 --prec-sloppy half --reliable-delta 0.1 --prec-precondition half \
      --inv-type pcg --precon-type cg --precon-schwarz-type additive --tol-precondition 1e-6 --maxiter-precondition 6 "
```
resulting in performance
```
...
CG: Convergence at 6 iterations, L2 relative residual: iterated = 8.296857e-02 (requested = 1.000000e-06)
PCG: 2206 iterations, <r,r> = 2.256522e-06, |r|/|b| = 9.870602e-08
PCG: Reliable updates = 9
Solution = 4.30911e+08
Reconstructed: CUDA solution = 3.01689e+08, CPU copy = 3.01689e+08
Done: 2206 iter / 322.422 secs = 207188 Gflops
```
Post Pull request performance, same args
```
...
CG: Convergence at 6 iterations, L2 relative residual: iterated = 8.406681e-02 (requested = 1.000000e-06)
PCG: 2205 iterations, <r,r> = 2.254062e-06, |r|/|b| = 9.865221e-08
CG: Reliable updates = 9
Solution = 4.30912e+08
Reconstructed: CUDA solution = 3.0169e+08, CPU copy = 3.0169e+08
Done: 2205 iter / 249.382 secs = 267757 Gflops
```
Currently compiling with `QUDA_PRECISION=15` so that we can test `--prec double --prec-sloppy half --prec-precondition quarter`.  Results posted (presumably) tomorrow.  So far (assuming no significant fluctuations in node performance), we have a speed compared to CG solve of 
```
419.571/249.382 = 1.68
```

hummingtree:
Thanks @walkloud for the test. The speed up number looks great, although it probably shows how slow Summit communication is. Several things here:
- ~~Do the timings here include the tuning times?~~ (@maddyscientist confirmed that the results are post-tuning)
- You might want to try with the argument `--recon 12` to have a faster reconstruction type.
- The `--maxiter-precondition` can be tuned to get maximum speed up.

maddyscientist (approved):

walkloud:
@hummingtree do you have a recommended change in `--maxiter-precondition`?

hummingtree:
> @hummingtree do you have a recommended change in `--maxiter-precondition`?

I would test `8` and see if that's a better one.

walkloud:
@hummingtree and @maddyscientist I have bad news and good news.  The bad news is that I somehow managed to use an old (pre-pull request) executable in my test for the post-pull-request.  What I fail to understand is how it was faster than the old one.  The good news is that when I re-ran the test today, with the actually freshly compiled executable, it was even a little faster
```
ARGS
--dslash-type mobius --dim 64 64 8 8 --gridsize 1 1 8 12 --Lsdim 12 --m5 -1.1 --b5 1.5 --c5 0.5 --mass 0.00152
 --load-gauge <some_gauge> --verbosity verbose 
--solution-type mat --matpc even-even --solve-type normop-pc 
--prec double --niter 14000 --prec-sloppy half --reliable-delta 0.1 --prec-precondition half 
--inv-type pcg --precon-type cg --precon-schwarz-type additive --tol-precondition 1e-6 --maxiter-precondition 6
...
CG: 6 iterations, <r,r> = 1.817484e-10, |r|/|b| = 8.289459e-02
CG: Reliable updates = 0
CG: Convergence at 6 iterations, L2 relative residual: iterated = 8.289459e-02 (requested = 1.000000e-06)
PCG: 2105 iterations, <r,r> = 2.294502e-06, |r|/|b| = 9.953323e-08
PCG: Reliable updates = 9
Solution = 4.30928e+08
Reconstructed: CUDA solution = 3.01722e+08, CPU copy = 3.01722e+08
Done: 2105 iter / 227.981 secs = 279610 Gflops
```
I also tried with maxiter-preconditioner=8, but that didn't seem to help - it brought the total iteration count down, but the time reduction is probably in the noise
```
CG: Convergence at 8 iterations, L2 relative residual: iterated = 5.048672e-02 (requested = 1.000000e-06)
PCG: 1940 iterations, <r,r> = 2.301722e-06, |r|/|b| = 9.968971e-08
PCG: Reliable updates = 8
Solution = 4.30943e+08
Reconstructed: CUDA solution = 3.01751e+08, CPU copy = 3.01751e+08
Done: 1940 iter / 223.753 secs = 344205 Gflops
```
Next - I'll try the `quarter` preconditioner-prec

walkloud:
@maddyscientist pointed out that my old run had FLOAT8 enabled which explains the speed up over the original MSPCG test at `322.422 seconds` (named pre pull-request) to `249.382 secs`

maddyscientist:
So in summary we have:
original CG, 6 steps: `5451 iter / 419.571 secs = 25973.1 Gflops`
MSPCG old default, 6 steps: `2206 iter / 322.422 secs = 207188 Gflops`
MSPCG old FLOAT8, 6 steps: `2205 iter / 249.382 secs = 267757 Gflops`
MSPCG new FLOAT8, 6 steps: `2105 iter / 227.981 secs = 279610 Gflops`
MSPCG new FLOAT8, 8 steps: `1940 iter / 223.753 secs = 344205 Gflops`

hummingtree:
@walkloud thanks for the test.

Within a reasonable range increasing `maxiter-preconditioner` reduces the (outer) iteration count, but it increases the per iteration cost. As a result you would just try different `maxiter-preconditioner` and pick the fastest one. Usually the optimal `maxiter-preconditioner` falls between `6` and `8`.

So even without `--recon 12` we are having a `1.9x` speed up over plain CG. This is really great. I would be really excited to see the results with quarter precision as well.

Just before I left Columbia I was able to get some further algorithmic improvement for MSPCG: the preconditioner can be run with reduced `Ls`, which makes it even faster. In order to do that you need to train a transition matrix that interpolates between different `Ls`'s (the original `Ls` and the reduced one). Once trained, the transition matrix can be used for any configuration.

I haven't got a chance to wrap the work up though.

walkloud:
ok, with recons 12, and doing a few solves with, looks like the optimal maxiter-precon is 5.  I've got it running up to 12.  Also - quarter precon-prec didn't pan out - solver couldn't get |r|/|b| below 9e-1.  I'll have to re-run the CG test with recons 12 for a baseline
```
mspcg_78_n16_test.new.maxprecon_2.stdout:5 solves, with mean solve time 252.413 (stddev = 18.2189), mean GFLOPS 184024 (stddev = 3814.8) [excluding first solve]
mspcg_78_n16_test.new.maxprecon_3.stdout:5 solves, with mean solve time 216.422 (stddev = 3.98378), mean GFLOPS 219442 (stddev = 3744.37) [excluding first solve]
mspcg_78_n16_test.new.maxprecon_4.stdout:5 solves, with mean solve time 193.673 (stddev = 3.47388), mean GFLOPS 252447 (stddev = 4376.48) [excluding first solve]
mspcg_78_n16_test.new.maxprecon_5.stdout:5 solves, with mean solve time 189.58 (stddev = 2.02506), mean GFLOPS 286402 (stddev = 2243.74) [excluding first solve]
mspcg_78_n16_test.new.maxprecon_6.stdout:5 solves, with mean solve time 199.885 (stddev = 0.799283), mean GFLOPS 319214 (stddev = 3337.02) [excluding first solve]
```

walkloud:
@hummingtree and @maddyscientist update - cg solve with float8
```
5 solves, with mean solve time 279.503 (stddev = 5.13788), mean GFLOPS 39231.5 (stddev = 639.303) [excluding first solve]
```
and the `maxprecon-iter=5` from the previous comment is the sweet spot for this set up with args
```
for maxprecon in $(seq 2 12); do
ARGS="--dslash-type mobius $dim_grid --nsrc 5 \
      --Lsdim 12 --m5 -1.1 --b5 1.5 --c5 0.5 --mass 0.00152 \
      --load-gauge ${production}/${ens}/cfgs_flow/${ens_long}.${CONF}_wflow1.0.lime \
      --verbosity verbose \
      --solution-type mat --matpc even-even --solve-type normop-pc \
      --prec double --recon 12 --niter 14000 --prec-sloppy half --reliable-delta 0.1 --prec-precondition half \
      --inv-type pcg --precon-type cg --precon-schwarz-type additive --tol-precondition 1e-6 --maxiter-precondition ${maxprecon} "
```

hummingtree:
@walkloud So a `279.503/189.58=1.47x` speed up: this is actually a less suspicious speed up number, at least from my experience. 

I see that the plain CG performance has increased a lot: is this all due to the use of `FLOAT8`?

Also the failure of using quarter precision for the preconditioner seems like a bug: from my previous experience it **should** converge, just a bit slower than using half precision. You are using the double-half-quarter (i.e. half precision for sloppy) setup, right?

maddyscientist:
Regarding CG being faster now: @walkloud is now using multiple source per run, and excluding the first solve time whereas before was just doing one solve per run.  This ensures that we don't count any memory allocation or MPI startup overheads.

cpviolator:
@walkloud Where can I find the run scrips for the above tests? I'll spin up https://github.com/lattice/quda/pull/1031 and load in some eigenvectors for deflation. We now have the ability to deflate a subset of the loaded eigenvectors, so that will be a double study to find the sweet spot in terms of deflation space size and node size.

walkloud:
@hummingtree args used for quarter
```
ARGS="--dslash-type mobius $dim_grid \
      --Lsdim 12 --m5 -1.1 --b5 1.5 --c5 0.5 --mass 0.00152 \
      --load-gauge ${production}/${ens}/cfgs_flow/${ens_long}.${CONF}_wflow1.0.lime \
      --verbosity verbose \
      --solution-type mat --matpc even-even --solve-type normop-pc \
      --prec double --niter 14000 --recon 12 --prec-sloppy half --reliable-delta 0.1 --prec-precondition quarter \
      --inv-type pcg --precon-type cg --precon-schwarz-type additive --tol-precondition 1e-6 --maxiter-precondition 6 "
```
@cpviolator script can be found here on summit
```
/ccs/proj/lgt100/c51/x_files/project_2/metaq/test_lanczos/mspcg_a09m135_cfg_78_n16_test_new.sh
```

PR #1028, opened 2020-07-02
Hotfix/multiblas

maddyscientist:
This is a hotfix PR that fixes multiple issues with the blas kernels, some of which were introduced in #1021 and some were long since present.
* Add mixed-precision support to blas_test for better test coverage
* blas_test is now more robust: use enums map to set which kernels to do as opposed to relying on some set index order 
* Fix `tuneKey` for mixed-precision blas kernels
* Fix mixed-precision multi-1d multi-blas kernels
* Fix mixed-precision multi-reduce kernels (cannot use transpose tuner for these cases)
* Fix multi-blas compilation for pure fixed-point compilation
* Fix broken mixed-precision `axpyCGNorm`
* Reduced some template bloat in multi-blas

### DIFF
--- a/include/multi_blas_helper.cuh
+++ b/include/multi_blas_helper.cuh
@@ -42,7 +42,13 @@ namespace quda
        @tparam fixed Whether we are using fixed point
        @return Max power of two
      */
+#if QUDA_PRECISION <= 3
+    // if we only have a fixed-point build then we need this WAR to avoid some invalid template instantiations
+    // this is temporary - can be removed once the norm and v pointers are fused
+    template <bool reducer, bool fixed> constexpr int max_NXZ_power2() { return reducer ? 16 : 64; }
+#else
     template <bool reducer, bool fixed> constexpr int max_NXZ_power2() { return reducer ? 16 : (fixed ? 64 : 128); }
+#endif
 
     /**
        @brief Return the maximum power of two enabled by default for
--- a/include/multi_blas_helper.cuh
+++ b/include/multi_blas_helper.cuh
@@ -92,7 +92,7 @@ namespace quda
       using SpinorX = Spinor<xType, 4>;
       using SpinorY = Spinor<yType, 4>;
       using SpinorZ = SpinorX;
-      using SpinorW = Spinor<xType, 4>;
+      using SpinorW = SpinorX;
 
       // compute the size remaining for the Y and W accessors
       constexpr int arg_size = (MAX_ARG_SIZE - sizeof(int)                                    // NYW parameter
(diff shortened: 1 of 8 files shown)


tests/blas_test.cpp
```diff
@@ -41,102 +65,146 @@ void display_test_info()
...
-                                        "caxpy_block",
-                                        "axpyBzpcx_block",
-                                        "caxpyBxpz",
-                                        "caxpyBzpx",
-                                        "cDotProductNorm_block",
-                                        "cDotProduct_block",
-                                        "reDotProductNorm_block",
-                                        "reDotProduct_block",
-                                        "axpy_block"};
+const std::map<Kernel, std::string> kernel_map
```
mathiaswagner, line 119: We already do have some maps like that for CLI11, I need to look up what. Maybe we can reuse part of that.

include/multi_blas_helper.cuh
```diff
@@ -80,9 +86,14 @@ namespace quda
...
        allows for a much larger NXZ (NYW) when NYW (NXZ) is small.
     */
-    template <int NXZ, typename SpinorX, typename SpinorY, typename SpinorZ, typename SpinorW, typename Functor>
+    template <int NXZ, typename xType, typename yType, typename Functor>
     inline constexpr int max_YW_size()
     {
+      using SpinorX = Spinor<xType, 4>;
+      using SpinorY = Spinor<yType, 4>;
+      using SpinorZ = SpinorX;
+      using SpinorW = Spinor<xType, 4>;
```
mathiaswagner: just style: why not use SpinorW=SpinorX here as well (as for SpinorZ) ?

tests/blas_test.cpp
```diff
@@ -256,72 +295,58 @@ void initFields(int prec)
...
-  delete lD;
+  delete voD;
+  delete woD;
+  delete xoD;
+  delete yoD;
+  delete zoD;
   delete xmD;
   delete ymD;
   delete zmD;
+  delete xmoD;
```
mathiaswagner, line 340: o'unique_ptr where art thou ;-)

mathiaswagner (approved):
Looks good and ctest passed for me.

tests/blas_test.cpp
maddyscientist: 😆

include/multi_blas_helper.cuh
maddyscientist: I guess I was tired.  Will push a quick fix for this.

tests/blas_test.cpp
maddyscientist: I just had a quick look at this.  I don't think CLI11 quite does what we want here.  The `CLI::TransformPairs` is a map that maps a string to an enum.  We need the other way round here, e.g., map an enum to a string.  Perhaps it could be be reused to do what we want, but for now I think sticking with the `std::map` is fine.

mathiaswagner: Agree, just want to avoid that we need to maintain multiple copies of the same maps.

weinbe2 (approved):

PR #1029, opened 2020-07-03
Added ColorSpinor templates and independent reductions

cpviolator:
This cherry-pick (and a commit) adds some tweaks needed for our the stochastic laph-NN code:

1. From @maddyscientist: Add an optional argument to ReduceArg where the argument is the number of independent reductions requested in the kernel and if this argument is non zero, there is a run-time check to ensure that the number of reductions is not greater than the maximum number possible.
2. Two minor additions to the ColorSpinor struct. One allows the user to call color contractions of spinors of different spin (eg, a spin 1 laplacian eigenvector and a spin 4 Wilson like vector) to give an open spin index contraction. The second is the addition of a color cross product `\epsilon_{ijk} A_{j} B_{k} = C_{i}` which is hardcoded to use 3 colors only.

### DIFF
--- a/include/color_spinor.h
+++ b/include/color_spinor.h
@@ -920,6 +920,28 @@ namespace quda {
   }
 
   /**
+     @brief Compute the color contraction over color at spin s
+     dot = \sum_s,c a(s,c) * b(s,c)
+     @param a Left-hand side ColorSpinor
+     @param b Right-hand side ColorSpinor
+     @return The color contraction
+  */
+  template <typename Float, int Nc, int Ns>
+  __device__ __host__ inline complex<Float> colorContract(const ColorSpinor<Float, Nc, Ns> &a,
+							  const ColorSpinor<Float, Nc, Ns> &b, int sa, int sb)
+  {
+    complex<Float> dot = 0;
+    for (int c = 0; c < Nc; c++) {
+      dot.x += a(sa, c).real() * b(sb, c).real();
+      dot.x -= a(sa, c).imag() * b(sb, c).imag();
+      dot.y += a(sa, c).real() * b(sb, c).imag();
+      dot.y += a(sa, c).imag() * b(sb, c).real();
+    }
+    
+    return dot;
+  }
+
+  /**
      Compute the inner product over color at spin s between two ColorSpinor fields
      dot = \sum_c conj(a(s,c)) * b(s,c)
      @param a Left-hand side ColorSpinor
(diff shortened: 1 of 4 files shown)


include/color_spinor.h
```diff
@@ -981,6 +1003,28 @@ namespace quda {
...
 
+  /**
+     Compute the cross product of two color vectors at spin sa and sb
+     cProd = \sum_{j,k} \epsilon_{i,j,k} a(s1,j) b(s2,k)     
+     @param a j ColorSpinor
+     @param b k ColorSpinor
+     @param sa j spin index
+     @param sb k spin index
+     @return The cross product
+  */
```
maddyscientist, line 1015: This function takes templated Nc but appears to be hard coded for Nc=3.

cpviolator: Good point! I'll remove the Nc=3 template... not in a hurry to abstract an Nc Levi-Civita.

include/color_spinor.h
```diff
@@ -919,6 +919,28 @@ namespace quda {
...
+  __device__ __host__ inline complex<Float> colorContract(const ColorSpinor<Float, Nc, Ns> &a,
+							  const ColorSpinor<Float, Nc, Ns> &b, int sa, int sb)
+  {
+    complex<Float> dot = 0;
+    for (int c = 0; c < Nc; c++) {
+      dot.x += a(sa, c).real() * b(sb, c).real();
+      dot.x -= a(sa, c).imag() * b(sb, c).imag();
+      dot.y += a(sa, c).real() * b(sb, c).imag();
+      dot.y += a(sa, c).imag() * b(sb, c).real();
+    }
```
maddyscientist, line 939: The correct way to write terms like this with modern C++ complex is something like the following.
```c++
      dot.real(dot.real() + a(sa, c).real() * b(sb, c).real());
      dot.real(dot.real() - a(sa, c).imag() * b(sb, c).imag());
      dot.imag(dot.imag() + a(sa, c).real() * b(sb, c).imag());
      dot.imag(dot.imag() + a(sa, c).imag() * b(sb, c).real());
```
It looks a bit messy, but probably good if we move away from the `.x` and `.y` in favour of more standard C++.

cpviolator: I'll change the other two instances that use `dot.x` and `dot.y` while I'm at it.

include/color_spinor.h
```diff
@@ -950,10 +972,10 @@ namespace quda {
...
 #pragma unroll
     for (int c = 0; c < Nc; c++) {
-      dot.x += a(sa, c).real() * b(sb, c).real();
-      dot.x += a(sa, c).imag() * b(sb, c).imag();
-      dot.y += a(sa, c).real() * b(sb, c).imag();
-      dot.y -= a(sa, c).imag() * b(sb, c).real();
+      dot.real(dot.real() + a(sa, c).real() * b(sb, c).real());
+      dot.real(dot.real() + a(sa, c).imag() * b(sb, c).imag());
+      dot.imag(dot.imag() + a(sa, c).real() * b(sb, c).imag());
+      dot.imag(dot.imag() - a(sa, c).imag() * b(sb, c).real());
```
maddyscientist: clang-format needed

maddyscientist:
This looks fine (beyond the Jenkins build failures).  I would correct the description above though.  This addition does not do  point 1.) above.  What this PR does is add an optional argument to `ReduceArg` where the argument is the number of independent reductions requested in the kernel and if this argument is non zero, there is a run-time check to ensure that the number of reductions is not greater than the maximum number possible.

maddyscientist (approved):

PR #1032, opened 2020-07-06
Hotfix/ldg

maddyscientist:
Only use `__ldg` on Kepler to Maxwell 1 architectures, else use generic loads.  On Maxwell 2 on upwards this will lead to cached loads, giving equivalent performance and behaviour to `__ldg`.


weinbe2 (approved):
I can confirm these changes fix the compile time issues (sanity restored) and fix the TransposeTuner issues that incidentally popped up while testing. For future reference, `blas_test --nsrc 8 --msrc 9` had triggered the issue, while `--nsrc 9 --msrc 8` did not.

Good to merge once the build tests have completed.

PR #1033, opened 2020-07-08
Fix bug in NXZ template instantiation in multi-blas and multi-reduce

maddyscientist:
Bug fix for multi-blas instantiation.


weinbe2:
FYI actively testing this hotfix, should be done decently soon.

weinbe2 (approved):
Fix confirmed, approving+merging now.

PR #1034, opened 2020-07-15
Feature/device

maddyscientist:
The PR continues the abstraction effort and cleanup some old experiments:
* Move CUDA specific initialization and destruction from `initQuda` and `endQuda`, moving to new target-specific file lib/targets/cuda/device.cpp
* Abstracted `cudaHostGetDevicePointer` to new function, `get_mapped_device_pointer` in malloc.cpp
* In preparation of merging NVSHMEM, remove old asynchronous dslash policy experiments

Notably this PR removes all CUDA specifics from interface_quda.cpp

### DIFF
--- a/lib/dslash_policy.cuh
+++ b/lib/dslash_policy.cuh
@@ -19,11 +19,6 @@ namespace quda
     // Auxiliary work that can be done while waiting on comms to finish
     extern Worker *aux_worker;
 
-#if CUDA_VERSION >= 8000
-    extern cuuint32_t *commsEnd_h;
-    extern CUdeviceptr commsEnd_d[Nstream];
-#endif
-
     // these variables are used for benchmarking the dslash components in isolation
     extern bool dslash_pack_compute;
     extern bool dslash_interior_compute;
@@ -234,8 +229,6 @@ namespace quda
      - if gdr or zero-copy for the receive buffer then we have nothing
        else to do, it is now safe to post halo kernel
 
-     - if staging with async, we release the scatter by setting the approriate commsEnd_h flag
-
      - if basic staging, we post the scatter (host to device memory copy)
 
      @param[in,out] in Field being commicated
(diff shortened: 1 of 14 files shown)


lib/dslash_policy.cuh
```diff
@@ -1862,8 +1603,6 @@ namespace quda
     QUDA_FUSED_ZERO_COPY_PACK_GDR_RECV_DSLASH,
     QUDA_DSLASH_FUSED_PACK,
     QUDA_DSLASH_FUSED_PACK_FUSED_HALO,
-    QUDA_DSLASH_ASYNC,
-    QUDA_FUSED_DSLASH_ASYNC,
     QUDA_DSLASH_NC,
```
weinbe2, line 1606: Are there any hardcoded comparisons of dslash policies with integers that may bork because the integer value of `QUDA_DSLASH_NC` has changed?

maddyscientist:
Nope

lib/interface_quda.cpp
```diff
@@ -289,28 +267,28 @@ static void profilerStart(const char *f) {
   static int target_count = 0;
   static unsigned int i = 0;
   if (do_not_profile_quda){
-    cudaProfilerStop();
+    device::profile::stop();
```
weinbe2, line 270: I'm a little confused about why this logic exists in the first place... if we're not profiling QUDA, why do we need to "stop" profiling QUDA in `profilerStart`? I could see a world where this maybe lets users update `QUDA_DO_NOT_PROFILE` from the environment in the middle of a run, but I still don't see how the logic of "stopping" when you're requesting a start makes sense.

lib/interface_quda.cpp
```diff
@@ -289,28 +267,28 @@ static void profilerStart(const char *f) {
...
     }
   }
 }
 }
 
 static void profilerStop(const char *f) {
-  if(do_not_profile_quda){
-    cudaProfilerStart();
+  if (do_not_profile_quda){
+    device::profile::start();
```
weinbe2, line 286: Same question as above, just flipped, why is this here in the first place?

weinbe2:
A few questions, but otherwise this PR looks reasonable. Happy to see all explicit CUDA references are gone from `interface_quda.cpp`. What testing has been done? (Though I guess this is a pretty compact "you'll know if you messed up quickly at runtime" PR.)

cpviolator (approved):
Evan raises some good points. As for the dslash policy enumeration, I know that we have some scripts that enable only a subset of policies to save on tuning time, but none of them go as high as the `NC` policy.

lib/dslash_policy.cuh
maddyscientist: Nope, and in fact `QUDA_DSLASH_NC` isn't used at all any more and is another legacy policy that could be deleted (It's from the original Möbius implementation that Hyung-Jin wrote).  I figured it could have some use at some point to have a no-comms policy though, so kept it for now.

maddyscientist:
Thanks for the reviews folks.  

@weinbe2 testing done is is both single and multi-GPU, and also with `QUDA_NVML` enabled.
@cpviolator the policy enumeration hasn't been changed for any polices that are actually used by anyone, so it won't affect any scripts.

weinbe2:
I'm going to test this with Chroma+QDPJIT, hold off on merging it at least until I give it a pass there.

weinbe2 (approved):
Passes the sanity tests. Good to merge.

PR #1035, opened 2020-07-17
Feature/milc multishift double

weinbe2:
This PR changes the behavior of `qudaMultishiftInvert` and `qudaInvert` to always use `double` as the outer precision, independent of if `float` or `double` is passed in as the outer precision. The environment variable `QUDA_MILC_OVERRIDE_DOUBLE_MULTISHIFT` can be set to override this behavior and return to the original behavior. This PR is meant to avoid wasteful single precision refinement solves when the requested tolerance is below single precision tolerance, ~1e-6 to 1e-7. In practice this PR should have no effect on outer `double` solves, which has been verified as noted below.

This PR also addresses what seems to be a bug in `qudaMultishiftInvert` where `invalidateGaugeQuda()` isn't called when `canReuseResidentGauge()` returns false. This check was already present in `qudaInvert`. I'm not exactly sure why this wasn't a problem before; perhaps it's because this is the first time in practice the fine precision is changed within a production workflow, in any case it seems like we'd been running on borrowed time. Not including this fix led to `errorQuda` being called due to precision mismatches between resident fields (some in single, some in double).

This PR has been tested against MILC RHMC in the following incantations:

- NERSC small in all combinations of overall precisions (double and single) and mixed precision settings (0 = no mixed precision, 1 = single precision sloppy, 2 = half precision sloppy), where there is no regression in performance except for (single, no mixed precision). This later case is reasonable because pure single does not have a refinement step while the implicit change to `double-single` triggers a refinement step within the multishift solve, which is arguably a rare case.
- NERSC medium in pure double to verify no regression, comparing `develop` with this branch both with and without `QUDA_MILC_OVERRIDE_DOUBLE_MULTISHIFT` set, demonstrating no regression above noise and no deviations in output observables (until the 14th digit but, I mean, come on)
- MILC single precision ECP RHMD benchmark, where a 144^3x288 volume, 96 node run goes from taking ~5500 seconds to ~5200 seconds

In the later case, there are no deviations in observables until the fourth digit, and delta H remains "sane".

One negative side effect is the time spend downloading the gauge field goes up for single precision outer solves (for ex, going from 15 sec to 90 sec in the ECP RHMD benchmark). This is explained by extra downloads required by the precision changes. This is more than amortized by the sharp reduction in single-source refinement iterations. This overhead will also disappear in future gauge residency work in MILC RHMD.

### DIFF
--- a/lib/milc_interface.cpp
+++ b/lib/milc_interface.cpp
@@ -977,8 +977,8 @@ void qudaInvert(int external_precision, int quda_precision, double mass, QudaInv
 
   // override fine precision to double, switch to mixed as necessary
   if (!do_not_force_double && device_precision == QUDA_SINGLE_PRECISION) {
+    // force outer double
     device_precision = QUDA_DOUBLE_PRECISION;
-    // invalidateGaugeQuda();
   }
 
   QudaGaugeParam fat_param = newQudaGaugeParam();
--- a/lib/milc_interface.cpp
+++ b/lib/milc_interface.cpp
@@ -858,17 +858,19 @@ void qudaMultishiftInvert(int external_precision, int quda_precision, int num_of
   static bool force_double_queried = false;
   static bool do_not_force_double = false;
   if (!force_double_queried) {
-    char* donotusedouble_env = getenv("QUDA_MILC_OVERRIDE_DOUBLE_MULTISHIFT"); // disable forcing outer double precision
-    if (donotusedouble_env && (!(strcmp(donotusedouble_env, "0") == 0)))  {
-      do_not_force_double=true;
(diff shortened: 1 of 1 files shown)


mathiaswagner:
Should we query the requested tolerance? I.e. as I understand it even if you just request 1e-4 this would force double?

weinbe2:
We discussed this and decided it was safest to always use double. The justification was it's not just the epsilon of the precision that matters, but also the conditioning of the matrix, so it's hard to just pick an epsilon and commit to it in a future+proof sense.

mathiaswagner (approved):
Visual review looks good to me.

PR #1025, opened 2020-06-26
Feature/clang2

bjoo:
This is my second attempt to change the Cmake setup to accommodate clang (duplicate of #1009)
I checked on summit that both NVCC and Clang builds (using suitably recent Makefile) compile. 
@mathiaswagner and I will reconcile with #1009 but I would like to see if it keeps Jenkins happy as is.

PR #1036, opened 2020-07-21
Feature/qio largefile

weinbe2:
This draft PR adds support for the `feature/largefile` branch of QIO (https://github.com/usqcd-software/qio/tree/feature/largefile) to QUDA. This branch of QIO is required for global volumes > 2^31-1 (current use case: MILC 192x192x192x384), though it works without issue for smaller volumes as well---it's mostly appropriate `int` -> `size_t` replacements.

The `feature/largefile` branch is QIO-interface breaking because it changes external API signatures, which I imagine is why it is not on the roadmap for a formal merge into the `master` branch.

This PR itself adds support for the `feature/largefile` branch of QIO via a new advanced boolean `cmake` variable `QUDA_QIOLARGEFILE`. This needs to be set when linking to an external build of the `feature/largefile` QIO branch. Alternatively, if you use QUDA's support for downloading and compiling QIO via `QUDA_DOWNLOAD_USQCD`, it'll download the `qio2-5-0` tag (existing behavior) or `feature/largefile` appropriately.

The `QUDA_QIOLARGEFILE` define effects a new `typedef` in `quda_define.h.in`, `qio_size_t`, which is set to `int` or `size_t` as appropriate. The reason why we can't blanket set it to `size_t` is because of the QIO APT changes.

This PR is a draft for now because there are outstanding tests noted above. I wanted to get this draft PR out for review before tests are completed because any changes may require redoing the tests (or perhaps additional tests).

- [x] Sanity check: test that gauge field io (via `heatbath_test`) works independent of/between largefile on/off builds
- [x] Sanity check: test that per-rank io with files > 2^31-1 works without issue
- [ ] Test that io works for global volumes > 2^31-1 (will use Summit to test)
- [ ] clang-format
- [ ] New: reconcile with changes in #1025 once it's merged

Pinging @mathiaswagner for general `cmake` review, @cpviolator for USQCD-download specific review, @maddyscientist for overall thoughts


CMakeLists.txt
```diff
@@ -540,7 +545,7 @@ if(QUDA_QMP)
       GIT_SHALLOW YES
       PREFIX usqcd
       CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} <SOURCE_DIR>/configure "INSTALL=${INSTALL_EXE} -C"
-                        --with-qmp-comms-type=MPI --prefix=<INSTALL_DIR>
+                        "CFLAGS=-Wall -O3 -std=c99 -g" --with-qmp-comms-type=MPI --prefix=<INSTALL_DIR>
```
maddyscientist, line 584: What's the motivation of `-g` by default here?

CMakeLists.txt
```diff
@@ -284,6 +287,8 @@ mark_as_advanced(QUDA_OPENMP)
 
 mark_as_advanced(QUDA_BACKWARDS)
 
+mark_as_advanced(QUDA_QIOLARGEFILE)
```
maddyscientist: I think this would be best written as `QUDA_QIO_LARGE_FILE` to give better readability.

maddyscientist (commented):
Look fine in its current form.  @weinbe2 you mentioned there are still issues here for saving, but it works for loading.  What's your plan on addressing those?

CMakeLists.txt
weinbe2: Ease of debugging when needed, combined with the generally negligible performance overheads of `-g`. (It's also what I've used in the past, plus what Carleton gave me for reference, though those aren't necessarily good long term reasons.)

I can add the logic to only include `-g` for non-`RELEASE` builds of QUDA, if that's preferred.

maddyscientist: That would be preferred, I think yes, if it's easy enough to add.  Thx

weinbe2:
I'll add `-g` and make the change to `QUDA_QIO_LARGE_FILE` tomorrow.

> Look fine in its current form. @weinbe2 you mentioned there are still issues here for saving, but it works for loading. What's your plan on addressing those?

I've only verified that files of all sizes work and that global volumes <= 2^31 work. I haven't gotten saving or loading of files corresponding to global volumes >= 2^31 working yet.

The most recent error I hit was related to `c-lime`, which doesn't have an analogous `largefile` branch (last I checked). I reached out to @detar but haven't had the chance to follow up. Once that's been addressed and I can verify io works through QUDA I can take this out of draft status.

maddyscientist:
Ah yes, thanks for the reminder of error, and that it was `c-lime` related.  In that case, once the fixes above are done, we'll be all set on this PR and we can merge.

weinbe2:
FYI: this work is on hold pending work by @jcosborn @detar and others on formalizing largefile support in QIO. Work on this PR will be resumed once that's completed.

weinbe2:
Closing this, will replace with a new PR momentarily.

weinbe2:
Actually closing this, will replace with a new PR momentarily.

PR #1039, opened 2020-07-28
Feature/atomic reduce

maddyscientist:
* Introduces new `std::atomic`-based signaling atomics, avoiding the need for explicit host-device synchronization for lower latency reductions (upto 1.45x faster reductions on POWER9).  These can be enabled on Pascal+ with CUDA 11.0 and above, with the cmake variable `-DQUDA_HETEROGENEOUS_ATOMIC=ON`.
* Remove prior unused reduction experiments
* Clean up of reductions: separated CUB and moved general reduction kernel to `reduction_helper.h`, moved generic reduction memory handling to new file reduction_helper.cu away from reduce_quda.cu 
* Extended `blas_test` to test asynchronous reductions with the `caxpyXmazMR` kernel
* Mixed-precision blas kernels now compute their sustained bandwidth correctly
* Fixes bug in mixed-precision BiCGStab introduced with #1021 
* Cleanup of `plaq_test`


mathiaswagner:
Are there any reasons not make `QUDA_HETEROGENEOUS_ATOMICS=ON` the default on CUDA 11 and pascal+?

maddyscientist:
At the moment a `SANITIZE` build will fail with heterogeneous atomics, since libcu++ doesn't support it.  Given the narrow scope of heterogeneous reductions, I'm probably ok with that, and enabling by default.

maddyscientist:
I have enabled it by default now, with cmake disabling it automatically if CUDA < 11 or a `SANITIZE` build is used.

cpviolator (approved):
The reduction procedure at the end of the kernel calls looks much neater, and great news for Power9 users!

maddyscientist:
Just to note that extended testing has revealed some issues that need to be fixed.  I will need to address these before we can merge.

CMakeLists.txt
```diff
@@ -218,6 +218,20 @@ option(QUDA_NUMA_NVML "experimental use of NVML to set numa affinity" OFF)
...
+if(QUDA_HETEROGENEOUS_ATOMIC AND CUDA_VERSION LESS 11.0)
+  message(
+  NOTICE
+  "Disabling heterogeneous atomic support (requires CUDA >= 11.0)")
+  set(QUDA_HETEROGENEOUS_ATOMIC OFF)
+endif()
+if(QUDA_HETEROGENEOUS_ATOMIC AND ${CMAKE_BUILD_TYPE} STREQUAL "SANITIZE")
+  message(
+  NOTICE
+  "Disabling heterogeneous atomic support (does not support SANITIZE build)")
```
weinbe2: I think we should make this an error so there are no surprises.

maddyscientist: Fixed

mathiaswagner: Not sure this is the best choice for Jenkins: 
```
- QUDA 1.0.0 (v0.9.0-3131-g0f8dc2108) **
-- cmake version: 3.15.7
-- Source location: <https://nvsocal2.caltech.edu:8443/job/qudabase_ubuntu1804_cuda102_cmake315/COMMS=-DQUDA_QMP=ON,GPUARCH=sm_70/ws/>
-- Build location: <https://nvsocal2.caltech.edu:8443/job/qudabase_ubuntu1804_cuda102_cmake315/COMMS=-DQUDA_QMP=ON,GPUARCH=sm_70/ws/native>
-- Build type: STRICT
-- QUDA target: CUDA
CMake Error at CMakeLists.txt:214 (message):
  QUDA_HETEROGENEOUS_ATOMIC=ON requires CUDA >= 11.0
```

mathiaswagner: And probably this will cause a lot of user support for those still on CUDA < 11 as with the default config they will run into this error.

include/reduce_helper.h
```diff
@@ -0,0 +1,237 @@
...
+#else
+      if (commAsyncReduction()) result_d = partial;
+#endif
+    }
+
+    template <typename host_t, typename device_t = host_t> void complete(host_t *result, const qudaStream_t stream = 0)
+    {
+#ifdef HETEROGENEOUS_ATOMIC
+      for (int i = 0; i < n_reduce * n_item; i++) {
+        while (result_h[i].load(cuda::std::memory_order_relaxed) == init_value<system_atomic_t>()) {}
```
mathiaswagner: could we use `result_h[i].wait(init_value<system_atomic_t>(),cuda::std::memory_order_relaxed)` here ?

lib/reduce_helper.cu
```diff
@@ -0,0 +1,115 @@
...
+    template <typename T> __global__ void init_count(T *counter, int n)
+    {
+      for (int i = 0; i < n; i++) new (counter + i) T {0};
+    }
+
+    void init()
+    {
+#ifdef HETEROGENEOUS_ATOMIC
+      // fast reductions use heterogeneous std::atomic to synchronize CPU and GPU as opposed to using CUDA events
+      warningQuda("Experimental heterogeneous-atomic reductions enabled");
```
mathiaswagner: guess that should go away when we no longer consider it experimental

maddyscientist: Yes, I should just delete that.  Will do so in next push.

include/reduce_helper.h
maddyscientist: I hadn't realized this was included in CUDA 11.0's libcu++ since it's a C++20 feature, but I see it's there.  Will fix this in next push.

maddyscientist:
Ok, comments above all addressed.  With luck Jenkins will now build for all configs 🤞

CMakeLists.txt
```diff
@@ -434,6 +435,28 @@ else()
   find_package(CUDAToolkit)
 endif()
 
+include(CMakeDependentOption)
+
+if(CUDAToolkit_VERSION VERSION_GREATER_EQUAL "11.0" AND CMAKE_CUDA_COMPILER_ID MATCHES "NVIDIA" AND NOT QUDA_JITIFY)
```
mathiaswagner: confused? why don't you use CMAKE_DEPENDENT_OPTION here?
That should make all the error if statements unnecessary

maddyscientist: It makes it more explicit to the user if they to try to enable it and it tells them that you can't.  So we have a working default for Jitify and clang cuda, but we don't secretly disable it if the user asks for it.  This seems like the right balance.

mathiaswagner: So in that case`include(CMakeDependentOption)` is unnecessary. 
Can't we just put NVCC and CUDA>= 11 in the documentation
I lean towards CMAKE_DEPENDENT_OPTION to keep the CMakeLIsts shorter.

include/blas_helper.cuh
```diff
@@ -10,35 +10,14 @@
 namespace quda
 {
 
-  template <bool X_ = false, bool Y_ = false, bool Z_ = false, bool W_ = false, bool V_ = false> struct write {
+  template <bool X_ = false, bool Y_ = false, bool Z_ = false, bool W_ = false, bool V_ = false> struct stream {
```
mathiaswagner: just wondering whether stream is the best name. I don't currently have a better suggestion.

mathiaswagner (approved):

include/blas_helper.cuh
weinbe2: `mem_out`? Not saying that's better, but I agree we should avoid using "stream" for anything that isn't a CUDA/etc stream...

maddyscientist: It's for both input and output, so `out` in the type name isn't appropriate.  `memory_stream` perhaps?

weinbe2: Sure, or `mem_io`. Depends on how pedantic we want to be about not using `stream`, even from a `grep` sense.

include/launch_kernel.cuh
```diff
@@ -309,37 +126,11 @@ default:								\
   case 144: kernel<144, __VA_ARGS__><<<tp.grid, tp.block, tp.shared_bytes, stream>>>(arg); break;                      \
   case 192: kernel<192, __VA_ARGS__><<<tp.grid, tp.block, tp.shared_bytes, stream>>>(arg); break;                      \
   case 200: kernel<200, __VA_ARGS__><<<tp.grid, tp.block, tp.shared_bytes, stream>>>(arg); break;                      \
+  case 250: kernel<250, __VA_ARGS__><<<tp.grid, tp.block, tp.shared_bytes, stream>>>(arg); break;                      \
```
weinbe2, line 129: Since we now know `5`s with `2`s are in play, we should also add `125` (`5 5 5 2`) and `50` (`5 5 2 2`). (`100` is already covered.) I guess I could make an argument for `40` and `80` as well.

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -10,12 +10,11 @@
 
 #include <cub_helper.cuh>
 
-#if (__CUDACC_VER_MAJOR__ >= 9 && __COMPUTE_CAPABILITY__ >= 700)
+#if (CUDA_VERSION >= 9000 && __COMPUTE_CAPABILITY__ >= 700)
```
weinbe2, line 13: @hummingtree FYI, not sure if this is reconciled in your branch yet

include/launch_kernel.cuh
maddyscientist: That is completely orthogonal to this PR and can be covered elsewhere .

include/blas_helper.cuh
mathiaswagner: how about calling it `memory_access` ? Would make something like `memory_access<1,1,0,0,0> read; memory_access<1,0,0,0,0> write` ? Guess if we would live in a FORTRAN world `INTENT` would be the proper name.

maddyscientist: Ok, `memory_access` it is.  Will plumb this in with next push.

include/mdw_dslash5_tensor_core.cuh
maddyscientist: Note the need for this is for compatibility with clang: it's not a CUDA compiler so doesn't identify as such.  The cross compiler mechanism is to use `CUDA_VERSION` instead.

PR #1040, opened 2020-07-29
fixed compiler error for kepler

alexstrel:
Fixed compiler error that required const qualifier.


maddyscientist (approved):

maddyscientist:
Thanks Alexei

PR #1046, opened 2020-08-11
Feature/abstraction

### LINKED ISSUE #985: gauge_alg_test is flaky with multi-GPU
Some issues seem to be present with `gauge_alg_test`

* with multi-gpu can hang when autotuning.
* the following has a failure but single precision does not (and doesn't fail if tuning is disabled)
```
mpirun -np 3 tests/gauge_alg_test --dim 8 8 8 8 --gridsize 1 1 1 3 --prec double
```

maddyscientist:
This PR continues the CUDA API abstraction, fixes various issues and extends the unit tests.
* Cleanup of the API abstraction, with all `quda` APIs now doing internal error checking with no external CUDA dependence.
* Added API abstraction for `qudaMemPrefetchAsync` 
* All kernel launches in QUDA now use `qudaLaunchKernel` abstraction as opposed to triple chevron syntax
* Improve robustness of heterogeneous atomics: added run-time checks to prevent possible deadlock; the members of `ReduceArg` are now private to prevent user contamination of the reduction buffers used for signaling (fixes race condition with plaquette and topological charge)
* Rename `TunableLocalParity` to `TunableLocalParityReduction` with only reduction kernels using this Tunable abstraction now.  All reduction kernels that use this now inherit the accelerated auto tuning previously deployed for the plaquette.
* Fix some old bugs with hangs in gauge fixing kernels (MPI_AllReduce performed during tuning), closes #985 
* Fix dynamic clover dslash_test and dslash_ctest when double-precision is not enabled
* Cleanup of gauge fixing and pure gauge evolution code: now respects `QUDA_PRECISION` and `QUDA_RECONSTRUCT` for optimized compilation trajectory
* Add `gauge_alg_test` to `ctest` list
* For `gauge_force_test` and `gauge_alg_test`, ctest will only double/single precision testing according to `QUDA_PRECISION` setting
* `ctest` usability improvements: will now only run on a single process if single-process mode is compiled, added `QUDA_TEST_NUMPROCS` environment variable which can be used to configure the number of processors used to run the unit tests (still requires rerunning `cmake` prior to `ctest`)
* Some general cleanup and dead-code removal


include/reduce_helper.h
```diff
@@ -99,21 +112,38 @@ namespace quda
 #endif
     }
 
-    template <typename host_t, typename device_t = host_t> void complete(host_t *result, const qudaStream_t stream = 0)
+    template <typename host_t, typename device_t = host_t>
```
weinbe2, line 122: doxygen

lib/gauge_fix_ovr.cu
```diff
@@ -1530,62 +1367,26 @@ public:
...
-        //printfQuda("QUDA_RECONSTRUCT_8\n");
-        numParams = 8;
-	typedef typename gauge_mapper<Float,QUDA_RECONSTRUCT_8>::type Gauge;
-        gaugefixingOVR<Float, 8>(Gauge(data), data, gauge_dir, Nsteps, verbose_interval, relax_boost, tolerance, reunit_interval, stopWtheta);
+  template <typename Float, int nColor, QudaReconstructType recon> struct GaugeFixingOVR {
+    GaugeFixingOVR(GaugeField& data, const int gauge_dir, const int Nsteps, const int verbose_interval,
+                   const Float relax_boost, const double tolerance, const int reunit_interval, const int stopWtheta)
+    {
+      using Gauge = typename gauge_mapper<Float, recon>::type;
+      if (gauge_dir != 3) {
```
weinbe2: Technically out of the scope of this PR, but I think this should throw an error if `gauge_dir` isn't strictly 3 or 4.

lib/pgauge_exchange.cu
```diff
@@ -3,358 +3,258 @@
...
-          comm_free(mh_recv_fwd[d]);
-          device_free(send_d[d]);
-          device_free(recv_d[d]);
-          device_free(sendg_d[d]);
-          device_free(recvg_d[d]);
-        #ifndef GPU_COMMS
-          host_free(hostbuffer_h[d]);
-        #endif
+      if (init) {
+        cudaStreamDestroy(GFStream[0]);
```
weinbe2, line 114: `qudaStreamDestroy`?

lib/reduce_helper.cu
```diff
@@ -78,8 +78,12 @@ namespace quda
...
-        init_count<<<1, 1>>>(reduce_count, QUDA_MAX_MULTI_REDUCE);
+        reduce_count = static_cast<count_t *>(device_malloc(max_n_reduce() * sizeof(decltype(*reduce_count))));
+        TuneParam tp;
+        tp.grid = dim3(1, 1, 1);
+        tp.block = dim3(1, 1, 1);
+
+        qudaLaunchKernel(init_count<count_t>, tp, 0, reduce_count);
       }
 
       cudaEventCreateWithFlags(&reduceEnd, cudaEventDisableTiming);
```
weinbe2, line 89: `qudaEventCreateWithFlags`? Or appropriate abstraction tbd?

lib/comm_common.cpp
```diff
@@ -261,12 +261,10 @@ void comm_peer2peer_init(const char* hostname_recv_buf)
 	  cudaDeviceCanAccessPeer(&canAccessPeer[1], neighbor_gpuid, gpuid);
 
 	  int accessRank[2] = { };
-#if CUDA_VERSION >= 8000  // this was introduced with CUDA 8
 	  if (canAccessPeer[0]*canAccessPeer[1] != 0) {
 	    cudaDeviceGetP2PAttribute(&accessRank[0], cudaDevP2PAttrPerformanceRank, gpuid, neighbor_gpuid);
```
cpviolator, line 265: Should this be abstracted?

lib/comm_common.cpp
```diff
@@ -261,12 +261,10 @@ void comm_peer2peer_init(const char* hostname_recv_buf)
...
 	    cudaDeviceGetP2PAttribute(&accessRank[1], cudaDevP2PAttrPerformanceRank, neighbor_gpuid, gpuid);
```
cpviolator, line 266: Should this be abstracted?

weinbe2:
TBD:

- [x] Test w/MILC RHMC (done by @maddyscientist)
- [ ] Test w/Chroma HMC (can do myself)
- [ ] Personal compile and test
- [ ] Iterate on PR comments

lib/gauge_fix_fft.cu
```diff
@@ -297,127 +254,74 @@ namespace quda {
...
-      : arg(arg){
-      cudaFuncSetCacheConfig( kernel_gauge_mult_norm_2D<Float>,   cudaFuncCachePreferL1);
-    }
-    ~GaugeFixINVPSP () {
+  public:
+    GaugeFixINVPSP(GaugeFixArg<Float> &arg, const GaugeField &meta) :
+      arg(arg),
+      meta(meta)
+    {
+      cudaFuncSetCacheConfig(kernel_gauge_mult_norm_2D<Float>, cudaFuncCachePreferL1);
```
cpviolator, line 302: Can we abstract this?

lib/gauge_fix_fft.cu
```diff
@@ -522,73 +426,55 @@ namespace quda {
...
+  public:
+    GaugeFixNEW(Gauge & dataOr, GaugeFixArg<Float> &arg, Float alpha, const GaugeField &meta) :
+      TunableVectorY(2),
+      dataOr(dataOr),
+      arg(arg),
+      meta(meta)
+    {
       half_alpha = alpha * 0.5;
-      cudaFuncSetCacheConfig( kernel_gauge_fix_U_EO_NEW<Float, Gauge>,   cudaFuncCachePreferL1);
+      cudaFuncSetCacheConfig(kernel_gauge_fix_U_EO_NEW<Float, Gauge>, cudaFuncCachePreferL1);
```
cpviolator, line 450: Abstraction?

lib/gauge_fix_fft.cu
```diff
@@ -622,74 +507,45 @@ namespace quda {
...
 
     public:
-    GaugeFix_GX(GaugeFixArg<Float> &arg, Float alpha)
-      : arg(arg) {
+    GaugeFix_GX(GaugeFixArg<Float> &arg, Float alpha, const GaugeField &meta) :
+      arg(arg),
+      meta(meta)
+    {
       half_alpha = alpha * 0.5;
       cudaFuncSetCacheConfig( kernel_gauge_GX<Elems, Float>,   cudaFuncCachePreferL1);
```
cpviolator, line 526: Abstraction.

lib/pgauge_exchange.cu
```diff
@@ -3,358 +3,258 @@
...
-        // look into storing these for later
-        mh_recv_back[d] = comm_declare_receive_relative(recv[d], d, -1, bytes[d]);
-        mh_recv_fwd[d]  = comm_declare_receive_relative(recvg[d], d, +1, bytes[d]);
-        mh_send_back[d] = comm_declare_send_relative(sendg[d], d, -1, bytes[d]);
-        mh_send_fwd[d]  = comm_declare_send_relative(send[d], d, +1, bytes[d]);
+      if (!init) {
+        X = (int*)safe_malloc(4 * sizeof(int));
+        for (int d = 0; d < 4; d++) X[d] = data.X()[d];
+
+        cudaStreamCreate(&GFStream[0]);
```
cpviolator, line 162: Abstraction.

lib/pgauge_exchange.cu
```diff
@@ -3,358 +3,258 @@
...
+        cudaStreamCreate(&GFStream[1]);
```
cpviolator, line 163: Abstraction.

lib/pgauge_exchange.cu
```diff
@@ -3,358 +3,258 @@
...
+        //extract bottom
+        arg.array = reinterpret_cast<complex<Float>*>(sendg_d[d]);
+        arg.borderid = data.R()[d];
+        qudaLaunchKernel(Kernel_UnPack<recon, Float, true, decltype(arg)>, tp, GFStream[1], arg);
 
-    #ifndef GPU_COMMS
-      comm_wait(mh_recv_fwd[d]);
-      cudaMemcpyAsync(recvg_d[d], recvg[d], bytes[d], cudaMemcpyHostToDevice, GFStream[1]);
-    #endif
+        cudaMemcpyAsync(send[d], send_d[d], bytes[d], cudaMemcpyDeviceToHost, GFStream[0]);
```
cpviolator, line 219: Abstraction.

cpviolator: Wait... you don't want these to tune, right? IIRC, when I tried to deal with this, I wrote some `qudaMemcpyAsyncNoTune` wrappers that would turn off the tuning for `qudaMemcpyAsync`

lib/pgauge_exchange.cu
```diff
@@ -3,358 +3,258 @@
...
-      } else {
-        errorQuda("Reconstruction type %d of gauge field not supported", data.Reconstruct());
+        comm_wait(mh_send_back[d]);
+        comm_wait(mh_send_fwd[d]);
+        qudaStreamSynchronize(GFStream[0]);
+        qudaStreamSynchronize(GFStream[1]);
       }
-    } else {
-      errorQuda("Invalid Gauge Order\n");
+      checkCudaError();
```
cpviolator, line 247: This could be abstracted.

lib/pgauge_init.cu
```diff
@@ -19,169 +20,94 @@
...
+    bool tuneGridDim() const { return false;}
+    unsigned int minThreads() const { return arg.threads; }
+
+  public:
+    InitGaugeCold(GaugeField &data) :
+      arg(data),
+      meta(data)
+    {
+      apply(0);
+      checkCudaError();
```
cpviolator, line 68: Abstraction.

tests/gauge_alg_test.cpp
```diff
@@ -42,30 +39,31 @@ class GaugeAlgTest : public ::testing::Test {
...
-    if(DABS(1.0 - detu.x) < prec_val && DABS(detu.y) < prec_val) return true;
+    if (prec == QUDA_DOUBLE_PRECISION) prec_val = 1.0e-15;
+    if (std::abs(1.0 - detu.x) < prec_val && std::abs(detu.y) < prec_val) return true;
     return false;
   }
 
-
   void CallUnitarizeLinks(cudaGaugeField *cudaInGauge){
     unitarizeLinks(*cudaInGauge, num_failures_dev);
     cudaMemcpy(&num_failures, num_failures_dev, sizeof(int), cudaMemcpyDeviceToHost);
```
cpviolator: Abstraction.

cpviolator (changes requested):
Should we try to deal with asynchronous memory copies without tuning on this pull, or leave it for another one? I haven't handed in any homework in a while... :)

include/reduce_helper.h
```diff
@@ -99,21 +112,38 @@ namespace quda
...
 #else
       auto event = reducer::get_event();
       qudaEventRecord(event, stream);
-      while (cudaSuccess != qudaEventQuery(event)) {}
+      while (!qudaEventQuery(event)) {}
 #endif
       // copy back result element by element and convert if necessary to host reduce type
       // unit size here may differ from system_atomic_t size, e.g., if doing double-double
       const int n_element = n_reduce * sizeof(T) / sizeof(device_t);
       for (int i = 0; i < n_element; i++) result[i] = reinterpret_cast<device_t *>(result_h)[i];
```
hummingtree, line 142: I might be asking for too much here, but is it possible to have some kind of bound check here so we don't get untraceable setfault? For example, having the interface as
```
void complete(std::vector<host_t> &result, const qudaStream_t stream = 0, bool reset = false)
```

include/tune_quda.h
```diff
@@ -428,18 +412,28 @@ namespace quda {
...
     unsigned int sharedBytesPerBlock(const TuneParam &param) const { return 0; }
 
-    // don't tune the grid dimension
-    virtual bool tuneGridDim() const { return false; }
+    /**
+       Reduction kernels require grid-size tuning, so enable this and
+       make it non-virtual so that if user accidentally disables it,
+       compiler will complain
+    */
+    bool tuneGridDim() const { return true; }
```
hummingtree: (I could be wrong here - this is based on my remote memory of how virtual functions work) I don't think dropping the `virtual` specifier here would make `tuneGridDim()` non-virtual: as long as `tuneGridDim()` is virtual in the base class `Tunable`, all `tuneGridDim()` in the derived classes are all considered virtual. I tried setting `tuneGridDim()` to return false in this [class,](https://github.com/lattice/quda/blob/8e323890793d36fd5a8f8887f9f84bffbbc72aa3/lib/contract.cu#L94) which inherits from `TunableLocalParityReduction`, and there is no compiler warning given.

Also in fact for the contraction code we do want to disable `gridDim` tuning since there `gridDim.z` is the number of time slices so we don't want to tune it. Actually in that case the `advanceBlockDim` here resets `gridDim.z` to `1` when ever `Tunable::advanceBlockDim` fails, which is not the intended behavior.

lib/pgauge_det_trace.cu
```diff
@@ -5,217 +5,151 @@
...
+        border[dir] = data.R()[dir];
+        X[dir] = data.X()[dir] - border[dir]*2;
+      }
     }
-    idx = linkIndex(x,X);
-  #endif
+  };
+
+  template <int blockSize, int type, typename Arg>
+  __global__ void compute_Value(Arg arg)
```
hummingtree:
```suggestion
  __global__ void compute_value(Arg arg)
```

hummingtree (changes requested):
The abstraction looks great. Mainly some comments on the `TunableLocalParityReduction` changes.

lib/comm_common.cpp
maddyscientist: This is left until next PR

lib/comm_common.cpp
maddyscientist: Next PR

lib/gauge_fix_fft.cu
maddyscientist: Next PR

lib/gauge_fix_fft.cu
maddyscientist: Next PR

lib/gauge_fix_fft.cu
maddyscientist: Next PR

lib/pgauge_exchange.cu
maddyscientist: Dealing with the `cudaMemcpy` vs `qudaMemcpy` and the autotuning is something I have a plan for, but will be for the next PR.

cpviolator (approved):

lib/reduce_helper.cu
maddyscientist: The streams and events will be dealt with in the next abstraction PR

lib/pgauge_exchange.cu
maddyscientist: The streams and events will be dealt with in the next PR

lib/pgauge_exchange.cu
maddyscientist: Next PR

lib/pgauge_det_trace.cu
maddyscientist: Fixed with https://github.com/lattice/quda/pull/1046/commits/3e634b0e8842b3a65f7f32da7ed588798a77a14d

tests/gauge_alg_test.cpp
maddyscientist: Fixed in https://github.com/lattice/quda/pull/1046/commits/74c730d46bc65883bdfa47b5b05e10d65aeb27d5.  I've cleaned up the test code, removing all use of `cudaMemcpy`, `cudaMemset` and `cudaMalloc`.  Moreover, I've also removed a bunch of unnecessary error checking in the test code (which is unneeded since error checking is done internally now) in https://github.com/lattice/quda/pull/1046/commits/3e8642f32fdecd0e3a02522e43b629521bc81c47.

lib/pgauge_exchange.cu
maddyscientist: Leaving this for now.  I think, once all API calls are moved their quda prefix equivalents, then we'll have no need at all for `checkCudaError` in the main library since all CUDA API calls will be guarded behind the abstraction, and equivalentl for other target backends.

lib/pgauge_init.cu
maddyscientist: Deferring, as noted above.

lib/gauge_fix_ovr.cu
maddyscientist: https://github.com/lattice/quda/pull/1046/commits/9a37a41912257a50bc038ae16f48544c4f301fb8

include/reduce_helper.h
maddyscientist: https://github.com/lattice/quda/pull/1046/commits/9a37a41912257a50bc038ae16f48544c4f301fb8

include/tune_quda.h
maddyscientist: You have good points here.
* I'll revert `tuneGridDim` to being virtual and remove the comment.
* Regarding `Tunable::advanceBlockDim` reseting `gridDim.y` and `gridDim.z`, this is actually a bug.  It's something that I fixed in Dean's ongoing experimental branch, and forgot to port to mainline QUDA.  I'll also add the fix for that in my next push.

include/reduce_helper.h
maddyscientist: Done with https://github.com/lattice/quda/pull/1046/commits/e78a953461c1c38d76e116a3f9e8f3deffcb3e09

include/tune_quda.h
maddyscientist: https://github.com/lattice/quda/pull/1046/commits/baff1c84ce93c925ce31470dd4d58b2471ceee17

maddyscientist:
All comments should be addressed now.  Thanks for the review comments @hummingtree, @cpviolator and @weinbe2.

hummingtree (approved):

weinbe2:
Do not merge yet: hit a snag on the coarsest level of an MG solve.

weinbe2 (approved):
Passes Chroma HMC, all looks good

PR #1047, opened 2020-08-14
test having launcher for each kernel

jcosborn:
This is mainly to get comments.  I would have put the launch* in dslash_pack.cuh instead of dslash_pack2.cu except for the setMaxDynamicSharedBytesPerBlock call.  Maybe that could be passed through TuneParam into the target specific launcher?  The qudaLaunchKernel still needs to be substituted, but at least it is localized.

### DIFF
--- a/include/kernels/dslash_pack.cuh
+++ b/include/kernels/dslash_pack.cuh
@@ -181,7 +181,8 @@ namespace quda
     }
   }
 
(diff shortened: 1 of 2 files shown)


mathiaswagner:

maddyscientist:
@jcosborn instead of creating a single launcher per kernel as you have done here, the cleaner alternative, as we discussed on the call would be pass a template template class parameter.  This would allow us to have a single launch function, instead of the four variants you have here.

maddyscientist:
Closing stale PR

PR #1049, opened 2020-08-19
switch eigen download url to gitlab

mathiaswagner:
bitbucket no longer supports mercurial and it seems that this killed the eigen download url.
Switch over to official gitlab repo, see https://eigen.tuxfamily.org/index.php?title=Main_Page


weinbe2 (approved):
I'm going to make the very case-specific judgement call that, because this is a narrow fix (just changing the download source of otherwise the same version of Eigen) of a code-breaking issue, we can skip waiting for the full Jenkins tests.

Nonetheless: I did do a quick download, build, and run of MG (to hit CA-solvers which use Eigen), and everything went through without surprise.

Issue #801, opened 2019-05-04
Jitify on Volta


cpviolator:
Compiling with `QUDA_JITYFY=ON` on summit (volta) causes runtime errors for all tests. Does the `jitify.hpp` file need updating for volta?

maddyscientist:
We're pretty sure this isn't a Volta issue - it's working fine on other Volta machines.  It looks to be a Summit / POWER9 issue.

@cpviolator can you do a `HOSTDEBUG` build and try again.  When built for debugging, jitify will output much more informatin and we'll be able to see what the problem is.

cpviolator:
Dozens of compile time errors with jitify on and `HOSTDEBUG` build type:

```
In file included from /ccs/home/howarth/QUDA_DEVELOP/quda/include/jitify_helper.cuh:34:0,
                 from /ccs/home/howarth/QUDA_DEVELOP/quda/lib/gauge_stout.cu:8:
/ccs/home/howarth/QUDA_DEVELOP/quda/lib/../include/externals/jitify.hpp:2945:226: error: _Pragma takes a parenthesized string literal
         <td id="LC564" class="blob-code blob-code-inner js-file-line">          (line.<span class="pl-c1">substr</span>(<span class="pl-c1">0</span>, pragma_beg) + <span class="pl-s"><span class="pl-pds">&quot;</span>_Pragma(<span class="pl-cce">\&quot;</span><span class="pl-pds">&quot;</span></span> + pragma_split[<span class="pl-c1">1</span>] + <span class="pl-s"><span class="pl-pds">&quot;</span><span class="pl-cce">\&quot;</span>)<span class="pl-pds">&quot;</span></span>);</td>
                                                                                                                                                                                                                                  ^
In file included from /ccs/home/howarth/QUDA_DEVELOP/quda/include/jitify_helper.cuh:34:0,
                 from /ccs/home/howarth/QUDA_DEVELOP/quda/lib/gauge_plaq.cu:3:
/ccs/home/howarth/QUDA_DEVELOP/quda/lib/../include/externals/jitify.hpp:2945:226: error: _Pragma takes a parenthesized string literal
[...]

PR #1050, opened 2020-08-20
Hotfix/p9 quarter

maddyscientist:
This is a bugfix PR
* Quarter precision is defined using the `int8_t` and not `char`.  This removes the ambiguity of the sign, and fixes quarter precision which was broken when QUDA when compiled on ARM and POWER platforms
* Deployed `instantiate` helper in more places to remove explicit template instantiation
* Moved copy_gauge.cu and copy_color_spinor.cu files to cpp since they contain no CUDA
* Improved performance of `computeCoarseClover` which was implemented inefficiently (@weinbe2 request)

I have done a full ctest run for all dslash types with quarter precision enabled, and all works fine on P9.  Also tested with Jitify as well, and things are good there.


include/instantiate.h
```diff
@@ -232,7 +232,118 @@ namespace quda
...
+     the precision
+     @param[in] field LatticeField we wish to instantiate
+     @param[in,out] args Any additional arguments required for the
+     computation at hand
+  */
+  template <template <typename> class Apply, typename F, typename... Args>
+  constexpr void instantiatePrecision(F &field, Args &&... args)
+  {
+    if (field.Precision() == QUDA_DOUBLE_PRECISION) {
+      // always instantiate double precision
```
weinbe2, line 268: Why is double always being instantiated?

include/instantiate.h
```diff
@@ -232,7 +232,118 @@ namespace quda
...
+#else
+      errorQuda("QUDA_PRECISION=%d does not enable quarter precision", QUDA_PRECISION);
+#endif
+    } else {
+      errorQuda("Unsupported precision %d\n", field.Precision());
+    }
+  }
+
+  /**
+     @brief The instantiatePrecision function is used to instantiate
```
weinbe2: Why do we need two `instantiatePrecision` functions (different template signatures noted)? If having two is unavoidable, is there a better naming option than `instantiatePrecision` and `instantiatePrecision2`?

include/instantiate.h
maddyscientist: This is a special case template: for the interface we need to always instantiate double precision for the copy in/out kernels.

include/instantiate.h
maddyscientist: Cleaning this up is easy with C++17 😜   This solution was expedient in the meantime....  Open to naming suggestions.  This is really just a placeholder.

include/instantiate.h
weinbe2: Understood, please add that to the doxygen then

include/instantiate.h
weinbe2: Fair, just please make the use for this version vs the other one clearer from the doxygen

maddyscientist: Will do.

weinbe2:
Open question, do we want to go down the rabbit hole/hell of replacing `short` with `int16_t` for aesthetic/consistency reasons?

maddyscientist:
I think there’s no rush to do this now.  Eventually, once we have `instantiate` everywhere, the number of shorts and int8_t actually in quda will be limited to just a few headers.  At that point, sure, but not now.

lib/copy_clover.cu
```diff
@@ -129,111 +128,67 @@ namespace quda {
...
 
-  // this is the function that is actually called, from here on down we instantiate all required templates
   void copyGenericClover(CloverField &out, const CloverField &in, bool inverse, QudaFieldLocation location,
-			void *Out, void *In, void *outNorm, void *inNorm) {
+                         void *Out, void *In, void *outNorm, void *inNorm)
+  {
 #ifdef GPU_CLOVER_DIRAC
-    if (out.Precision() == QUDA_HALF_PRECISION && out.Order() > 4) 
+    if (out.Precision() < QUDA_SINGLE_PRECISION && out.Order() > 4) 
       errorQuda("Half precision not supported for order %d", out.Order());
```
weinbe2: Perhaps change this to say "Fixed precision..." (here and below)

hummingtree (approved):
The only request is maybe we can have something like a `instantiateGaugeOrder` template available, similar to `instantiatePrecision`.

maddyscientist:
I have pushed some updates, so we should be good with this PR now
* Addressed @weinbe2's review comments
* Updated to the latest version of Jitify - this closes an old bug #801 (Jitify didn't work on POWER9)
* Added a workaround for CUDA 9.x Jenkins failure 

@hummingtree to address your question: `instantiateGaugeOrder` has limited use since it would only be in the copy_gauge routines.  In most places we simply infer the gauge order for native fields from the precision and reconstruct values.

weinbe2 (approved):
I've given this a few iterations of visual reviews, plus asked @maddyscientist to do a few due-diligence tests related to checking jitify+MG, etc, and all looks good. Assuming nothing weird turns up during the current set of Jenkins tests this will be good to merge.

PR #1053, opened 2020-09-08
Abstract `__habs2` to a customized function to handle the CUDA < 10.2 case.

hummingtree:
`__habs2` is a `CUDA >= 10.2` only feature so the use of it breaks on 10.1. This hotfix fixes it.


maddyscientist (approved):

include/mdw_dslash5_tensor_core.cuh
```diff
@@ -234,16 +236,7 @@ namespace quda
 
   __device__ inline void __half_max_abs_half2__(half &max, const half2 &input)
   {
-#if CUDA_VERSION >= 10200
```
weinbe2, line 237: Should this be `10020` instead?

PR #1031, opened 2020-07-05
Feature/mspcg deflation

cpviolator:
This PR adds a new Dirac operator to QUDA, matEig. Those solvers with the capability to deflate will now use the operator constructed by `gauge(Long/Fat)Eigensolver (and clover field too). New command lines have been added:
```
 --prec-eigensolver ENUM:value in {double,single,half,quarter}=-2147483648
 --recon-eigensolver ENUM:value in {18,13,12,9,8}=-2147483648
```
to dictate the precision of the eigensolver when used as a deflation operator. Two command line commands have been tested:
```
mpirun -np 2 ./invert_test --sdim 16 --tdim 32 --tgridsize 2 --inv-deflate true --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --mass -0.42 --verbosity verbose --tol 1e-10 --inv-type gcr --anisotropy 2.38 --solve-type direct-pc --prec double --prec-sloppy half --prec-precondition half --prec-eigensolver single --ngcrkrylov 50 --niter 5000 --verify false --eig-tol 1e-3 --eig-poly-deg 50 --eig-amin 1e-3 --eig-n-ev 32 --eig-n-kr 64
```
and
```
mpirun -np 2 ./invert_test --sdim 16 --tdim 32 --tgridsize 2 --inv-deflate true --load-gauge /scratch/lattices/wl_16_64_5p5_x2p38_um0p4086_cfg_1000.lime --mass -0.42 --verbosity verbose --tol 1e-10 --inv-type cg --anisotropy 2.38 --solve-type normop-pc --prec double --prec-sloppy half --prec-precondition half --prec-eigensolver single --ngcrkrylov 50 --niter 5000 --verify false --eig-tol 1e-3 --eig-poly-deg 50 --eig-amin 1e-3 --eig-n-ev 32 --eig-n-kr 64
```
which test the SVD and regular deflation code paths.

@walkloud This will now allow us to deflate a solve using the MSPCG inverter.

PR #1057, opened 2020-09-14
Correct logic for gauge deletion in staggered gauges, add/fix comments

cpviolator:
This fixes a memory leak introduced by #1031 in which the longSloppy gauge field was not deleted.


weinbe2 (approved):
Thanks for catching this, along with the additional comments you added as part of the fix. Good to go once Jenkins completes.

maddyscientist:
Confirmed: this PR fixes a memory leak that occurred with MILC RHMC and now does not.

maddyscientist (approved):

PR #1059, opened 2020-09-21
hotfix/eigensolver_gamma_basis

cpviolator:
Ensure the eigensolver rotates the Krylov space from/to the host gamma basis.

Previous behaviour had the UKQCD gamma basis hardcoded in the `eigensolve_test` routine, so the CPU and GPU bases always matched. We now pass Degrand Rossi for the CPU basis, and ensure that a gamma basis rotation is applied during vector CPU->GPU and GPU->CPU transfer.


weinbe2:
Did you test this with `staggered_eigensolve_test`? I can't test myself now, but I think staggered (sort of arbitrarily, of course) requires you to specify some specific gamma basis as the GPU basis.

Otherwise, pending Jenkins, this seems like a non-controversial PR.

cpviolator:
@weinbe2 Good call on the staggered check: if you try to rotate the basis for nSpin == 1 you get runtime errors.

weinbe2:
I'm going to do a quick build+test now, then odds are I'll be quickly hitting merge.

weinbe2 (approved):
Just tested Wilson and staggered deflations as a sanity check, everything looks good. Thanks for this fix @cpviolator !

PR #1062, opened 2020-09-29
Add NVSHMEM communication policies for Dslashes

mathiaswagner:
This PR enables NVSHMEM support in QUDA.

NVSHMEM in QUDA is only supported for
* Volta and later NVIDIA GPUs
* CUDA 11.0+
* NVSHMEM 2.0+

~For now the communication policies are only enabled for 4d-Dslashes. Support for 5d and multi-rhs will be added at a later stage.~

NVSHMEM support can be enabled in cmake by setting QUDA_NVSHMEM and QUDA_NVSHMEM_HOME. 

~I'll add build instructions to the wiki in the next days.~ WIP: https://github.com/lattice/quda/wiki/Multi-GPU-with-NVSHMEM


lib/comm_common.cpp
```diff
@@ -308,6 +308,8 @@ bool comm_peer2peer_enabled(int dir, int dim){
   return enable_p2p ? peer2peer_enabled[dir][dim] : false;
 }
 
+bool comm_peer2peer_possible(int dir, int dim) { return peer2peer_enabled[dir][dim]; }
```
hummingtree: Might be a dumb question but what's the purpose of this function?

lib/lattice_field.cpp
```diff
@@ -461,11 +488,15 @@ namespace quda {
...
-	  cudaIpcOpenMemHandle(ghostDest, ipcRemoteGhostDestHandle[b][dir][dim],
-			       cudaIpcMemLazyEnablePeerAccess);
-	}
-	if (num_dir == 1) ghost_remote_send_buffer_d[b][dim][1] = ghost_remote_send_buffer_d[b][dim][0];
+#ifndef NVSHMEM_COMMS
+          void **ghostDest = &(ghost_remote_send_buffer_d[b][dim][dir]);
+          cudaIpcOpenMemHandle(ghostDest, ipcRemoteGhostDestHandle[b][dir][dim], cudaIpcMemLazyEnablePeerAccess);
+#else
+          ghost_remote_send_buffer_d[b][dim][dir]
+            = nvshmem_ptr(static_cast<char *>(ghost_recv_buffer_d[b]), comm_neighbor_rank(dir, dim));
```
hummingtree, line 479: We need to destroy and then re-construct the pointer whenever we change the communicator - the same peer2peer handles need to be refreshed.

lib/interface_quda.cpp
```diff
@@ -521,6 +527,15 @@ void initQudaMemory()
...
   device::create_context();
+#if defined(NVSHMEM_COMMS)
+  // MPI_Init(&argc, &argv);
+  MPI_Comm tmp = MPI_COMM_WORLD;
+  printfQuda("Init NVSHMEM\n");
+  nvshmemx_init_attr_t attr;
+  attr.mpi_comm = &tmp;
+  nvshmemx_init_attr(NVSHMEMX_INIT_WITH_MPI_COMM, &attr);
+#endif
+
```
maddyscientist, line 511: This code needs to be in `targets/cuda`

lib/lattice_field.cpp
```diff
@@ -236,32 +240,46 @@ namespace quda {
           qudaDeviceSynchronize();
           comm_barrier();
           for (int b=0; b<2; b++) {
-	    device_pinned_free(ghost_recv_buffer_d[b]);
-	    device_pinned_free(ghost_send_buffer_d[b]);
-	    host_free(ghost_pinned_send_buffer_h[b]);
-	    host_free(ghost_pinned_recv_buffer_h[b]);
-	  }
+#ifndef NVSHMEM_COMMS
```
maddyscientist: This should moved to targets/cuda/malloc.cpp and in a special allocator `device_comms_pinned_malloc` etc.

include/tune_quda.h
```diff
@@ -602,6 +602,21 @@ namespace quda {
...
 
+  /**
+   * @brief Query whether we are currently tuning a policy
+   */
+  bool policyTuning();
+
+  /**
+   * @brief Enable / disable whether we are tuning an uber kernel
+   */
+  void setUberTuning(bool);
```
maddyscientist, line 622: Why do we need a separate `uberTuning`?  What function does this serve that `policyTuning` doesn't?

lib/CMakeLists.txt
```diff
@@ -213,15 +221,31 @@ target_compile_definitions(quda_cpp PRIVATE $<TARGET_PROPERTY:quda,COMPILE_DEFIN
 target_include_directories(quda_cpp PRIVATE $<TARGET_PROPERTY:quda,INCLUDE_DIRECTORIES>)
 target_compile_options(quda_cpp PRIVATE $<TARGET_PROPERTY:quda,COMPILE_OPTIONS>)
 
-# add target specific files
-if(${QUDA_TARGET_TYPE} STREQUAL "CUDA")
-  add_subdirectory(targets/cuda)
+# nvshmem enabled parts need CUDA_SEPARABLE_COMPILATION ...
```
maddyscientist, line 228: How does this dslash object separation affect the order of how things are built?

lib/comm_common.cpp
```diff
@@ -696,6 +698,25 @@ bool comm_gdr_blacklist() {
...
 }
 
+bool comm_nvshmem_enabled()
+{
+#if (defined MULTI_GPU) && (defined NVSHMEM_COMMS)
+  static bool nvshmem_enabled = true;
+#else
+  static bool nvshmem_enabled = false;
+#endif
+#ifdef MULTI_GPU
```
maddyscientist: shouldn't this also check for `NVSHMEM_COMMS` being defined?

lib/dslash_pack2.cu
```diff
@@ -2,17 +2,24 @@
...
+
 #include <dslash_quda.h>
 #include <kernels/dslash_pack.cuh>
 #include <instantiate.h>
 
 namespace quda
 {
 
-  static int commDim[QUDA_MAX_DIM];
+  // static int commDim[QUDA_MAX_DIM];
```
maddyscientist: delete this

lib/dslash_pack2.cu
```diff
@@ -2,17 +2,24 @@
...
 // if not STRIPED then this means we assign one thread block per direction / dimension
 
+#ifdef NVSHMEM_COMMS
+#include <nvshmem.h>
+#include <nvshmemx.h>
+#else
+// MWTODO: how to handle striped ...
+#define STRIPED 1
+#endif
+
```
maddyscientist, line 10: so we don't have `STRIPED` support now when NVSHMEM is enabled?

lib/dslash_pack2.cu
```diff
@@ -2,17 +2,24 @@
 
 // STRIPED - spread the blocks throughout the workload to ensure we
 // work on all directions/dimensions simultanesouly to maximize NVLink saturation
-#define STRIPED
 // if not STRIPED then this means we assign one thread block per direction / dimension
 
+#ifdef NVSHMEM_COMMS
```
maddyscientist: Style: the direct `NVSHMEM` includes should be hidden in a `shmem_helper.h` file.

lib/interface_quda.cpp
```diff
@@ -38,6 +38,12 @@
 #endif
 #include <gauge_update_quda.h>
 
+#if defined(NVSHMEM_COMMS)
```
maddyscientist: `shmem_helper.h`

maddyscientist: and following on from my comment below, this nvshmem include probably isn't even needed here since the nvshmem code should be moved to `targets/cuda/`

lib/lattice_field.cpp
```diff
@@ -236,32 +240,46 @@ namespace quda {
...
-	  ghost_pinned_send_buffer_h[b] = mapped_malloc(ghost_bytes);
-
-	  // set the matching device-mapped pointer
-	  ghost_pinned_send_buffer_hd[b] = get_mapped_device_pointer(ghost_pinned_send_buffer_h[b]);
-
-	  // pinned buffer used for receiving
-	  ghost_pinned_recv_buffer_h[b] = mapped_malloc(ghost_bytes);
-
-	  // set the matching device-mapped pointer
+#ifndef NVSHMEM_COMMS
```
maddyscientist: with a `device_comms_pinned_malloc` function these ifdefs won't be needed in the cpp code

lib/lattice_field.cpp
```diff
@@ -236,32 +240,46 @@ namespace quda {
...
+          ghost_send_buffer_d[b] = device_pinned_malloc(ghost_bytes);
+#else
+          ghost_send_buffer_d[b] = shmem_pinned_malloc(ghost_bytes);
+#endif
+
+          // pinned buffer used for sending
+          ghost_pinned_send_buffer_h[b] = mapped_malloc(ghost_bytes);
+
+          // set the matching device-mapped pointer
+          cudaHostGetDevicePointer(&ghost_pinned_send_buffer_hd[b], ghost_pinned_send_buffer_h[b], 0);
```
maddyscientist: this needs to be reverted back to 
`ghost_pinned_send_buffer_hd[b] = get_mapped_device_pointer(ghost_pinned_send_buffer_h[b]);`

include/kernels/dslash_pack.cuh
```diff
@@ -5,10 +5,31 @@
...
 #include <dslash_helper.cuh>
 
+#ifdef NVSHMEM_COMMS
+#include <nvshmem.h>
+#include <nvshmemx.h>
+#include <cuda/atomic>
+
+namespace
+{
+  __device__ cuda::atomic<int, cuda::thread_scope_system> retcount_intra[] {
```
maddyscientist: doxygen

lib/dslash_policy.cuh
```diff
@@ -435,7 +441,72 @@ namespace quda
...
+
+      dslash::synccounter++;
+      in->bufferIndex = (1 - in->bufferIndex);
+      profile.TPSTOP(QUDA_PROFILE_TOTAL);
+#else
+      errorQuda("NVSHMEM Dslash policies not build.");
+#endif
+    }
+  };
+
```
maddyscientist, line 501: I like this composable style here for the nvshmem policies 👍

maddyscientist (commented):
Initial visual review and left various comments.  A wiki guide on how to configure and run with NVSHMEM will be needed before we merge this in.

lib/CMakeLists.txt
mathiaswagner: Not a lot for Ninja. For make it creates a somewhat artificial barrier IIRC. Unfortunately we cannot set SEPARABLE_COMPILATION on individual source files, so we need to have an additional target.
I currently do not have a better idea how to handle that without applying it to all of QUDA.

lib/comm_common.cpp
mathiaswagner: done

lib/interface_quda.cpp
mathiaswagner: Thoughts on whether to include that in `device.cpp` or a separate `comms.cpp` file?

maddyscientist: I think a separate file would be the way to go here.

lib/CMakeLists.txt
maddyscientist: Ok, understood.  We'll just tell folks to use `ninja` for NVSHMEM for now.

lib/dslash_pack2.cu
mathiaswagner: yes, at least for now. We can open an issue to track adding support.

lib/comm_common.cpp
mathiaswagner: I needed it at some point. Basically nvshmem does not care about whether we enable P2P via the env variable. But we exploit the knowledge that we can directly pack into the remote buffer with nvshmem when p2p is possible.
Now that I think of it again there is a potential we don't need that function.

include/tune_quda.h
mathiaswagner: The uber kernel is not a policy but needs to run on all GPUs so that flags are set correctly. I added that to avoid an infinite spin at some point but good to reconsider whether we can get away without it.

maddyscientist: Yeh, I realize that this isn't a policy and is a separate use case.  This may be useful in the future for other uber kernels that aren't part of a policy at all, e.g., if we were to implement the gauge force as an uber kernel for example, or more conservatively, implementing the gauge field extension using NVSHMEM.  So this is good to keep.

weinbe2: I think we should include @mathiaswagner 's reply here for clarity. An inline comment that explicitly says "An uber kernel is not a policy but it needs to run on all GPUs so that flags are set correctly." would be a useful clarifier here and elsewhere.

lib/comm_common.cpp
mathiaswagner: I have removed this now.

lib/lattice_field.cpp
mathiaswagner: Yes, but we might just be fine here. Will need to check that later.

lib/dslash_pack2.cu
mathiaswagner: Tracked in #1084

lib/lattice_field.cpp
hummingtree: We can probably address this in #1091.

mathiaswagner:
I have updated the title and description to reflect that initial support for 5d Dslashes has been added.

CMakeLists.txt
```diff
@@ -556,6 +572,41 @@ if(QUDA_DOWNLOAD_USQCD)
...
 endif()
 
+if(QUDA_NVSHMEM)
+  if(QUDA_DOWNLOAD_NVSHMEM)
+    ExternalProject_Add(NVSHMEM
+                        URL ${QUDA_DOWNLOAD_NVSHMEM_TAR}
+                        PREFIX nvshmem
+                        CONFIGURE_COMMAND ""
+                        BUILD_IN_SOURCE ON
+                        BUILD_COMMAND  make NVSHMEM_PREFIX=<INSTALL_DIR> NVSHMEM_MPI_SUPPORT=1 GDRCOPY_HOME=/usr/local/gdrcopy/ install
```
weinbe2: Add a cmake flag for overriding the default `gdrcopy` path

include/color_spinor_field.h
```diff
@@ -493,6 +493,7 @@ namespace quda {
     size_t NormBytes() const { return norm_bytes; }
     size_t TotalBytes() const { return bytes + norm_bytes; }
     size_t GhostBytes() const { return ghost_bytes; }
+    size_t GhostFaceBytes(int i) const { return ghost_face_bytes[i]; }
```
weinbe2, line 496: Looks like this should be squared with `0028fd019c65bdf200bbaf1ff832c56637cca185`

weinbe2: https://github.com/lattice/quda/commit/0028fd019c65bdf200bbaf1ff832c56637cca185

include/dslash.h
```diff
@@ -140,14 +185,24 @@ namespace quda
 
     virtual void initTuneParam(TuneParam &param) const
     {
+      if(arg.shmem & 64){  //FIXME later: For the uber kernel temporarily limit the grid y anz dimension to 1
```
weinbe2: is this FIXME still relevant?

include/dslash.h
```diff
@@ -125,12 +160,22 @@ namespace quda
       if (arg.pack_threads && arg.kernel_type == INTERIOR_KERNEL) {
         // if doing the fused kernel we tune how many blocks to use for
         // communication
-        constexpr int max_blocks_per_dir = 4;
+        constexpr int max_blocks_per_dir = 6;
```
weinbe2: How carefully have we verified that some corners of the tuning space don't cause lockups b/c not enough blocks are active? Is this even a relevant question based on the design?

include/dslash_helper.cuh
```diff
@@ -4,10 +4,27 @@
...
+  __device__ cuda::atomic<long, cuda::thread_scope_system> intersync {0};
+  __device__ cuda::atomic<long, cuda::thread_scope_device> interior_done {0};
+  __device__ cuda::atomic<int, cuda::thread_scope_block> interior_count {0};
+} // namespace
+#endif
 
 namespace quda
 {
 
+  namespace dslash
```
weinbe2: Can this be wrapped in `#ifdef NVSHMEM_COMMS`?

lib/dslash_policy.cuh
```diff
@@ -1904,6 +2023,7 @@ public:
         }                                     // p2p policies
 
         enableProfileCount();
+        // dslash::synccounter = 5000000;
```
hummingtree: Is this still needed?

lib/dslash_policy.cuh
```diff
@@ -435,7 +442,72 @@ namespace quda
...
+      dslashParam.kernel_type = INTERIOR_KERNEL;
+      dslashParam.threads = volume;
+      dslash.setShmem(shmem);
+      dslashParam.setExt(shmem & 64);
+
+      // record start of the dslash
+      const int packIndex = Nstream - 1;
+      constexpr MemoryLocation location = static_cast<MemoryLocation>(Shmem);
+
+      if (!(((shmem & 2)) and (shmem & 1))) {
```
hummingtree:
```suggestion
      if (!((shmem & 2) and (shmem & 1))) {
```

lib/lattice_field.cpp
```diff
@@ -565,24 +574,31 @@ namespace quda {
...
 	if (comm_peer2peer_enabled(1,dim)) {
 	  if (mh_send_p2p_fwd[b][dim] || mh_recv_p2p_fwd[b][dim]) {
 	    cudaEventDestroy(ipcCopyEvent[b][1][dim]);
 	    // only close this handle if it doesn't alias the back ghost
-	    if (num_dir == 2) cudaIpcCloseMemHandle(ghost_remote_send_buffer_d[b][dim][1]);
-	  }
+
+#ifndef NVSHMEM_COMMS
+            if (num_dir == 2) cudaIpcCloseMemHandle(ghost_remote_send_buffer_d[b][dim][1]);
+#endif
```
hummingtree, line 593: A note to myself: I realize that split grid might just work with nvshmem, since one doesn't need to destroy the shmem pointer - once the init flag is set to false the pointers will be set correctly next time `allocateGhostBuffer` is called.

hummingtree: #1091.

include/dslash_helper.cuh
```diff
@@ -429,6 +491,129 @@ namespace quda
...
+
+        if (shmem_interiordone && threadIdx.x == blockDim.x - 1 && threadIdx.y == 0 && threadIdx.z == 0) {
+          long tst_val = interior_done.load(cuda::std::memory_order_relaxed);
+          while (tst_val < arg.counter - 1) {
+            interior_done.compare_exchange_strong(tst_val, arg.counter - 1, cuda::std::memory_order_relaxed,
+                                                  cuda::std::memory_order_relaxed);
+          }
+          interior_done.wait(arg.counter - 1, cuda::std::memory_order_acquire);
+        }
+        if (threadIdx.x < 8 && threadIdx.y == 0 && threadIdx.z == 0) {
```
hummingtree: I would add a comment here saying something like use the first `8` threads to spin.

hummingtree (approved):

hummingtree:
The PR looks great to me. I have left a few minor comments. I will run a few tests with NVSHMEM.

include/color_spinor_field.h
maddyscientist: The addition of the local surfaces shouldn't really intersect with this.  That just adds some helpers to return the non-extended surface of a given extended field.  For a field that isn't extended, the local surface and regular surface are the same.

include/dslash.h
```diff
@@ -77,9 +78,31 @@ namespace quda
...
+      if (arg.kernel_type == EXTERIOR_KERNEL_ALL && arg.shmem > 0) {
+        int nDimComms = 0;
+        for (int d = 0; d < in.Ndim(); d++) nDimComms += arg.commDim[d];
+        return ((deviceProp.multiProcessorCount) / (2 * nDimComms)) * (2 * nDimComms);
+      } else {
+        return TunableVectorYZ::minGridSize();
+      }
+    }
+
+    virtual int gridStep() const
```
maddyscientist, line 100: Might be nice to add some comments here to document how the grid size is tuned, and what the constraints are when nvshmem is enabled.

include/dslash_helper.cuh
```diff
@@ -251,6 +268,7 @@ namespace quda
     bool remote_write;      // used by the autotuner to switch on/off remote writing vs using copy engines
 
     int_fastdiv threads; // number of threads in x-thread dimension
+    int_fastdiv ext_threads; //  number of threads in x-thread dimension for fused exteriot dslash
```
maddyscientist: exteriot -> exterior

include/dslash_helper.cuh
```diff
@@ -264,13 +282,25 @@ namespace quda
...
     int pack_blocks; // total number of blocks used for packing in the dslash
+    int ext_dims;    // dimension to run in the exterior Dslash
+    int ext_blocks;
+
+    // for shmem ...
+    static constexpr bool packkernel = false;
+    void *packBuffer[4 * QUDA_MAX_DIM];
+    int neighbor_ranks[2 * QUDA_MAX_DIM];
+    int bytes[2 * QUDA_MAX_DIM];
+    volatile long *sync_arr;
```
maddyscientist: Why is this volatile?  Can we replace with `std::atomic`?

include/dslash_helper.cuh
```diff
@@ -364,7 +425,8 @@ namespace quda
     out << std::endl;
     out << "active_dims = " << arg.active_dims << std::endl;
     out << "pack_blocks = " << arg.pack_blocks << std::endl;
-
+    out << "ext_threads = " << arg.ext_threads;
+    out << "ext_blocks " << arg.ext_blocks;
```
maddyscientist: missing `=`

include/dslash_helper.cuh
```diff
@@ -429,6 +491,129 @@ namespace quda
       // first few blocks do packing kernel
       P<dagger, dslash.pc_type(), Arg> packer;
       packer(arg, s, 1 - parity, dslash.twist_pack()); // flip parity since pack is on input
+
+      // we use that when running the exterior -- this is either
+      // * an explicit call to the exterior when not merged with the interior or
+      // * the interior with ext_blocks > 0
+#ifdef NVSHMEM_COMMS
+    } else if (arg.shmem > 0
```
maddyscientist, line 643: Can we move the contents of this to a separate function to make `dslashGPU` more readable from a high level?

include/dslash_helper.cuh
```diff
@@ -441,6 +626,22 @@ namespace quda
...
+          if (threadIdx.x == 0 && threadIdx.y == 0 && threadIdx.z == 0) {
+            int amlast = interior_count.fetch_add(1, cuda::std::memory_order_acq_rel); // ensure that my block is done
+            if (amlast == (gridDim.x - arg.pack_blocks - arg.ext_blocks) * gridDim.y * gridDim.z - 1) {
+              interior_done.store(arg.counter, cuda::std::memory_order_release);
+              interior_done.notify_all();
+              interior_count.store(0, cuda::std::memory_order_relaxed);
+            }
+          }
+        }
+      }
```
maddyscientist: similarly, can we move the above to something like an `epilogue` function?

include/kernels/dslash_wilson.cuh
```diff
@@ -31,6 +31,7 @@ namespace quda
 
     F out;        /** output vector field */
     const F in;   /** input vector field */
+    const F in_pack; /** input vector field */
```
maddyscientist: Expand to explain how `in` is different from `in_pack`

include/malloc_quda.h
```diff
@@ -45,15 +45,22 @@ namespace quda {
...
+  void device_comms_pinned_free_(const char *func, const char *file, int line, void *ptr);
   void managed_free_(const char *func, const char *file, int line, void *ptr);
   void host_free_(const char *func, const char *file, int line, void *ptr);
-
+#ifdef NVSHMEM_COMMS
+  void *shmem_malloc_(const char *func, const char *file, int line, size_t size);
+  void *shmem_pinned_malloc_(const char *func, const char *file, int line, size_t size);
+  void shmem_free_(const char *func, const char *file, int line, void *ptr);
+  void shmem_pinned_free_(const char *func, const char *file, int line, void *ptr);
+#endif
```
maddyscientist: These don't look like they are used outside of malloc, with the exception of the `sync_arr` allocation.  Can that not just use the abstracted `device_comms_pinned_malloc`, and not expose these to the outside world?

maddyscientist:
I've completed a visual review of the changes, and left some comments.  Will continue testing.

include/dslash_helper.cuh
```diff
@@ -4,17 +4,34 @@
...
 #include <index_helper.cuh>
+#include <shmem_helper.cuh>
+
+#ifdef NVSHMEM_COMMS
+namespace
+{
+  __device__ cuda::atomic<long, cuda::thread_scope_system> intersync {0};
+  __device__ cuda::atomic<long, cuda::thread_scope_device> interior_done {0};
+  __device__ cuda::atomic<int, cuda::thread_scope_block> interior_count {0};
+} // namespace
```
maddyscientist: Since we require `cuda::atomic`, I think we need to exclude `SANITIZE` builds like we do with heterogeneous atomics, since libcu++ doesn't play nicely with ASAN.

include/kernels/dslash_pack.cuh
```diff
@@ -360,48 +548,115 @@ namespace quda
...
-        while (local_tid < arg.nFace * arg.dc.ghostFaceCB[1]) {
-          int ghost_idx = dir * arg.nFace * arg.dc.ghostFaceCB[1] + local_tid;
-          if (arg.nFace == 1)
-            packStaggered<1, 1>(arg, ghost_idx, s, parity);
-          else
-            packStaggered<1, 3>(arg, ghost_idx, s, parity);
-          local_tid += arg.blocks_per_dir * blockDim.x;
+      }
+
+#ifdef NVSHMEM_COMMS
```
maddyscientist, line 596: This is the same code as above for the non-staggered kernel.  Can you move this to its own function to avoid the repeated boilerplate?

include/dslash_helper.cuh
```diff
@@ -429,6 +491,129 @@ namespace quda
...
+            if (arg.commDim[1]) {
+              if (arg.commDim[0]) spin = threadIdx.x / 2 < 1 ? true : spin;
+            }
+            break;
+          case 0: break;
+          }
+
+          if (getNeighborRank(threadIdx.x, arg) >= 0) {
+            if (spin) {
+              while (arg.counter > *(arg.sync_arr + threadIdx.x)) { __nanosleep(10); };
```
maddyscientist: Curious, how did you arrive at a value of 10 here?

include/dslash_helper.cuh
maddyscientist: Another thought: since these are statically declared per compilation unit, we end up with this allocation overhead repeated for every dslash file.  Is there any reason not to allocate these dynamically, say in dslash_quda.cu, and reuse these atomics for all stencils to avoid this overhead?  Initialization could be done with a placement new constructor similar to the atomics used with the heterogeneous atomic reductions.

CMakeLists.txt
```diff
@@ -272,6 +276,9 @@ option(QUDA_DOWNLOAD_USQCD "Download USQCD software as requested by QUDA_QMP / Q
 option(QUDA_DOWNLOAD_ARPACK "Download ARPACK-NG software as requested by QUDA_ARPACK" OFF)
 option(QUDA_DOWNLOAD_OPENBLAS "Download OpenBLAS software as requested by QUDA_OPENBLAS" OFF)
 option(QUDA_JITIFY "build QUDA using Jitify" OFF)
+option(QUDA_DOWNLOAD_NVSHMEM "Download NVSHMEM" OFF)
+set(QUDA_DOWNLOAD_NVSHMEM_TAR "" CACHE STRING "location of NVSHMEM tarball")
```
weinbe2, line 280: What's your reason for passing in a location of the `NVSHMEM` tarball as opposed to having QUDA automatically download a fixed one, similar to what's done with QMP/QIO? I can see argument in either direction, so I'm just curious.

include/kernels/dslash_pack.cuh
mathiaswagner: What is wrong with the doxygen in line 14-15?

include/dslash.h
mathiaswagner: no

include/color_spinor_field.h
mathiaswagner: Please reopen but for now I am marking this as resolved as above commit is from the gk branch.

CMakeLists.txt
mathiaswagner: done.

include/dslash_helper.cuh
mathiaswagner: THat does not seem to work for me if I open the two namespace again it still complains ...

include/dslash_helper.cuh
mathiaswagner: I blocked the SANITIZE build for now.

include/kernels/dslash_pack.cuh
mathiaswagner: Did that refer to the shmem_signal you already extracted?

maddyscientist: Yep

include/dslash_helper.cuh
maddyscientist: Best we can do for now until libcu++ fixes this

include/dslash_helper.cuh
mathiaswagner: Guess we could get this when using placement new, but it might prevent us better `nvshmem_wait_<whatever>` things in the future. I'd defer that to future optimization. But getting this to be atomic would be nicer.

include/dslash_helper.cuh
mathiaswagner: I'll look into placement new for that. I guess the same applies to the ones in `dslash_pack.cuh`. Should we put that in now or for future optimization? I don't have a feeling for the actual overhead here.

include/dslash_helper.cuh
mathiaswagner: No strong motivation for that here. Worth looking tuning that a bit.

include/malloc_quda.h
mathiaswagner: Good idea. I think so.

include/dslash_helper.cuh
maddyscientist: Looking at the output from nvcc, the overhead is 64 KiB per translation unit I think.  It should be fairly easy to make it a single instance with placement new, so my preference would be to do this now.

include/dslash.h
mathiaswagner: added a few more comments on this.

include/dslash_helper.cuh
mathiaswagner: Just commit this. Please have a look.

maddyscientist: Ok, this looks mostly good now.  I would though prefer having a getter method rather than having the variables extern, and made global.  We've got enough globals in QUDA 😉

CMakeLists.txt
```diff
@@ -469,7 +486,7 @@ endif()
 include(CMakeDependentOption)
 CMAKE_DEPENDENT_OPTION(QUDA_HETEROGENEOUS_ATOMIC "enable heterogeneous atomic support (CUDA >= 11.0)?" ON "QUDA_HETEROGENEOUS_ATOMIC_SUPPORT" OFF)
 
-if(QUDA_HETEROGENEOUS_ATOMIC AND ${CMAKE_BUILD_TYPE} STREQUAL "SANITIZE")
+if((QUDA_HETEROGENEOUS_ATOMIC OR QUDA_NVSHMEM) AND ${CMAKE_BUILD_TYPE} STREQUAL "SANITIZE")
   message(SEND_ERROR "QUDA_HETEROGENEOUS_ATOMIC=ON does not support SANITIZE build)")
```
weinbe2: Update error message

include/dslash.h
```diff
@@ -77,9 +78,36 @@ namespace quda
...
+    virtual unsigned int minGridSize() const
+    {
+      /* when using nvshmem we perform the exterior Dslash using a grid strided loop and uniquely assign communication
+       * directions to CUDA block and have all communication directions resident. We therefore figure out the number of
+       * communicating dimensions and make sure that the number of blocks is a multiple of the communicating directions (2*dim)
+       */
+      if (arg.kernel_type == EXTERIOR_KERNEL_ALL && arg.shmem > 0) {
+        int nDimComms = 0;
+        for (int d = 0; d < in.Ndim(); d++) nDimComms += arg.commDim[d];
+        return ((deviceProp.multiProcessorCount) / (2 * nDimComms)) * (2 * nDimComms);
```
weinbe2, line 94: Sanity check: you can never hit this codepath when `nDimComms` is zero, correct? Is this still true with Schwarz preconditioners/the 3-d laplace op (for covariant derivatives)? @maddyscientist

include/dslash_helper.cuh
```diff
@@ -264,13 +266,31 @@ namespace quda
 
     int pack_threads; // really number of face sites we have to pack
     int_fastdiv blocks_per_dir;
+    int sites_per_block;
     int dim_map[4];
     int active_dims;
     int pack_blocks; // total number of blocks used for packing in the dslash
+    int ext_dims;    // dimension to run in the exterior Dslash
+    int ext_blocks;
```
weinbe2: Can this be renamed to `exterior_blocks`, etc for other `ext_*` variables for code readability?

include/dslash.h
```diff
@@ -323,23 +394,39 @@ namespace quda
...
+      }
 
       for (int dim = 0; dim < 4; dim++) {
         for (int dir = 0; dir < 2; dir++) {
           if ((location & Remote) && comm_peer2peer_enabled(dir, dim)) { // pack to p2p remote
             packBuffer[2 * dim + dir] = static_cast<char *>(in.remoteFace_d(dir, dim)) + in.GhostOffset(dim, 1 - dir);
           } else if (location & Host && !comm_peer2peer_enabled(dir, dim)) { // pack to cpu memory
             packBuffer[2 * dim + dir] = in.myFace_hd(dir, dim);
+          } else if (location & Shmem) {
+            packBuffer[2 * dim + dir] = in.remoteFace_d(dir, dim) ?
```
weinbe2, line 440: What the logic for querying `remoteFace_d` here? Is it to support `--partition`?

include/dslash_helper.cuh
```diff
@@ -292,7 +313,22 @@ namespace quda
...
+      ext_dims(0),
+      ext_blocks(0),
+#ifndef NVSHMEM_COMMS
+      shmem(0),
+      counter(0)
+#else
+      shmem(shmem_),
+      counter(dslash::get_shmem_sync_counter()),
+      sync_arr(dslash::get_shmem_sync_arr()),
+      interior_done(*dslash::get_shmem_interior_done()),
```
weinbe2, line 328: ordering, swap `interior_*` with `retcount_*`

include/dslash_helper.cuh
```diff
@@ -321,12 +363,31 @@ namespace quda
...
+          if (i == 3 && !getKernelPackT()) continue;
+          dim_map[d++] = i;
+        }
         pack_threads = 0;
         pack_blocks = 0;
         active_dims = 0;
       }
     }
+
+    void setExt(bool ext)
```
weinbe2: Ext -> Exterior (please)

include/dslash_helper.cuh
```diff
@@ -401,6 +463,148 @@ namespace quda
...
+
+    // figure out some details on blocks
+    const bool shmem_interiordone = (arg.shmem & 64);
+    const int myblockidx = arg.ext_blocks > 0 ? blockIdx.x - (gridDim.x - arg.ext_blocks) : blockIdx.x;
+    const int nComm = arg.commDim[0] + arg.commDim[1] + arg.commDim[2] + arg.commDim[3];
+    const int blocks_per_dim = (arg.ext_blocks > 0 ? arg.ext_blocks : gridDim.x) / (nComm);
+    const int blocks_per_dir = blocks_per_dim / 2;
+
+    int dir = (myblockidx % blocks_per_dim) / (blocks_per_dir);
+    // this id the dimdir we are working on ...
```
weinbe2: typo `id` -> `is` (?)

include/dslash_quda.h
```diff
@@ -36,6 +39,28 @@ namespace quda {
...
+    using shmem_retcount_intra_t = cuda::atomic<int, cuda::thread_scope_system>;
+    using shmem_retcount_inter_t = cuda::atomic<int, cuda::thread_scope_device>;
+    using shmem_interior_done_t = cuda::atomic<long, cuda::thread_scope_device>;
+    using shmem_interior_count_t = cuda::atomic<long, cuda::thread_scope_block>;
+    shmem_sync_t *get_shmem_sync_arr();
+    shmem_retcount_intra_t *get_shmem_retcount_intra();
+    shmem_retcount_inter_t *get_shmem_retcount_inter();
+    shmem_interior_done_t *get_shmem_interior_done();
+    shmem_interior_count_t *get_shmem_interior_count();
+#endif
```
maddyscientist, line 110: Doxygen for these functions please

include/dslash.h
maddyscientist: This won't be a problem for a Schwarz preconditioner since we never have exterior kernels for that.  It will be fine also for 3-d Laplace since `commDim` will be set appropriately.

CMakeLists.txt
mathiaswagner: Done

include/dslash_helper.cuh
mathiaswagner: swapped above

include/dslash.h
mathiaswagner: if `remoteFace_d` then we can directly pack to the remote. The packing destination is `packBuffer[2* dim + dir]`. If not the packing writes to `in.myFace_d(dir, dim);` and then we use `nvshmem_putmem` to get the data to `static_cast<char *>(in.remoteFace_r()) + in.GhostOffset(dim, 1 - dir);` (copy from `packBuffer[2* dim + dir]` to `packBuffer[2*QUDA_MAX_Dim + 2* dim + dir]`

include/dslash_helper.cuh
mathiaswagner: done

include/kernels/covDev.cuh
```diff
@@ -31,6 +31,7 @@ namespace quda
 
     F out;      /** output vector field */
     const F in; /** input vector field */
+    const F in_pack; /** input vector field used in packing to be able to independently resetGhost */
```
weinbe2, line 34: Why is this necessary/is there a cleaner way to do this? I see `in_pack` is always initialized to `in`, so I'm not so worried about memory overhead bloat, but it seems a little hacky.

mathiaswagner: In `dslash.h` around line 140 
we may need different ghosts for packing 
```cpp
      arg.in.resetGhost(in, ghost);

      if (arg.pack_threads && arg.kernel_type == INTERIOR_KERNEL) {
        arg.blocks_per_dir = tp.aux.x;
        arg.setPack(true, this->packBuffer); // need to recompute for updated block_per_dir
        arg.in_pack.resetGhost(in, this->packBuffer);
```

weinbe2: Sure, my thought is related to including different/multiple ghosts within a single `ColorSpinorField` as opposed to using two separate `ColorSpinorField`s with different ghosts.

include/lattice_field.h
```diff
@@ -664,6 +664,8 @@ namespace quda {
     */
     void *remoteFace_d(int dir, int dim) const { return ghost_remote_send_buffer_d[bufferIndex][dim][dir]; }
 
+    void *remoteFace_r() const { return ghost_recv_buffer_d[bufferIndex]; }
```
weinbe2, line 673: doxygen

include/shmem_helper.cuh
```diff
@@ -0,0 +1,16 @@
...
+   Include this file as opposed to nvshmem headers directly to ensure
+   correct compilation with NVSHMEM
+ */
+
+#if defined(NVSHMEM_COMMS)
+#include <mpi.h>
+#include <nvshmem.h>
+#include <nvshmemx.h>
+#include <cuda/atomic>
+#endif
```
weinbe2: Not sure if it's in our style guides, but add extra blank line to end of file

include/tune_quda.h
```diff
@@ -610,6 +611,21 @@ namespace quda {
    */
   void setPolicyTuning(bool);
 
+  /**
+   * @brief Query whether we are currently tuning a policy
```
weinbe2, line 615: Just for cleanliness/consistency, is there a reason why we couldn't add this routine to the base `Tuneable` class instead and let it get inherited downstream?

lib/CMakeLists.txt
weinbe2: ^ has this note been added to the wiki?

lib/dslash_pack2.cu
```diff
@@ -63,55 +64,53 @@ protected:
...
+      if (location & Host || location & Shmem) {
 #ifdef STRIPED
         // if zero-copy policy then set a maximum number of blocks to be
         // the 3 * number of dimensions we are communicating
         int max = 3;
 #else
         // if zero-copy policy then assign exactly up to four thread blocks
         // per direction per dimension (effectively no grid-size tuning)
-        int max = 2 * 4;
+        int max = ((shmem & 32) || (location & Host)) ? 4 : -1;
```
weinbe2: What's the reason for the `|| (location & Host)`? Was some issue with the previous default of `8` when `STRIPED` was undefined uncovered with `Host` specifically?

weinbe2: Alternatively, is in related to the changes in `minGridSize` below that I just noticed?

lib/dslash_pack2.cu
```diff
@@ -150,46 +148,44 @@ protected:
...
-        parity(parity),
-        nParity(in.SiteSubset()),
-        threads(0),
-        a(a),
-        b(b),
-        c(c)
-    {
-      fillAux();
+  Pack(void *ghost[], const ColorSpinorField &in, MemoryLocation location, int nFace, bool dagger, int parity, double a,
+       double b, double c, int shmem_) :
```
weinbe2: style: `shmem_` -> `shmem`, matching other args

lib/dslash_policy.cuh
```diff
@@ -435,7 +439,72 @@ namespace quda
     }
   };
 
-/**
+  /**
+     Generic shmem dslash
+      // shmem bitfield encodes
```
weinbe2, line 444: Readability thought, we should make this an `enum` in `enum_quda.h` to remove the random magic numbers in the code.

lib/dslash_policy.cuh
```diff
@@ -435,7 +439,72 @@ namespace quda
...
+        if (!(shmem & 64)) {
+          PROFILE(if (dslash_exterior_compute) dslash.apply(streams[Nstream - 1]), profile, QUDA_PROFILE_DSLASH_KERNEL);
+        }
+      }
+
+      dslash::inc_shmem_sync_counter();
+      in->bufferIndex = (1 - in->bufferIndex);
+      profile.TPSTOP(QUDA_PROFILE_TOTAL);
+#else
+      errorQuda("NVSHMEM Dslash policies not build.");
```
weinbe2: build -> built

lib/dslash_quda.cu
```diff
@@ -10,6 +10,9 @@
 #include <linalg.cuh>
 #include <dslash_policy.cuh>
 #include <instantiate.h>
+#ifdef NVSHMEM_COMMS
```
weinbe2, line 13: Random thought I had here, should we put this in the `shmem_helper`/some other common relevant header so we don't have this in multiple places? (I imagine this will happen one way or another in `generic_kernels` anyway.)

lib/dslash_quda.cu
```diff
@@ -98,6 +125,27 @@ namespace quda {
       cudaEventCreateWithFlags(&packEnd[i], cudaEventDisableTiming);
       cudaEventCreateWithFlags(&dslashStart[i], cudaEventDisableTiming);
     }
+#ifdef NVSHMEM_COMMS
+    sync_arr = (shmem_sync_t *)device_comms_pinned_malloc(2 * QUDA_MAX_DIM * sizeof(shmem_sync_t));
```
weinbe2: `(shmem_sync_t *)` -> `static_cast<shmem_sync_t *>`, matching convention below

lib/dslash_quda.cu
```diff
@@ -98,6 +125,27 @@ namespace quda {
...
+    cudaMemset(sync_arr, 0, 2 * QUDA_MAX_DIM * sizeof(shmem_sync_t));
```
weinbe2: `cudaMemset` -> `qudaMemset`? (I feel like there was some fuzziness about `qudaMemset` here or there, thus the "?")

lib/targets/cuda/malloc.cpp
```diff
@@ -58,6 +60,13 @@ namespace quda
     }
   };
 
+#ifdef NVSHMEM_COMMS
+  // FIXME: Temporary value, based on DGX-1V layout
```
weinbe2: Any updates?

lib/targets/cuda/malloc.cpp
```diff
@@ -359,6 +368,61 @@ namespace quda
...
+    a.size = a.base_size = size;
+
+    auto dev = 0;
+    auto err = cudaGetDevice(&dev);
+    if (err != cudaSuccess) {
+      printfQuda("ERROR: cudaGetDevice failed in shmem_malloc_ (%s:%d in %s())\n", file, line, func);
+      errorQuda("Aborting");
+    }
+
+    if (!has_texture_alignment[dev]) {
```
weinbe2: Should we be setting `texture_alignment` and `has_texture_alignment` for all devices as opposed to just device 0? If not, can we make these values scalars instead of arrays?

lib/targets/cuda/malloc.cpp
```diff
@@ -573,6 +685,13 @@ namespace quda
         }
         pool_init = true;
       }
+#if defined(NVSHMEM_COMMS)
+      MPI_Comm tmp = MPI_COMM_WORLD;
```
weinbe2, line 648: One thing to be mindful of---the default MPI space may not be `MPI_COMM_WORLD`, and that doesn't necessarily mean split grid is being used. This is exposed in the QUDA interface here: https://github.com/lattice/quda/blob/develop/lib/interface_quda.cpp#L361

I'm not sure if there's a great solution to this, since a user passing in their own custom handle _could_ mean they're managing split-grid on their own, and I don't know how that'll play with `nvshmem`... ugh.

weinbe2: This gets leveraged in MILC (via the MILC interface exposure) here: https://github.com/milc-qcd/milc_qcd/blob/develop/generic/milc_to_quda_utilities.c#L35 ; tracing the routine `mycomm` becomes a bit of a rabbit hole.

lib/targets/cuda/malloc.cpp
```diff
@@ -679,6 +840,22 @@ namespace quda
       }
     }
 
+#ifdef NVSHMEM_COMMS
+#define shmem_free(ptr) quda::shmem_free_(__func__, quda::file_name(__FILE__), __LINE__, ptr)
+    void flush_shmem()
+    {
+      if (shmem_memory_pool) {
+        std::multimap<size_t, void *>::iterator it;
```
weinbe2: Since `multimap`s can have multiple keys pointing to the same value, shouldn't we need some guarantee that a value hasn't already been `shmem_free`'d? Alternatively, does the shmem memory pool actually need to be a `multimap` (I think it's unique among the pools as being one), or would a `map` suffice?

weinbe2:
Done with my first full visual pass through the branch, littered the code with comments/questions. I'll get this tested with some sort of `MILC` measurement (as opposed to RHMD) workflow.

include/kernels/covDev.cuh
mathiaswagner: I am inclined to put this to a fixed later and file an issue for looking into this later. But will think about it more.

lib/targets/cuda/malloc.cpp
mathiaswagner: This is the same thing we use for the other pools as well. But it actually brings up the question whether we do need a pool for symmetric allocations. Since we move to hide all shmem allocations behind `device_comms_pinned_` which falls back to `device_pinned_` if shmem is not used maybe just using a pure shmem_malloc is sufficient here.

lib/dslash_quda.cu
mathiaswagner: The `#ifdef NVSHMEM_COMMS` ?

lib/dslash_policy.cuh
mathiaswagner: Yes, I was waiting for C++17 to clean that up using something like https://github.com/Neargye/magic_enum

include/tune_quda.h
mathiaswagner: Not sure I follow. This is a global property here...

lib/targets/cuda/malloc.cpp
mathiaswagner: We can pass in MPI_COMM_HANDLE_USER here if this has been set. Or we can disallow setting a different handle here for now with nvshmem. We need to do some more investigation with split_grid anyway.

lib/targets/cuda/malloc.cpp
mathiaswagner: might be no longer necessary (see other comment)

lib/targets/cuda/malloc.cpp
mathiaswagner: guess with texture gone from QUDA we no longer need that at all ...

include/kernels/covDev.cuh
maddyscientist: Yes, I'd be happy deferring this.  It could be reexamined circa multi-RHS.

mathiaswagner: #1117

lib/targets/cuda/malloc.cpp
mathiaswagner: fixed.

lib/targets/cuda/malloc.cpp
mathiaswagner: no longer there

include/tune_quda.h
weinbe2: Oh, yep, this is my bad, and what happens when I assume too much about what github's web interface squishes away.

lib/dslash_policy.cuh
weinbe2: Fair, and I'm happy deferring it, but it'd be nice to clean up the random magic numbers sooner as opposed to later in my opinion, even though abusing enum <-> int casts can be headache inducing.

lib/dslash_quda.cu
weinbe2: The `#include <cuda/atomic>`

weinbe2: (and corresponding ifdef wrapper)

lib/targets/cuda/malloc.cpp
weinbe2: For the time being we can just add a warning around using NVSHMEM with split grid, either within QUDA or as done by other apps. Something along the lines of QUDA's NVSHMEM work assumes your MPI world == MPI_COMM_WORLD?

lib/dslash_quda.cu
mathiaswagner: No, we don't want to include the shmem_helper in device code without device-linking ()i.e. code that just needs the atomic) as it will warn about `extern` being treated as static (in the nvshmem headers). So I'd like to keep that separate for now. But worth looking into further cleanup and see how things evolve with atomic usage.

mathiaswagner: #1118

lib/CMakeLists.txt
mathiaswagner: That they should use Ninja?
No, this only affects the parallelization fo the compile. Ninja does a better job in this case than make, but make works. No need to confuse users. Ninjas will use ninja anyway.

lib/dslash_pack2.cu
mathiaswagner: Above if has changed (now line 73). This no longer uniquely is always host and we only limit the maxGridsize for zero copy and nvshmem uber kernel (shmem & 32).

lib/CMakeLists.txt
```diff
@@ -100,6 +90,24 @@ set (QUDA_OBJS
...
+  dslash_staggered.cu dslash_improved_staggered.cu
+  dslash_wilson.cu dslash_wilson_clover.cu dslash5_domain_wall.cu
+  dslash_wilson_clover_preconditioned.cu 
+  dslash_twisted_mass.cu dslash_twisted_mass_preconditioned.cu
+  dslash_ndeg_twisted_mass.cu dslash_ndeg_twisted_mass_preconditioned.cu
+  dslash_twisted_clover.cu dslash_twisted_clover_preconditioned.cu
+  dslash_wilson_clover_hasenbusch_twist.cu
+  dslash_wilson_clover_hasenbusch_twist_preconditioned.cu
+  dslash_domain_wall_4d.cu  dslash_domain_wall_5d.cu
+  mdw_fused_dslash.cu dslash5_mobius_eofa.cu
```
maddyscientist: I don't think there's any need for mdw_fused_dslash.cu and dslash5_mobius_eofa.cu to be listed here.  There are no communications in these kernels.

include/dslash_helper.cuh
```diff
@@ -364,7 +427,8 @@ namespace quda
     out << std::endl;
     out << "active_dims = " << arg.active_dims << std::endl;
     out << "pack_blocks = " << arg.pack_blocks << std::endl;
-
+    out << "exterior_threads = " << arg.exterior_threads;
+    out << "exterior_blocks =" << arg.exterior_blocks;
```
maddyscientist: "exterior_blocks =" -> "exterior_blocks = "

maddyscientist (approved):
Subject to the minor couple of issues I just pointed out, approving this PR.  This is an excellent addition to QUDA, thank you @mathiaswagner for this fantastic work.

lib/CMakeLists.txt
mathiaswagner: Looking through that it seems that `mdw_fused_dslash.cu` triggers some warnings (nvshmem extern being treated as static). I wonder if cleaning up the `instantiate` there can help. @hummingtree Do you think it is possible to move the `instantiatePreconditioner` out of `dslash.h`?

I think if we can remove the include `dslash.h` and only include what is actually need we might be able to clean that up.

mathiaswagner: If I change the includes for `mdw_fused_dslash.cu` like

```
-#include <dslash.h>
+
+#include <typeinfo>
+
+#include <tune_quda.h>
+#include <dslash_quda.h>
+#include <jitify_helper.cuh>
+#include <instantiate.h>
```

and add all the `instantiate` from dslash.h explicitly I can pull it out of the `dslash_pack` target. 

Do we want to move them into a new header? Does this conflict with GK somehow? In that case we might want to fix that later and leave it as it is for now.

hummingtree: I will try cleaning this up (after re-compile).

mathiaswagner: The critical thing is that it should not include `dslash_helper.cuh` in any way since this pull in nvshmem headers which trigger the warning when compiled without `rdc=true` (which is the case for the dslash*.cu files in the list above).

include/kernels/dslash_pack.cuh
```diff
@@ -47,9 +52,29 @@ namespace quda
...
+    char *packBuffer[4 * QUDA_MAX_DIM];
+    int neighbor_ranks[2 * QUDA_MAX_DIM];
+    int bytes[2 * QUDA_MAX_DIM];
+    // shmem bitfield encodes
+    // 0 - no shmem
+    // 1 - pack P2P
+    // 2 - pack IB
+    // 3 - pack P2P + IB
+    // 8 - barrier part I (just the put part)
+    // 16 - barrier part II (wait on shmem to complete, all directions) -- not implemented
```
weinbe2, line 64: `16` seems to be noted as implemented in `lib/dslash_policy.cuh`---are these different bitfields? Or is this wrong/I'm misunderstanding the comment in that file?

lib/dslash_policy.cuh
```diff
@@ -1946,7 +2060,7 @@ public:
      popKernelPackT();
    }
 
-   int tuningIter() const { return 10; }
+   int tuningIter() const { return 20; }
```
weinbe2, line 2063: Can this be reduced back down to 10? I'm thinking about how long the nvshmem policies take to tune as it is. If this is empirically worth it for stability I'm happy to keep it this way.

weinbe2 (approved):
Pending the naive staggered tests and a few small questions I've added, I'm very happy with this PR and looking forward to breaking it out in the real world---thanks @mathiaswagner !

lib/dslash_policy.cuh
mathiaswagner: This is only for the final policy tuning. What takes time is the exploded tune space for the fused kernels.

include/kernels/dslash_pack.cuh
mathiaswagner: thanks. The comment was outdated.

Issue #1066, opened 2020-10-20
CPU preconditioned coarse link construction is broken when communications (and/or partition != 0) is enabled


weinbe2:
This is more to record the issue for the time being, I'll go on a `git bisect` hunt this week. GPU construct is fine so this is not a blocking issue.

The issue can be roughly traced to the bi-directional link exchange here: https://github.com/lattice/quda/blob/develop/lib/coarse_op_preconditioned.cu#L234

The evidence suggesting this is the `norm2` of `Yhat` with `--partition 15` deviates from the `norm2` with `--partition 0` by a factor that can be explained by surface-area-to-volume considerations. The construction of `X`, `Y`, and `Xhat` are all ship-shape, and there doesn't appear to be any reason why the `Yhat` construct kernel itself would have a problem gathering from the ghost buffer.

This can be reproduced in `develop` for Wilson fermions, as well as in `feature/optimized-kd` for both Wilson and staggered construct.

weinbe2:
I've looked into this across a few volumes and partitionings, and @hummingtree has looked into this for formal multi-GPU builds, and it seems to have fixed itself. Closing this.

PR #1067, opened 2020-10-20
Staggered aggregation coarsening and optimized Kahler-Dirac operator

weinbe2:
This PR is the first of two or three related to further optimizing staggered MG setup and solve performance. The key features of this first PR are re-implementing support for coarsening the staggered and (fat-link part of the) HISQ operator, implementing an optimized Kahler-Dirac preconditioned operator which leverages the fine-level staggered/HISQ op directly, as well as adding an enum for the `Transfer` object which, for now, is used to specify the form of the KD operator, with the intention of being the interface to specify different types of DWF multigrid in the future.

This PR also fixes a few sneaky bugs: some Dirac matrix types overwriting temporary vectors (see `dirac_quda.h`, changes to `DiracMdagM`, for example), and there were issues generating appropriately phased free fields in the staggered test routines. I also added some doxygen in a few parts of the code that were giving me headaches for maybe no good reason.

The new staggered and improved staggered KD operators are defined in `dirac_(improved_)staggered_kd.cpp`. They inherit directly from the `Dirac(Improved)Staggered` classes, and as such act on `Nspin ==1, Ncolor == 3` `ColorSpinorField`s. They also carry around an extra field, `Xinv`, which is the inverse of the "super-site" local Kahler-Dirac block. This field is identical to the `Xinv` generated by the existing staggered MG implementation. Two new special routines were introduced to build (`BuildStaggeredKahlerDiracInverse`, defined in `staggered_kd_build_xinv.cu`) and apply (`ApplyStaggeredKahlerDiracInverse`, defined in `staggered_kd_apply_xinv.cu`) the Kahler-Dirac block, in the latter case to fine `ColorSpinorField`s. Of note, the precision of `Xinv` does not have to match the precision of the input vector when applying the Kahler-Dirac inverse, in line with MG coarse operators always acting on single precision `ColorSpinorField`s even though the fields themselves may be stored as `float` or `half`. 

The new operators are not exposed outside of an MG solve. In many regards this is intentional because they live in a weird space as operators, being preconditioned but still being full parity. This isn't an inherent problem, but a lot of QUDA assumes a preconditioned operator is single-parity, and fixing that everywhere is outside the scope of this PR. This did require a few hacks to get the right behavior within MG preconditioned solves, as can be seen in the changes to `PreconditionedSolver` in `invert_quda.h`. In a similar vein, this operator has the pieces in place for left versus right preconditioning, it's just not exposed yet because the plumbing isn't there in QUDA (yet).

This PR does not support coarsening the new operator; that is the scope of the next PR. That said, I put a lot of the plumbing in place to pass `Xinv` through, as can be seen in the updates to `staggered_coarse_op.cu`.

The interface parameter bool `is_staggered` has been replaced by a new enum, `QudaTransferType transfer_type`, which implies different prescriptions for coarsening/prolongating/restricting. It can take on the values:
* `QUDA_TRANSFER_AGGREGATE` (traditional aggregation, replaces `is_staggered == false` on the fine level, and is the default for all coarse levels)
* `QUDA_TRANSFER_COARSE_KD` (existing staggered MG unitary transformation between `Nc = 3, Ns = 1` and `Nc = 24, Ns = 2` on the KD lattice, replaces `is_staggered == true` on the fine level)
* `QUDA_TRANSFER_OPTIMIZED_KD` (new staggered MG, which is just a copy).

New parameters can be added in the future for different domain wall prescriptions, etc.

While outside of the scope of this PR, this interface suggests a way to exploit the MG interface to precondition any fine operator with other fine operators, for example `DiracImprovedStaggered` with `DiracStaggered` or `DiracClover` with `DiracWilson`, to name a few.

This PR does a little bit of compile-time cleanup by taking out some combinations of coarse staggered Nc that are not used in practice: `3 -> 96` and `96 -> 96`. This forces a workflow of `3` -> (`24` ->) `64` -> `64`/`96`, which is fine. Compiling still takes forever, so I recommend testing with a jitify build.

As a side note before getting into example commands, I discovered over the course of this PR that there are standing issues with CPU setup for multi-GPU/single-GPU w/partition != 0. This has been reported in #1066 and will be fixed separately. CPU setup goes through with `--partition 0` without issue.

### Tests

Aggregation of the staggered operator (in case you needed a reminder that coarsening the staggered operator directly is a bad idea). Note the new command line parameter `--mg-staggered-coarsen-type aggregate`.

```
ARGSTR="--dim 16 16 16 16 --prec double --prec-sloppy single --prec-precondition half --prec-null half --dslash-type asqtad --solve-type direct --compute-fat-long true --partition 0 --verbosity verbose --nsrc 3 --inv-multigrid true "
ARGSTR=$ARGSTR"--load-gauge l16t16b7p0 --mass 0.01 --tadpole-coeff 0.9 "
ARGSTR=$ARGSTR"--mg-levels 3 --mg-coarse-solve-type 0 direct --mg-verbosity 0 verbose --mg-staggered-coarsen-type aggregate "
ARGSTR=$ARGSTR"--mg-setup-inv 0 cgnr --mg-setup-maxiter 0 1000 --mg-setup-tol 0 1e-5 "
ARGSTR=$ARGSTR"--mg-setup-inv 1 cgnr --mg-setup-maxiter 1 1000 --mg-setup-tol 1 1e-5 "
ARGSTR=$ARGSTR"--mg-block-size 0 4 4 4 4 --mg-nvec 0 64 --mg-n-block-ortho 0 2 --mg-coarse-solve-type 1 direct-pc --mg-smoother-solve-type 0 direct --mg-smoother 0 ca-gcr --mg-nu-pre 0 0 --mg-nu-post 0 4 --mg-smoother-tol 0 1e-10 --mg-coarse-solver-tol 1 0.25 --mg-coarse-solver-maxiter 1 16   --mg-coarse-solver 1 gcr --mg-verbosity 1 verbose "
ARGSTR=$ARGSTR"--mg-block-size 1 2 2 2 2 --mg-nvec 1 96 --mg-n-block-ortho 1 2 --mg-coarse-solve-type 2 direct-pc --mg-smoother-solve-type 1 direct-pc --mg-smoother 1 ca-gcr --mg-nu-pre 1 0 --mg-nu-post 1 4 --mg-smoother-tol 1 1e-10 --mg-coarse-solver-tol 2 0.25 --mg-coarse-solver-maxiter 2 12   --mg-coarse-solver 2 ca-gcr --mg-coarse-solver-ca-basis-size 2 12 --mg-verbosity 2 verbose "
./staggered_invert_test $ARGSTR
```

The "traditional" implementation of staggered MG. This is still the default, and indeed the only way to get a functional staggered MG algorithm

```
ARGSTR="--dim 16 16 16 16 --prec double --prec-sloppy single --prec-precondition half --prec-null half --dslash-type asqtad --solve-type direct --compute-fat-long true --partition 0 --verbosity verbose --nsrc 3 --inv-multigrid true "
ARGSTR=$ARGSTR"--load-gauge l16t16b7p0 --mass 0.01 --tadpole-coeff 0.9 "
ARGSTR=$ARGSTR"--mg-levels 4 --mg-coarse-solve-type 0 direct --mg-verbosity 0 verbose "
ARGSTR=$ARGSTR"--mg-setup-inv 1 cgnr --mg-setup-maxiter 1 1000 --mg-setup-tol 1 1e-5 "
ARGSTR=$ARGSTR"--mg-setup-inv 2 cgnr --mg-setup-maxiter 2 1000 --mg-setup-tol 2 1e-5 "
ARGSTR=$ARGSTR"--mg-block-size 0 2 2 2 2 --mg-nvec 0 24 --mg-n-block-ortho 0 2 --mg-coarse-solve-type 1 direct-pc --mg-smoother-solve-type 0 direct    --mg-smoother 0 ca-gcr --mg-nu-pre 0 0 --mg-nu-post 0 4 --mg-smoother-tol 0 1e-10 --mg-coarse-solver-tol 1 0.25 --mg-coarse-solver-maxiter 1 16   --mg-coarse-solver 1 gcr --mg-verbosity 1 verbose "
ARGSTR=$ARGSTR"--mg-block-size 1 2 2 2 2 --mg-nvec 1 64 --mg-n-block-ortho 1 2 --mg-coarse-solve-type 2 direct-pc --mg-smoother-solve-type 1 direct-pc --mg-smoother 1 ca-gcr --mg-nu-pre 1 0 --mg-nu-post 1 4 --mg-smoother-tol 1 1e-10 --mg-coarse-solver-tol 2 0.25 --mg-coarse-solver-maxiter 2 16   --mg-coarse-solver 2 gcr --mg-verbosity 2 verbose "
ARGSTR=$ARGSTR"--mg-block-size 2 2 2 2 2 --mg-nvec 2 96 --mg-n-block-ortho 2 2 --mg-coarse-solve-type 3 direct-pc --mg-smoother-solve-type 2 direct-pc --mg-smoother 2 ca-gcr --mg-nu-pre 2 0 --mg-nu-post 2 4 --mg-smoother-tol 2 1e-10 --mg-coarse-solver-tol 3 5e-2 --mg-coarse-solver-maxiter 3 16   --mg-coarse-solver 3 ca-gcr --mg-coarse-solver-ca-basis-size 3 16 --mg-verbosity 3 verbose "
ARGSTR=$ARGSTR"--mg-eig 3 true --mg-eig-type 3 trlm --mg-eig-use-dagger 3 false --mg-eig-use-normop 3 true --mg-eig-use-poly-acc 3 true --mg-eig-spectrum 3 LR "
ARGSTR=$ARGSTR"--mg-nvec 3 24 --mg-eig-n-ev 3 24 --mg-eig-n-kr 3 64 --mg-eig-tol 3 1e-6 --mg-eig-poly-deg 3 50 --mg-eig-amin 3 1e-1 "
./staggered_invert_test $ARGSTR
```

"Optimized" Kahler-Dirac "MG", which in reality is just preconditioning the staggered op with the new Kahler-Dirac operator.  Note the parameter `--mg-staggered-coarsen-type kd-optimized`. Attempting a 3-level solve will throw an error appropriately.

```
ARGSTR="--dim 16 16 16 16 --prec double --prec-sloppy single --prec-precondition half --prec-null half --dslash-type asqtad --solve-type direct --compute-fat-long true --partition 0 --verbosity verbose --nsrc 3 --inv-multigrid true "
ARGSTR=$ARGSTR"--load-gauge l16t16b7p0 --mass 0.01 --tadpole-coeff 0.9 "
ARGSTR=$ARGSTR"--mg-levels 2 --mg-coarse-solve-type 0 direct --mg-verbosity 0 verbose --mg-staggered-coarsen-type kd-optimized "
ARGSTR=$ARGSTR"--mg-block-size 0 1 1 1 1 --mg-nvec 0 3  --mg-coarse-solve-type 1 direct    --mg-smoother-solve-type 0 direct    --mg-smoother-solve-type 1 direct --mg-smoother 0 ca-gcr --mg-nu-pre 0 0 --mg-nu-post 0 4 --mg-smoother-tol 0 1e-10 --mg-coarse-solver-tol 1 0.025 --mg-coarse-solver-maxiter 1 1024   --mg-coarse-solver 1 cgnr --mg-coarse-solver-ca-basis-size 1 16 --mg-verbosity 1 verbose "
./staggered_invert_test $ARGSTR
```

### To do

Just a few small things left, but nothing blocking the review of this PR

- [x] Test PR against the MILC HISQ MG interface
- [x] Large battery of Wilson-type MG tests based on https://github.com/lattice/quda/wiki/Validating-Large-Multigrid-Changes ---- ongoing
- [x] Due-diligence test of Chroma MG-accelerated HMC
- [x] Updating the wiki
- [x] Sanitize build
- [x] Doxygen (will wait until review is done)


include/index_helper.cuh
```diff
@@ -336,6 +336,29 @@ namespace quda {
...
+      the full linear lattice (not what's used for
+      indexing into fields, re:padding argument for getIndexFull).
+
+      @param[out] out_parity Output site parity
+      @param[in] X full lattice dimensions
+      @param[in] full_index full linear lattice index
+      @return 1-d checkerboard index
+   */
+  template <typename I>
+  __device__ __host__ inline int getParityCBFromFull(int& out_parity, const I X[4], const int full_index) {
```
hummingtree, line 349: Just a comment: I believe there are more index utility functions in this file than necessary, it might take some cleaning in the future.

weinbe2:
A numerical pathology of note: the naively coarsened operator (corresponding to the aggregation mode above) sometimes generates `nan` for the inner-most CA-GCR solve. After some investigation, it's clear this is from a numerical instability that manifests when inverting the dense matrix inside CA-GCR and not because of a bug. There are a few pieces of evidence for this:

* You can tune this effect by increasing the `N` for CA-GCR(N) for "CA" basis size `n` == `N`
* You can eliminate this effect by decreasing basis size `n` relative to `N`
* Most concretely, you can inject printing the result of the Cholesky solve in CA-GCR and see the `nan`s, for example, adding the following code below line 149 of `inv_ca_gcr.cpp`:
```
for (int i = 0; i < N; i++) {
  printf("phi component %d value %e + I %e, psi component value %e + I %e\n", i, real(phi(i)), imag(phi(i)), real(psi(i)), imag(psi(i)));
}
```

In some regards, this isn't a surprise, and you can probably trigger this for any operator if you try hard enough---the coarsened staggered op is just such a train wreck it's not that hard to do.

Anyway: if you hit this issue in testing, the easiest solution is to change the coarsest level max iters and CA basis size to 12 from the current 16.

weinbe2:
FYI, found a new bug in half precision (turns out even a buggy operator is a good enough operator sometimes), investigating now.

weinbe2:
All's good now!

For everyone's entertainment, for a 24^4 configuration, here's the total memory req'd for a solve with a two level, half precision link solve:
Explicit coarse operator: 6183.9 MB
And with the optimized operator: 2876.8 MB

Where the difference can near-spot-on be explained by not needing the coarse clover `X`, coarse links `Y`, and preconditioned coarse links `Yhat`.

include/gauge_field_order.h
```diff
@@ -905,17 +905,22 @@ namespace quda {
...
-    template <typename Float, int nColor, int nSpinCoarse, QudaGaugeFieldOrder order, bool native_ghost = true,
-              typename storeFloat = Float>
+    template <typename Float_, int nColor, int nSpinCoarse, QudaGaugeFieldOrder order, bool native_ghost = true,
+              typename storeFloat_ = Float_>
     struct FieldOrder {
 
+      /** Convenient types */
+      using Float = Float_;
+      using storeFloat = storeFloat_;
+
```
hummingtree, line 923: Some noise here: in the MMA PR I used `type` and `store_type` [here](https://github.com/lattice/quda/blob/e3612009273c0c3a947f67b6faca675551edc3e4/include/gauge_field_order.h#L241). We should probably unify the namings at some point.

hummingtree:
Looks good after a visual review. Next will build and run a few tests.

hummingtree (approved):
Tested with the three MG tests shown in the description of this PR, with both CPU and GPU setup, and all are good.

include/gauge_field_order.h
maddyscientist: ...and I am now favouring `store_t` 😜

hummingtree: So `store_t` and `compute_t`?

maddyscientist: I think that would be fine.  But please no global edits until generic kernels is merged in.  One global edit at a time is already one too many....

include/kernels/staggered_kd_build_xinv_kernel.cuh
```diff
@@ -0,0 +1,140 @@
...
+        for (int ic_f=0; ic_f<fineColor; ic_f++) {
+          for (int jc_f=0; jc_f<fineColor; jc_f++) {
+            ComputeStaggeredKDBlock<Float,fineColor,coarseSpin,coarseColor>(arg, parity, x_cb, ic_f, jc_f);
+          } // coarse color columns
+        } // coarse color rows
+      } // c/b volume
+    } // parity
+  }
+
+  template<typename Float, int fineColor, int coarseSpin, int coarseColor, typename Arg>
```
maddyscientist: For minimal changes for generic kernels, please remove all the template parameters except `Arg` and make all the other template parameters members of `Arg`

include/multigrid.h
```diff
@@ -245,6 +245,9 @@ namespace quda {
     /** Coarse temporary vector */
     ColorSpinorField *tmp2_coarse;
 
+    /** Kahler-Dirac Xinv */
+    cudaGaugeField *xInvKD;
```
maddyscientist, line 249: Why is this a `cudaGaugeField` and not just `GaugeField`?

lib/coarse_op.cu
```diff
@@ -2,14 +2,16 @@
 #include <color_spinor_field.h>
 #include <gauge_field.h>
 #include <clover_field.h>
+
+#define WILSONCOARSE
```
maddyscientist, line 9: add a comment here perhaps to explain what this does

lib/color_spinor_field.cpp
```diff
@@ -776,9 +776,22 @@ namespace quda {
                                                    QudaMemoryType new_mem_type) {
     ColorSpinorParam coarseParam(*this);
     for (int d=0; d<nDim; d++) coarseParam.x[d] = x[d]/geoBlockSize[d];
-    coarseParam.nSpin = (nSpin == 1) ? 2 : (nSpin / spinBlockSize); // coarsening staggered check
 
-    coarseParam.nColor = Nvec;
+    // check for optimized-KD build by seeing if total geoBlockSize == 1
```
maddyscientist: This looks like it will prevent coarsening the degrees of freedom for a Wilson MG lattice at constant lattice geometry.  It's not something we presently care about, but I don't want to hard code against this possibility with the below logic.

I think a more robust enablement of KD optimized is required here.

lib/cpu_gauge_field.cpp
```diff
@@ -65,8 +65,8 @@ namespace quda {
       }
 
       if (create == QUDA_NULL_FIELD_CREATE || create == QUDA_ZERO_FIELD_CREATE) {
-	gauge = (void **) safe_malloc(bytes);
-	if(create == QUDA_ZERO_FIELD_CREATE) memset(gauge, 0, bytes);
+	gauge = bytes ? (void **) safe_malloc(bytes) : nullptr;
+	if(create == QUDA_ZERO_FIELD_CREATE && bytes) memset(gauge, 0, bytes);
```
maddyscientist: Look like this PR needs a clang-format.

lib/dirac_improved_staggered.cpp
```diff
@@ -133,7 +133,12 @@ namespace quda {
   void DiracImprovedStaggered::createCoarseOp(GaugeField &Y, GaugeField &X, const Transfer &T, double kappa,
                                               double mass, double mu, double mu_factor) const
   {
-    StaggeredCoarseOp(Y, X, T, *fatGauge, mass, QUDA_ASQTAD_DIRAC, QUDA_MATPC_INVALID);
+    if (T.getTransferType() == QUDA_TRANSFER_OPTIMIZED_KD)
+      errorQuda("The optimized improved Kahler-Dirac operator is not built through createCoarseOp");
+
+    // nullptr == no Kahler-Dirac Xinv
+    const cudaGaugeField* XinvKD = nullptr;
```
maddyscientist: Should not use `cudaGaugeField` and instead should be `GaugeField`

lib/dirac_staggered.cpp
```diff
@@ -129,7 +129,12 @@ namespace quda {
 
   void DiracStaggered::createCoarseOp(GaugeField &Y, GaugeField &X, const Transfer &T,
            double kappa, double mass, double mu, double mu_factor) const {
-    StaggeredCoarseOp(Y, X, T, *gauge, mass, QUDA_STAGGERED_DIRAC, QUDA_MATPC_INVALID);
+    if (T.getTransferType() == QUDA_TRANSFER_OPTIMIZED_KD)
+      errorQuda("The optimized Kahler-Dirac operator is not built through createCoarseOp");
+
+    // nullptr == no Kahler-Dirac Xinv
+    const cudaGaugeField* XinvKD = nullptr;
```
maddyscientist: `cudaGaugeField` -> `GaugeField`

lib/staggered_kd_apply_xinv.cu
```diff
@@ -0,0 +1,254 @@
...
+    // the staggered KD block inverse can only be applied to a full field
+    if (out.SiteSubset() != QUDA_FULL_SITE_SUBSET || out.SiteSubset() != QUDA_FULL_SITE_SUBSET)
+      errorQuda("There is no meaning to applying the KD inverse to a single parity field");
+
+    // Instantiate based on ColorSpinor precision
+    // We don't have a constraint on the precision of Xinv matching
+    // the precision of the spinors.
+    auto precision = checkPrecision(out, in);
+
+#if QUDA_PRECISION & 8
```
maddyscientist: If you turn `applyStaggeredKDBlock` into a class, you can use `instantiatePrecision` here to clean up this `if` / `else` precision boilerplate

lib/staggered_kd_build_xinv.cu
```diff
@@ -0,0 +1,389 @@
...
+      if (meta.Location() == QUDA_CPU_FIELD_LOCATION) {
+        ComputeStaggeredKDBlockCPU<Float,fineColor,coarseSpin,coarseColor>(arg);
+      } else {
+#ifdef JITIFY
+        using namespace jitify::reflection;
+        jitify_error = program->kernel("quda::ComputeStaggeredKDBlockGPU")
+          .instantiate(Type<Float>(),fineColor,coarseSpin,coarseColor,Type<Arg>())
+          .configure(tp.grid,tp.block,tp.shared_bytes,stream).launch(arg);
+#else // not jitify
+        qudaLaunchKernel(ComputeStaggeredKDBlockGPU<Float,fineColor,coarseSpin,coarseColor,Arg>, tp, stream, arg);
```
maddyscientist: as noted elsewhere, need to put these template arguments into the `Arg`

lib/transfer.cpp
```diff
@@ -79,12 +79,26 @@ namespace quda {
       total_block_size *= geo_bs[d];
     }
 
-    if (total_block_size == 1) errorQuda("Total geometric block size is 1");
+    // Aggregation size is "technically" 1 for optimized KD
+    if (total_block_size == 1 && transfer_type != QUDA_TRANSFER_OPTIMIZED_KD) errorQuda("Total geometric block size is 1 but transfer type isn't optimized-kd");
```
maddyscientist: As noted elsewhere, we want to preserve the ability to thin the degrees of freedom at constant geometry without needing KD optimized build.

lib/staggered_coarse_op.cu
```diff
@@ -207,28 +242,181 @@ namespace quda {
...
+      if (T.Vectors(Y.Location()).FieldOrder() != csOrder)
+        errorQuda("Unsupported field order %d\n", T.Vectors(Y.Location()).FieldOrder());
+      if (g.FieldOrder() != gOrder) errorQuda("Unsupported field order %d\n", g.FieldOrder());
+
+      typedef typename colorspinor::FieldOrderCB<Float,fineSpin,fineColor,coarseColor,csOrder,vFloat> V;
+      typedef typename colorspinor::FieldOrderCB<Float,fineSpin,fineColor,coarseColor,csOrder,vFloat> F; // will need 2x the spin components for the KD op
+      typedef typename gauge::FieldOrder<Float,fineColor,1,gOrder> gFine;
+      typedef typename gauge::FieldOrder<typename mapper<vFloatXinv>::type,xinvColor,xinvSpin,xinvOrder,true,vFloatXinv> xinvFine;
+      typedef typename gauge::FieldOrder<Float,coarseColor*coarseSpin,coarseSpin,gOrder,true,vFloat> gCoarse;
+      typedef typename gauge::FieldOrder<Float,coarseColor*coarseSpin,coarseSpin,gOrder,true,storeType> gCoarseAtomic;
```
maddyscientist: General style comment, I think we should be preferring `using` over `typedef` since it's easier to read, e.g., delineation between what we're defining to be what, `using` has the advantage of `=` between.

include/staggered_kd_build_xinv.h
```diff
@@ -0,0 +1,28 @@
...
+  void BuildStaggeredKahlerDiracInverse(GaugeField &Xinv, const cudaGaugeField &gauge, const double mass);
+
+  /**
+     @brief Allocate and build the Kahler-Dirac inverse block for KD operators
+     @param[in] in gauge original fine gauge field
+     @param[in] in mass the mass of the original staggered operator w/out factor of 2 convention
+     @param[in] in precision of Xinv field
+     @return constructed Xinv, which needs to be deleted manually
+  */
+  cudaGaugeField* AllocateAndBuildStaggeredKahlerDiracInverse(const cudaGaugeField &gauge, const double mass, const QudaPrecision override_prec);
```
maddyscientist: As noted elsewhere, why `cudaGaugeField` and not `GaugeField`?

maddyscientist:
First pass review done, and I've left comments to be addressed.

include/multigrid.h
weinbe2: To a point, no good reason, but it's more a declaration that the optimized KD operator acts more like the current fine ops than a coarse op. The current fine ops contain `cudaGaugeField` or `cudaCloverField` as opposed to the more general types, reflecting (for now) that the fine ops can only be applied to device fields, though I know that's going away soon.

In any case, I can update this.

lib/cpu_gauge_field.cpp
weinbe2: Discussed offline, will be the last step before merging to make it easier to track commit histories.

include/gauge_field_order.h
weinbe2: ...I'm going to wait to change this until we/you make a global decision in a later PR.

maddyscientist: Agreed!

lib/color_spinor_field.cpp
weinbe2: One approach would be flipping the logic, checking if `(geoBlockSize != 1 && transfer_type == QUDA_TRANSFER_OPTIMIZED_KD)`. Thoughts @maddyscientist ?

lib/staggered_coarse_op.cu
weinbe2: Done.

weinbe2: I changed it everywhere I added a `typedef`, and for the time being left other instances unchanged.

lib/coarse_op.cu
weinbe2: Done here, as well as in `lib/coarsecoarse_op.cu` and `lib/staggered_coarse_op.cu`.

lib/color_spinor_field.cpp
maddyscientist: That should work I think.

weinbe2: Ah, well, this'll work in the other case you tagged. This routine doesn't know about the transfer_type. We can make this check more robust though: if the geometric block size is 1 and the `ColorSpinorField` invoking `CreateCoarse` is `nSpin == 1` and the input `nVec == nColor` (the convention for the optimized KD op), then we know we're creating something that preserves volume and dof. If anything that motivates replacing `is_optimized_kd` with `is_mg_copy` or something.

weinbe2: For now we could go with `is_staggered_layout_preserving` in lieu of thinking through just using `ColorSpinorParam` copies, etc.

weinbe2: Or the current name, just with more comments...

lib/transfer.cpp
weinbe2: Fixed via logic noted in other comment, `if (total_block_size != 1 && transfer_type == QUDA_TRANSFER_OPTIMIZED_KD)`. Updated the error appropriately to note that geometric block size must be 1 for optimized KD op.

include/staggered_kd_build_xinv.h
weinbe2: Fixed return type. At this point the input fine gauge field will always be on the GPU; once that changes there will be a lot of places were we need to figure out the same logic of how `gauge` needs to be transformed.

lib/dirac_improved_staggered.cpp
weinbe2: Postponing to next PR.

lib/dirac_staggered.cpp
weinbe2: Postponing to next PR

lib/staggered_kd_apply_xinv.cu
weinbe2: We should discuss this on a call---this file has the quirk that Xinv can be stored in a separate precision from the ColorSpinorField, but that may just be a separate flavor of `instantiatePrecision`.

maddyscientist: That shouldn't matter, you just need to pass a field with the correct precision to be instantiated as the first argument to `instantiatePrecision`.  Subsequent field arguments can have different precisions which can be unwrapper later.

include/kernels/staggered_kd_build_xinv_kernel.cuh
weinbe2: Done

lib/staggered_kd_apply_xinv.cu
weinbe2: Done

weinbe2:
@maddyscientist pending Jenkins and barring further review comments, this PR is good to go

include/invert_quda.h
```diff
@@ -1198,9 +1198,20 @@ namespace quda {
       ColorSpinorField *out=nullptr;
       ColorSpinorField *in=nullptr;
 
-      dirac.prepare(in, out, x, b, solution_type);
+      bool special_prepare_recon
+        = (dirac.getDiracType() == QUDA_STAGGEREDKD_DIRAC || dirac.getDiracType() == QUDA_ASQTADKD_DIRAC);
+
```
maddyscientist: Since this is a property of the Dirac operator, why can't this be hidden inside the Dirac operator implementation rather than exposing this if / else condition to all?

lib/transfer.cpp
```diff
@@ -79,12 +79,27 @@ namespace quda {
       total_block_size *= geo_bs[d];
     }
 
-    if (total_block_size == 1) errorQuda("Total geometric block size is 1");
+    // Aggregation size is "technically" 1 for optimized KD
+    if (total_block_size != 1 && transfer_type == QUDA_TRANSFER_OPTIMIZED_KD)
+      errorQuda("Total geometric block size must be 1 for transfer type optimized-kd");
+    if (transfer_type == QUDA_TRANSFER_OPTIMIZED_KD) {
+      if (total_block_size != 1)
+        errorQuda("Invalid geometric block size %d for optimized-kd aggregation, must be 1", total_block_size);
```
maddyscientist: Looks like you're catching this error (`total_block_size != 1`) twice here

include/invert_quda.h
weinbe2: No good reason, fixed

lib/transfer.cpp
weinbe2: Good catch, fixed

maddyscientist (approved):
Ok, this looks good now.  Subject to the latest fixes being pushed, and a clang-format being applied, we can merge this.

weinbe2:
> Ok, this looks good now. Subject to the latest fixes being pushed, and a clang-format being applied, we can merge this.

Doing a due-diligence build now!

PR #1074, opened 2020-11-03
Feature/gamma5 op eigensolver

cpviolator:
A minor PR that allows the user to pre multiply the eigen operator with \gamma_5. Useful for DWF investigations. One must use the IRAM solver (which reduces to IRLM if the operator is hermitian) if one wants the Smallest Magnitude spectrum.

Example command:
```
./eigensolve_test --dim 8 8 8 8 --eig-use-poly-acc false --eig-type iram --eig-use-normop false --verbosity verbose --eig-spectrum SM --eig-n-conv 10 --eig-n-kr 96 --eig-compute-gamma5 true
```
will compute the smallest magnitude of a Wilson operator pre multiplied with gamma_5.


include/dirac_quda.h
```diff
@@ -2238,6 +2250,48 @@ namespace quda {
...
+    void operator()(ColorSpinorField &out, const ColorSpinorField &in) const
+    {
+      dirac->M(out, in);
+      if (shift != 0.0) blas::axpy(shift, const_cast<ColorSpinorField &>(in), out);
+      gamma5(out, out);
+    }
+
+    void operator()(ColorSpinorField &out, const ColorSpinorField &in, ColorSpinorField &tmp) const
+    {
+      dirac->tmp1 = &tmp;
```
weinbe2: This needs the `reset` logic present, for ex, in `DiracMdag` (which I added recently). This prevents an existing `dirac->tmp1`, etc, from being overriden if it's already set, which happens within multigrid. (I know this PR doesn't support gamma5 w/coarse ops yet, so this is more being proactive.)

cpviolator: https://github.com/lattice/quda/commit/78d633b6807d1ae4f9d33583ae7c30b145e3fe65

include/dirac_quda.h
```diff
@@ -2238,6 +2250,69 @@ namespace quda {
...
+
+  public:
+    DiracG5M(const Dirac &d) : DiracMatrix(d) { }
+    DiracG5M(const Dirac *d) : DiracMatrix(d) { }
+
+    void operator()(ColorSpinorField &out, const ColorSpinorField &in) const
+    {
+      dirac->M(out, in);
+      if (shift != 0.0) blas::axpy(shift, const_cast<ColorSpinorField &>(in), out);
+      gamma5(out, out);
```
weinbe2: If you want to extend this to staggered, this should work---staggered gamma5 corresponds to flipping the sign on odd sites. (I can test this later if you pop it in, or I can pop it in myself and test it, or whatever works best!)
```
if (isStaggered()) {
  if (out.Subset() == QUDA_FULL_SITE_SUBSET) {
    blas::ax(-1.0, out.Odd());
  } else if (out.Subset() == QUDA_PARITY_SITE_SUBSET && out.SuggestedParity() == QUDA_ODD_PARITY) {
    // harmless if suggested parity is unset
    blas::ax(-1.0, out.Odd());
  }
}
```

cpviolator: You best put that in, I'm not even sure at which line to place it!

weinbe2:
~@cpviolator I just pushed some changes, they're currently completely untested and thus may be buggy, but I wanted to see if you had any feedback on/qualms with the choices I made.~ Wait for a bit, I have a different implementation thought.

include/dirac_quda.h
```diff
@@ -2238,6 +2288,95 @@ namespace quda {
...
+      }
+      if (reset1) {
+        dirac->tmp1 = NULL;
+        reset1 = false;
+      }
+    }
+
+    int getStencilSteps() const { return dirac->getStencilSteps(); }
+
+    virtual bool hermitian() const { return true; } // gamma5 op is always Hermitian
```
weinbe2: FYI, removing this: it's not necessarily Hermitian, and even if so it's not strictly HPD. This also hits issues in the eigensolver class, where the Hermitian check enforces smallest real and largest real only, which means you can't extract the interior spectrum of a gamma5 operator.

weinbe2: Well, correction, modifying it, because some combinations are still technically HPD.

weinbe2:
I added support for staggered-type operators, and I got a bit creative and added support for asymmetric preconditioned clover operators as well since they're gamma5 Hermitian. Please note that, in the case of the clover op and asqtad op, you need to explicitly pass in `--compute-clover true` and `--compute-fat-long true`, respectively.

Some examples:

Staggered (need to load a gauge field for stability, it seems):
```
./staggered_eigensolve_test --dim 16 16 16 16 --eig-use-poly-acc false --eig-type iram --eig-use-normop false --verbosity verbose --eig-spectrum SM --eig-n-conv 10 --eig-n-kr 96 --eig-compute-gamma5 true --dslash-type staggered --compute-fat-long true --load-gauge l16t16b7p0
```

Asymmetric preconditioned clover:
```
./eigensolve_test --dim 16 16 16 16 --eig-use-poly-acc false --eig-type iram --eig-use-normop false --verbosity verbose --eig-spectrum SM --eig-n-conv 10 --eig-n-kr 96 --eig-compute-gamma5 true --dslash-type clover --compute-clover true --load-gauge --solution-type mat-pc --solve-type direct-pc --matpc even-even-asym
```

Furthermore, while it's not technically gamma5 _Hermitian_, I maintained support for the gamma5 versions of twisted operators because the spectrum becomes `[Hermitian indefinite] + i mu`, up to normalization, which also remains true for the asymmetric preconditioned version of the operator.

For example, non-preconditioned:
```
> ./eigensolve_test --dim 8 8 8 8 --eig-use-poly-acc false --eig-type iram --eig-use-normop false --verbosity verbose --eig-spectrum SM --eig-n-conv 10 --eig-n-kr 96 --eig-compute-gamma5 true --dslash-type twisted-clover --compute-clover true
[...]
0011 converged eigenvalues at iter 35
Resizing kSpace to 192 vectors
Eval[0000] = (+2.2074195998518328e-01,+2.4390235594406475e-02) residual = +6.2537944228732114e-07
Eval[0001] = (-2.2272257651295577e-01,+2.4390243167254695e-02) residual = +5.8926165651571871e-07
Eval[0002] = (+2.2324950071035538e-01,+2.4390242110027889e-02) residual = +5.9616980039758174e-07
Eval[0003] = (-2.2394925376872274e-01,+2.4390241674485266e-02) residual = +5.3720774020934465e-07
```

And odd-odd-asymmetric preconditioned:
```
> ./eigensolve_test --dim 16 16 16 16 --eig-use-poly-acc false --eig-type iram --eig-use-normop false --verbosity verbose --eig-spectrum SM --eig-n-conv 10 --eig-n-kr 96 --eig-compute-gamma5 true --dslash-type twisted-clover --compute-clover true --solution-type mat-pc --solve-type direct-pc --matpc odd-odd-asym
```

This PR does not have support for domain wall operators, since we don't have the 5-d "flip" operator yet (and the Mobius version requires an inversion). It also does not support coarse operators, which are technically "sigma_3"-Hermitian so long as you don't precondition any of the operators.

weinbe2:
Added to the wiki page as well: https://github.com/lattice/quda/wiki/QUDA%27s-eigensolvers#computing-the-gamma5-hermitian-spectrum

cpviolator:
@weinbe2 Nice Wiki addition. Is that all you wanted to push? All looks good me if so.

weinbe2:
@cpviolator yeah, that's all for this time around. I'll get Shamir DWF and the coarse op going in a later pass. I also slipped in a fix for #1097 . @maddyscientist or @hummingtree since both of us have had our hands in this PR, do you mind giving it a quick independent review before merging?

include/dirac_quda.h
```diff
@@ -1878,29 +1891,63 @@ namespace quda {
...
+    bool isWilsonType() const
+    {
+      return (Type() == typeid(DiracWilson).name() || Type() == typeid(DiracWilsonPC).name()
+              || Type() == typeid(DiracClover).name() || Type() == typeid(DiracCloverPC).name()
+              || Type() == typeid(DiracCloverHasenbuschTwist).name()
+              || Type() == typeid(DiracCloverHasenbuschTwistPC).name() || Type() == typeid(DiracTwistedMass).name()
+              || Type() == typeid(DiracTwistedMassPC).name() || Type() == typeid(DiracTwistedClover).name()
+              || Type() == typeid(DiracTwistedCloverPC).name()) ?
+        true :
+        false;
```
hummingtree, line 1918: A non-blocking comment here: in the future if someone adds a new dirac type, but do not update the list here, things could go wrong. A possible solution is to have a pure virtual `isWilsonType()` for `Dirac`, and all derived classes of `Dirac` are forced to have a definition of this method.

hummingtree: The same applies to `isStaggerd` and `isDwf` as well.

hummingtree (approved):
I don't think there is anything controversial here. Left a code style comment.

include/dirac_quda.h
weinbe2: @hummingtree that's a good suggestion---I'll file an issue about it and take care of it later this week.

weinbe2:
Thanks for the review @hummingtree --- I created an issue for implementing your suggestion, I'll get to it soon, but since it's non-blocking let's get this merged in now.

PR #1075, opened 2020-11-10
Hotfix/bug fixes

cpviolator:
The PR fixes some bugs in the test utils spinor construction code, as well as cleanly deleting the `cloverEigensolver` field.

### DIFF
--- a/utils/host_utils.cpp
+++ b/utils/host_utils.cpp
@@ -259,7 +259,7 @@ void constructRandomSpinorSource(void *v, int nSpin, int nColor, QudaPrecision p
   for (int d = 0; d < 4; d++) param.x[d] = x[d];
   if (isPCSolution(sol_type)) param.x[0] /= 2;
   quda::cpuColorSpinorField spinor_in(param);
-
+  quda::spinorNoise(spinor_in, rng, QUDA_NOISE_UNIFORM);
 }
 
 void initComms(int argc, char **argv, std::array<int, 4> &commDims) { initComms(argc, argv, commDims.data()); }
--- a/utils/host_utils.cpp
+++ b/utils/host_utils.cpp
@@ -146,13 +146,6 @@ void setQudaDefaultMgTestParams()
   }
 }
 
-bool isPCSolution(QudaSolutionType solution_type) {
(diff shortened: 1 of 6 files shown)


weinbe2:
Reviewing now. Can you please hit this with a `clang-format` @cpviolator ?

utils/host_utils.cpp
```diff
@@ -221,10 +228,10 @@ void constructWilsonTestSpinorParam(quda::ColorSpinorParam *cs_param, const Quda
     cs_param->nDim = 4;
   }
   for (int d = 0; d < 4; d++) cs_param->x[d] = gauge_param->X[d];
-  bool pc = (inv_param->solution_type == QUDA_MATPC_SOLUTION || inv_param->solution_type == QUDA_MATPCDAG_MATPC_SOLUTION);
+  bool pc = isPCSolution(inv_param->solution_type);
```
weinbe2, line 224: This also needs to be propagated into `utils/staggered_host_utils.cpp`

utils/host_utils.cpp
```diff
@@ -146,6 +146,13 @@ void setQudaDefaultMgTestParams()
   }
 }
 
+bool isPCSolution(QudaSolutionType solution_type) {
```
weinbe2: This definition could be moved into `utils/host_utils.h` and then get used in `utils/staggered_host_utils.cpp`

weinbe2:
Minus my couple of comments on propagating this to `utils/staggered_host_utils.cpp` this looks good. Great catches all around.

We should propagate this into `lib/interface_quda.cpp` in the future, but for now let's get these bugfixes in.

weinbe2 (approved):
Approved pending staggered requests, clang-format, and Jenkins.

utils/host_utils.cpp
cpviolator: 7052e04

utils/host_utils.cpp
cpviolator: 7052e04

cpviolator:
@weinbe2 @maddyscientist Thank you both!

weinbe2:
Thanks for the quick fix @cpviolator , looks good!

PR #1076, opened 2020-11-10
Template cleanup of coarsening routines

weinbe2:
This PR does a bunch of template parameter cleanup between `coarse_op.cuh` and `coarse_op_kernel.cuh`. In short, any instance of `from_coarse`, `Float`, `coarseSpin`, `fineSpin`, `coarseColor`, and `fineColor` are replaced with appropriate `static constexpr [type] [name] = Arg::[name];` and `using Float = typename Arg::Float;`.

This does not (currently) attempt to eliminate template arguments for dimensions and directions from `compute[V]UV`, or the `parity_flip` and `shared_atomic` flags from `computeVUV`. It also leaves in the `max` check for `CloverInvMax`. It also doesn't touch any MMA routines.

This PR is currently marked as a draft PR because I need to do a full sweep over multigrid parameters to check for unintentional mistakes, both with and without Jitify.

I opened this PR to begin a visual design and style review. I intentionally haven't run a `clang-format` yet to avoid "flattening" the history of any one particular step for now.


maddyscientist:
Looks all good at this stage.

weinbe2:
This PR successfully passed a battery of standard MG tests, both jitify and non-jitify, and various combinations of CPU setups. I'm going a last compile after merging in `develop` but otherwise it's good to go.

@maddyscientist , for whatever reason `clang-format` didn't turn up any changes in its `diff` with develop---do you mind also trying? I don't believe I'm "that" good at sticking to the formatting rules.

maddyscientist (approved):

PR #1077, opened 2020-11-11
Move singularity of recon-8 to {{0,1,0},{1,0,0},{0,0,-1}}

### LINKED ISSUE #978: Modify recon-8 convention so the unit gauge isn't the singular point
The current recon-8 has a singularity for unit gauge fields which leads to nan's, which, among other quirks, can lead to false positives during debugging using the free field. Changing the location of the singularity would be a great benefit for debugging.

weinbe2:
This ~(draft)~ PR moves the singular point of recon-8 (and by extension -9) from the identity matrix to {{0,1,0},{1,0,0},{0,0,-1}}. In practice, this avoids issues with unit gauge fields that may come up in free-field experiments or during the first trajectory of (R)HMC from a cold start.

In its current state the PR is functionally correct and passes a performance "laugh test", but still needs a more exhaustive sweep of volumes, precisions, etc.

As part of this test I found that something has gone awry with generating a unit staggered field. Building the free HISQ fat/long links works just fine.

This PR will close #978 once it's fully tested.

Representative sanity check with Wilson:
```
for r in 18 12 8; do echo -n "recon $r , "; ./invert_test --verbosity verbose --prec double --prec-sloppy single --recon 18 --recon-sloppy $r --nsrc 20 --unit-gauge 1 2>&1 | grep "mean solve" | awk ' { print $11,$12,$13,$14; } '; done
```

Representative sanity check with HISQ:
```
for r in 18 13 9; do echo -n "recon $r , "; ./staggered_invert_test --verbosity verbose --dslash-type asqtad --compute-fat-long true --mass 0.3 --test 1 --prec double --prec-sloppy single --recon 18 --recon-sloppy $r --nsrc 20 --unit-gauge 1 2>&1 | grep "mean solve" | awk ' { print $11,$12,$13,$14; } '; done
```

### DIFF
--- a/host_reference/dslash_reference.cpp
+++ b/host_reference/dslash_reference.cpp
@@ -172,10 +172,10 @@ void verifyWilsonTypeInversion(void *spinorOut, void **spinorOutMulti, void *spi
       errorQuda("Mass normalization %s not implemented", get_mass_normalization_str(inv_param.mass_normalization));
     }
 
-    void *spinorTmp = malloc(V * spinor_site_size * host_spinor_data_type_size * inv_param.Ls);
+    void *spinorTmp = malloc(Vh * spinor_site_size * host_spinor_data_type_size * inv_param.Ls);
     printfQuda("Host residuum checks: \n");
     for (int i = 0; i < inv_param.num_offset; i++) {
-      ax(0, spinorCheck, V * spinor_site_size, inv_param.cpu_prec);
+      ax(0, spinorCheck, Vh * spinor_site_size, inv_param.cpu_prec);
 
       if (dslash_type == QUDA_TWISTED_MASS_DSLASH) {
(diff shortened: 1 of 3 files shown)


maddyscientist:
As part of this PR, are there any improvements in Nan / inf detection we should be including?

weinbe2:
Added the heavy-handed `errorQuda`s inside the `converged` functions in `lib/solver.cpp`.

I confirmed these changes go through the NERSC Small benchmark without issue, and it can successfully handle an initial free field with recon 9.

maddyscientist (approved):
Great addition.  I've done some testing, and all looks good (modulo one small fix needed (pushed) since the test host code was cleaned up).

mathiaswagner:
Has someone checked perf?

maddyscientist:
Perf is fine, no change as expected.

PR #1078, opened 2020-11-13
Rework handling of targets in cmake

mathiaswagner:
This switches from object libraries to using target sources which seems to be a better solution for handling dependencies and include paths.

Putting it out as a draft so people can take a look at and comment on the approach.


maddyscientist (approved):
This looks good to me.

cpviolator (approved):
Much neater :)

PR #1079, opened 2020-11-30
Add interface compilation tests

mathiaswagner:
Make sure `quda.h` compiles with a C compiler.


mathiaswagner:
Maybe we should add Fortran as well, but by default we don't even compile that in Jenkins.

maddyscientist:
Closing this PR, since the fix for this will be included in #1081 and this will include this additional test.

Issue #926, opened 2019-11-13
Include test for C compilation of quda.h


mathiaswagner:
Make sure the header is valid C. This is not covered currently. Maybe this can also be done for FORTRAN.

PR #1081, opened 2020-11-30
Expose additional MG params plus C interface compilation test

weinbe2:
This PR:
* Exposes two additional parameters to the MILC HISQ MG interface: (1) coarse link/near-null precision (half or single instead of hard-coded half), and (2) coarsening the full or preconditioned operator on intermediate and coarsest levels (instead of hard-coded preconditioned)
* Adds @mathiaswagner 's C interface compile test
* Fixes an issue in the C interface introduced by having a `bool` in the signature of the BLAS interface.

Closes #1079 and #926

### DIFF
--- a/lib/milc_interface.cpp
+++ b/lib/milc_interface.cpp
@@ -1287,6 +1287,7 @@ struct mgInputStruct {
 
   int mg_levels;
   bool verify_results;
+  QudaPrecision preconditioner_precision; // precision for near-nulls, coarse links
 
   // Setup
   int nvec[QUDA_MAX_MG_LEVEL];                   // ignored on first level, if non-zero on last level we deflate
@@ -1321,6 +1322,7 @@ struct mgInputStruct {
   mgInputStruct() :
     mg_levels(4),
     verify_results(true),
+    preconditioner_precision(QUDA_HALF_PRECISION),
     deflate_n_ev(66),
     deflate_n_kr(128),
     deflate_max_restarts(50),
@@ -1429,6 +1431,17 @@ struct mgInputStruct {
     }
   }
 
+  QudaPrecision getQudaPrecision(const char *name)
+  {
+    if (strcmp(name, "single") == 0) {
+      return QUDA_SINGLE_PRECISION;
+    } else if (strcmp(name, "half") == 0) {
+      return QUDA_HALF_PRECISION;
+    } else {
+      return QUDA_INVALID_PRECISION;
+    }
+  }
+
   // parse out a line
   bool update(std::vector<std::string> &input_line)
   {
@@ -1448,6 +1461,13 @@ struct mgInputStruct {
         error_code = 1;
       } else {
         verify_results = input_line[1][0] == 't' ? true : false;
+      }
+
+    } else if (strcmp(input_line[0].c_str(), "preconditioner_precision") == 0) {
+      if (input_line.size() < 2) {
+        error_code = 1;
+      } else {
+        preconditioner_precision = getQudaPrecision(input_line[1].c_str());
       }
 
     } else if (strcmp(input_line[0].c_str(), "mg_verbosity") == 0) {
@@ -1622,6 +1642,7 @@ struct milcMultigridPack {
   QudaMultigridParam mg_param;
   QudaInvertParam mg_inv_param;
   QudaEigParam mg_eig_param[QUDA_MAX_MG_LEVEL];
+  QudaPrecision preconditioner_precision;
   double last_mass;
   void *mg_preconditioner;
 };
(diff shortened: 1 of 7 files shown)


weinbe2:
@mathiaswagner your C test appears to have angered the gods of `clang` testing:

```
[185/252] Building C object tests/CMakeFiles/c_interface_test.dir/c_interface_test.c.o
FAILED: tests/CMakeFiles/c_interface_test.dir/c_interface_test.c.o 
/usr/bin/cc -DBUILD_MILC_INTERFACE -DBUILD_QDP_INTERFACE -DGPU_CLOVER_DIRAC -DGPU_CLOVER_HASENBUSCH_TWIST -DGPU_DOMAIN_WALL_DIRAC -DGPU_FATLINK -DGPU_GAUGE_TOOLS -DGPU_NDEG_TWISTED_MASS_DIRAC -DGPU_STAGGERED_DIRAC -DGPU_TWISTED_CLOVER_DIRAC -DGPU_TWISTED_MASS_DIRAC -DGPU_UNITARIZE -DGPU_WILSON_DIRAC -DMPI_COMMS -DMULTI_GPU -DQUDA_PRECISION=14 -DQUDA_RECONSTRUCT=7 -Itests -I../tests -I../tests/utils -I../tests/host_reference -I../lib/../include -Iinclude -isystem ../tests/googletest/include -isystem ../tests/googletest -isystem externals/eigen/src/Eigen -isystem /usr/local/cuda/include -isystem /usr/local/openmpi/include -Wall -O3 -Werror -Wno-error=unused-private-field -pthread -MD -MT tests/CMakeFiles/c_interface_test.dir/c_interface_test.c.o -MF tests/CMakeFiles/c_interface_test.dir/c_interface_test.c.o.d -o tests/CMakeFiles/c_interface_test.dir/c_interface_test.c.o -c ../tests/c_interface_test.c
cc1: error: '-Wno-error=unused-private-field': no option '-Wunused-private-field'
```

weinbe2:
My guess is you have to add `-DCMAKE_C_COMPILER=clang` or something like that along with the other flags...

mathiaswagner:
I cleaned up the CMAKE_C_FLAGS to be consistent with CMAKE_CXX_FLAGS and also made sure to pass CMAKE_C_COMPILER in Jenkins when we also use CMAKE_CXX_COMPILER.

weinbe2:
> I cleaned up the CMAKE_C_FLAGS to be consistent with CMAKE_CXX_FLAGS and also made sure to pass CMAKE_C_COMPILER in Jenkins when we also use CMAKE_CXX_COMPILER.

Thanks for the quick turnover @mathiaswagner !

tests/CMakeLists.txt
```diff
@@ -57,8 +58,11 @@ if(QUDA_ARPACK)
 endif()
 
 # define tests
+add_executable(c_interface_test c_interface_test.c)
+  target_link_libraries(c_interface_test ${TEST_LIBS})
+  quda_checkbuildtest(c_interface_test QUDA_BUILD_ALL_TESTS)
 
-# if we build with QDP JIT the tests cannot run anyway
+  # if we build with QDP JIT the tests cannot run anyway
```
maddyscientist: Space inserted before `#`

lib/milc_interface.cpp
```diff
@@ -1972,7 +2018,10 @@ void milcSetMultigridParam(milcMultigridPack *mg_pack, QudaPrecision host_precis
 
   inv_param.verbosity = verbosity;
 
-  inv_param.verbosity_precondition = verbosity;
+  inv_param.verbosity = input_struct.mg_verbosity[0] ? QUDA_VERBOSE : QUDA_SUMMARIZE;
```
maddyscientist: This appears that there's no way to enable debug verbosity.  This might be good to add.

maddyscientist:
Apart from my minor comments, this PR looks good to go.

maddyscientist (approved):

weinbe2:
Thanks for the comments @maddyscientist --- I can merge in `develop` after merging in the `computeCoarseClover` PR and address them in one go.

tests/CMakeLists.txt
weinbe2: Is there a particular reason that's a problem in the context of `CMakeLists.txt`, since it's a comment (and thus should arguably match indentation), not a `pragma` in C++?

lib/milc_interface.cpp
weinbe2: Good call---I added the full battery of silent/summary/verbose/debug, default summary, so we have it!

tests/CMakeLists.txt
weinbe2: We indent comments elsewhere in the file, so I'm going to opt to leave it as-is.

maddyscientist: My point was this is an unnecessary change made in this PR.  We should be striving to minimizing unnecessary diffs.

weinbe2: Ah, fair. It looks like it got picked up because it was neighboring an addition Mathias made. I'll revert it.

weinbe2: Actually, it looks like the issue is more that the two lines above got indented for no good reason.

weinbe2: Fixed

weinbe2:
This will be good to go once the build tests complete!

PR #1082, opened 2020-12-01
Optimized coarse coarse clover workflow

weinbe2:
This PR reduces the compile time of a full Wilson+staggered MG build to about an hour.

...More substantively, this PR leverages `computeUV` and `computeVUV` to coarsen the coarse clover, as opposed to using the existing `computeCoarseClover` routine. The key observation is that `computeUV` as it stood applied the coarse links `Y` to the near-nulls `V` in the +/- `dir` direction. This can be simply reformulated to apply the coarse clover `X` to the near-nulls `V` at the same site instead. `computeVUV` is then broadly unchanged, since it already completes the inner-product in a site-local sense, not requiring any gather from a neighboring site.

With this modification, we can remove the code paths in `computeCoarseClover` relevant for coarsening the coarse operator. The routine still exists for coarsening the fine clover, which isn't a real compile time or performance limiter. 

These changes can be trivially (in a textual sense) ported to the MMA routines as well. That is intentionally being deferred to a later PR---I'll implement it, then (politely) ask @hummingtree to do performance verifications and tweaks as is necessary.

This PR is functionally complete, I'm just doing a full parameter sweep across fermion types, precisions, etc, to make sure everything's ship-shape. Once that's done I'll take it out of draft PR status.

### DIFF
--- a/include/kernels/coarse_op_kernel.cuh
+++ b/include/kernels/coarse_op_kernel.cuh
@@ -149,6 +149,7 @@ namespace quda {
   /**
      Calculates the matrix UV^{s,c'}_mu(x) = \sum_c U^{c}_mu(x) * V^{s,c}_mu(x+mu)
      Where: mu = dir, s = fine spin, c' = coarse color, c = fine color
+     or, if dir == QUDA_IN_PLACE, UV^{s,c'}(x) = \sum_c C^{c}_mu(x) * V^{s,c}_mu(x+mu)
   */
   template<int dim, QudaDirection dir, typename Wtype, typename Arg>
   __device__ __host__ inline void computeUV(Arg &arg, const Wtype &Wacc, int parity, int x_cb, int i0, int j0)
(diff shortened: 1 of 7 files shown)


include/kernels/coarse_op_kernel.cuh
```diff
@@ -786,22 +801,22 @@ namespace quda {
 #pragma unroll
       for (int j = 0; j < TileType::N; j++) {
 
-        if (!isDiagonal) {
+        if (dir == QUDA_IN_PLACE || isDiagonal) {
```
maddyscientist, line 804: Any reason not just to set `isDiagonal = true` always for `QUDA_IN_PLACE`, perhaps simplifying a few of the changes here?

weinbe2: Logically no, but `dir == QUDA_IN_PLACE` is compile-time knowledge while `isDiagonal` is runtime, so it's meant to give the compiler more info.

maddyscientist:
This all looks great.  Nice to get the compilation time improved.  Subject to @weinbe2 confirming all tests are passing as expected this is an easy approval.

maddyscientist (approved):

include/kernels/coarse_op_kernel.cuh
maddyscientist: Yeh, I figured you'd say that 😉   Although `isDiagonal` isn't `constexpr`, it will be evaluated at compile time for the in-place case since it's unconditionally always true for in-place.

weinbe2: There are some poor logic reasons why I didn't have `dir == QUDA_IN_PLACE` directly imply `isDiagonal` (related to the `-arg.kappa` rescale, see around line 975)---let me know if you'd like me to rework it. It's ultimately a trade-off in where you want your headaches :) 

If you want me to rework this, I'll plug it into my last updates to the MG params PR, if that's alright.

weinbe2:
All tests are ship-shape.

include/kernels/coarse_op_kernel.cuh
maddyscientist: No need to do any more updates in this PR with respect to `isDiagonal`, and any such work can be done later.

PR #1085, opened 2020-12-03
Staggered cleanup and bugfixes

weinbe2:
* Fixed a tunekey goof from the coarse coarse clover PR---technically didn't lead to a tunekey degeneracy hazard, but better to fix it than not.
* Fixed issues with `Nsrc != 1` in `staggered_dslash_[c]test` by enforcing `Nsrc = 1`; can be reverted once multi-rhs is here.
* Bugfix: `staggered_invert_test` was giving garbage for the host residual if there were comms in any direction (incl those from `--partition != 0`). The issue was roughly that the CPU fat link halo depth wasn't being properly set; the fix was a one-liner setting `gauge_param.type`.
* Tweaked `staggered_invert_test` to allocate all `Nsrc` sources up front to simplify @hummingtree 's work
* Added support for performing `Nsrc > 1` multishift solves...
* ~...which uncovered some static variable linking issues within the multishift interface. Downstream effect: values that were being modified in place within the `QudaInvertParam` weren't being restored properly. Threw in a quick fix that removed the static variable.~
* Just kidding: there was a massive bug in staggered multishift solves that by sheer luck has never been an issue. The bug appears if you run more than one `invertMultiShiftQuda` with staggered-type fermions _without_ explicitly resetting the shifts each time (iirc MILC does reset the shifts each time, which is why this bug never manifested there). For Wilson-type fermions, shifts were "cached" to a static array, rescaled in place to match `kappa` conventions, and then restored after the solves. For staggered-type fermions no rescalings were necessary so no caching was done, but shifts were still "restored" from the uninitialized static array, giving garbage. This has now been fixed.

@hummingtree this PR should address all the issues you uncovered, some of which had been kicking around for a while! Let me know if you hit any other woes, or if there's anything I can clarify to help you with your split grid work.

### DIFF
--- a/tests/staggered_invert_test.cpp
+++ b/tests/staggered_invert_test.cpp
@@ -270,13 +270,15 @@ int main(int argc, char **argv)
 
   // Staggered vector construct START
   //-----------------------------------------------------------------------------------
-  quda::ColorSpinorField *in;
+  std::vector<quda::ColorSpinorField *> in;
   quda::ColorSpinorField *out;
   quda::ColorSpinorField *ref;
   quda::ColorSpinorField *tmp;
   quda::ColorSpinorParam cs_param;
   constructStaggeredTestSpinorParam(&cs_param, &inv_param, &gauge_param);
-  in = quda::ColorSpinorField::Create(cs_param);
+  for (int k = 0; k < Nsrc; k++) {
+    in.emplace_back(quda::ColorSpinorField::Create(cs_param));
+  }
   out = quda::ColorSpinorField::Create(cs_param);
   ref = quda::ColorSpinorField::Create(cs_param);
   tmp = quda::ColorSpinorField::Create(cs_param);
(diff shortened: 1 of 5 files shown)


tests/staggered_invert_test.cpp
```diff
@@ -269,13 +272,13 @@ int main(int argc, char **argv)
 
   // Staggered vector construct START
   //-----------------------------------------------------------------------------------
-  quda::ColorSpinorField *in;
+  std::vector<quda::ColorSpinorField *> in;
```
hummingtree, line 275: We will also need the `out` fields to be present simultaneously, but I will fix it in the split grid PR.

tests/staggered_invert_test.cpp
```diff
@@ -243,6 +243,9 @@ int main(int argc, char **argv)
   }
 
   // Create ghost gauge fields in case of multi GPU builds.
+  gauge_param.type = (dslash_type == QUDA_STAGGERED_DSLASH || dslash_type == QUDA_LAPLACE_DSLASH) ?
+    QUDA_SU3_LINKS :
+    QUDA_ASQTAD_FAT_LINKS;
```
hummingtree, line 248: Just curious, (and sorry for the duplicate comment/question), Was this breaked at some point, or it's broken from day one?

hummingtree (commented):
I don't think there is anything controversial here - will run the tests to confirm and it should be good to go.

tests/staggered_invert_test.cpp
maddyscientist: From Evan's comments to me last night, I believe this has been broken for years.

maddyscientist (approved):
Looks good to me.  Thanks for the fixes @weinbe2.

We can merge this once @hummingtree signs off on this with his tests.

hummingtree (approved):
`staggered_invert_test` is now behaving correctly. Thanks @weinbe2 for the quick fix.

Issue #1087, opened 2020-12-05
Replace explicit '0' stream with device::get_default_stream() in cudaEventRecord() 


bjoo:
Something I came accross during merging. In places like blas_test.cpp (lines 368, 532) also in dslash_test.cpp and dslash_ctest.cpp  `cudaEventSynchronize` makes reference still to stream=0, in 
places like
```
cudaEventRecord(start,0);
```
under the new scheme these should probably be:
```
cudaEventRecord(start, device::get_default_stream());
```

bjoo:
Also in `multigrid_benchmark_test.cpp:`

bjoo:
also in `covdev_test.cpp`, `staggered_dslash_ctest.cpp` and `staggered_dslash_test.cpp`

bjoo:
Fixed in PR #1088

bjoo:
PR #1088 has been merged so I will close this..

PR #1088, opened 2020-12-05
Feature/generic kernel hotfixes

bjoo:
Fixes issue #1087  and also a compiler error I encountered with lib/targets/cuda/device.cpp where the code 

``` 
 switch (deviceProp.major) {
      case 2:
	return 8;
      case 3:
	return 16;
      case 5:
      case 6: return 32;
      case 7:
        switch (deviceProp.minor) {
        case 0: return 32;
        case 2: return 32;
        case 5: return 16;
        }
      default:
        warningQuda("Unknown SM architecture %d.%d - assuming limit of 32 blocks per SM\n",
                    deviceProp.major, deviceProp.minor);
        return 32;
      }
```
caused the compiler to complain that fallthrough is possible from the internal switch statement. This happens only on CUDA version < 11 (i.e. the 10.2 I was working with)

### DIFF
--- a/lib/targets/cuda/device.cpp
+++ b/lib/targets/cuda/device.cpp
@@ -283,16 +283,26 @@ namespace quda
       switch (deviceProp.major) {
       case 2:
 	return 8;
+	break;
       case 3:
 	return 16;
+	break;
       case 5:
-      case 6: return 32;
+	 return 32;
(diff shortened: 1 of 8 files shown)


maddyscientist (approved):
Thank you @bjoo

PR #1093, opened 2020-12-28
Hotfix/milc qio support

weinbe2:
This hotfix branch adds support for @jcosborn 's new QIO branch `extended`: https://github.com/usqcd-software/qio/tree/extended . This new branch supports IO for volumes larger than 2^31 sites via an extension to the API, as well as callback routines (not used from QUDA). It does not break the original API. This hotfix branch is backwards compatible with old builds of QIO using the `develop` or `qio2-5-0` branch as determined by compile time (that is, no `cmake` flag needs to be set).

The reason why this has been elevated to the status of `hotfix` is because the MILC `develop` branch now requires the `extended` QIO branch as enforced here: https://github.com/milc-qcd/milc_qcd/blob/develop/generic/io_scidac.c#L18

The `extended` QIO branch provides a `#define QIO_HAS_EXTENDED_LAYOUT` to indicate largefile support. I've taken advantage of this in the PR to provide compatibility for both the `develop`/`qio2-5-0` branch of QIO as well as the new `extended` branch. That said, because the `extended` branch preserves support for the legacy API, I have made it the default when `QUDA_DOWNLOAD_USQCD` is set to `yes` during `cmake` configuration. (Another benefit of this is there's no need to update any QUDA+MILC build instructions.) This default can be overridden by setting the advanced `cmake` flag `QUDA_DOWNLOAD_QIO_LEGACY` to `YES`.

To reiterate a point further above, `QUDA_DOWNLOAD_QIO_LEGACY` _only_ needs to be set if you want QUDA to download the `qio2-5-0` branch when `QUDA_DOWNLOAD_USQCD` is set to true. You do not need to set this `cmake` flag when linking QUDA against a separate build of QMP or QIO as it will automatically query the define `QIO_HAS_EXTENDED_LAYOUT`.

Completed tests:
* @detar has separately tested the new API in QIO's `extended` branch on Summit with large volumes and has found no issues.
* Experiments myself with smaller volumes, confirming compatibility between lattices loaded/saved using the QIO `qio2-5-0` and `extended` branches.
* Running MILC RHMC on one GPUs, for due diligence
* Confirming QUDA properly links to and supports an external build of the `develop` branch of QIO. This is meant to confirm these changes don't break other workflows that build QMP and QIO separately, such as Chroma, etc

I have not yet properly tested support for loading and saving large files directly through QUDA (loading/saving gauge fields, eigenvectors, etc). I don't consider this a blocker for this hotfix because QUDA didn't support this before anyway, but I'll file an issue for this and get to testing it sooner as opposed to later.


weinbe2:
@mathiaswagner @cpviolator @cpviolator @stevengottlieb @detar @bjoo @chulwoo1 for awareness

maddyscientist (approved):

maddyscientist:
Great to get large-volume IO working 😜

PR #1096, opened 2021-01-08
Have 'make install' copy 'usqcd' directory if QUDA_DOWNLOAD_USQCD is enabled.

### LINKED ISSUE #1094: Have `make install` copy `usqcd` directory if QUDA_DOWNLOAD_USQCD is enabled
Self-explanatory via title

cpviolator:
Closes #1094

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -594,6 +594,12 @@ install(
 
 install(DIRECTORY ${CMAKE_SOURCE_DIR}/include/ DESTINATION include)
 
+# If the USQCD stack was automatically downloaded, this will copy the usqcd library
+# to the install path
+if(QUDA_DOWNLOAD_USQCD) 
+  install(DIRECTORY ${CMAKE_BINARY_DIR}/usqcd DESTINATION ${CMAKE_INSTALL_PREFIX})
+endif()
+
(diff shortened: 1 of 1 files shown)


weinbe2 (approved):
This looks great, I confirmed it works visually and via testing using an `install` instead of `build` directory when linking MILC to QUDA's QIO and QMP.

Once Jenkins goes through this is good to merge. Thank you for the quick addition @cpviolator

mathiaswagner:
Ugh, has anyone checked that just plain copying using `install` in cmake is equivalent to running the `make install` step of QIO/QMP to that location?
@weinbe2 @cpviolator?

Also remind me why this is needed. I don't necessary want to turn QUDA into a package manager for USQCD ;-)

weinbe2:
Can't speak to your first question @mathiaswagner beyond it didn't come to mind.

As to your second question: this came up in the context of the wiki; we say that when building MILC with QUDA it suffices to point `$PATH_TO_QUDA` to either the build or install directory, the latter of which broke when we added `usqcd` support. It seemed like a good best-practices thing to fix (especially after we got a few questions about it in one week).

Don't worry, this isn't going to turn into a complete USQCD package manager if I have any say in it 🙂

PR #1102, opened 2021-01-14
Hotfix/misc

maddyscientist:
Bugfix PR
* Fix random.cu: large global volumes would cause overflow
* Fix `cloverQuda` and `computeStaggeredForceQuda` declarations in quda.h to match definitions in interface_quda.cpp (back port from feature/generic_kernels)
* Improve error reporting in quda_api.cpp (back port from feature/generic_kernels)

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -1198,7 +1198,7 @@ extern "C" {
    * @param inverse Whether to apply the inverse of the clover term
    */
   void cloverQuda(void *h_out, void *h_in, QudaInvertParam *inv_param,
-      QudaParity *parity, int inverse);
+                  QudaParity parity, int inverse);
 
   /**
    * Apply the full Dslash matrix, possibly even/odd preconditioned.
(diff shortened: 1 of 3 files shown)


weinbe2:
All looks good to me, thanks @maddyscientist 

One question, though, why weren't the incorrect declarations an issue in the first place? I'd assume, if not the compiler, the linker would've complained somewhere.

weinbe2 (approved):
Approved pending Jenkins

maddyscientist:
> One question, though, why weren't the incorrect declarations an issue in the first place? I'd assume, if not the compiler, the linker would've complained somewhere.

It confused me too for a minute 😄  They weren't an issue because to date, they have not been called directly by outside applications.  The only users of these functions have been through the Fortran interface, which is defined in the same compilation unit (interface_quda.cpp), and so they just used the file-local definitions and ignored the declaration in quda.h.  I only found this bug when moving the Fortran interface definitions into their own file in the generic_kernel branch.

weinbe2:
> > One question, though, why weren't the incorrect declarations an issue in the first place? I'd assume, if not the compiler, the linker would've complained somewhere.
> 
> It confused me too for a minute smile They weren't an issue because to date, they have not been called directly by outside applications. The only users of these functions have been through the Fortran interface, which is defined in the same compilation unit (interface_quda.cpp), and so they just used the file-local definitions and ignored the declaration in quda.h. I only found this bug when moving the Fortran interface definitions into their own file in the generic_kernel branch.

That's insufferably confusing... but at least it makes sense, I guess. Thanks for the explainer!

PR #1020, opened 2020-06-19
Allow QUDA to have multiple communicators and Split Grid

hummingtree:
This PR put the previously global communication-related variables (say, `Topology` and `MPI_COMM_HANDLE`) into classes that are called `Communicators`, which allows QUDA to have multiple communicators, therefore enabling performing communication based on different grid partitions.

Current design is to have an `std::map<std::array<int,4>, Communicator> communicator_stack` that contains the different communicators that correspond to different partitions of the whole grid, e.g. the un-partitioned grid comes with a split key of `1,1,1,1`. One of communicators can be pushed to the top, which becomes the *current* communicator, and all communication and reduction are performed according to this current communicator. This makes the way to call the communication routines unchanged.

To change communicator, one uses the interface `push_communicator(const CommKey &split_key)`.

~This would be the first step towards supporting split-grid in QUDA.~

Now this PR also adds the **split grid** feature/functionality.

Split grid is one of the answers to the multi-rhs problem for domain wall/M\"obius fermion, for which the multi-rhs performance is typically limited by communication bandwidth. It re-distribute the right-hand-sides into sub-partitions of the node/GPU grid, within which the communication bandwidth is usually larger. Each sub-partition of the grid performs inversion for one or several of the rhs. The re-distribution also reduces the surface over volume ratio since the local volume on each node/GPU is smaller, thus reduces the total amount of communication needed.

This PR also attempts to reduce the code duplication in `dslash_(c)test` and `staggered_dslash_(c)test`.

Split grid typically fits into scenarios part of which needs consumes large chunk of memory, e.g. deflation with large number of eigenvectors. Large memory usage typically requires strong scaling, which typically renders the problem bottle-necked by communication bandwidth.

The split grid interface looks like
```
  void invertMultiSrcQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param, void *h_gauge, QudaGaugeParam *gauge_param);
```
(There are other new interfaces for clover-style and staggered-style fermions.)
and a new option is added to `QudaInvertParam`,
```
    /**< The grid of sub-partition according to which the processor grid will be partitioned
    The product of the four integers should be equal to `num_src`. */
    int split_grid[QUDA_MAX_DIM];
```
as well as the command line option `--grid-partition` for `invert_test`.

The routines that are used to split the fields (single partition to multiple partitions) and join the field (multiple partitions to single partition) are in `include/split_grid.h`:
```
  template <class Field>
[...]

PR #1107, opened 2021-02-09
Fixes for QMP communicators

mathiaswagner:
Mostly fixes issues observed when running Chroma after #1020.

### DIFF
--- a/include/communicator_quda.h
+++ b/include/communicator_quda.h
@@ -680,6 +680,9 @@ struct Communicator {
 
 #if defined(QMP_COMMS)
   QMP_comm_t QMP_COMM_HANDLE;
+  // A bool indicating if the QMP handle here is the default one, which we should not free at the end,
+  // or a one that QUDA creates through `QMP_comm_split`, which we should free at the end.
+  bool is_qmp_handle_default;
 #endif
 
   int rank = -1;
--- a/include/communicator_quda.h
+++ b/include/communicator_quda.h
@@ -680,8 +680,11 @@ struct Communicator {
 
 #if defined(QMP_COMMS)
   QMP_comm_t QMP_COMM_HANDLE;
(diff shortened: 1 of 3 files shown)


weinbe2:
Changes are straightforward and look good.

Some things I'm confused about @hummingtree, and should've asked about in your initial PR:

1. Why does QMP's `comm_rank_global` -> `QMP_get_node_number` work to always pick the _global_ rank zero, as opposed to `comm_rank` -> `QMP_comm_get_node_number`? I'd assume `QMP_get_node_number` would grab the current communicator as opposed to the default. Have you looked under the hood to make sure it'll always behave as expected? (And/or should I?)

2. I think there's a potential risk in the MPI backend: if `user_set_comm_handle` in the constructor is set to `true`, the "default" MPI world won't be `MPI_COMM_WORLD`, which runs afoul of the MPI backend's `comm_rank_global`baking in `MPI_COMM_WORLD`. A simple solution could be caching the default MPI world on the MPI backend `Communicator` constructor, similar to how we're saving some default state with the QMP backend. (This may also be helpful for the QMP backend re:my concern above.)

weinbe2:
On a different note, checking against NERSC small looks good---all the deviations from the sample (CPU) output look familiar.

include/communicator_quda.h
```diff
@@ -680,6 +680,9 @@ struct Communicator {
 
 #if defined(QMP_COMMS)
   QMP_comm_t QMP_COMM_HANDLE;
+  // A bool indicating if the QMP handle here is the default one, which we should not free at the end,
```
maddyscientist: Comment here should be in doxygen markup

maddyscientist (approved):
Aside from my trivial comment on comment, this looks good.  I've done some testing with split-grid solves and also tested that `QUDA_RANK_VERBOSITY` works as expected here, and it does.  Signing off on this.

PR #1109, opened 2021-02-12
Add support for optionally running some benchmarks from ctest

mathiaswagner:
This add some basic dslash_test benchmarks to ctest so we can easily do some perf sweeps. 
This is controlled by the `QUDA_CTEST_DISABLE_BENCHMARKS` variable and by default benchmarks are disabled.

To selectively only run the benchmarks something like
```
ctest  -R "benchmark*" -V
```
can be used.

Output is store in a `.json` file.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -593,6 +593,7 @@ foreach(pol IN LISTS DSLASH_POLICIES)
             --dslash-type ${DIRAC_NAME}
             --test 0
             --dim 20 20 20 20
+            --Lsdim 12
             --gtest_output=json:dslash_${DIRAC_NAME}_benchmark_pol${pol2}.json
             --gtest_filter=*benchmark/*n0)
     set_tests_properties(benchmark_dslash_${DIRAC_NAME}-policy${pol2} PROPERTIES DISABLED ${QUDA_CTEST_DISABLE_BENCHMARKS})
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -537,6 +632,19 @@ foreach(pol IN LISTS DSLASH_POLICIES)
                      --dim 2 4 6 8
                      --Lsdim 4
                      --gtest_output=xml:dslash_mobius_test_pol${pol2}.xml)
+    
+    add_test(NAME benchmark_dslash_${DIRAC_NAME}-policy${pol2}
+    COMMAND ${QUDA_CTEST_LAUNCH} $<TARGET_FILE:dslash_ctest> ${MPIEXEC_POSTFLAGS}
+            --dslash-type ${DIRAC_NAME}
+            --test 0
+            --dim 20 20 20 20
```
weinbe2, line 642: Add a specific `--Lsdim` --- maybe 16? @hummingtree

weinbe2: (applicable to other dwf types too)

weinbe2:
Remind me, will these do both a tuning run and a true perf benchmarking run?

mathiaswagner:
It will do tuning. The Gflops extracted are from the output in dslash_test as there is not yet a way to ge that directly reported from the tunecache. Thus it is mostly intended for catching notable perf differences.

tests/CMakeLists.txt
mathiaswagner: I used 12 here for now.

hummingtree: `12` should be good here.

mathiaswagner:
Just to clarify: This is not really intended to do some thorough benchmarking, but more to be able to have a quick way to check for regressions.

maddyscientist:
I had a quick whirl of this, and it all looks good to me.  

An aside: at some point in the future, we may want to look at [google benchmark](https://github.com/google/benchmark) for our benchmarking needs.

maddyscientist (approved):

PR #1110, opened 2021-02-16
Fix performance of heterogeneous atomic reductions

maddyscientist:
Fix the performance of heterogeneous atomic reductions:
 * use a busy-wait loop instead of `std::atomic::wait`, since the latter uses nanosleep which adds latency bubbles

### DIFF
--- a/include/reduce_helper.h
+++ b/include/reduce_helper.h
@@ -130,7 +130,7 @@ namespace quda
       if (consumed) errorQuda("Cannot call complete more than once for each construction");
 
       for (int i = 0; i < n_reduce * n_item; i++) {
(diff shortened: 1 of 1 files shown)


mathiaswagner (approved):
Looks good to me.

weinbe2:
Looks good! I'll get a full `blas_test` sweep going momentarily just to make sure there are no other surprises.

weinbe2 (approved):
I ran a regression test comparing Wilson, 32^4, volume between CUDA 10.2 (where there aren't heterogeneous atomics) and 11.0 and saw no regressions. This is good to go.

PR #1111, opened 2021-02-23
Minor CG clean-up, marginally improved HQ residual stability, QIO tweaks

weinbe2:
While this PR was originally intended to be a stability hotfix, it has turned into a misc clean-up/small improvements PR.

This PR:
* Moves some `numerical_limits` epsilon heuristics from the CG solver into base `Solver` class routines
* Adds a stability "catch" for heavy quark residual solves where the solve hard restarts if `|r2|/|b2|` drops "unreasonably" low, in response to observations that `|r2|/|b2|` can go far below the numerically reasonable ~1e-16 after `L2breakdown` is set to `true`
* ...disables the above catch for pure `double` solves, since if pure `double` breaks down there's no hope. (Note: this _doesn't_ disable the solve automatically restarting if it hits the requested accumulated residual but the "true" residual isn't sufficiently small.)
* Slips in a change to the `configure` flags to QIO based on feedback from @jcosborn

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -649,8 +649,8 @@ if(QUDA_QIO)
       GIT_SHALLOW YES
       PREFIX usqcd
       CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} <SOURCE_DIR>/configure "INSTALL=${INSTALL_EXE} -C"
-                        "CFLAGS=-Wall -O3 -std=c99 -DHAVE_BGL " --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
-                        --enable-largefile --disable-qmp-route --enable-dml-output-buffering
+                        "CFLAGS=-Wall -O3 -std=c99 " --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
+                        --enable-largefile --disable-qmp-route --enable-dml-output-buffering --enable-dml-bufsize=33554432
       BUILD_COMMAND make
       INSTALL_COMMAND make install
       DEPENDS QMP
--- a/include/invert_quda.h
+++ b/include/invert_quda.h
@@ -564,19 +564,11 @@ namespace quda {
     void PrintSummary(const char *name, int k, double r2, double b2, double r2_tol, double hq_tol);
 
     /**
-       @brief Helper function which returns the epsilon tolerance for a given precision
+       @brief Returns the epsilon tolerance for a given precision, by default returns
+       the solver precision.
+       @param[in] prec Input precision, default value is solver precision
     */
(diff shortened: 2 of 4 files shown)


mathiaswagner (approved):

include/invert_quda.h
```diff
@@ -563,6 +563,21 @@ namespace quda {
...
+    /**
+       @brief Returns the epsilon tolerance of the solver precision
+    */
+    const double solverPrecisionEpsilon() const;
+
+    /**
+       @brief Returns the epsilon tolerance of the solver sloppy precision
+    */
+    const double solverPrecisionSloppyEpsilon() const;
+
```
maddyscientist, line 572: Some comments here:
* The `solver` prefixes are redundant since we are in the `Solver` class here, so I think these should be removed.
* These functions return by value, so the `const` is redundant.  In fact the additional `const` on the return type will lead to warnings with some compilers when additional warnings are enabled.
* Instead of having separate `solverPrecisionEpsilonHelper` and `solverPrecisionEpsilon` functions, how about fusing these into a single function `double precisionEpsilon(QudaPrecision prec = QUDA_INVALID_PRECISION)`, and if the default then return the epsilon tolerance of the solver precision else return the epsilon of the argument.

CMakeLists.txt
```diff
@@ -649,8 +649,8 @@ if(QUDA_QIO)
       GIT_SHALLOW YES
       PREFIX usqcd
       CONFIGURE_COMMAND CC=${MPI_C_COMPILER} CXX=${MPI_CXX_COMPILER} <SOURCE_DIR>/configure "INSTALL=${INSTALL_EXE} -C"
-                        "CFLAGS=-Wall -O3 -std=c99 -DHAVE_BGL " --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
-                        --enable-largefile --disable-qmp-route --enable-dml-output-buffering
+                        "CFLAGS=-Wall -O3 -std=c99 " --with-qmp=${QUDA_QMPHOME} --prefix=<INSTALL_DIR>
+                        --enable-largefile --disable-qmp-route --enable-dml-output-buffering --enable-dml-bufsize=33554432
```
maddyscientist, line 653: Just curious what the motivation for this is?

weinbe2: Purely based on an e-mail from @jcosborn about the `configure` flags he uses (motivated by some unrelated QIO perf headaches I was seeing on Summit), which I took as a good enough reason to make the change---unclear if this actually addresses that issue.

include/invert_quda.h
weinbe2: https://github.com/lattice/quda/pull/1111/commits/c64a21fc736d8d6e413267d63d02601cb5d52002 and https://github.com/lattice/quda/pull/1111/commits/408e985252d476e40932774b3a0bd33d44e4fb66

maddyscientist (approved):

PR #1112, opened 2021-02-23
Remove variable-size arrays from dslash test code

maddyscientist:
This PR expunges the global variable `my_spinor_site_size` variable from the test code, and replaces the associated variable-size arrays, which are undefined behaviour in C++, with static arrays using `spinor_site_size` and `stag_spinor_site_size` for Wilson-type and staggered-type test codes, respectively.

Note this fixes the test codes when compiled using nvc++ as the host compiler.

### DIFF
--- a/tests/blas_interface_test.cpp
+++ b/tests/blas_interface_test.cpp
@@ -366,7 +366,6 @@ int main(int argc, char **argv)
   initQuda(device_ordinal);
   int X[4] = {xdim, ydim, zdim, tdim};
   setDims(X);
-  setSpinorSiteSize(24);
   //-----------------------------------------------------------------------------
 
   int result = 0;
--- a/tests/blas_test.cpp
+++ b/tests/blas_test.cpp
@@ -1021,7 +1021,6 @@ int main(int argc, char** argv)
     Ncolor = 3;
   }
 
-  setSpinorSiteSize(24);
   initComms(argc, argv, gridsize_from_cmdline);
   display_test_info();
   initQuda(device_ordinal);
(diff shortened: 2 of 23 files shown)


mathiaswagner (approved):

weinbe2:
Based on some `grep`ing for the number `24`, there are still a lot of magic 24s (and 12s, 6s, etc...) in the depths of `tests/host_reference/wilson_dslash_reference.cpp` and `tests/host_reference/domain_wall_dslash_reference.cpp`. Do we want to address some of those here, or file that under future work (for ex. the generic Nc efforts)?

Edit: I ask only because this PR already does address a few magic 24s---more specifically `4*3*2`s---just not all of them.

weinbe2:
I'm running this through `ctest` now (amusingly to verify the CPU code is consistent with the GPU code, as opposed to the other way around). Assuming that goes through I'll approve+merge this.

cpviolator:
In the Nc agnostic branch I have I rooted out every instance of magic 3, as well as a 4 for spin in both the CPU code and the GPU code, so in the timeline for getting that branch merged this issue will be dealt with too.

weinbe2:
> In the Nc agnostic branch I have I rooted out every instance of magic 3, as well as a 4 for spin in both the CPU code and the GPU code, so in the timeline for getting that branch merged this issue will be dealt with too.

That's great to know, thanks @cpviolator . In that case we can ignore my question about cleaning up more 24s in this PR.

maddyscientist:
> I'm running this through `ctest` now (amusingly to verify the CPU code is consistent with the GPU code, as opposed to the other way around). Assuming that goes through I'll approve+merge this.

I did a full build with everything enabled and did a full ctest run, with it all looking good.  Another `ctest` run is always good though.

Regarding 24 and 6 remaining: yes I was mindful of the fact that @cpviolator had the agnostic branch in the works, so I didn't try to catch all of those.

weinbe2 (approved):
`ctest` passed, merging

PR #1130, opened 2021-03-30
Fusing stencil kernels for Mobius fermion with immediately followed fifth dimension operators 

### LINKED ISSUE #193: Optimize mobius dslash kernels and strong scaling
The present Mobius Dslash kernels are not optimal in either raw performance or with respect to strong scaling.  The current implementation of the asymmetrically preconditioned Mobius matrix-vector application looks like this:

```
    Dslash4pre(*tmp1, in, QUDA_ODD_PARITY);
    Dslash4(out, *tmp1, QUDA_EVEN_PARITY);
    Dslash5inv(*tmp1, out, QUDA_ODD_PARITY, kappa5);
    Dslash4pre(out, *tmp1, QUDA_EVEN_PARITY);
    Dslash4(*tmp1, out, QUDA_ODD_PARITY);
    Dslash5Xpay(out, in, QUDA_EVEN_PARITY, *tmp1, 1.0);
```

where Dslash4 is the 4-d dslash, Dslash4pre and Dslash5 are variations on the dslash on the fifth dimension only and Dslash5inv is the analytically inverted fifth dimension dslash.  In general, the fifth dimension is kept local to a GPU, and communication only occurs over the four space-time dimensions.  There exists multiple opportunities for optimization here:

**Multi-GPU communication**: this only occurs in Dslash4, with all other kernels being local to a GPU.  However, when we try to hide the communication for dslash4, we only attempt to overlap it with the computation on the interior of dslash4.  We could just as well overlap this communication with the computation of Dslash4pre, Dslash5inv and Dslash5 since all these kernels use only local computation.  This optimization would drastically increase the amount of local-computation with which we can overlap with giving significantly better strong scaling.  This refactoring is much easier if we are able to do kernel fusion, since there it essentially comes for free.  If not, we may have to introduce a new dslash policy, since the present multi-GPU strategies rely on a single kernel for the interior region with which to overlap with.

**Optimization of Dslash4pre and Dslash5**: these kernels currently sustain just a bit north of 120 Gflops on a K40, and are the slowest kernels being completely memory bandwidth bound.  This is easy to see with some back-of-the-envelope modeling: Dslash4pre does 168 flops per site, and requires the loading of 3 spinors and saving of 1.  This gives an arithmetic intensity of 0.44, and given a sustained memory bandwidth of 220 GB/s we get 100 Gflops.  We see a bit more performance because the texture cache gives us some caching across Ls.  However, this caching is highly limited because of texture cache size:  The smallest extent of the space-time dimension is 32 sites (minimum thread block x-size of a warp), so for Ls=16, we would need 48 KiB to get perfect caching (in single precision).  On Kepler each individual the texture cache pool is 12 KiB, of which there are four per SM (on Maxwell it is 2x 24 KiB per SM): neither give enough to give perfect caching.  If we had perfect caching over Ls, the arithmetic intensity increases to 0.9 (one spinor load and spinor save), and the performance would increase to 200 Gflops.  

There are a couple of options to get better caching across Ls here:
1. Decrease the thread block x dimension to 8 or 16.  This will reduce the width of the block, allowing for a greater height to be cached.  This is not necessarily a win since the x dimension will have less threads than a warp, but it should not affect memory coalescing since 128-bit memory transaction occur at quarter-warp granularity.
2. Utilize shared memory to cache across the Ls dimension.  This is motivated by the fact that the shared memory pool is 48 KiB, thus it could store space-time width of 32 and the full Ls=16 height.  This could be used in congestion with decreased x thread block size as well.

**Kernel fusion**: there exists a lot of scope for kernel fusion in these kernels since Dslash4 is the only kernel that has space-time dependence.  For a given space-time volume, if we are able to keep the entire extent of the Ls dimension local to a thread block, then we could in principle merge Dslash4pre / Dslash4 / Dslash5inv and Dslash4pre / Dslash4 / Dslash5Xpay into a single kernel.  We would use the shared memory to easily facilitate this fusion.  With this optimization in place, we would be relying on L2 cache for doing all space-time locality caching.

For how best to approach these optimizations we should discuss as to what the expected sizes of Ls are for Mobius dwf in the near future.  For example if Ls <= 16 then we could utilize extremely aggressive optimizations.  For larger Ls, some of the optimizations will become much trickier.

If we are able to implement all of these optimizations, especially kernel fusion, we obtain an estimated doubling in performance of the overall matrix-vector operation, which would likely see a single K40 achieve 1 Tflops in single precision, as well as having superior strong scaling.  Due to the high amount of locality in the Mobius operator, it may turn out to be the ideal operator for GPUs.

hummingtree:
This PR is based on #1056, and should be merged after #1056.

This PR adds the option to run the even-odd preconditioned `MdagM`, `M`, and `Mdag` operators for Mobius fermion with fused kernels. These fused kernels fuse the Wilson stencil kernel `Dslash4` with immediately followed fifth dimension operators, such as `m5inv`, `m5pre`, `m5mobius`, or any combinations of these.

To run the fused kernels, an option `use_mobius_fused_kernel` (default to `true`) is added to `InvertParam`. A corresponding command line option `--mobius-fused-kernel <true/false>` (default to `true`) is also added to the various tests.

More tests are ongoing, and single GPU performance of the Mobius `MdagM` operators has increased by 20-30% as a result of using the fused kernels.

Closes #193.

A few remaining pieces:
- [x] A scaling plot with and without the fused kernel.
- [ ] See how this works with `NVSHMEM`.
- [x] And remember to restore the max Ls.

A few scaling captured on selene.

single-32
single-24
half-32
half-24

Weak scaling on Summit captured by @cpviolator.

| Precision-L-Nnodes | non-fused-GFLOPS | fused-GFLOPS | speed up |
| :-: |:-:| :-:| :-: |
|   half-24- 1 |   4742 |   5569 | 1.17 |
|   half-24- 2 |   3478 |   3687 | 1.06 |
|   half-24- 3 |   3541 |   3646 | 1.03 |
|   half-24- 4 |   3609 |   3772 | 1.05 |
|   half-24- 8 |   2387 |   2566 | 1.07 |
|   half-24-16 |   2278 |   2371 | 1.04 |
|   half-24-32 |   2243 |   2482 | 1.11 |
|   half-32- 1 |   4710 |   5522 | 1.17 |
|   half-32- 2 |   3569 |   3784 | 1.06 |
|   half-32- 3 |   3600 |   3775 | 1.05 |
|   half-32- 4 |   3274 |   3560 | 1.09 |
|   half-32- 8 |   2289 |   2546 | 1.11 |
|   half-32-16 |   2260 |   2490 | 1.10 |
|   half-32-32 |   2254 |   2532 | 1.12 |
| single-24- 1 |   2838 |   4058 | 1.43 |
| single-24- 2 |   2064 |   2155 | 1.04 |
| single-24- 3 |   2110 |   2201 | 1.04 |
| single-24- 4 |   2073 |   2181 | 1.05 |
| single-24- 8 |   1380 |   1517 | 1.10 |
| single-24-16 |   1281 |   1369 | 1.07 |
| single-24-32 |   1281 |   1447 | 1.13 |
| single-32- 1 |   2854 |   4014 | 1.41 |
| single-32- 2 |   2117 |   2187 | 1.03 |
| single-32- 3 |   2163 |   2184 | 1.01 |
| single-32- 4 |   1933 |   2106 | 1.09 |
| single-32- 8 |   1355 |   1465 | 1.08 |
| single-32-16 |   1327 |   1451 | 1.09 |
| single-32-32 |   1333 |   1453 | 1.09 |
| double-24- 1 |   1299 |   1781 | 1.37 |
| double-24- 2 |    927 |   1077 | 1.16 |
| double-24- 3 |   1070 |   1097 | 1.03 |
| double-24- 4 |   1054 |   1084 | 1.03 |
| double-24- 8 |    678 |    757 | 1.12 |
| double-24-16 |    644 |    688 | 1.07 |
| double-24-32 |    647 |    718 | 1.11 |
| double-32- 1 |   1307 |   1799 | 1.38 |
| double-32- 2 |   1036 |   1106 | 1.07 |
| double-32- 3 |   1066 |   1101 | 1.03 |
| double-32- 4 |    940 |   1037 | 1.10 |
| double-32- 8 |    690 |    745 | 1.08 |
| double-32-16 |    653 |    718 | 1.10 |
| double-32-32 |    685 |    754 | 1.10 |


maddyscientist:
Great work @hummingtree.  I guess the first thing I would love to see here is some strong scaling curve comparing the fused and unfused paths.

FWIW: @walkloud and @cpviolator, this should give a nice improvement to Mobius CG solver, and improve the scaling.

cpviolator:
@maddyscientist @hummingtree @walkloud I can dig up my old testing scripts on SUMMIT and test this out. My correlation function branch (GK based, so an easy merge) has fermion and gauge smearing enabled so I can prepare a production quality solve with a CG of around 1400 iterations.

maddyscientist:
Sounds like a great idea @cpviolator, thanks

include/quda_constants.h
```diff
@@ -46,7 +46,7 @@
  * @def   QUDA_MAX_DWF_LS
  * @brief Maximum length of the Ls dimension for domain-wall fermions
  */
-#define QUDA_MAX_DWF_LS 32
+#define QUDA_MAX_DWF_LS 16
```
maddyscientist: This needs to be reverted to 32

hummingtree:
Superseded by #1163.

PR #1136, opened 2021-04-05
Compute gauge path

sbacchio:
This PR extends the gauge force kernel to compute the product of gauge links along a given path.
It is useful for computing Gauge actions.

**NOTE:** commit 08110b2 fixes a bug in the current implementation: if `path_coeff` is not aligned as `double`, CUDA would raise a `misaligned address` error. For this reason I added a padding.

For its usage in lyncs  I'm happy with the current implementation (https://github.com/Lyncs-API/lyncs.quda/blob/bae5bf21769f2c1bc3724137bba8ce8d796c2e32/lyncs_quda/gauge_field.py#L317).

But I suppose you would like more features like tests and function in interface_quda?

### DIFF
--- a/include/kernels/gauge_force.cuh
+++ b/include/kernels/gauge_force.cuh
@@ -51,10 +51,11 @@ namespace quda {
     }
   };
 
-  template <typename Float_, int nColor_, QudaReconstructType recon_u, QudaReconstructType recon_m>
+  template <typename Float_, int nColor_, QudaReconstructType recon_u, QudaReconstructType recon_m, bool force_>
   struct GaugeForceArg {
     using Float = Float_;
     static constexpr int nColor = nColor_;
+    static constexpr bool force = force_;
     static_assert(nColor == 3, "Only nColor=3 enabled at this time");
     typedef typename gauge_mapper<Float,recon_u>::type Gauge;
     typedef typename gauge_mapper<Float,recon_m>::type Mom;
@@ -153,8 +154,13 @@ namespace quda {
 
     // update mom(x)
     Link mom = arg.mom(dir, idx, parity);
-    mom = mom - arg.epsilon * linkA;
-    makeAntiHerm(mom);
+    if(arg.force) {
+      mom = mom - arg.epsilon * linkA;
+      makeAntiHerm(mom);
+    }
+    else {
+      mom = mom + arg.epsilon * linkA;
+    }
     arg.mom(dir, idx, parity) = mom;
   }
 
--- a/include/kernels/gauge_force.cuh
+++ b/include/kernels/gauge_force.cuh
@@ -14,40 +14,49 @@ namespace quda {
     int *input_path[4];
     const int *length;
     const double *path_coeff;
+    int *buffer;
     int count;
 
-    paths(void *buffer, size_t bytes, int pad, int ***input_path, int *length_h, double *path_coeff_h, int num_paths, int max_length) :
+    paths(int ***input_path, int *length_h, double *path_coeff_h, int num_paths, int max_length) :
       num_paths(num_paths),
       max_length(max_length),
       count(0)
     {
-      void *path_h = safe_malloc(bytes);
+      // create path struct in a single allocation
+      size_t bytes = 4 * num_paths * max_length * sizeof(int) + num_paths * sizeof(int);
+      int pad = ((sizeof(double) - bytes % sizeof(double)) % sizeof(double))/sizeof(int);
+      bytes += pad*sizeof(int) + num_paths*sizeof(double);
+
+      buffer = static_cast<int*>(pool_device_malloc(bytes));
+      int *path_h = static_cast<int*>(safe_malloc(bytes));
       memset(path_h, 0, bytes);
 
-      int *input_path_h = (int*)path_h;
       for (int dir=0; dir<4; dir++) {
         // flatten the input_path array for copying to the device
         for (int i=0; i < num_paths; i++) {
           for (int j=0; j < length_h[i]; j++) {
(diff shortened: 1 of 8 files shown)


maddyscientist:
Thanks for the contribution @sbacchio!
* I haven't seen the issue with misaligned address in the develop branch.  Is there an easy way to reproduce this issue?
* For the unit test, I imagine it would be easy to modify the gauge_force_test to also test this functionality?

sbacchio:
Hi @maddyscientist!
I got into this issue when testing the path for the plaquette. In that case I used
`path = [ [1, 7, 6], [2, 7, 5], [3, 7, 4]]`
so it has `max_length=3` and `num_path=3`.
This gives a starting point for `path_coeff` of `39*sizeof(int)` which is not multiple of `sizeof(double)`.

Ok I will work on a test tomorrow and push it. I would do a crosscheck against the plaquette.

sbacchio:
@maddyscientist I have extended the test and added a function in the interface.
Please let me know if anything more is needed!

maddyscientist:
This all looks great.  I ran out of time before testing this today, but will do so tomorrow.

include/kernels/gauge_force.cuh
```diff
@@ -39,22 +39,23 @@ namespace quda {
...
+      memcpy((char*)path_h + 4 * num_paths * max_length * sizeof(int) + num_paths*sizeof(int) + pad, path_coeff_h, num_paths*sizeof(double));
 
       qudaMemcpy(buffer, path_h, bytes, qudaMemcpyHostToDevice);
       host_free(path_h);
 
       // finally set the pointers to the correct offsets in the buffer
       for (int d=0; d < 4; d++) this->input_path[d] = (int*)((char*)buffer + d*num_paths*max_length*sizeof(int));
       length = (int*)((char*)buffer + 4*num_paths*max_length*sizeof(int));
-      path_coeff = (double*)((char*)buffer + 4 * num_paths * max_length * sizeof(int) + num_paths*sizeof(int));
+      path_coeff = (double*)((char*)buffer + 4 * num_paths * max_length * sizeof(int) + num_paths*sizeof(int) + pad);
```
mathiaswagner: while touching that code can we please change the casts to C++ ones?

include/kernels/gauge_force.cuh
```diff
@@ -16,7 +16,7 @@ namespace quda {
     const double *path_coeff;
     int count;
 
-    paths(void *buffer, size_t bytes, int ***input_path, int *length_h, double *path_coeff_h, int num_paths, int max_length) :
+    paths(void *buffer, size_t bytes, int pad, int ***input_path, int *length_h, double *path_coeff_h, int num_paths, int max_length) :
```
mathiaswagner: does pad needs to be exposed here or can we do all calculation inside of the `paths` function?

tests/gauge_force_test.cpp
```diff
@@ -236,7 +236,7 @@ int path_dir_t[][5] = {
 static double force_check;
 static double deviation;
 
-void gauge_force_test(void)
+void gauge_force_test(bool force)
```
mathiaswagner: `force` is not self-explanatory. Do we _force_ something or do you calculate _the force_ ? Can we rename that and also add a comment / doxygen ?

mathiaswagner (changes requested):
Looks good to me just some small comments. I did not do a proper check.

include/kernels/gauge_force.cuh
sbacchio: Since I had to get it for `bytes`, then I preferred to not repeat calculations.
On the other hand there is also no reason to have `buffer` and `bytes` outside.
I would suggest then to move everything in `paths` and make the other functions lighter.

tests/gauge_force_test.cpp
sbacchio: Ok I renamed to `compute_force`

include/kernels/gauge_force.cuh
sbacchio: done

sbacchio:
@mathiaswagner thank you for reviewing. I think I took into account your comments

include/kernels/gauge_force.cuh
```diff
@@ -14,47 +14,57 @@ namespace quda {
...
 
       // finally set the pointers to the correct offsets in the buffer
-      for (int d=0; d < 4; d++) this->input_path[d] = (int*)((char*)buffer + d*num_paths*max_length*sizeof(int));
-      length = (int*)((char*)buffer + 4*num_paths*max_length*sizeof(int));
-      path_coeff = (double*)((char*)buffer + 4 * num_paths * max_length * sizeof(int) + num_paths*sizeof(int));
+      for (int d=0; d < 4; d++) this->input_path[d] = buffer + d*num_paths*max_length;
+      length = buffer + 4*num_paths*max_length;
+      path_coeff = reinterpret_cast<double*>(buffer + 4 * num_paths * max_length + num_paths + pad);
+    }
+    void free() {
```
sbacchio, line 58: I know I could use a destructor, but the class is copied at some point. I also tried using `shared_ptr` but without success...

kostrzewa:
Is the test for this correct? In particular, I'm surprised by `max_length=6`:

https://github.com/lattice/quda/blob/f48a9ffa3c4273734b430bb6ca4c637264f58507/tests/gauge_force_test.cpp#L241-L244

Does `max_length` then refer to the path including the current link or just the length of the staples?

maddyscientist:
Looks like this got auto closed since GK is merged finally.  You should be able to retarget the PR at develop and reopen this.

PR #1137, opened 2021-04-06
fix shmem argument not being passed for Nspin=1

mathiaswagner:
also remove default value in constructor to prevent hits from happening again


maddyscientist (approved):
Could this bug have affected NERSC?

mathiaswagner:
> Could this bug have affected NERSC?

maybe. Depends on what the autotuner likes.

PR #1138, opened 2021-04-07
MILC HISQ MG struct initialization hotfix

weinbe2:
I was a bit sloppy when I wrote the MILC HISQ MG interface and didn't initialize all values in the custom `struct`, which ended up biting me in gcc 9. This PR fixes this, adding some dummy initialization values to all of the array fields. A point of future work is (still) cleaning up this messy interface.

### DIFF
--- a/lib/milc_interface.cpp
+++ b/lib/milc_interface.cpp
@@ -1317,7 +1317,8 @@ struct mgInputStruct {
   double deflate_a_min; // ignored if no polynomial acceleration
   int deflate_poly_deg; // ignored if no polynomial acceleration
 
-  void setArrayDefaults() {
+  void setArrayDefaults()
+  {
     // set dummy values so all elements are initialized
     // some of these values get immediately overriden in the
     // constructor, in some cases with identical values:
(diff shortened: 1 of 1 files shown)


lib/milc_interface.cpp
```diff
@@ -1317,6 +1317,33 @@ struct mgInputStruct {
   double deflate_a_min; // ignored if no polynomial acceleration
   int deflate_poly_deg; // ignored if no polynomial acceleration
 
+  void setArrayDefaults()
+  {
+    // set dummy values so all elements are initialized
+    // some of these values get immediately overriden in the
+    // constructor, in some cases with identical values:
+    // this is to separate "initialzing" with "best practices"
```
maddyscientist: typo with initialzing

maddyscientist (approved):

lib/milc_interface.cpp
weinbe2: thanks, fixed!

mathiaswagner (approved):

PR #1140, opened 2021-04-08
potential size reuction for libquda

mathiaswagner:
always enable binary compression for fatbin


maddyscientist:
Do you have any before / after numbers to demonstrate the effect of this?

mathiaswagner:
I saw 2.7G vs 2.6G, so nothing huge. It might however sometimes prevent us from tripping over library size issues. It is a bit tricky to get a feeling for that as you need to do builds with most things enabled in QUDA and compare with and without that. And it requires a full recompile ...

maddyscientist:
Any feel if there's any impact on module load time?

mathiaswagner:
No, not really.

maddyscientist (approved):

PR #1146, opened 2021-05-01
Cleanup/gf test interface

cpviolator:
This PR adds CL params for the gauge fixing routines so that one can control the parameters. The original gauge alg test is now a `ctest` with the Google testing suite, and the file `gauge_alg_test.cpp` performs only a single gauge fix test specified by the user, with the option to save the gauge field.


maddyscientist (commented):
This is a solid PR, but given we're working on merging GK into develop soon, would you mind rebasing this against GK?  Apologies for the churn.

cpviolator:
Will do, no problem at all.

cpviolator:
Rebased this against GK, builds and runs fine. I moved the `gauge_alg_test.cpp` to `gauge_alg_ctest.cpp` as it is incorporated into the google testing suite.

maddyscientist:
This PR needs develop merged in, and then we can review this.

cpviolator:
I see some regression in the FFT gauge fixing. It is present in the current develop branch too for both tuned and untuned runs. The reported data in the FFT gauge fixing run does not make sense in that the action is incorrect and there is no convergence. If allowed to complete to the max iterations, the plaquette value is the same as it was prior to gauge fixing, so the routine is not leaving the gauge orbit. I've pasted a command and the full output is attached. If I can't find a solution I'll open an issue for the develop branch.

```
./gauge_alg_test --dim 8 8 8 8 --gf-tol 1e-10 --gf-maxiter 1000
```

[github_gf_output.txt](https://github.com/lattice/quda/files/7324604/github_gf_output.txt)

cpviolator:
The issue is that the default value passed for `--gf-fft-autotune` should be `true` else the default `alpha` value for the lattice may be poor. Fixed with https://github.com/lattice/quda/commit/bb9c0de72f285c002fb140f9a2bae100d80f3cd6

tests/CMakeLists.txt
```diff
@@ -812,8 +812,8 @@ foreach(prec IN LISTS TEST_PRECS)
 
   if(QUDA_GAUGE_ALG)
     add_test(NAME gauge_alg_${prec}
-             COMMAND ${QUDA_CTEST_LAUNCH} $<TARGET_FILE:gauge_alg_test> ${MPIEXEC_POSTFLAGS}
-                     --dim 4 6 8 10 --prec ${prec}
+             COMMAND ${QUDA_CTEST_LAUNCH} $<TARGET_FILE:gauge_alg ctest> ${MPIEXEC_POSTFLAGS}
+                     --dim 2 4 6 8 --prec ${prec}
```
maddyscientist: This size needs to be reverted I think: I intentionally made the ctest problem size 4x6x8x10.  If I recall correctly, one cannot run multi-GPU with a local dimension of size 2: the issue is that there is no interior, since everything is halo, and you get failures as a result.

maddyscientist: Also, this looks like a type: `gauge_alg ctest` and not `gauge_alg_ctest`

tests/gauge_alg_ctest.cpp
```diff
@@ -293,7 +328,7 @@ class GaugeAlgTest : public ::testing::Test
     setWilsonGaugeParam(gauge_param);
 
     void *cpu_gauge[4];
-    for (int dir = 0; dir < 4; dir++) { cpu_gauge[dir] = safe_malloc(V * gauge_site_size * gauge_param.cpu_prec); }
+    for (int dir = 0; dir < 4; dir++) { cpu_gauge[dir] = malloc(V * gauge_site_size * gauge_param.cpu_prec); }
```
maddyscientist: This should be changed back to `safe_malloc`: we should be tracking allocations in the tests

tests/gauge_alg_ctest.cpp
```diff
@@ -69,37 +60,75 @@ void display_test_info()
...
+}
+
+class GaugeAlgTest : public ::testing::Test {
+
 protected:
   QudaGaugeParam param;
-
-  Timer<false> a0, a1;
+  
+  Timer<false> a0,a1;
```
maddyscientist: Perhaps we should switch this use of `Timer` to `host_timer_t`.  See dslash_test for example.

lib/gauge_fix_fft.cu
```diff
@@ -185,14 +185,26 @@ namespace quda {
...
                       double alpha0, int autotune, double tolerance, int stopWtheta)
   {
+    // We hardcode the value of autotune to 1 at this point as it is the last
+    // point before computation begins. This ensures that the the user cannot
+    // override alpha autotuning. This is done because it is very easy for the
+    // FFT gauge fixing to fail with a poorly chosen value of alpha, but
+    // autotuning ensures optimal behaviour.
+    // Users who wish to change this behaviour may remove the follwing line
+    // of code and recompile to regain control of alpha autotuning.
+    autotune = 1;
```
maddyscientist: I think this hard coding of the autotune parameter makes me uneasy.

What we could do is replace the interface for `computeGaugeFixingFFTQuda` and `computeGaugeFixingOVRQuda`:  instead of passing the arguments all separately to the interface function, we should make a interface param, e.g., `QudaGaugeFixParam` and place all the gauge fixing parameters in such a struct.  This would allow us to have default values if we create partner helper functions `initQudaGaugeFixParam` (as well as `printQudaGaugeFixParam` and `checkQudaGaugeFixParam`) in check_param.h.

What do you think @cpviolator ?

maddyscientist (changes requested):
Thanks for this PR @cpviolator! I've given it a visual review, and left comments inline.  This would be a nice update to the gauge fixing.

cpviolator:
Thanks Kate, I'll fix all of these and clang tidy. I agree RE the gauge fixing param, it's a neat way to address bloating function signatures and control defaults.

tests/CMakeLists.txt
cpviolator: https://github.com/lattice/quda/commit/040be93004b847b2092280c1b79016904943a511

lib/gauge_fix_fft.cu
cpviolator: What is the best approach with regards to the interface function signatures for gauge fixing? If we go ahead and change them, we may as well consolidate the two functions into a single call to `computeQudaGaugeFixing`, add a `QUDA_GAUGEFIX_TYPE` and direct the workflow within it. We could preserve the two existing function signatures too as wrappers to this consolidated function for back compatibility, or force downstream modifications.

cpviolator: As far as I can tell, the MILC interface would be largely unaffected as we simply change the inner plumbings of
```C++
void qudaGaugeFixingOVR(...)
```
and
```C++
void qudaGaugeFixingFFT(...)
```

maddyscientist: Having a single gauge fixing interface function sounds like a great idea.  Agree regarding MILC itself not needing to be changed, since it goes through milc_interface.cpp, so we can keep these changes internal to QUDA.

Since this breaking change will align with QUDA 2.0, e.g., a major version number change, I think there's no need to maintain the prior legacy interface as a fallback.  So I think we can just remove the prior interface.

cpviolator: https://github.com/lattice/quda/commit/aed7763076cef871a73157f6c358e2a053ec6c83

tests/gauge_alg_ctest.cpp
cpviolator: I used `device_timer_t` as there is no host cross check code.

tests/gauge_alg_ctest.cpp
cpviolator: Done

cpviolator:
All comments addressed. I use `QudaBoolean` for the bools in the `fix.cu` files to remain consistent with the interface, but I can convert them to `bool` prior to the device function call is that is not the desired method.

cpviolator:
BTW, this will need checking with the MILC interface for completeness. Google testing:
```
mpirun -np 2 ./gauge_alg_ctest --dim 8 8 8 8 --test 0 --gridsize 1 1 1 2
```
and spot testing
```
mpirun -np 2 ./gauge_alg_ctest --dim 8 8 8 8 --save-gauge test --test [1/2] --gridsize 1 1 1 2
```

PR #1148, opened 2021-05-06
Hotfix/dslash ndimcomms

maddyscientist:
This fixes a few issues that were recently introduced
* UB fix for overrun of `arg.commDim` in dslash.h (loop limit should be 4 not `in.Ndim()`)
* Fix for ndeg-tm fermions with cmake benchmarks
* Revert initialization in complex default constructor (inhibits composability of complex numbers in `__shared__` objects)

Update: I've also now improved the robustness of the dslash and blas unit tests to check for Nan propagation, to prevent erroneous false positive passes.

### DIFF
--- a/include/complex_quda.h
+++ b/include/complex_quda.h
@@ -459,7 +459,7 @@ template<>
 {
 public:
   typedef float value_type;
-  __host__ __device__ inline complex<float>() : float2() {};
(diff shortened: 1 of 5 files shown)


mathiaswagner (approved):
Looks good to me.

weinbe2 (approved):

PR #1160, opened 2021-06-06
Add csw parameter to invert and clover param

cpviolator:
The PR addresses the recent push to CHROMA master: https://github.com/JeffersonLab/chroma/commit/8af1b2d48836724ca2601dbb0c8428c1dc190737 . In that branch there is reference to a `QudaInvertParam` member `clover_csw` which is used only in branches of QUDA that are in development. In order to ensure that a stable branch of QUDA exists that will compile with CHROMA master, the parameter has been added to to this branch.

Up to now, users are expected to pass `C_sw * kappa` to QUDA via the invert parameter `clover_coeff`. If the user passes any non-zero value of `clover_coeff` to QUDA (regardless of the value `clover_csw`) the `clover_coeff` value will be used and the `clover_csw` value ignored. If the user passes `clover_coeff = 0.0` and a non-zero value of `clover_csw` then QUDA will infer the value of `clover_coeff = clover_csw * kappa`. If both `clover_coeff = 0` and `clover_csw = 0` then an error is thrown.

This ensures back compatibility with CHROMA and other libraries that use  `clover_coeff = clover_csw * kappa` while also adding the functionality to set the clover coefficient via the `clover_csw` value, and not breaking builds with the current master branch of CHROMA.


maddyscientist:
Thanks for the PR @cpviolator.  I'm glad we caught this before we release 1.1.  I presume you have tested this PR against both Chroma and the various internal tests we have?

cpviolator:
@maddyscientist Tested against chroma master with Nc=4. Internals for Nc=3 and Nc=4 are both good as far as I can see, passing the invert test and eigensolve test with `--dslash-type clover`.

I recall there being a `rho` parameter in the `cloverParam` struct that I'm not familiar with, and I don't see it used anywhere. Is it still needed?

maddyscientist:
Yes, rho is still needed: it is used by BQCD to add a fictitious mass term for Hasenbusch preconditioning.

maddyscientist:
I’ll test this against BQCD to make sure nothing is awry there.  @kostrzewa would you mind testing this PR against tmLQCD?  Thx

kostrzewa:
@maddyscientist will do within today

kostrzewa:
~This seems to brea inversions from within tmLQCD:~

```
# QUDA: ERROR: Parameter clover_coeff undefined (rank 0, host cassiopeia, check_params.h:302 in checkCloverParam())
# QUDA:        last kernel called was (name=N4quda9CopyGaugeINS_12CopyGaugeArgIsdLi18ENS_5gauge11FloatNOrderIsLi18ELi2ELi18EL20QudaStaggeredPhase_s0ELb1EL19QudaGhostExchange_sn2147483648ELb0EEENS3_IdLi18ELi2ELi18ELS4_0ELb1ELS5_n2147483648ELb0EEEEEEE,volume=24x24x24x24,aux=GPU-offline,vol=331776,stride=172800,precision=8,geometry=4,Nc=3vol=331776,stride=172800,precision=2,geometry=4,Nc=3,ghost)
```

~it seems to me that `clover_coeff` is not given a sensible default value, which is surprising since the `INIT_PARAM` branch which calls `newQudaCloverParam` seems to set it to zero and we use `newQudaInvertParam()` in tmLQCD to create the struct with sensible defaults to be modified further down the line.~ (see https://github.com/lattice/quda/pull/1160#issuecomment-870471589 below)

It's also a bit problematic (although certainly more consistent) that the meaning of `csw` in `CloverField` has changed, apparently. While this used to be `kappa*csw`, it is now just `csw`, correct?

kostrzewa:
Further, if the meaning of `csw` has changed, what about `rho`?

kostrzewa:
Actually, we explicitly set `clover_coeff = csw * kappa` when we initialise an operator with a clover term in our QUDA interface and we never touch `clover_csw` (clearly, since it is new). I'm a bit surprised at the ERROR that I observe..

tests/utils/set_params.cpp
```diff
@@ -328,6 +334,14 @@ void setMultigridParam(QudaMultigridParam &mg_param)
   inv_param.gamma_basis = QUDA_DEGRAND_ROSSI_GAMMA_BASIS;
   inv_param.dirac_order = QUDA_DIRAC_ORDER;
 
+  if (kappa == -1.0) {
```
kostrzewa, line 347: is this here by accident?

kostrzewa: ah, no I see that kappa needs to be set

kostrzewa:
> This seems to brea inversions from within tmLQCD:
> 
> ```
> # QUDA: ERROR: Parameter clover_coeff undefined (rank 0, host cassiopeia, check_params.h:302 in checkCloverParam())
> # QUDA:        last kernel called was (name=N4quda9CopyGaugeINS_12CopyGaugeArgIsdLi18ENS_5gauge11FloatNOrderIsLi18ELi2ELi18EL20QudaStaggeredPhase_s0ELb1EL19QudaGhostExchange_sn2147483648ELb0EEENS3_IdLi18ELi2ELi18ELS4_0ELb1ELS5_n2147483648ELb0EEEEEEE,volume=24x24x24x24,aux=GPU-offline,vol=331776,stride=172800,precision=8,geometry=4,Nc=3vol=331776,stride=172800,precision=2,geometry=4,Nc=3,ghost)
> ```
> 
> it seems to me that `clover_coeff` is not given a sensible default value, which is surprising since the `INIT_PARAM` branch which calls `newQudaCloverParam` seems to set it to zero and we use `newQudaInvertParam()` in tmLQCD to create the struct with sensible defaults to be modified further down the line.
> 
> It's also a bit problematic (although certainly more consistent) that the meaning of `csw` in `CloverField` has changed, apparently. While this used to be `kappa*csw`, it is now just `csw`, correct?

Mea culpa! I had a stray `LD_LIBRARY_PATH` set to a different version of `libquda.so` ...

As far as I can tell, this works fine.

The only question which I would have is to the change of meaning of `clover_param.csw`. We don't use it directly in tmLQCD (yet?), but @sbacchio and co. might rely on it explicitly in PLEGMA and lyncs.

cpviolator:
@kostrzewa I wrote this update so that users can either:
1. Pass `kappa` and `csw` separately, and QUDA will infer the necessary clover coefficient.
2. Pass an explicit value of the coefficient (e.g. CHROMA's use case) and that will override the above inference.

Hopefully, this will preserve existing workflows and be more natural for new users.

@maddyscientist ~I see in the PR that I actually removed the instantiation of `rho` in the clover param, I'll reinstate it.~ It was never removed, just moved to a new line.

maddyscientist:
Yep, thanks for the `rho` catch.  This would have broken BQCD.

@kostrzewa given Dean's explanation, are you happy with this?

@sbacchio anything to add here?

Getting this `csw` / `clover_coeff` behaviour working properly with clearly defined behaviour is highly desirable given we about to make a new release.

lib/interface_quda.cpp
```diff
@@ -899,7 +899,14 @@ void loadCloverQuda(void *h_clover, void *h_clovinv, QudaInvertParam *inv_param)
 
   CloverFieldParam clover_param;
   clover_param.nDim = 4;
-  clover_param.csw = inv_param->clover_coeff;
```
kostrzewa, line 902: @cpviolator In response to https://github.com/lattice/quda/pull/1160#issuecomment-870927989 I get that external codes can now specify either `csw` and `kappa` or `clover_coeff` directly (as we do in tmLQCD). I'm wondering, however, about the change of meaning of `clover_param.csw` here. Clearly, this is an improvement as `csw` should always just be `csw` for consistency.

I'm just not sure if anyone possibly relies on `clover_param.csw` being `csw*kappa`. If that were the case, the current change (which I fully agree with) would break that. I would argue, however, that it's okay to break this since it wasn't a very consistent situation to begin with.

cpviolator:
@kostrzewa Ah, I see what you mean. I agree that if a user is constructing a `cloverParam` struct for themselves then they are going to run into trouble. Thankfully, not a difficult problem to fix. @maddyscientist perhaps this is worth a sentence or two in the NEWS, just as a CYA?

maddyscientist:
Adding an explicit note in the `NEWS` file sounds like a good plan.  Since we have a separate PR for updating the documentation for the 1.1 release, I'll update the `NEWS` there.

sbacchio:
> Yep, thanks for the `rho` catch. This would have broken BQCD.
> 
> @kostrzewa given Dean's explanation, are you happy with this?
> 
> @sbacchio anything to add here?
> 
> Getting this `csw` / `clover_coeff` behaviour working properly with clearly defined behaviour is highly desirable given we about to make a new release.

Hi @maddyscientist!
Today I tried out the release 1.1, merging all the other open PRs.
Everything works fine also from our side!

**Just a small issue**... some functions defined in `comm_quda.h` are not implemented anymore.
E.g. `comm_default_topology` and `comm_coords_from_rank` were missing for us, but probably also others are...

`comm_coords_from_rank` is then implemented in `communicator_quda.h` as inline function
and  `comm_default_topology` is easy to implement exposing `get_current_communicator`...
so not a big issue to fix, but would be nice to have them (or remove them) for completeness.

mathiaswagner:
Ping here. @kostrzewa it seems you are fine with a readme change?

kostrzewa:
My apologies, I missed the `@`'s, yes!

maddyscientist (approved):

maddyscientist:
I've verified this against BQCD: there was a missing addition to the Fortran interface, but I've fixed that.  As part of that, I've also made the struct_size checking more robust, to better catch these breakages in the Fortran interfaces.

Great work @cpviolator.  I've approved the PR, but as you noted in slack, you were going to a do final check with Chroma. Once you've done that, we can merge this PR.

mathiaswagner:
I verified this with latest Chroma devel.

mathiaswagner (approved):

PR #1161, opened 2021-06-09
remove fatbin.ld

mathiaswagner:
This seems to cause more troubles than it saves space. And for QUDA 2.x we will use better options to control the library.

gold seems to completely fail with it. And It doesn't seem to make a significant impact. I'd rather play it safe for releases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ find_package(PythonInterp)
 if(DEFINED ENV{QUDA_BUILD_TYPE})
   set(DEFBUILD $ENV{QUDA_BUILD_TYPE})
 else()
-  set(DEFBUILD "DEVEL")
+  set(DEFBUILD "RELEASE")
 endif()
 
 set(VALID_BUILD_TYPES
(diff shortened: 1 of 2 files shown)


mathiaswagner:
Also wondering if we should change the default build type to RELEASE for `release/` branches.

maddyscientist:
Switching to RELEASE by default sounds good to me for release branches.  Since it likely doesn’t buy anything, we could at the same time remove the `-march=native` since it all it mainly result in is headaches for users who have a different head node from compute node.

maddyscientist (approved):

weinbe2:
> Switching to RELEASE by default sounds good to me for release branches. Since it likely doesn’t buy anything, we could at the same time remove the `-march=native` since it all it mainly result in is headaches for users who have a different head node from compute node.

I second all parts of this, especially getting rid `-march=native`. I support plugging that into this PR (or slipping it into one of @maddyscientist 's PRs on the horizon).

weinbe2 (approved):

mathiaswagner:
thoughts on `mtune` instead of `march` ?

weinbe2:
> thoughts on `mtune` instead of `march` ?

Ah, I always lose track of the differences between the two---my answer is "whatever choice will let me build a container on a machine with AVX512 and run on a machine without it".

mathiaswagner:
https://stackoverflow.com/questions/10559275/gcc-how-is-march-different-from-mtune

maddyscientist:
I'm happy to go with `mtune`.  If we ever care about CPU vectorization performance, this is something we can refine in the future as needed.

hummingtree (approved):

PR #1166, opened 2021-07-10
use nvhsmem 2.2.1 with CUDA 11.3 and later to prevent potential issue…

mathiaswagner:
… with UCX


weinbe2:
I assume nvshmem 2.2.1 has a hard requirement of CUDA 11.3, which is why we can't just update to that across CUDA versions?

maddyscientist (approved):

mathiaswagner:
No, 2.2.1 includes a fix for an issue that breaks compatibility with UCX if nvshmem 2.1 is compiled with CUDA >= 11.3. But 2.2.1 requires you to put a bootstrap shared library in the LD_LIBRARY_PATH so 2.1 requires less user intervention.

PR #1167, opened 2021-07-15
Hotfix/block ortho float8

maddyscientist:
Some fixes for block orthogonalize that should be included in the 1.1 release
* Fix local memory overhead with using float-8 ordering in the block orthogonalize kernel (do not pass raw `chirality` value to `load` function else this results in dynamic local indexing
* Remove UB from `load` function in color_spinor_field_order.h (replace raw local array cast with `memcpy`).  This fixes the block orthogonalize kernel when compiled in debug mode.

### DIFF
--- a/include/color_spinor_field_order.h
+++ b/include/color_spinor_field_order.h
@@ -317,7 +317,8 @@ namespace quda {
         constexpr int M = nSpinBlock * nColor * nVec;
 #pragma unroll
         for (int i = 0; i < M; i++) {
-          vec_t tmp = vector_load<vec_t>(reinterpret_cast<const vec_t *>(in + parity * offset_cb), (chi * M + i) * stride + x_cb);
+          vec_t tmp = vector_load<vec_t>(reinterpret_cast<const vec_t *>(in + parity * offset_cb),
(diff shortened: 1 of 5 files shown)


weinbe2:
This PR looks non-controversial. Are there other places in `color_spinor_field_order.h` where we should pre-emptively make the same change you made in c63abd9?

maddyscientist:
I don't believe there's any more illegal type-punning like this elsewhere.  The `load` functions used here, are only used in the block orthogonalize, hence probably why I hadn't noticed this issue before.  Generally recasting arrays like this works fine until you compile in debug mode, upon which, the compiler can insert padding etc., for example such that `sizeof(T[2*n]) != sizeof(complex<T>[n])` even if `2 * sizeof(T) = sizeof(complex<T>)`.

maddyscientist:
Noting that I've pulled in a couple of other fixes to this PR:
* Fix jitify compilation (broke with NVSHMEM merge)
* Apply @bjoo's local rank identification fix for systems that fail to correctly differentiate their hostnames

weinbe2 (approved):
Looks good!

PR #1175, opened 2021-08-23
Backport of bug fix from GK branch

maddyscientist:


mathiaswagner (approved):

weinbe2 (approved):
Looks good. This is residual of maintaining backwards compatibility with older MILC (old MILC works with new QUDA)---the reason why this slipped through was `qudaUpdateUPhased` was never called, only `qudaUpdateU` which is defined to have `phase_in == 0`, thus the typo in `qudaUpdateUPhased` was incidentally consistent with all usage.

weinbe2 (approved):

PR #1186, opened 2021-09-10
Feature/spin taste operator

AlexVaq:
Two modifications:

- A core one, the covariant derivative now supports nSpin = 1 fields. The packing was changed to the staggered version, which seems fine with both types of fermions (after all, there is no packing/projecting in the covariant derivative).
-
- An interface one, there are new interfaces for both the covariant derivative and the shift required to construct the spin-taste operator in staggered-like fermions. This is important, because the way the covariant derivative was applied before was horrible. There are also modifications in the MILC interface to be able to apply the spin-taste operator.

It would be nice if someone reviewed carefully the first part: I didn't have too much time to figure out how the new instantiation mechanism works, so I did "old school" instantiation. But most likely this can be streamlined like every other dslash.

For the second part, the two new functions share 90% of their code with dslashQuda. My original idea (to select within dslashQuda what you wanted to do) was discarded, because I couldn't use default parameters in the inputs (the interface is mainly C, not C++), and I didn't want to break existing code. But Im sure there is a more elegant solution.

So for now this code is ok, but I'm sure there is room for improvement.


maddyscientist:
Thanks for the contribution @AlexVaq.  It looks like you haven't derived this work from the generic kernel branch.  Can you rebase against this?  We are imminently about to merge in this branch (e.g., next week), which is a complete rewrite, using an architecture-agnostic abstraction.

AlexVaq:
I have rebased the merge in feature/generic_kernel, as per other people suggestions. The only conflicting file was lib/covDev.cu, with the question on whether I should call policy.apply(0) or not at the end of the apply function. I did not, since in the feature/generic_kernel branch there is not such a call (and actually it was removed one month ago).

include/quda.h
```diff
@@ -1152,6 +1152,35 @@ extern "C" {
...
+
+  /**
+   * Apply the spin-taste operator.
+   * @param h_out  Result spinor field
+   * @param h_in   Input spinor field
+   * @param spin   Spin gamma structure
+   * @param taste  Taste gamma structure
+   * @param param  Metadata for host and device storage
+   */
+  void spinTasteQuda(void *h_out, void *h_in, int spin_, int taste, QudaInvertParam inv_param);
```
maddyscientist, line 1182: `spin_` -> `spin` to be consistent with the Doxygen comment

include/kernels/spin_taste.cuh
```diff
@@ -0,0 +1,106 @@
...
+    SpinTasteArg(ColorSpinorField &out_, const ColorSpinorField &in_) :
+      kernel_param(dim3(in_.VolumeCB(), in_.SiteSubset(), 1)),
+      out(out_),
+      in(in_)
+    {
+      checkOrder(out_, in_);     // check all orders match
+      checkPrecision(out_, in_); // check all precisions match
+      checkLocation(out_, in_);  // check all locations match
+      if (!in_.isNative())
+        errorQuda("Unsupported field order colorspinor= %d\n", in_.FieldOrder());
```
maddyscientist, line 31: no need for "\n" in `errorQuda` calls

include/kernels/spin_taste.cuh
```diff
@@ -0,0 +1,106 @@
...
+      if (!out_.isNative())
+        errorQuda("Unsupported field order colorspinor= %d\n", out_.FieldOrder());
+      #pragma unroll
+      for (int i=0; i<4; i++) { X[i] = in_.X()[i]; }
+    }
+  };
+
+  // FIXME only works with even local volumes
```
maddyscientist, line 39: Is this a constraint that matters?

include/kernels/spin_taste.cuh
```diff
@@ -0,0 +1,106 @@
...
+      sign = 1.0 - 2.0 * (y % 2);
+    } else if (Arg::gamma == QUDA_SPIN_TASTE_G5GZ) {
+      sign = 1.0 - 2.0 * (z % 2);
+    } else if (Arg::gamma == QUDA_SPIN_TASTE_G5GT) {
+      sign = 1.0 - 2.0 * (t % 2);
+    }
+    return sign;
+  }
+
+  template <typename Arg>
```
maddyscientist, line 78: Why not just have the contents of `applySign` in the `SpinTastePhase::operator()` method?

include/kernels/spin_taste.cuh
```diff
@@ -0,0 +1,106 @@
...
+    getCoords(x, idx, arg.X, parity);
+
+    real sign = getSign(x[0], x[1], x[2], x[3], arg);
+
+    Vector out = arg.in(idx, parity);
+
+    arg.out(idx, parity) = sign * out;
+  }
+
+  template <typename Arg> struct SpinTastePhase
```
maddyscientist, line 94: We should have comments about what this functor does exactly.

lib/covDev.cu
```diff
@@ -140,12 +144,24 @@ namespace quda
 
     {
       constexpr int nDim = 4;
-      CovDevArg<Float, nColor, recon, nDim> arg(out, in, U, mu, parity, dagger, comm_override);
-      CovDev<decltype(arg)> covDev(arg, out, in);
 
-      dslash::DslashPolicyTune<decltype(covDev)> policy(
-        covDev, const_cast<cudaColorSpinorField *>(static_cast<const cudaColorSpinorField *>(&in)), in.VolumeCB(),
-        in.GhostFaceCB(), profile);
+      if (in.Nspin() == 4) {
```
maddyscientist, line 148: This needs to be updated since we no longer have `cudaColorSpinorField` objects anymore

lib/covDev.cu
```diff
@@ -37,7 +37,7 @@ namespace quda
 
       constexpr bool xpay = false;
       constexpr int nParity = 2;
-      Dslash::template instantiate<packShmem, nParity, xpay>(tp, stream);
+      Dslash::template instantiate<packStaggeredShmem, nParity, xpay>(tp, stream);
```
maddyscientist, line 40: Calling `packStaggeredShmem` here regardless of the spin?

lib/interface_quda.cpp
```diff
@@ -2027,6 +2026,451 @@ void dslashQuda(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity
...
+  ColorSpinorParam cpuParam(h_in, inv_param, gauge.X(), false, inv_param.input_location);
+  ColorSpinorField *in_h = ColorSpinorField::Create(cpuParam);
+  ColorSpinorParam cudaParam(cpuParam, inv_param);
+
+  cpuParam.v = h_out;
+  cpuParam.location = inv_param.output_location;
+  ColorSpinorField *out_h = ColorSpinorField::Create(cpuParam);
+
+  cudaParam.create = QUDA_NULL_FIELD_CREATE;
+  cudaColorSpinorField in(*in_h, cudaParam);
```
maddyscientist, line 2054: Needs to be updated for `ColorSpinorField`.  Moreover, since we now have move assignment and move constructors, as well as a default constructor for the `ColorSpinorField` class, all these pointers to fields can be made fields and not pointers to fields.

include/quda.h
maddyscientist: Also, instead of making this an int, shouldn't it be of type `QudaSpinTasteGamma`?

maddyscientist (changes requested):
Thanks for the PR @AlexVaq.  This adds some good features to QUDA.  There are, however, a number of issues to be addressed.  I've left a variety of comments in the PR, but the main points are:
* Merge in latest develop and replace the `cudaColorspinorField` with `ColorSpinorField`
* We need tests for these new interface functions.  I assume that the current `covdev_test` can be repurposed for some of this
* Said tests needs to be added to the CMake test list (`tests/CMakeLists.txt`) so that they are run automtically when `ctest` is invoked
* clang-format of the diff

PR #1187, opened 2021-09-11
Fix for multi-blas tuneKey degeneracy

maddyscientist:
Fix degeneracy in multi-blas/multi-reduce which could lead to incorrect results, e.g., Nxz=16/Nyw=1 and Nxz=1/Nyw=61 both used the same tuneKey.  This simple fix prevents this.


mathiaswagner (approved):
good catch

PR #1194, opened 2021-10-14
Feature/profile cumulative

### LINKED ISSUE #1193: Remove some unnecessary unrolls
In the block_orthogonalize.cuh
```
#pragma unroll	
      for (int tx = 0; tx < n_sites_per_thread; tx++)
```
On the host `n_sites_per_thread` can be up to 1024, and an excessive unroll can lead to LLVM having a fit.  On the device this is trivially of length one and will always unroll.  We should just remove these unrolls.

maddyscientist:
Simple PR that:
* Adds cumulative percentage to profile output (feature request from @bjoo)
* Remove some unnecessary unrolls which causes LLVM to choke when compiling for the host (closes #1193)

I have confirmed that the removal of the loop unrolling has no effect on device performance as expected (loop is length 1 on the device).

### DIFF
--- a/include/atomic.cuh
+++ b/include/atomic.cuh
@@ -10,6 +10,34 @@
    avoid confusion when resolving the native atomicAdd functions.
  */
 
+#if defined(__NVCC__) && defined(__CUDA_ARCH__) && (__COMPUTE_CAPABILITY__ < 600)
+/**
+   @brief Implementation of double-precision atomic addition using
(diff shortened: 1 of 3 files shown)


maddyscientist:
@bjoo can you review this please and confirm this fulfills your needs?

bjoo (approved):
include/atomic.cuh brings arch specific compilation into an arch non-specific header, to deal with Maxwell. @maddyscientist will fix in follow on PR

PR #1061, opened 2020-09-28
Add dynamically trained MADWF blocking to MSPCG

hummingtree:
This PR adds an exploratory feature that enables using a Mobius DWF operator with reduced `Ls` for MSPCG.

The idea aligns with the M\"obius Accelerated Domain Wall Fermion (MADWF, [arXiv:1111.5059](https://arxiv.org/abs/1111.5059)): having a M\"obius operator with reduced `Ls` that is *close* to the original operator to speed up inversions. The original MADWF algorithm is based on the domain wall - overlap transformation to transform between different `Ls`. In principle for MSPCG we can use the same trick for the preconditioner, but the application of MADWF will not be fast enough due to the multiple inversions involved.

Instead we apply a simple transfer matrix to transform between different `Ls`. For the preconditioning inversion, the rhs is first transformed to reduced `Ls`, after which a inversion is performed. The solution, which lives in the reduce `Ls` space, is transformed back into the original `Ls` space with the very same transfer matrix.

And the transfer matrix is *trained* with a straight forward loss function.

See section 3.3 of [arXiv:2104.05615](https://arxiv.org/abs/2104.05615).

See the [wiki page](https://github.com/lattice/quda/wiki/Preconditioning) for more info.

A few remaining TODOs:
- [x] Remove the duplicate (alternative) reliable update routines in both PCG and CG: create a reliable update `struct` to handle this.
- [x] Add issue for creating preconditioning solvers.
- [x] Update corresponding sections of wiki.
- [x] Clang-format.

### `16` nodes, grid layout `2x2x4x6`
***
- CG, double-half
```
6 solves, mean iteration count 11383.4 (stddev = 31.1095), with mean solve time 217.133 (stddev = 0.908744), mean GFLOPS 106630 (stddev = 170.277) [excluding first solve]
```

- MSPCG, suppressor = `0.848`, `double-half-half`, `6` inner iteration, `Ls`: `12->4`
```
6 solves, mean iteration count 4778 (stddev = 23.5266), with mean solve time 172.357 (stddev = 0.927253), mean GFLOPS 169292 (stddev = 180.042) [excluding first solve]
```
`speed up = 217.133/172.357 = 1.26`

### `64` nodes, grid layout `4x4x4x6`
***
- CG, double-half
```
6 solves, mean iteration count 11384.2 (stddev = 30.9629), with mean solve time 92.531 (stddev = 0.766309), mean GFLOPS 250247 (stddev = 1464.39) [excluding first solve]
```
[...]

PR #1197, opened 2021-10-21
Feature/cub abstraction

maddyscientist:
This PR applies the abstractions that were missed in #1056.
* Abstract all use of cub from QUDA.  To this end, we now have a generic interface for `BlockReduce` and `WarpReduce` modeled on cub's interface, with cub deployed as the CUDA implementation of this in `block_reduce_helper.h`
* Cleanup and abstract the atomic operations, into a target-specific atomic_helper.h
* Some fixes for Jitify compilation, and we now use the runtime CUDA version to determine the compilation options
* Introduction of the quda_arch.h file from #1092 to remove explicit inclusion of CUDA headers
* Renamed `vector_type` to `array` since this is a more apt description of this class
* Adopt the power-of-two array from #1092 
* Adopt various minor fixes from #1092

Once this PR is merged, and the changes propagated into #1092, we should be ready to review #1092.

@bjoo can you review this and confirm that I've captured all of your non-HIP specific additions?
@hummingtree can you perform a general review, but especially to check if the `BlockReduce` abstraction suits your needs?


include/kernels/dslash_mdw_fused.cuh
```diff
@@ -305,9 +305,7 @@ namespace quda {
         constexpr int N_sm = N + sm_n_pad_size(N);
         constexpr int M_sm = M + sm_m_pad_size(M);
 
-        float *smem_scale = shared_memory_data;
-
-        half2 *sm_b = reinterpret_cast<half2 *>(smem_scale + 32);
+        half2 *sm_b = reinterpret_cast<half2 *>(shared_memory_data + 32);
```
hummingtree: Since the shared memory scale has been moved to thread-owned, we do not need the `32` here. I will push the changes to this branch - does that sound good?

maddyscientist: Sounds good

include/targets/cuda/mma_tensor_op/gemm.cuh
```diff
@@ -129,8 +130,8 @@ namespace quda
                 } else {
 
                   using store_type = typename GmemAccessor::store_type;
-                  using store_vector_type = typename VectorType<store_type, 4>::type;
-                  store_vector_type v = *reinterpret_cast<store_vector_type *>(&p[n_idx * ld + m_idx]);
+                  using store_array = typename VectorType<store_type, 4>::type;
+                  store_array v = *reinterpret_cast<store_array *>(&p[n_idx * ld + m_idx]);
```
hummingtree, line 134: Do we need to change from `VectorType` to `array` to match the style in other places?

include/targets/cuda/mma_tensor_op/hmma_m16n16k16_sm70.cuh
```diff
@@ -373,8 +374,8 @@ namespace quda
 
       constexpr bool fixed = GmemOperandC::fixed;
 
-      using vector_type = vector_type<store_type, 2>;
-      auto ptr = reinterpret_cast<vector_type *>(cc.data());
+      using array = array<store_type, 2>;
+      auto ptr = reinterpret_cast<array *>(cc.data());
```
hummingtree, line 378: Have you checked the performance implications of changing from `VectorType` to `array`? When I wrote the code the compiler did not generate the 64-bit `STG` instruction unless the type is explicitly `float2`. :(

include/array.h
```diff
@@ -0,0 +1,57 @@
...
+#ifdef QUAD_SUM
+  template <> __device__ __host__ inline doubledouble zero() { return doubledouble(); }
+  template <> __device__ __host__ inline doubledouble2 zero() { return doubledouble2(); }
+  template <> __device__ __host__ inline doubledouble3 zero() { return doubledouble3(); }
+#endif
+
+  /**
+     Simple array object which mimics std::array
+   */
+  template <typename T, int n> struct array {
```
weinbe2, line 9: re:a comment about LDS from @hummingtree elsewhere---since we know the size of this at compile time, we should stick an `alignas(sizeof(T) * n)` on this struct. I've found the explicit `alignas` is helpful, even though the compiler should be able to exactly deduce the type and alignment anyway.

include/kernels/dslash_mdw_fused.cuh
hummingtree: [715aa4e](https://github.com/lattice/quda/pull/1197/commits/715aa4efdfc6adb26d3df1d4ac3055860309cbab).

include/targets/cuda/mma_tensor_op/hmma_m16n16k16_sm70.cuh
hummingtree: This is all good now - I confused `vector_type` with `VectorType` ...

include/targets/cuda/mma_tensor_op/gemm.cuh
hummingtree: Discussed offline. We should keep `VectorType`.

include/array.h
maddyscientist: Since we're not using `array` for load/store directly in performance critical code (that's what `VectorType` is for) I think we should leave this without the `alignas`.  Otherwise adding the `alignas` could have unexpected effects with respect to `sizeof(array)`.

include/targets/generic/block_reduce_helper.h
```diff
@@ -0,0 +1,251 @@
...
+    template <bool async = false> __device__ __host__ inline T AllMin(const T &value)
+    {
+      static_assert(param_t::batch_size == 1, "Cannot do AllMin with batch_size > 1");
+      return target::dispatch<block_reduce>(value, async, batch, true, quda::minimum<T>(), param_t());
+    }
+
+    /**
+       @brief Perform a block-wide custom reduction
+       @param[in] value Thread-local value to be reduced
+       @param[in] r The reduction operation we which to apply
```
hummingtree:
```suggestion
       @param[in] r The reduction operation we want to apply
```

include/targets/generic/block_reduce_helper.h
```diff
@@ -0,0 +1,251 @@
...
+    template <bool async = false, typename reducer_t>
+    __device__ __host__ inline T Reduce(const T &value, const reducer_t &r)
+    {
+      return target::dispatch<block_reduce>(value, async, batch, false, r, param_t());
+    }
+
+    /**
+       @brief Perform a block-wide custom reduction
+       @param[in] value Thread-local value to be reduced
+       @param[in] r The reduction operation we which to apply
```
hummingtree:
```suggestion
       @param[in] r The reduction operation we want to apply
```

hummingtree (approved):
Nice and clean PR @maddyscientist. This makes QUDA truly _generic_, and together with the GK PR, is really fantastic work. As a side note, I have verified this works well with #1061.

include/targets/generic/block_reduce_helper.h
maddyscientist: c3b901c

include/targets/generic/block_reduce_helper.h
maddyscientist: c3b901c

bjoo (approved):
Looks visually OK from point of view of HIP. Please merge and I'll have to catch up. Don't want to hold this back esp. if it is already well tested on NV testbeds.

PR #1198, opened 2021-10-26
non-degenerate twisted clover and gauge path

kostrzewa:
enables rudimentary tmLQCD-based Nf=2+1+1 HMC including gauge derivative


include/kernels/clover_invert.cuh
```diff
@@ -54,9 +56,11 @@ namespace quda
         Mat A = arg.clover(x_cb, parity, ch);
         A *= static_cast<real>(2.0); // factor of two is inherent to QUDA clover storage
 
-        if (Arg::twist) { // Compute (T^2 + mu2) first, then invert
+        if (Arg::twist) { // Compute (T^2 + mu2 - epsilon2) first, then invert
           A = A.square();
-          A += arg.mu2;
+          // not sure what's better, a scalar -0.0 for the generate case or another
+          // if condition and a change to the template arguments of CloverInvertArg
+          A += (arg.mu2 - arg.epsilon2);
```
maddyscientist: Perhaps add a new parameter in the arg which stores the difference `mu2 - epsilon2`, changing this line to
```c++
A += arg.mu2_minus_epsilon2;
```

include/kernels/dslash_clover_helper.cuh
```diff
@@ -1,10 +1,11 @@
...
-#include "clover_field_order.h"
-#include "color_spinor.h"
-#include "linalg.cuh"
-#include "kernel.h"
+#include <color_spinor_field_order.h>
+#include <clover_field_order.h>
+#include <index_helper.cuh>
+#include <color_spinor.h>
+#include <linalg.cuh>
+#include <kernel.h>
```
maddyscientist: This seems an unnecessary change.  Tangentially, I've been considering whether we should switch to using `#include " "` for internal includes as is used in many other places, but haven't really decided one way or another.

include/kernels/dslash_clover_helper.cuh
```diff
@@ -1,10 +1,11 @@
 #pragma once
 
-#include "color_spinor_field_order.h"
-#include "clover_field_order.h"
-#include "color_spinor.h"
-#include "linalg.cuh"
-#include "kernel.h"
+#include <color_spinor_field_order.h>
+#include <clover_field_order.h>
+#include <index_helper.cuh>
```
maddyscientist: I suspect this is a copy-paste addition: I don't think the index_helper header is needed in this file.

include/kernels/gauge_force.cuh
```diff
@@ -14,48 +14,58 @@ namespace quda {
...
 
       qudaMemcpy(buffer, path_h, bytes, qudaMemcpyHostToDevice);
       host_free(path_h);
 
       // finally set the pointers to the correct offsets in the buffer
-      // FIXME we should clean up this pointer and casting mess
-      for (int d=0; d < 4; d++) this->input_path[d] = (int*)((char*)buffer + d*num_paths*max_length*sizeof(int));
-      length = (int*)((char*)buffer + 4*num_paths*max_length*sizeof(int));
-      path_coeff = (double*)((char*)buffer + 4 * num_paths * max_length * sizeof(int) + num_paths*sizeof(int));
+      for (int d=0; d < 4; d++) this->input_path[d] = buffer + d*num_paths*max_length;
```
maddyscientist, line 54: Nice to see this mess cleaned up 😄

include/quda.h
```diff
@@ -77,7 +77,8 @@ extern "C" {
 
     int overlap; /**< Width of overlapping domains */
 
-    int overwrite_mom; /**< When computing momentum, should we overwrite it or accumulate to to */
+    int overwrite_gauge; /**< When computing gauge, should we overwrite it or accumulate to it */
+    int overwrite_mom; /**< When computing momentum, should we overwrite it or accumulate to it */
```
maddyscientist: This needs to be added to the Fortran interface as well: lib/quda_fortran.F90

include/quda.h
```diff
@@ -1265,6 +1267,23 @@ extern "C" {
 			    double* loop_coeff, int num_paths, int max_length, double dt,
 			    QudaGaugeParam* qudaGaugeParam);
 
+  /**
+   * Compute the product of gauge along a path and add to/overwrite the output field
+   *
+   * @param out The output field to be updated
```
maddyscientist: Add the in/out descriptors to these `@param` items

include/quda.h
```diff
@@ -129,6 +130,7 @@ extern "C" {
     double mq3;
 
     double mu;    /**< Twisted mass parameter */
+    double tm_rho; /**< Hasenbusch mass shift applied like twisted mass to diagonal (but not inverse) */
```
maddyscientist: Add to quda_fortran.F90

tests/gauge_force_test.cpp
```diff
@@ -236,7 +236,9 @@ int path_dir_t[][5] = {
 static double force_check;
 static double deviation;
 
-void gauge_force_test(void)
+// The same function is used to test computePath.
+// If compute_force is false then a path is computed
+void gauge_force_test(bool compute_force=true)
```
maddyscientist: Great to see that arbitrary path link calculation through extension of the gauge force, and also that the gauge_force_test has been augmented to support testing this.  Can you add this variation of `gauge_force_test` to the test list?  You just need add the appropriate gauge_force invocation to `tests/CMakeLists.txt`, such that it will be run when one calls `ctest`.

lib/dslash_wilson.cu
```diff
@@ -35,13 +40,25 @@ namespace quda
...
+        WilsonArg<Float, nColor, nDim, recon> arg(out, in, U, a, x, parity, dagger, comm_override);
+        Wilson<decltype(arg)> wilson(arg, out, in);
 
-      dslash::DslashPolicyTune<decltype(wilson)> policy(
-        wilson, const_cast<cudaColorSpinorField *>(static_cast<const cudaColorSpinorField *>(&in)), in.VolumeCB(),
-        in.GhostFaceCB(), profile);
+        dslash::DslashPolicyTune<decltype(wilson)> policy(
+          wilson, const_cast<cudaColorSpinorField *>(static_cast<const cudaColorSpinorField *>(&in)),
+          in.getDslashConstant().volume_4d_cb, in.getDslashConstant().ghostFaceCB, profile);
+      }
```
maddyscientist: This if / else with respect to dimension will double the number of templates instantiated for `dslash_wilson.cu`.  I want to consider this.  Remind me, why do we need the `nDim=5` template for the `WilsonArg`?  (I realize I probably wrote the underlying code originally 😉 )

include/kernels/dslash_wilson.cuh
```diff
@@ -141,16 +141,16 @@ namespace quda
...
     {
       typedef typename mapper<typename Arg::Float>::type real;
       typedef ColorSpinor<real, Arg::nColor, 4> Vector;
 
       bool active
         = mykernel_type == EXTERIOR_KERNEL_ALL ? false : true; // is thread active (non-trival for fused kernel only)
       int thread_dim;                                        // which dimension is thread working on (fused kernel only)
-
-      auto coord = getCoords<QUDA_4D_PC, mykernel_type>(arg, idx, 0, parity, thread_dim);
+
```
maddyscientist, line 152: This explicit change from `0` to `s` may have a small performance regression for plain Wilson fermions.  We should check if this is the case.  I think we could rewrite this as `Arg::nDim == 5 ? s : 0`, to avoid any regression here since at present the Wilson operator is instantiated separately for the non-degenerate TMC operator.

include/quda.h
maddyscientist: Wondering if we can unify this with the `clover_rho` member which has the same purpose.  Replace them both with `hasenbusch_rho` ?

maddyscientist:
Thanks for this contribution @urbach and @kostrzewa (and it looks like with help from @Marcogarofalo @pittlerf @sbacchio) 😄   I've undertaken an initial visual review of the PR, and left various inline comments that need to be addressed.  Overall it looks great though.

A couple of other things:
* Add some non-deg tmc test invocations to the test list on tests/CMakeLists.txt, as well as for the gauge path invocations of `gauge_force_test`
* `clang-format` needs to be applied to the diff of this PR.  See instructions here: https://github.com/lattice/quda/wiki/Coding-Conventions-and-Style#coding-style-and-clang-format.  If you want a short cut for this, there is a helper script on nvsocal2: `/usr/bin/clang-format-quda-11`.  To use this, you do something like (making sure that your local repository is upto date for both develop and the target branch (feature/ndeg-twisted-clover):
```bash
$ clang-format-quda-11 develop feature/ndeg-twisted-clover
```
* Please add any new names to the contributor list on the README that contributed to this PR

Once these are taken care of, I'll do some testing and then we can get this merged into develop.

kostrzewa:
@maddyscientist thanks a lot for looking at this and the list of todos to get it ready for inclusion. I apologise that it currently takes me a while to react to things. I'll try to take care of the remaining points soon, hopefully by the end of next week.

maddyscientist:
Sounds good.  😄

include/kernels/dslash_clover_helper.cuh
kostrzewa: This might be a merge artefact or an overzealous inclusion of headers, I'll take a look and keep only what is really necessary.

include/kernels/clover_invert.cuh
kostrzewa: 02dc5afaf605f4a63f49d3324cd96c5c0eadf6d3

include/kernels/dslash_clover_helper.cuh
kostrzewa: 55e34781a879bbd5afc42ae82e95f374a99fb5d0

include/kernels/dslash_clover_helper.cuh
kostrzewa: 55e34781a879bbd5afc42ae82e95f374a99fb5d0

include/quda.h
kostrzewa: I absolutely wanted to follow this path in the first place but realised that unfortunately, `clover_rho` and `tm_rho` are used differently. `rho` is added to the clover term as a real number on the diagonal in the clover accessor (see `clover_field_order.h`). `tm_rho`, instead, is added in operators and comes with an imaginary unit, a gamma_5 and a tau_3. Hence, instead of locating it in the clover accessor, the only place where it comes into play is in the call of `ApplyTwistedClover` in `DiracTwistedClover::DslashXpay` by shifting the twisted quark mass.

In this sense, the duplication isn't really between `clover_rho` and `tm_rho` but between `DiracTwistedClover` and `DiracCloverHasenbuschTwist`, the latter being basically Twisted Clover but not quite, if I understood correctly.

kostrzewa: The complication then arises in differentiating whether the proposed `hasenbusch_rho` should be added to the diagonal in the clover accessor or whether it should take the place of `tm_rho` for `DiracTwistedClover` or `mu` for `DiracCloverHasenbuschTwist`.

IIRC it was nontrivial to prevent the `add_rho=true` instance of the clover accessors being instantiated (and used in the operators) when `clover_rho` (after unification `hasenbusch_rho`) is set non-zero. I don't remember the logic chain that I followed to conclude this and I might well have the wrong impression.

kostrzewa: Hmm. Having `clover_rho != 0` -> `cloverPrecise->setRho(inv_param->clover_rho);` (`interface_quda`). This in turn leads to `rho(clover.Rho())` in the accessor constructor but, if it is instantiated with `add_rho==false`, this shouldn't have any effect.

One issue that we certainly have from the tmLQCD side is that we use `tm_rho` also for Hasenbusch preconditioning regular untwisted clover. In that case, we almost certainly would run into problems (so in practice for untwisted clover HMC with preconditioning, we employ `TwistedClover` with `mu` set to zero and `tm_rho` set accordingly.) 

I'm also not a fan of the lack of unification of these three very similar things... It should be noted that at least in principle, `clover_rho` and `tm_rho` (or, equivalently, `mu` of `DiractCloverHasenbuschTwist`) are different and one might argue that at least two of the three should be allowed to coexist.

lib/dslash_wilson.cu
kostrzewa: IIRC, the reason is to be found in `include/kernels/dslash_ndeg_twisted_clover_preconditioned.cuh`, where I call `applyWilson` on a 5D spinor. This is different than in the `dslash_ndeg_twisted_mass_preconditioned.cuh` case, where the two *input* flavours are taken care of by different threads and the output is kept in shared memory.

I didn't know how to achieve this call of `WilsonApply` onto a 5D spinor inside of `nDegTwistedCloverPreconditioned` otherwise but I presume that it would be possible to modify this to work in the same way as `nDegTwistedMassPreconditioned` with, I think, the drawback that the clover field would have to be loaded twice (once for each flavour).

kostrzewa:
sorry about the force-push but I had left a typo in https://github.com/lattice/quda/pull/1198/commits/456a840ace3d4ff3e69cb71d4ce52816a46b0b9c where I had actually intended to clean something up...

include/kernels/dslash_wilson.cuh
kostrzewa: 7105c0efe8d16d44f125da9a08384c4ca2a944f7

include/quda.h
kostrzewa: What would you like to see here specifically? In principle these are present, although for the "in" (`sitelink`) the descriptor refers to a force, which is incorrect of course. I've updated these in 8751d044c34cd864d8c139e65e7500046368ab27 but it's unclear to me if this is sufficient.

include/quda.h
kostrzewa: ce9d8e0a47f6bb76a2ea725f19a67aa5af6a8b64

include/quda.h
kostrzewa: > Add to quda_fortran.F90

9d6b5c58b57e4eedbd2cd32fb80ff9d1516573a4

include/quda.h
maddyscientist: What I meant here is to add [in,out] declarations to the doxygen parameter items.  This isn't practiced throughout this file, but it is best practice to apply this to all new functions.
e.g.
```
  /**
   * Compute the product of gauge along a path and add to/overwrite the output field
   *
   * @param[in,out] out The output field to be updated
   * @param[in] sitelink The gauge field from which we compute the force
   * @param[in] input_path_buf[dim][num_paths][path_length]
 ```

Comment (maddyscientist) [reply include/quda.h]: Maybe you could add this to the doxygen for `computeGaugeForceQuda` at the same time 😉

File: include/quda.h (hunk 7 above)
Comment (maddyscientist) [reply include/quda.h]: IRC, I added `add_rho` and the `clover_rho` parameter as an expedient hack to support the BQCD style Hasenbusch preconditioning employed for clover fermions.  The fact that it's buried into the accessor as opposed tot he operator isn't for any good reason.  It probably makes more logical sense to have it applied as part of the operator.

You make a good point about duplication between `DiracTwistedClover` and `DiracCloverHasenbuschTwist`.  That's something I had on my list to look at, and understand if we can pare back on some of these kernels.

File: lib/dslash_wilson.cu (hunk 9 above)
Comment (maddyscientist) [reply lib/dslash_wilson.cu]: I think we should follow the same strategy as `nDegTwistedMassPreconditioned` here and split the flavours onto separate threads. The autotuner will care of this, and place both threads for the same site on the same thread block, and the double load of the clover field will be avoided with the second load being sourced from L1 cache with essentially negligible overhead.  This will also help reduce register pressure, improving occupancy, and likely improving performance.

File: tests/gauge_force_test.cpp (hunk 8 above)
Comment (kostrzewa) [reply tests/gauge_force_test.cpp]: Is 934bac3e796517137bd0b031be772f38261cd5ab okay?

**Comment by kostrzewa** [comment]
> Add some non-deg tmc test invocations to the test list on tests/CMakeLists.txt

I've made the following modif: c6ad97a4d0f5927acff457d3bbf875f2f7770b42, I hope that was the right thing to do.

File: include/quda.h (hunk 6 above)
Comment (kostrzewa) [reply include/quda.h]: Ah, got it!

Comment (kostrzewa) [reply include/quda.h]: 5e97b26d6be77d0dc22c53e77734ffe298e110d4

**Comment by kostrzewa** [comment]
Save for the clang-format I think I've addressed all the mentioned points, please let me know if I've missed something.

Should we discuss the `clover_rho` / `tm_rho` / Hasenbusch twist `mu` issue?

If / when you're happy with the modifications I'll take care of synching with the latest develop and perform the clang-format.

**Comment by maddyscientist** [comment]
> > Add some non-deg tmc test invocations to the test list on tests/CMakeLists.txt
> 
> I've made the following modif: [c6ad97a](https://github.com/lattice/quda/commit/c6ad97a4d0f5927acff457d3bbf875f2f7770b42), I hope that was the right thing to do.

This looks good, thanks.

File: tests/gauge_force_test.cpp (hunk 8 above)
Comment (maddyscientist) [reply tests/gauge_force_test.cpp]: This looks good, thanks.

**Comment by maddyscientist** [comment]
> Save for the clang-format I think I've addressed all the mentioned points, please let me know if I've missed something.
> 
> Should we discuss the `clover_rho` / `tm_rho` / Hasenbusch twist `mu` issue?
> 
> If / when you're happy with the modifications I'll take care of synching with the latest develop and perform the clang-format.

I think the next step is now for me to just do some testing, and then we can discuss the open items.  I'll try to get this done this week.

**Comment by maddyscientist** [comment]
Ok, I've done some initial testing and pushed a couple of CMake fixes.  

One thing I do see is that the dslash kernels are producing local memory: this shouldn't be happening for these kernels, and will impact the performance.  I'll take a look at what's going on here in the next few days.  Likely once this is addressed, performance will increase.

**Comment by maddyscientist** [comment]
Ok, I've fixed the stack frame usage in the dslash_ndeg_twisted_clover.cuh kernel.  I'll push this fix shortly once I've done more testing (don't want to overload nvsocal2 while I'm testing).

Issue was this line:
```c++
Vector x = flavor == 0 ? x0 : x1;
```
To the compiler's point of view, this is dynamic indexing (`Vector x = x_[flavor]`, and it spills `x0` and `x1` onto the stack as opposed to keeping them in registers.  The fix is simple:
```c++
Vector x;
if (flavor == 0) x = x0;
else             x = x1;
```
The end result is a large speedup (on my Quadro GV100 I saw 24^4 single precision performance jump from 1250 to 1550 GFLOPS).

maddyscientist:
Some updates from my end
* I confirmed that there is no need in principle to compiler both 4-d and 5-d variants of the Wilson-Dslash.  The 4-d variant is sufficient, as long as the thread count and offsets are set correctly.
* However, I did verify that the overhead of using a generalized batched Wilson Dslash for regular Wilson fermions, e.g., the compiler doesn't know `s=0`.  On my workstation I see ~1% reduction in performance.  Not something to worry about, but enough to give me pause.
* I then realized that the batched Wilson-Dslash operator you need for ndeg-TMC is exactly the same as the Dslash4 operator from 4-d preconditioned Domain-Wall.  So we can use that instead, and remove the doubling from the Wilson-Dslash compilation unit.  We still have to compile it of course, but this means it's done in a separate unit for parallel compilation, and we avoid the small hit to plain Wilson performance.

So, I have pushed a couple of commits:
* Switch the two flavor TMC to use DWF-4d dslash instead of the Wilson-Dslash
* I've applied clang format since I figured I might as well 😉 

I will revisit this with the forthcoming multi-RHS work that is commencing.  Ideally we'll be bale to utilize the same Wilson-Dslash for plain Wilson, ndeg-TMC and DWF-4d.

So I think the final thing here to deal with is the issue of the `rho` parameter.  Let me dwell on that for a bit longer and then we can discuss.

kostrzewa:
> I then realized that the batched Wilson-Dslash operator you need for ndeg-TMC is exactly the same as the Dslash4 operator from 4-d preconditioned Domain-Wall. So we can use that instead, and remove the doubling from the Wilson-Dslash compilation unit. We still have to compile it of course, but this means it's done in a separate unit for parallel compilation, and we avoid the small hit to plain Wilson performance.

Ah, perfect! I was wondering how this was done for DWF but stopped short of actually checking...

kostrzewa:
> I've applied clang format since I figured I might as well wink

Thanks a lot!

kostrzewa:
> Ok, I've fixed the stack frame usage in the dslash_ndeg_twisted_clover.cuh kernel. I'll push this fix shortly once I've done more testing (don't want to overload nvsocal2 while I'm testing).
> 
> Issue was this line:
> 
> ```c++
> Vector x = flavor == 0 ? x0 : x1;
> ```
> 
> To the compiler's point of view, this is dynamic indexing (`Vector x = x_[flavor]`, and it spills `x0` and `x1` onto the stack as opposed to keeping them in registers. The fix is simple:
> 
> ```c++
> Vector x;
> if (flavor == 0) x = x0;
> else             x = x1;
> ```
> 
> The end result is a large speedup (on my Quadro GV100 I saw 24^4 single precision performance jump from 1250 to 1550 GFLOPS).

This is a great pointer. I would not have guessed the consequences of the ternary operator there to be that severe...

maddyscientist:
> This is a great pointer. I would not have guessed the consequences of the ternary operator there to be that severe...

Yeah, usually the ternary operator is the way to go when dealing with simple types.  In this case though, since we have an object of 96 bytes size, and the compiler is forced to put the object into addressable space, e.g., the stack.  If you look at the regular ndeg twisted mass kernels, you'll see I had to do something similar there as well.

maddyscientist (approved):
This PR is a great contribution to QUDA, thanks to all that worked on this.

I think keeping the `tm_rho` and `clover_rho` separate makes sense, since as stated have different meanings.  At least for now anyway.  We may want to revisit this in the future.

@weinbe2 @mathiaswagner @cpviolator @hummingtree would one of you do a review here please?  I've been through this, done a lot of testing, but having a second eye would be good in case I missed something.  Noting that I've verified the Jitify path is all ok here after merging in develop, and applying a patch from #1213.

Given that we have the #1213 PR, which touches may of the same files, once we've merged this, I reverify that all is ok with ndeg twisted clover fermions when clover reconstruction is enabled.  It should give a nice perf boost.

include/dslash_quda.h
```diff
@@ -522,6 +523,81 @@ namespace quda
...
+  /**
+     @brief Driver for applying the non-degenerate preconditioned twisted-clover stencil
+
+     out = a * (C + i*b*gamma_5*tau_3 - c*tau_1)^{-1} * D * in + x
+         = a * (C - i*b*gamma_5*tau_3 + c*tau_1)/(C^2 + b^2 - c^2) * D * in + x
+         = A^{-1} * D * in + x
+
+     where D is the gauged Wilson linear operator and C is the clover
+     field.  This operator can (at present) be applied to only single
+     parity (checker-boarded) fields.  When the dagger operator is
```
weinbe2, line 567: While I certainly appreciate the logic noted below (kernel fusion), `dagger` not meaning `dagger` does bother me---in particular I think it implies `DiracTwistedClover::MdagM` isn't a Hermitian positive-definite operator in the non-degenerate case?

To be clear, I'm not saying "throw out kernel fusion" because obviously this workflow is working, but I'm wondering if there's a way we can change the plumbing so `dagger` means `dagger` and the fused "dagger without transpose" operator is exposed another way.

maddyscientist: `MdagM` is a Hermitian operator for ndeg TMC.  It's just not composed using `M`, followed by `Mdag`.  This is not unique to nondeg TMC, and we've got this style for other operators.

weinbe2: That's not what it looks like in the body of `DiracTwistedClover::MdagM`---it looks like it's indeed the sequence of `M` then `Mdag`, and I don't see a custom implementation of `Mdag`.

maddyscientist: Oops, my brain crossed wires there.  Agreed, it is written as `M` and `Mdag` product.

weinbe2: Correct what I have from `DiracTwistedClover` to `DiracTwistedCloverPC` --- I think `DiracTwistedClover` itself is fine.

maddyscientist: Let me try again: the operator `MdagM` is Hermitian, and the components `M` and `Mdag` are what they claim to be.  The issue is just that the components are put together to enable kernel fusion.

maddyscientist: And this isn't an issue, it's just that for kernel fusion, all operators must appear to the left of D_4, so when we call something daggered, we only mean we are daggering the operator components, and not the entire operator (e.g., reversing its order).  That's the convention that we've had for a long time.

maddyscientist: Where my brain crossed wires is that we could fuse across `M` / `Mdag` boundaries in order to give more latency hiding, in the same way that's already been done for DWF-4d / Möbius fermions.  But that optimization lives in my head only (a subtle, but important distinction 😉 ).

weinbe2: Okay, I see where I went astray. I was comparing against `DiracCloverPC` where the mathematical dagger vs non-dagger is in `DiracCloverPC::M`, so when I didn't see that pattern in `DiracTwistedCloverPC` I got confused. What I missed is that pattern is in `DiracTwistedCloverPC::Dslash` instead.

tl;dr I was thrown off by an inconsistency of convention between regular clover and twisted clover, and now everything's clear.

maddyscientist: All is good.  This discussion reminded me of wanting to fuse across the Mdag / M boundaries for symmetric clover preconditioned kernels, so it was worthwhile for me 😄

kostrzewa:
I'm sorry for remaining silent for a week again. End of the year report hell. I'll try to get back into this ASAP.

maddyscientist:
All good @kostrzewa.  The clover reconstruction PR (#1213) is now approved, and since this touches the same files, the plan is to merge that into develop and then merge into here.  I'll handle that, and I'll update here when I've confirmed all is ok, and I'll report updated performance numbers as well.

maddyscientist:
I've merged in develop, which now includes the clover reconstruction optimization (72 -> 56 numbers per clover field site).  All looks ok, and as expected, with clover reconstruction enabled we get a nice speedup.

## Performance in GFLOPS, CUDA 11.2, Quadro GV100, V = 32^4

### Full operator
| Precision | Gauge Recon | Standard |  Clover Recon | 
| --------- | --------------| --------- | -------- |
| double     | 18                   | 782          |  804        |
| double     | 12                   | 881          |  887        |
| double     | 8                   | 750          |   782       |
| single     | 18                   | 1395          |  1502        |
| single     | 12                   | 1900          |  2007        |
| single     | 8                   | 2039          |   2188       |
| half     | 18                   | 2800          |  2854        |
| half     | 12                   | 2977          |  3043        |
| half     | 8                   | 2961          |   3076       |

### Preconditioned Dslash operator
| Precision | Gauge Recon | Standard |  Dynamic Clover | Clover Recon | 
| --------- | --------------| --------- | --- | -------- |
| double     | 18                   | 590          |  707 | 718        |
| double     | 12                   | 650          |  805 | 815        |
| double     | 8                   | 650          |  821   | 853 |
| single     | 18                   | 1181 | 1465          |  1513        |
| single     | 12                   | 1425 | 1683          |  1778        |
| single     | 8                   | 1513          |   1861 | 1936       |
| half     | 18                   | 2362          |  2411        | 2470 |
| half     | 12                   | 2655          |  2582       | 2617 |
| half     | 8                   | 2718          |   2584       | 2618 |

maddyscientist:
I have applied a few optimizations, as well as some cleanup.
 * I noticed that the prior local memory observed when using `ColorSpinor` objects in conjunction with the ternary operator was caused by the underlying object not being trivially copyable.  Rectifying this (letting the compile synthesis the copy constructor and assignment operator) allows the compiler to properly optimize now, and allows for the restoration of using the ternary operator. (218b003)
* For the preconditioned operator, with dynamic-clover, the overhead of both flavor threads inverting the clover term was significant, as could be seen in the results above, where non-dynamic clover was actually faster, despite having to load both the inverse and direct clover matrix.  To rectify this, we now "swizzle" the flavor parallelism with the per-thread chirality.  In other words, each flavor thread is now responsible for applying the clover matrix for one chirality only, and it does so for both flavors.  To do this, we perform a shared memory exchange prior to clover application, and then restore afterwards prior to writing out.  (855ec7a)
* For the direct operator, I noticed the register pressure was rather high, and indeed in fp64 we had register spills.  This was because each thread was loading both flavors of the accumulator field (`x`) simultaneously, and keeping them in registers throughout the clover application.  After some experimentation, I found the best approach was for each thread to load its own vector, and then store it in shared memory for consumption by the other flavor thread after the clover term has been applied.  (c7a48d8)
* Optimization of the plain NdegTwistClover kernel: again register pressure was high here, with spills in fp64.  The optimum approach here does seem to be to keep both flavors in the same thread, but I found it was best to let the compiler deal with the loop unrolling rather than the effective manual unroll previously employed (001d880)
* Consistent will all other clover operators with dynamic inversion, the precision of the Cholesky decomposition is given by a trait `clover::cholesky_t`. (3cdfa2b)
 * Following on from the above, I have added the CMake parameter `QUDA_CLOVER_CHOLESKY_PROMOTE`, to determine the whether the type returned by `clover::cholesky_t` should promote the precision or not.  To be explicit, with `QUDA_CLOVER_CHOLESKY_PROMOTE=ON`, the Cholesky decomposition used when inverting the clover term will use double precision for both double and single precision kernels, and with it `OFF` will use double for double and single for single precision.  This option was added since we noticed that this can sometimes help the stability of mixed precision solvers.  For half and quarter precision we always use single precision. (b84ae09)

 @kostrzewa please have a read through my changes, and let me know if you think this all fine.
 I'm handing over the review for this PR now to @weinbe2.

include/gauge_force_quda.h
```diff
@@ -16,6 +16,20 @@ namespace quda {
    */
   void gaugeForce(GaugeField& mom, const GaugeField& u, double coeff, int ***input_path,
 		  int *length, double *path_coeff, int num_paths, int max_length);
+
+  /**
+     @brief Compute the product of gauge-links along the given path
+     @param[out] out Gauge field which the result is added to
+     @param[in] u Gauge field (extended when running no multiple GPUs)
```
weinbe2: `no` -> `on`

include/kernels/dslash_clover_helper.cuh
```diff
@@ -162,5 +170,78 @@ namespace quda {
...
+        const complex<real> a(0.0, chirality == 0 ? arg.a : -arg.a);
+
+        Mat A = arg.clover(x_cb, clover_parity, chirality);
+
+        half_fermion in_chi[n_flavor] = {in[0].chiral_project(chirality), in[1].chiral_project(chirality)};
+        half_fermion out_chi[n_flavor];
+
+#pragma unroll
+        for (int flavor = 0; flavor < n_flavor; flavor++) {
+          out_chi[flavor] += A * in_chi[flavor];
```
weinbe2: This `+=` can just be an `=`

maddyscientist: 3840f73a

include/gauge_force_quda.h
maddyscientist: 3840f73a

lib/clover_invert.cu
```diff
@@ -17,7 +17,8 @@ namespace quda {
       compute_tr_log(compute_tr_log)
     {
       strcat(aux, compute_tr_log ? ",trlog=true" : "trlog=false");
-      strcat(aux, clover.Twisted() ? ",twist=true" : ",twisted=false");
+      strcat(aux, clover.TwistFlavor() == QUDA_TWIST_SINGLET || clover.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ?
+             ",twist=true" : ",twisted=false");
```
weinbe2: Not relevant for correctness, but `twist` vs `twisted` should be reconciled

lib/dirac_twisted_clover.cpp
```diff
@@ -71,28 +75,46 @@ namespace quda {
...
+  // apply full operator
   void DiracTwistedClover::M(ColorSpinorField &out, const ColorSpinorField &in) const
   {
     checkFullSpinor(out, in);
     if (in.TwistFlavor() != out.TwistFlavor())
       errorQuda("Twist flavors %d %d don't match", in.TwistFlavor(), out.TwistFlavor());
 
     if (in.TwistFlavor() == QUDA_TWIST_NO || in.TwistFlavor() == QUDA_TWIST_INVALID) {
-      errorQuda("Twist flavor not set %d", in.TwistFlavor());
+      errorQuda("Twist flavor not set %d\n", in.TwistFlavor());
```
weinbe2: Added `\n` isn't necessary b/c `errorQuda` adds a newline on its own

lib/dirac_twisted_clover.cpp
```diff
@@ -169,6 +191,25 @@ namespace quda {
     twistedCloverApply(out, in, QUDA_TWIST_GAMMA5_INVERSE, parity);
   }
 
+  void DiracTwistedCloverPC::WilsonDslash(ColorSpinorField &out, const ColorSpinorField &in, QudaParity parity) const
+  {
+    if (in.TwistFlavor() == QUDA_TWIST_SINGLET) {
+      DiracWilson::DslashXpay(out, in, parity, in, 0.0);
+    } else {
+      ApplyDomainWall4D(out, in, *gauge, 0.0, 0.0, nullptr, nullptr, in, parity, dagger, commDim, profile);
```
weinbe2, line 200: I'd appreciate a brief comment on why this is being used, just something simple like "Reusing domain wall machinery to apply WilsonDslash to both chiralities" or w/e is most clear

include/clover_field_order.h
```diff
@@ -585,8 +587,9 @@ namespace quda {
         const AllocInt offset; // offset can be 32-bit or 64-bit
         const int volumeCB;
 
-	const bool twisted;
-	const real mu2;
+        const QudaTwistFlavorType twist_flavor;
+        const real mu2;
```
weinbe2, line 591: All of these parameters --- `mu2`, `epsilon2`, `rho`, as well as `a`, `b`, `c` further below --- have gotten confusing. Can these please be documented on the wiki in some fashion?

weinbe2:
All of these parameters --- `mu2`, `epsilon2`, `rho`, as well as `a`, `b`, `c` --- have gotten confusing. Some of them get documented explicitly in the code (in particular the roman letters get covered in ` include/kernels/dslash_ndeg_twisted_clover.cuh`), but it's scattered. Can these please be documented on the wiki in some fashion? It doesn't need to block this PR, but we can at least file an issue about it.

lib/dirac.cpp
```diff
@@ -210,18 +210,16 @@ namespace quda {
...
       if (getVerbosity() >= QUDA_DEBUG_VERBOSE) printfQuda("Creating a DiracTwistedClover operator (%d flavor(s))\n", param.Ls);
-      if (param.Ls == 1) {
-	return new DiracTwistedClover(param, 4);
-      } else { 
-	errorQuda("Cannot create DiracTwistedClover operator for %d flavors\n", param.Ls);
-      }
+      if (param.Ls == 1)
+        return new DiracTwistedClover(param, 4);
+      else
+        return new DiracTwistedClover(param, 5);
```
weinbe2: Let's reintroduce a `errorQuda` path for `param.Ls != 1` and `2`, just to be safe.

lib/dslash_ndeg_twisted_clover_preconditioned.cu
```diff
@@ -0,0 +1,133 @@
...
+        return 2 * in.Ncolor() * 2 * sizeof(typename mapper<typename Arg::Float>::type);
+      }
+      
+    public:
+    NdegTwistedCloverPreconditioned(Arg &arg, const ColorSpinorField &out,
+                                    const ColorSpinorField &in) :
+      Dslash(arg, out, in)
+      {
+        TunableKernel3D::resizeVector(2, arg.nParity);
+        // this will force flavor to be contained in the block
```
weinbe2, line 34: Is there a good way to enforce this with an `errorQuda` somewhere---perhaps by providing a custom tuning advance routine that just makes sure the `y` blocksize is always 2?

lib/gauge_force.cu
```diff
@@ -47,20 +51,35 @@ namespace quda {
...
 
-    // create path struct in a single allocation
-    size_t bytes = 4 * num_paths * path_max_length * sizeof(int) + num_paths*sizeof(int) + num_paths*sizeof(double);
-    void *buffer = pool_device_malloc(bytes);
-    paths p(buffer, bytes, input_path, length_h, path_coeff_h, num_paths, path_max_length);
+    paths p(input_path, length_h, path_coeff_h, num_paths, path_max_length);
+
+    // gauge field must be passed as first argument so we peel off its reconstruct type
+    instantiate<GaugeForce_,ReconstructNo12>(u, mom, epsilon, p);
+    p.free();
```
weinbe2, line 58: Can the `free` be enforced in the destructor?

lib/interface_quda.cpp
```diff
@@ -4148,6 +4152,88 @@ int computeGaugeForceQuda(void* mom, void* siteLink,  int*** input_path_buf, int
...
+  profileGaugePath.TPSTOP(QUDA_PROFILE_INIT);
+  if (!qudaGaugeParam->overwrite_gauge) {
+    profileGaugePath.TPSTART(QUDA_PROFILE_H2D);
+    cudaOut->loadCPUField(*cpuOut);
+    profileGaugePath.TPSTOP(QUDA_PROFILE_H2D);
+  }
+
+  cudaGaugeField *cudaGauge = createExtendedGauge(*cudaSiteLink, R, profileGaugePath);
+  // apply / remove phase as appropriate
+  if (cudaGauge->StaggeredPhaseApplied()) cudaGauge->removeStaggeredPhase();
```
weinbe2, line 4205: We should add a `bool` tracking if the phases were removed so they can be reapplied at the end of the function

tests/host_reference/clover_reference.cpp
```diff
@@ -242,6 +242,66 @@ void twistCloverGamma5(void *out, void *in, void *clover, void *cInv, const int
   host_free(tmp1);
 }
 
+// Apply (A + i*mu*gamma_5*tau3 - epsilon*tau1) for QUDA_TWIST_GAMMA5_DIRECT
+// and   (A - i*mu*gamma_5*tau3 + epsilon*tau1)/(A^2 + mu^2 - epsilon^2) for QUDA_TWIST_GAMMA5_INVERSE
+void ndegTwistCloverGamma5(void *out1, void *out2, void *in1, void *in2, void *clover, void *cInv, const int dagger,
+                           const double kappa, const double mu, const double epsilon, const int parity,
+                           QudaTwistGamma5Type twist, QudaPrecision precision)
+{
+  void *tmp1 = malloc(Vh * spinor_site_size * precision);
```
weinbe2: `safe_malloc`

tests/host_reference/clover_reference.cpp
```diff
@@ -351,6 +411,207 @@ void tmc_matpc(void *out, void **gauge, void *in, void *clover, void *cInv, doub
...
+  void *inOdd1 = (char *)inEven2 + precision * Vh * spinor_site_size;
+  void *inOdd2 = (char *)inOdd1 + precision * Vh * spinor_site_size;
+
+  void *outEven1 = out;
+  void *outEven2 = (char *)outEven1 + precision * Vh * spinor_site_size;
+
+  void *outOdd1 = (char *)outEven2 + precision * Vh * spinor_site_size;
+  void *outOdd2 = (char *)outOdd1 + precision * Vh * spinor_site_size;
+
+  void *tmpEven1 = malloc(Vh * spinor_site_size * precision);
```
weinbe2: `safe_malloc`

tests/host_reference/clover_reference.cpp
```diff
@@ -242,6 +242,66 @@ void twistCloverGamma5(void *out, void *in, void *clover, void *cInv, const int
...
+    apply_clover(out2, cInv, tmptmp2, parity, precision);
+
+    free(tmptmp1);
+    free(tmptmp2);
+  } else {
+    printf("Twist type %d not defined\n", twist);
+    exit(0);
+  }
+
+  free(tmp2);
```
weinbe2: `host_free`

tests/host_reference/clover_reference.cpp
```diff
@@ -351,6 +411,207 @@ void tmc_matpc(void *out, void **gauge, void *in, void *clover, void *cInv, doub
...
+                        QUDA_TWIST_GAMMA5_DIRECT, precision);
+
+  // combine
+  xpay(tmpOdd1, -kappa, outOdd1, Vh * spinor_site_size, precision);
+  xpay(tmpOdd2, -kappa, outOdd2, Vh * spinor_site_size, precision);
+
+  xpay(tmpEven1, -kappa, outEven1, Vh * spinor_site_size, precision);
+  xpay(tmpEven2, -kappa, outEven2, Vh * spinor_site_size, precision);
+
+  free(tmpOdd1);
```
weinbe2: `host_free`

tests/host_reference/clover_reference.cpp
```diff
@@ -351,6 +411,207 @@ void tmc_matpc(void *out, void **gauge, void *in, void *clover, void *cInv, doub
...
+                     QudaGaugeParam &gauge_param)
+{
+  // V-4d volume and Vh=V/2, see tests/utils/host_utils.cpp -> setDims()
+  void *in1 = in;
+  void *in2 = (char *)in1 + precision * Vh * spinor_site_size;
+
+  void *out1 = out;
+  void *out2 = (char *)out1 + precision * Vh * spinor_site_size;
+
+  void *tmp1 = malloc(Vh * spinor_site_size * precision);
```
weinbe2: `safe_malloc`

tests/host_reference/clover_reference.cpp
```diff
@@ -351,6 +411,207 @@ void tmc_matpc(void *out, void **gauge, void *in, void *clover, void *cInv, doub
...
+      wil_dslash(out1, gauge, tmp1, oddBit, daggerBit, precision, gauge_param);
+      wil_dslash(out2, gauge, tmp2, oddBit, daggerBit, precision, gauge_param);
+    }
+  } else {
+    wil_dslash(tmp1, gauge, in1, oddBit, daggerBit, precision, gauge_param);
+    wil_dslash(tmp2, gauge, in2, oddBit, daggerBit, precision, gauge_param);
+    ndegTwistCloverGamma5(out1, out2, tmp1, tmp2, clover, cInv, daggerBit, kappa, mu, epsilon, oddBit,
+                          QUDA_TWIST_GAMMA5_INVERSE, precision);
+  }
+  free(tmp1);
```
weinbe2: `host_free` (continue further down in the file)

lib/gauge_force.cu
maddyscientist: No, otherwise the copy made by the driver that is made when calling the kernel would free the buffer when it goes out of scope.

lib/dslash_ndeg_twisted_clover_preconditioned.cu
maddyscientist: I'm not sure entirely what is meant here.  The `resizeStep` method was introduced to achieve exactly this, e.g., enforce the block size is advanced in united of the step size = 2, and is what is used in other operators that need this locality (non-deg twisted mass, domain-wall M5, etc).  I don't really see a need for a custom routine when the general one does fine.

I'm not adverse to further customized additions though, but I'm not entirely convinced of the need for it here 😜

tests/host_reference/clover_reference.cpp
maddyscientist: bfa3da9516449da73099c1ad2d9fe869d0d4ea7e

tests/host_reference/clover_reference.cpp
maddyscientist: bfa3da9516449da73099c1ad2d9fe869d0d4ea7e

tests/host_reference/clover_reference.cpp
maddyscientist: bfa3da9516449da73099c1ad2d9fe869d0d4ea7e

tests/host_reference/clover_reference.cpp
maddyscientist: bfa3da9516449da73099c1ad2d9fe869d0d4ea7e

tests/host_reference/clover_reference.cpp
maddyscientist: bfa3da9516449da73099c1ad2d9fe869d0d4ea7e

tests/host_reference/clover_reference.cpp
maddyscientist: bfa3da9516449da73099c1ad2d9fe869d0d4ea7e

lib/dirac.cpp
maddyscientist: e0185c7fefda80fac0284473e25ee4d9a010b7d9

lib/dirac_twisted_clover.cpp
maddyscientist: d83e46f5ca2fdc4fec6cba4e396d3ef11e8e6917

lib/dirac_twisted_clover.cpp
maddyscientist: d83e46f5ca2fdc4fec6cba4e396d3ef11e8e6917

lib/clover_invert.cu
maddyscientist: d83e46f5ca2fdc4fec6cba4e396d3ef11e8e6917

kostrzewa:
@maddyscientist @weinbe2 
Many thanks for the continued work on this. Last week I got around to testing the efficiency improvements for dynamic clover and in particular the chiral<->flavour swizzling for the preconditioned nd-operator, for which I see an improvement of about 10% in the relevant parts in the HMC when using full double precision multi-shift, as in our rational approx correction term. The same is true in the parts where I employ multi-shift CG in single precision with mixed-CG refinement for the 3-4 partial fraction terms in our MD action. Awesome!

I noticed that `sp_pad` and `cl_pad` were removed in 50864ffde1bd8f46fd4a2a2b2e6d44a5a588e2c2. This is something which breaks our interface in terms of backwards-compatibility, but I'll add a work-around. Not having to deal with these is great! (we always set them to zero anyway)

Just now testing the latest changes. Hopefully I can report today but I'm not sure if I will manage before the waterfall of meetings this afternoon and tomorrow...

kostrzewa:
> Last week I got around to testing the efficiency improvements for dynamic clover and in particular the chiral<->flavour swizzling for the preconditioned nd-operator, for which I see an improvement of about 10% in the relevant parts in the HMC when using full double precision multi-shift, as in our rational approx correction term. The same is true in the parts where I employ multi-shift CG in single precision with mixed-CG refinement for the 3-4 partial fraction terms in our MD action. Awesome!

Actually, sorry I got this wrong. The speedup was due to the stack usage fix (and clover reconstruct?). Performance between 5faa4fbb44d001e8bdebdc6e032708a8bf7d153b and now (on quad-A100 single node) has not changed much.

lib/interface_quda.cpp
weinbe2: Discussed offline, didn't realize this was some copy+pasted boilerplate, this can be saved for another day because it doesn't seem to be an issue -> out of scope for this PR

lib/dslash_ndeg_twisted_clover_preconditioned.cu
weinbe2: Ah, fair enough, I hadn't noticed there was precedence for that approach (and it was literally the intended design)---I'm a bit paranoid about corner-cases, as you can probably tell. We can discuss it outside of the scope of this PR.

weinbe2:
My visual review is all done, comments addressed, so I'm content there. I'm doing some cursory checks, and assuming there are no surprises, and assuming all looks good to @kostrzewa after his last checks, we can get this merged.

maddyscientist:
@kostrzewa great to hear of the speedup (whatever the source is).  The improvement in the preconditioned operator performance (855ec7a) would only really be seen in half precision, since the other precisions weren't really compute limited.

And thank you for catching the `sp_pad` and `cl_pad` removal.  These parameters haven't been useful for 10 years (since the Fermi architecture), so it was good to nuke this.

Once you give final approval we should be good to merge this 🍾

kostrzewa:
All good from my side, thanks a lot for all the effort. All residuals in the HMC come out correct, so I'm a happy camper ;)

kostrzewa:
>  The improvement in the preconditioned operator performance (855ec7a) would only really be seen in half precision, since the other precisions weren't really compute limited.

That makes sense. This means that I should get around to exposing the refinement precision so we can make best use of this (using double-half to refine the single-prec multishift solves). Thanks!

weinbe2 (approved):
I did some hands on testing of both the ndeg twisted clover, as well as sanity checking the gauge force (since the gauge path modifications technically perturb that code), and all looks well. Awesome contribution @kostrzewa + optimizations @maddyscientist

PR #1199, opened 2021-10-27
Remove QUDA's internal copy of cub

maddyscientist:
Removes the internal copy of cub, and we will now rely on the toolkit version of cub.


maddyscientist:
Closing this PR, since we can't remove `cub` until we drop support for CUDA 10.x

PR #1200, opened 2021-10-28
Hotfix/cub legacy

maddyscientist:
Use the toolkit version of cub from CUDA 11.0 onwards, else use the legacy variant included with QUDA.  This is a stepping stone to removing CUB completely from QUDA.


mathiaswagner (approved):
Looks good.

PR #1202, opened 2021-10-29
Adds Zenodo DOI badge to readme on main page

mathiaswagner:


weinbe2:
Looking at the Zenodo link: I see some names are full names while others are usernames, etc, for example @maddyscientist is literally "maddyscientist". I'm guessing this is due to some automatic scraping, is there a way to change it by hand?

mathiaswagner:
Good point. I would need to check. It is probably scraped from the GitHub porfiles.

There is https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files but I have not managed to give that a closer look.

maddyscientist (approved):

PR #1204, opened 2021-10-30
Cleanup/gf format fixes

cpviolator:
This PR adds CL params for the gauge fixing routines so that one can control the parameters. The original gauge alg test is now a ctest with the Google testing suite, and the file gauge_alg_test.cpp performs only a single gauge fix test specified by the user, with the option to save the gauge field.

It also consolidates the gauge fixing routines using a QudaGaugeFixParam struct to define the workflow.

This PR escapes the formatting issues in https://github.com/lattice/quda/tree/cleanup/gf_test_interface and supersedes it.


maddyscientist:
Thanks @cpviolator.  This looks like a great cleanup of the gauge fixing interface.

A few outstanding issues:
* Can you revert the name of the gauge_alg_test: this name change serve little purpose other than to break prior scripts
* The test doesn't actually utilize the new interface, and is calling the underlying functions directly.  Can you update the test code to actually use the interface?
* I think `setVerbosity` probably should be moved to just before `initQuda` in `gauge_alg_test` to ensure that we have visibility of anything occurring in the library set up.

cpviolator:
The gauge fixing interface does two things:

1. `--test 0` will perform Google testing. Both algorithms (OVR/FFT) both Landau/Coulomb are tested with the given CL parameters. The GPU functions are used for this route and the field is on the GPU.
2. `--test 1` will perform a specific test. In this case, the gauge field is evolved on the device and copied to the CPU. It is then passed to the QUDA gauge fixing interface function and the desired gauge fixing is performed. This instructs the user how to use a CPU gauge field and tests the QUDA interface function.

In both cases, the fixed gauge field observables are compared to the unfixed values.

cpviolator:
I've addressed online and offline comments and suggestions. Both the GPU kernel and interface functions now use a new parameter struct that wholly defines the gauge fixing. The Google Test Suite tests the GPU kernel function calls directly, and the individual test uses the `interface_quda.cpp` function, which is now accepts CPU gauges only.

I also introduced a new CL parameter `--seed` so that seeds can be passed to the test suite, as well as a few extra sanity checks in the GF test. The is now ready for another look @maddyscientist

cpviolator:
@maddyscientist Actually, hold off for a bit. Some precisions are not being set correctly.

cpviolator:
Some test commands:

To test an individual test type (FFT, Landau) in single prec, save the gauge, then load the fixed gauge and compute in double.
```
./gauge_alg_test --dim 8 8 8 8 --test 1 --prec single --gf-tol 1e-6 --gf-reunit-interval 1000 --gf-fft-autotune true --verbosity verbose --seed 8526 --gf-fix-type fft --save-gauge test_gauge

./gauge_alg_test --dim 8 8 8 8 --test 1 --prec double --gf-tol 1e-12 --gf-reunit-interval 1000 --gf-fft-autotune true --verbosity verbose --seed 8526 --gf-fix-type fft --load-gauge test_gauge
```

To perform the Google test.
```
./gauge_alg_test --dim 8 8 8 8 --test 0 --prec single --gf-tol 1e-6 --gf-reunit-interval 1000 --gf-fft-autotune true --verbosity verbose --seed 8526
```

@maddyscientist The above are working well, which shows that the Google test for GPU functions, spot test for `interface_quda.cpp` function, and IO functionality are working as expected.

PR #1205, opened 2021-10-31
Suppress warnings for non DWF builds

cpviolator:
This PR adds function definitions with blank signatures to the MDWF files to suppress `unused-variable` warnings when not building with MDWF support.


hummingtree (approved):
Thanks Dean for the clean up. @cpviolator Can you please do a clang-format to these files?

maddyscientist (approved):

PR #1208, opened 2021-11-05
rely on CUDA_ARCHITECTURES from cmake to specify cuda gpu architecture

mathiaswagner:
Add QUDA_GPU_ARCH_SUFFIX to allow suffix to use real and virtual architectures.
RELEASE builds will always build for `-real` architectures. 

Only building for real reduces binary size.
Building for virtual only can be useful for compilation tests but should not be used  for execution.


mathiaswagner:
I have enabled `QUDA_GPU_ARCH_SUFFIX=virtual` for the 11.4 jenkins builds.

maddyscientist (approved):
Great addition.  Thanks @mathiaswagner

PR #1211, opened 2021-11-10
Feature/nvcc nvrtc atomic

maddyscientist:
Cleanup and enablement PR
* Fix nvrtc compilation (must not use `static` shared memory)
* Enable heterogenous atomics for nvc++, clang, and nvrtc (Jitify)
* Remove legacy cub
* Some cleanup and warnings fixes needed for these compilers


weinbe2:
This PR seems pretty uncontroversial but I don't really have a stake in the changes. My only open question is if this will have any issues with CUDA 11.0, since that's what's officially supported on Summit (and newer flavors of CUDA 11.* have various issues with GDR, etc). Did that get tested?

maddyscientist:
Discussed offline with @weinbe2, but addressing here for posterity:
* No effect with CUDA 11.0: libcu++ was shipped starting with CUDA 10.2, and we've always enabled heterogenous atomics by default on CUDA 11.0+.
* Similarly, no effect on Summit due to the removal of the legacy CUB, since we have utilized the toolkit version of CUB by default for CUDA 11.0+.

tests/utils/command_line_params.cpp
```diff
@@ -145,7 +145,7 @@ QudaTransferType staggered_transfer_type = QUDA_TRANSFER_COARSE_KD;
 // we only actually support 4 here currently
 quda::mgarray<std::array<int, 4>> geo_block_size = {};
 
-#if (CUDA_VERSION >= 10010 && __COMPUTE_CAPABILITY__ >= 700)
+#ifdef QUDA_MMA_AVAILABLE
```
hummingtree, line 148: Just a thought: we are removing `CUDA_VERSION >= 10010` as we now require CUDA 11.0, should we also remove it in the definition of `QUDA_MMA_AVAILABLE`?

hummingtree (approved):
This is not controversial and it is a nice clean up.

tests/utils/command_line_params.cpp
weinbe2: IIRC we still technically support Pascal? I've lost track, tbh.

mathiaswagner: Yes, but can and should remove the CUDA_VERSION here.

maddyscientist: 7c0a6bd

weinbe2 (approved):
Verified this compiles with CUDA 11.0 on Summit, we're all good.

tests/utils/command_line_params.cpp
mathiaswagner: What happens on sm60 here?

mathiaswagner (approved):

PR #1213, opened 2021-11-11
Feature/clover reconstruct

maddyscientist:
Feature-add branch
* Add clover-reconstruct CMake option `QUDA_CLOVER_RECONSTRUCT` to QUDA (enabled by default) which compresses the clover field when using the native field order.  This is only supported when `QUDA_CLOVER_DYNAMIC_INVERSE=ON`
* This reduces the memory footprint and improves performance for all kernels that utilize clover fields
* Clean up of handling when dynamic-inverse is enabled to remove hacks in the interface
* Dynamic-clover now utilizes double precision for the Cholesky for double and single precision fields, and single precision for half and quarter precision fields.  This improves solver stability.
* Unified `cpuCloverField` and `cudaCloverField`: we now only have `CloverField`
* Add support for `FLOAT8` ordering for half and quarter precision clover fields
* Utilize global norm scale for fixed-point clover field instead of per-site normalization
* Fixes to tunable_reduction.h with respect to correct setting of the `aux` key
* Remove legacy `writeAuxString` method

PR #1217, opened 2021-11-18
Feature/color spinor unity

maddyscientist:
This PR addresses a long-standing concern in QUDA, namely subsuming the `cpuColorSpinorField` and `cudaColorSpinorField` classes into the parent `ColorSpinorField` class.  As a result, `ColorSpinorField` is no longer a vritual class and can be instantiated directly.  This PR depends on #1213, so will not be merged until that is merged.
* `ColorSpinorField` is no longer virtual, and it now used for all quark field allocation regardless of location
* Removed legacy `cudaColorSpinorField` and `cpuColorSpinorField` classes
* Significant cleanup and removal of hacks implemented due to prior class structure
* Fixes a bug in `dslash_test_util.h` regarding correct Dirac operator instantiation when using non-preconditioned Mobius and DWF-4d operators

Every Dirac operator has been tested, both single and multi-GPU builds.  In addition I have also tested multi-GPU multi-shift and multi-GPU multi-grid solvers.

There is much left to cleanup left to be done after this PR.  Namely:
* Unification of communication handling within `ColorSpinorField`.  These are presently disparate paths, and intentionally left untouched here.
* Solver cleanup: one of the goals of this work is allow us to utilize heap-based allocation of vector objects, specifically in the solvers.  This will be deferred to a subsequent follow-up PR.


include/dslash_helper.cuh
```diff
@@ -352,8 +352,7 @@ namespace quda
 
       if (in.Location() == QUDA_CUDA_FIELD_LOCATION) {
         // create comms buffers - need to do this before we grab the dslash constants
-        ColorSpinorField *in_ = const_cast<ColorSpinorField *>(&in);
-        static_cast<cudaColorSpinorField *>(in_)->createComms(nFace, spin_project);
+        const_cast<ColorSpinorField &>(in).createComms(nFace, spin_project);
```
mathiaswagner, line 355: would it be worth thinking about `createComms(...) const` and using mutable instead of having to `const_cast` (I know this is unrelated to the spirit of this PR)

maddyscientist: Completely agree.  I had thought about fixing this aspect, but then we decided this PR won't touch the comms, I figured to leave this.

mathiaswagner:
We should try to move to stack allocate most ColorSpinorFields and consider smart ptrs where not possible. But I am fine with doing that in a follow-up PR to keep this one simple and clean.

lib/eigensolve_quda.cpp
```diff
@@ -337,11 +337,8 @@ namespace quda
     // C_1 is the current 'out' vector.
 
     // Clone 'in' to two temporary vectors.
-    ColorSpinorField *tmp1 = ColorSpinorField::Create(in);
-    ColorSpinorField *tmp2 = ColorSpinorField::Create(in);
-
-    blas::copy(*tmp1, in);
-    blas::copy(*tmp2, out);
+    auto tmp1 = std::make_unique<ColorSpinorField>(in);
```
mathiaswagner, line 340: stack?

lib/color_spinor_field.cpp
```diff
@@ -27,20 +27,40 @@ namespace quda {
     dslash_constant(static_cast<DslashConstant *>(safe_malloc(sizeof(DslashConstant)))),
     bytes(0),
     norm_bytes(0),
+    location(param.location),
     even(0),
     odd(0),
     composite_descr(param.is_composite, param.composite_dim, param.is_component, param.component_id),
     components(0)
   {
+    if (param.location == QUDA_INVALID_FIELD_LOCATION) errorQuda("Location not set");
```
mathiaswagner, line 35: might  be worth to do a full clang-format on this file given how much has changed

lib/color_spinor_field.cpp
```diff
@@ -50,19 +70,291 @@ namespace quda {
...
+          pool_device_free(v);
+          if (norm_bytes) pool_device_free(norm);
+          break;
+        case QUDA_MEMORY_MAPPED:
+          host_free(v_h);
+          if (norm_bytes) host_free(norm_h);
+          break;
+        default: errorQuda("Unsupported memory type %d", mem_type);
+        }
+      }
```
mathiaswagner, line 252: should is also set alloc=false here?

lib/color_spinor_util.cu
```diff
@@ -443,9 +446,9 @@ else if (a.Ncolor() == 96 && a.Nspin() == 2) {
   // the above host print function.
 
   template <typename StoreType, int Ns, int Nc, QudaFieldOrder FieldOrder>
-  void genericCudaPrintVector(const cudaColorSpinorField &field, unsigned int i)
+  void genericCudaPrintVector(const ColorSpinorField &field, unsigned int i)
   {
-
+#warning "fix or delete"
```
mathiaswagner: commenting here so we don't forget

lib/color_spinor_util.cu
```diff
@@ -352,190 +354,85 @@ namespace quda {
...
   {
+    if (rank != comm_rank()) return;
 
-    typedef colorspinor::AccessorCB<StoreType, Ns, Nc, 1, FieldOrder> AccessorType;
+    ColorSpinorParam param(a);
+    param.location = QUDA_CPU_FIELD_LOCATION;
+    param.create = QUDA_COPY_FIELD_CREATE;
+    // if field is a pinned device field then we need to clone it on the host
+    bool host_clone = (a.Location() == QUDA_CUDA_FIELD_LOCATION && a.MemType() == QUDA_MEMORY_DEVICE && !use_managed_memory()) ? true : false;
+    const ColorSpinorField *a_ = !host_clone ? &a : new ColorSpinorField(param);
```
mathiaswagner: no more new ;-)
```
std::unique_ptr<ColorSpinorField> clone_a = !host_clone ? nullptr : std::make_unique<ColorSpinorField>(param);
const ColorSpinorField *a_ = !host_clone ? &a : clone_a.get();
```
and delete the delete.

include/color_spinor_field.h
```diff
@@ -736,23 +501,8 @@ namespace quda {
...
-
-    /**
-      Initiate the cpu to gpu send of the ghost zone (halo)
-      @param ghost_spinor Source of the ghost zone
-      @param dim The lattice dimension we are sending
-      @param dir The direction (QUDA_BACKWARDS or QUDA_FORWARDS)
-      @param stream The array of streams to use
-      */
-    void unpackGhost(const void *ghost_spinor, const int dim, const QudaDirection dir, const qudaStream_t &stream);
+    // fuse with above
```
mathiaswagner, line 496: should this be taken care of in this PR?

lib/color_spinor_field.cpp
```diff
@@ -876,19 +882,697 @@ namespace quda {
...
+                                  MemoryLocation location[2 * QUDA_MAX_DIM], MemoryLocation location_label,
+                                  bool spin_project, double a, double b, double c, int shmem)
+  {
+    if (Location() == QUDA_CPU_FIELD_LOCATION) errorQuda("Host field not supported");
+    createComms(nFace, spin_project); // must call this first
+
+    packGhost(nFace, (QudaParity)parity, dagger, stream, location, location_label, spin_project, a, b, c, shmem);
+  }
+
+  // FIXME reconcile with above
```
mathiaswagner, line 1039: for this PR?

mathiaswagner (changes requested):
I generally like that and it is a great cleanup. Some comments should still be addressed.

Also we should clang-format the full file for color_spinor_field now that we touched so much of it.

And final comment: I am still seeing some issues with mobius in ctest. Not sure they are new.

lib/color_spinor_util.cu
maddyscientist: ad913bb deals with this

lib/color_spinor_field.cpp
maddyscientist: Will do that when I apply clang-format to the PR.  That'll make @weinbe2 happy 😉 .

lib/eigensolve_quda.cpp
maddyscientist: I tried this initially, but the present of the `std::swap` makes this less trivial.  Since the point of this PR is not to clean up the solvers completely, I figured this could be deferred.

include/color_spinor_field.h
maddyscientist: I consider the packing aspects as part of the comms cleanup, which as we've discussed, is deferred for a follow up PR

lib/color_spinor_field.cpp
maddyscientist: I consider the packing aspects as part of the comms cleanup, which as we've discussed, is deferred for a follow up PR

lib/color_spinor_util.cu
maddyscientist: 46efcd94949f10364106d0f3058e977a707aa8ef

lib/color_spinor_field.cpp
maddyscientist: 4a4dab7df8d4b1fede2f7a50b314181c8433e4c5

maddyscientist:
I have address all the outstanding comments so we should be good here now.

mathiaswagner (approved):

include/color_spinor_field_order.h
```diff
@@ -264,11 +275,10 @@ namespace quda {
       const int stride;
       const int offset_cb;
       AccessorCB(const ColorSpinorField &field) :
-        stride(field.Stride()),
-        offset_cb((field.Bytes() >> 1) / sizeof(complex<Float>))
+        stride(field.VolumeCB()), offset_cb((field.Bytes() >> 1) / sizeof(complex<Float>))
```
weinbe2, line 278: Should we change the variable name `stride` to `volume_cb`?

include/color_spinor_field_order.h
```diff
@@ -668,17 +684,16 @@ namespace quda {
 #endif
 
         if (block_float) {
-          // only if we have block_float format do we set these (only block_orthogonalize.cu at present)
           norm = static_cast<norm_t *>(const_cast<void *>(field.Norm()));
-          norm_offset = field.NormBytes() / (2 * sizeof(norm_t));
+          norm_offset = field.Bytes() / (2 * sizeof(norm_t));
```
weinbe2, line 688: Should this be divided by `2 * sizeof(ghostFloat)`? I feel like I've seen this pattern elsewhere, too.

include/color_spinor_field_order.h
```diff
@@ -951,20 +972,19 @@ namespace quda {
...
-      FloatNOrder(const ColorSpinorField &a, int nFace = 1, Float *field_ = 0, norm_type *norm_ = 0, Float **ghost_ = 0) :
-        field(field_ ? field_ : (Float *)a.V()),
-        norm(norm_ ? norm_ : (norm_type *)a.Norm()),
+      FloatNOrder(const ColorSpinorField &a, int nFace = 1, Float *buffer = 0, Float **ghost_ = 0) :
+        field(buffer ? buffer : (Float *)a.V()),
+        norm(buffer ? reinterpret_cast<norm_type *>(reinterpret_cast<char *>(buffer) + a.NormOffset()) :
+                      const_cast<norm_type *>(reinterpret_cast<const norm_type *>(a.Norm()))),
         offset(a.Bytes() / (2 * sizeof(Float) * N)),
-        norm_offset(a.NormBytes() / (2 * sizeof(norm_type))),
+        norm_offset(a.Bytes() / (2 * sizeof(norm_type))),
```
weinbe2, line 986: Similar to above, should this be divided by `2 * sizeof(Float))` instead of `2 * sizeof(norm_type)`?

weinbe2: Nevermind, misunderstood the purpose of the variable. All good here.

include/color_spinor_field_order.h
weinbe2: Nevermind, misunderstood the purpose of the variable. All good here.

include/color_spinor_field_order.h
maddyscientist: We could do.  I left the `stride` variable there, since that's what it is at this point and the physical meaning isn't so important.  Aside: we'll likely address these variables again when we address large volume / 64-bit indexing.

tests/blas_test.cpp
```diff
@@ -317,29 +318,29 @@ void initFields(prec_pair_t prec_pair)
...
-  delete xoD;
-  delete yoD;
-  delete zoD;
-  delete xmD;
-  delete ymD;
-  delete zmD;
-  delete xmoD;
-  delete ymoD;
-  delete zmoD;
+  vD.reset();
```
weinbe2, line 321: Is there a particular reason we're explicitly calling `reset()` as opposed to letting `unique_ptr` "do its thing"?

include/color_spinor_field_order.h
weinbe2: Fair enough---I think I saw `volume_cb` some other places, so this was also a comment on consistency, but I'm fine punting it until large volume / 64-bit indexing.

weinbe2:
I've completed my visual review and everything looks good, insofar as this PR does what's promised (which I'm a big fan of). I'll get hands on with some compilation tests, and once that's done (and assuming all is ship-shape) I'll give this an approval.

tests/blas_test.cpp
mathiaswagner: Yes. `unique_ptr` does not go out of scope before `endQuda` and that will complain about memory leaks. 
So for now we still need to reset this to avoid the warning.

weinbe2 (approved):
After doing some hands-on tests all looks good to me! Awesome PR @maddyscientist

hummingtree (approved):

PR #1218, opened 2021-11-19
rely on findpackage for QDPJIT now that it uses cmake

mathiaswagner:


mathiaswagner:
@bjoo I have tested this with the recent development version of QDPJIT and Chroma. All looks good.
Do you mind having a look at this?

mathiaswagner:
ping

lib/CMakeLists.txt
```diff
@@ -480,18 +480,9 @@ if(QUDA_QIO)
...
-              ${QDP_LIB}
-              ${QDP_LIBS}
-              ${QIO_LIB}
-              ${LIME_LIB}
-              ${QUDA_QMP_LDFLAGS}
-              ${QMP_LIB}
-              MPI::MPI_CXX)
+   target_compile_definitions(quda PRIVATE USE_QDPJIT)
+   # this avoids issues when qdpjit is build as a static library we should eventually switch to target_link_libraries(quda PRIVATE QDPXX::qdp)	
+   target_include_directories(quda PRIVATE $<TARGET_PROPERTY:QDPXX::qdp,INTERFACE_INCLUDE_DIRECTORIES>)
```
bjoo, line 485: Why not just:

target_link_libraries(quda PUBLIC QDPXX::qdp)
?

mathiaswagner: I used that initially but when build qdpjit as a static library cmake still passed that to the linker command for libquda.so and that will fail as mentioned in the comment.
Using this workaround is fine as we never link any quda tests with qdpjit enabled and the link step in Chroma gets everything resolved.
Not sure this is an issue cmake should be able to handle better or there is a better solution for this.

maddyscientist: We should not preclude being able to build the tests with qdpxx enabled.  Adding support for doing so is highly desirable for testing, and it’s something Balint and I discussed as getting fixed.

mathiaswagner: We can fix that when we get there.

maddyscientist: If it's a simply case of adding `QDP_initialize()` and `QDP_finalize()` to the QUDA tests to enable the tests to work with `QUDA_QDPJIT` enabled, can we not add that to this PR, and resolve @bjoo's concern now?

bjoo: So ok I guess the issue is that we can't link the .a to a .so if QDP-JIT is built statically. That is a reasonable comment. However, it seems like this is a bit odd use of CMake, since now we in principle don't propagate linking transitively. The client needs to explicitly do it. That seems odd to me. Right now we are saved by the fact that the only QDP-JIT client is Chroma which over QDP-JIT links things anyway.  I guess we need to choose which ugliness we take: users mixing static/dynamic linking and failing, or not propagating dependency information in the CMake way to clients (of which there is only one and may be one forever...)

bjoo (approved):
Ultimately while Chroma is the only QDP-JIT based client the discussion is in some sense academic... So I'll approve.

lib/CMakeLists.txt
mathiaswagner: Yes, this is not _how it should be_. Will look into fixing this but for now it allows us to compile recent Chroma.

maddyscientist (approved):

lib/CMakeLists.txt
mathiaswagner: About the tests: I don't actually know what needs to be done here. I have created #1222

mathiaswagner:
Follow-up issue for the workaround #1223

PR #1220, opened 2021-11-22
Improve managed memory perf with QPPJIT's managed feature

mathiaswagner:
With this PR we let QUDA use managed memory out of QDPJIT's managed memory pool (and no longer use pinnend device memory only). Also add some prefetch commands to multi shift inverter.

This requires #1218 to be merged first.


weinbe2 (approved):
_Note: this doesn't cover the QDPJIT side of things, I don't have the relevant experience there._

The prefetching calls look good to me. I wonder if it'd be worth it to prefetch the CUDA `ColorSpinorField`s after they're "created"? I know that seems like a weird suggestion, but if the "allocated" memory is in reality coming out of a pre-allocated managed memory pool it might be worth making sure that's also device resident going into the solve.

mathiaswagner:
@weinbe2 like adding a prefetch after we allocate the memory in create?

weinbe2:
> @weinbe2 like adding a prefetch after we allocate the memory in create?

No, just calling the `ColorSpinorField` `prefetch` routine (which maybe needs to be implemented quickly). Or is that what you meant by your question?

mathiaswagner:
`device_malloc_` includes a prefetch so wouldn't that prefetch the allocation to the GPU already? I think src and solution vector should already be GPU resident when things get called from Chroma.

weinbe2:
If I'm understanding `device_malloc_` correctly, doesn't that only get hit for a fresh `cudaMalloc`, as opposed to when an allocation is given a pre-existing chunk of memory from the pool?

In any case---in general, sure, those managed memory chunks should already be device resident. I'm coming from my experience with MILC HISQ MG where I'm oversubscribing like mad, and every bit of prefetching can help.

weinbe2:
We can certainly punt on this for now---I'm not intending this convo to hold up merging this PR. We can revisit it once I get more hands-on with a MILC workflow specifically.

maddyscientist:
@bjoo any thoughts on this PR?  @weinbe2 has already approved it, but since this is relevant for Chroma, just checking if you want to weigh in.  Thanks.

mathiaswagner:
I have tested this again with the latest fixes to QDPJIT.
@bjoo any thoughts on that before we merge?

weinbe2 (approved):
Looks great!

mathiaswagner:
I resolved the merge conflict and once builds passed this is good to merge.

maddyscientist (approved):

PR #1221, opened 2021-12-02
Optimized Kahler-Dirac operator for staggered/HISQ MG

weinbe2:
This (draft) PR adds support for applying and coarsening an optimized Kahler-Dirac operator for use in staggered and HISQ multigrid. Relative to the previous support for staggered multigrid:

- The “optimized” KD operator is now applying a staggered/HISQ stencil, then applying the rank-local Kahler-Dirac inverse
- The Kahler-Dirac field is now a `GaugeField` with a custom 16-way geometry, allowing a far more efficient implementation with coalescing “for free,” as opposed to the previous implementation as a QDP-ordered 48x48 coarse matrix.
- This new form avoids explicitly building the full Nc=24 coarse operator, in particular the “Y” field, which greatly reduces memory overheads.
- The optimized KD operator naturally supports applying the long links, as opposed to the previous formulation which truncated them.
- We now support coarsening the long links, ~apart from when an aggregation dimension is smaller than 3---in that case we drop the links just in that direction (and appropriately disable the operator verification).~ Edit: now opt-in, otherwise it errors out
- We've threaded in and verified support for preconditioning the staggered Schur operator: this requires reconstructing the full Dirac linear solve from the Schur solve when computing the error correction.
- Added (restored) the command line options for configuring heatbath gauge field generation in `heatbath_test` and `multigrid_evolve_test`

Various commands for testing staggered/HISQ MG can be extracted from here: https://github.com/lattice/quda/wiki/Validating-Large-Multigrid-Changes

A few outstanding items:
- [x] Double checking and updating the flops/bytes count for coarsening the staggered op vs coarsening the KD operator
- [x] Verifying correctness and performance of Wilson-type MG---I don’t expect any surprises given aggressive use of SFINAE.
- [x] Adding support for intentionally dropping the long links when forming the KD operator: `--mg-staggered-coarsen-type kd-optimized-drop-long` (as opposed to the default `--mg-staggered-coarsen-type kd-optimized`)
- [x] Adding support for opting into dropping the long links for aggregation dimensions smaller than 3, otherwise error out: `--mg-staggered-allow-drop-long true`
- [x] ~Ripping out the naively coarsened KD op (i.e., explicitly forming the full Nc=24 coarse operator) because, once we add support for intentionally dropping the long links, we can get something functionally equivalent~ Edit: Not functionally equivalent: the Nc=24 coarse operator enables Schur-preconditioning in terms of Kahler-Dirac blocks, which would be painful to implement for the KD operator
- [x] Verifying the accuracy of the documentation that was changed/added: https://github.com/lattice/quda/wiki/Staggered-Multigrid-Solver
- [x] Pulling out placeholders for right block jacobi preconditioning, it’s just confusing and not used in practice
- [x] Change defaults for staggered MG test routines from coarse KD to optimized KD
- [x] Expose new options in the MILC HISQ MG interface
- [x] Confirmed it works with jitify
- [x] clang-format as a last action


weinbe2:
Performance/correctness of Wilson-type Nc = 3 -> Nc = 24, 32 looks good (insofar as I poked some of the insides of `computeUV/computeVUV`, so there was a risk of regression). Solves are unchanged (as to be expected), and performance changes to `computeUV` and `computeVUV` are in the noise: mostly sub-percentile, a few up to +/- 4% but with no real systematic pattern.

weinbe2:
This is now ready for review @maddyscientist @mathiaswagner

include/dirac_quda.h
```diff
@@ -61,6 +61,7 @@ namespace quda {
     Dirac *dirac;
     bool need_bidirectional; // whether or not we need to force a bi-directional build
     bool use_mma;            // whether to use tensor cores where applicable
+    bool allow_drop_long;    // whether or not we can drop long links for small aggregate dimensions
```
maddyscientist: I wonder if a more generic name should be used here, e.g., `allow_truncation` or similar?  The `allow_drop_long` is very improved-staggered-centric, and I wonder if we'll ever need to allow something similar for other improved operators?

include/kernels/coarse_op_kernel.cuh
```diff
@@ -308,24 +319,211 @@ namespace quda {
...
+        // loading from V == AV, which has nSpin == 1
+        W.loadCS(Wacc, 0, 0, (parity+1)&1, y_cb, 0, k, j0);
+        UV[0].mma_nn(U, W);
+      } // fine color columns
+    }
+
+    real uv_max = static_cast<real>(0.0);
+#pragma unroll
+    for (int s = 0; s < uvSpin; s++) {
+      IF_CONSTEXPR (Arg::compute_max) {
```
maddyscientist: aside: I guess we can remove this macro now that we're C++17 and above now

tests/multigrid_evolve_test.cpp
```diff
@@ -102,6 +102,7 @@ int main(int argc, char **argv)
   setQudaDefaultMgTestParams();
   // command line options
   auto app = make_app();
+  add_heatbath_option_group(app);
```
maddyscientist, line 105: Make a note of this fix in the PR notes

lib/staggered_kd_reorder_xinv.cu
```diff
@@ -0,0 +1,234 @@
...
+    unsigned int minThreads() const { return arg.fineVolumeCB; }
+
+  public:
+    CalculateStaggeredGeometryReorder(Arg &arg, GaugeField &meta, const GaugeField &xInvCoarse) :
+      TunableKernel3D(meta, Arg::kdBlockSize, 2),
+      arg(arg),
+      xInvCoarse(xInvCoarse),
+      meta(meta)
+    {
+      if (meta.Location() == QUDA_CPU_FIELD_LOCATION) strcat(aux, getOmpThreadStr());
```
maddyscientist: These strcat's can be removed that label the location: this is now now done automatically in the parent tuning class.  Only thing that could be left if it is still relevant is the `MemType` check.

lib/staggered_kd_reorder_xinv.cu
```diff
@@ -0,0 +1,234 @@
...
+      strcat(aux,",computeStaggeredGeometryReorder");
+      strcat(aux, (meta.Location()==QUDA_CUDA_FIELD_LOCATION && xInvCoarse.MemType() == QUDA_MEMORY_MAPPED) ? ",GPU-mapped," :
+             meta.Location()==QUDA_CUDA_FIELD_LOCATION ? ",GPU-device," : ",CPU,");
+      if (dagger_approximation) strcat(aux, "dagger_approximation");
+    }
+
+    void apply(const qudaStream_t &stream)
+    {
+      TuneParam tp = tuneLaunch(*this, getTuning(), getVerbosity());
+
```
maddyscientist: Since you are instantiating both host and device below with a run-time if/else, this means both host and device field order are being compiled despite the order-based instantiation below.  To remedy this, you need to pass the field location as a template when the order is being instantiated and then make this into `if constexpr`.

Alternatively, since the long term plan is to use the same order on host or device, you could simply write the below as
```c++
launch<ComputeStaggeredGeometryReorder, true>(tp, stream, arg);
```
with no enclosing if statement.

lib/staggered_kd_reorder_xinv.cu
```diff
@@ -0,0 +1,234 @@
...
+     @param xInvCoarse[in] KD inverse coarse lattice field accessor
+     @param xInvFine_[out] KD inverse fine gauge in KD geometry
+     @param xInvCoarse_[in] KD inverse coarse lattice field
+     @param scale[in] Scaling factor for the reorder
+   */
+  template<typename Float, int fineColor, int coarseSpin, int coarseColor, bool dagger_approximation, typename fineXinv, typename coarseXinv>
+  void calculateStaggeredGeometryReorder(fineXinv &xInvFine, coarseXinv &xInvCoarse, GaugeField &xInvFine_, const GaugeField &xInvCoarse_, const Float scale)
+  {
+    // sanity checks
+    if (fineColor != 3)
```
maddyscientist: General comment: the more that general non-template specific code (like these run-time checks) can be moved out of templated code, the less bloat we have.

lib/staggered_kd_reorder_xinv.cu
```diff
@@ -0,0 +1,234 @@
...
+      xc_size[i] = xInvCoarse_.X()[i];
+      // check that local volumes are consistent
+      if (2 * xc_size[i] != x_size[i]) {
+        errorQuda("Inconsistent fine dimension %d and coarse KD dimension %d", x_size[i], xc_size[i]);
+      }
+    }
+    x_size[4] = xc_size[4] = 1;
+
+    // Calculate X (KD block), which is really just a permutation of the gauge fields w/in a KD block
+    using Arg = CalculateStaggeredGeometryReorderArg<Float,coarseSpin,fineColor,coarseColor,dagger_approximation,fineXinv,coarseXinv>;
```
maddyscientist: Could the `Arg` instantiation be delayed until the `apply` function here?  E.g., as is done in staggered_kd_build_xinv.cu ?

lib/dirac_improved_staggered_kd.cpp
```diff
@@ -231,19 +181,14 @@ namespace quda
     Dirac::updateFields(fat_gauge_in, nullptr, nullptr, nullptr);
     fatGauge = fat_gauge_in;
     longGauge = long_gauge_in;
-
-    // Recompute Xinv (I guess we should do that here?)
```
maddyscientist, line 235: What's the question here for?

include/dirac_quda.h
```diff
@@ -1477,14 +1512,39 @@ namespace quda {
...
   };
 
   // Kahler-Dirac preconditioned staggered
   class DiracImprovedStaggeredKD : public DiracImprovedStaggered
   {
 
   protected:
-    mutable cudaGaugeField *Xinv; /** inverse Kahler-Dirac matrix */
+    mutable GaugeField* Xinv; /** inverse Kahler-Dirac matrix */
+    const Dirac *parent_dirac; /** Parent Dirac operator, needed to determine if parent is pc'd or not */
```
maddyscientist: I think this is dangerous, to make a pointer to this Dirac operator.  While the general case will have the so-called operator always exist through an MG cycle, I think for robustness it would be better to simply make a local copy of the type of this operator, for decision making in the MG solver.

include/staggered_kd_build_xinv.h
```diff
@@ -13,17 +15,21 @@ namespace quda
...
      @brief Allocate and build the Kahler-Dirac inverse block for KD operators
      @param[in] in gauge original fine gauge field
      @param[in] in mass the mass of the original staggered operator w/out factor of 2 convention
-     @param[in] in precision of Xinv field
-     @return constructed Xinv, which needs to be deleted manually
+     @return constructed Xinv
   */
-  cudaGaugeField *AllocateAndBuildStaggeredKahlerDiracInverse(const cudaGaugeField &gauge, const double mass,
-                                                              const QudaPrecision override_prec);
+  std::unique_ptr<GaugeField> AllocateAndBuildStaggeredKahlerDiracInverse(const cudaGaugeField &gauge, const double mass);
```
maddyscientist: I assume not, but just checking that this header is not included in device code?  The presence of smart pointers will break Jitify,

include/kernels/staggered_kd_apply_xinv_kernel.cuh
```diff
@@ -1,219 +1,110 @@
...
+    static constexpr bool dagger = dagger_;
 
-    const int fineVolumeCB;
-    const int_fastdiv coarseVolumeCB;   /** Coarse grid volume */
+    F out;               // output vector field
+    const F in;          // input vector field
+    const X xInv;        // Kahler-Dirac inverse gauge field
+    int_fastdiv X0h;     // One-half of X dimension length
+    int_fastdiv dim[4];  // full lattice dimensions
+    const int volumeCB;  // checkerboarded volume
```
maddyscientist: All these variable comments should include doxygen markup

include/kernels/staggered_kd_apply_xinv_kernel.cuh
```diff
@@ -1,219 +1,110 @@
...
-
-      // Type for gauge/compute
-      using complex = complex<typename Arg::real>;
-
-      /////////////////////////////////
-      // Figure out some identifiers //
-      /////////////////////////////////
-
-      // Decompose into factors of 16.
-      const unsigned int fast_idx = tid & 0b1111;
```
maddyscientist, line 79: Nothing like human readable code 😛 .  The compiler is smart enough to decipher if written in terms of `/` and `%`, please do so 😄

maddyscientist (changes requested):
Initial visual review complete.  @weinbe2 this looks to mesh well with the GK framework, and will prove to be a great addition to QUDA's evolving MG coverage.  There's some things to be addressed, but nothing too major.

include/kernels/staggered_kd_apply_xinv_kernel.cuh
weinbe2: This is in code that I deleted, not added, so I guess already done :)

include/staggered_kd_build_xinv.h
weinbe2: Can confirm, I've run this PR through jitify (well, at least a version after I made this update) and there are no issues related to this file.

include/kernels/staggered_kd_apply_xinv_kernel.cuh
maddyscientist: 🤣

lib/dirac_improved_staggered_kd.cpp
weinbe2: That was an old thought I had and have since discarded (or at least discarded implementing here).

The idea was you could imagine rebuilding _just_ `Xinv` when you update the fields as part of a "thin" MG update, since that's relatively inexpensive compared to a full rebuild of the MG preconditioner. I ended up discarded putting that in here because `updateFields` is essentially meant to update pointers, so slipping a "real" kernel in here would be inappropriate.

I might revisit the idea of rebuilding just `Xinv` in the future, but that'd be through some other interface and not here.

include/dirac_quda.h
weinbe2: Noted---the (sloppy) motivation was to just re-use the `dirac` member of `DiracParams`, but I could just copy in the type of `dirac` instead.

include/kernels/coarse_op_kernel.cuh
weinbe2: Want me to take care of that (at least in this file)?

tests/multigrid_evolve_test.cpp
weinbe2: Done

include/kernels/coarse_op_kernel.cuh
maddyscientist: 👍

include/dirac_quda.h
weinbe2: https://github.com/lattice/quda/commit/4139ae2e3733e3f08528262b4231957df056794f

include/kernels/coarse_op_kernel.cuh
weinbe2: https://github.com/lattice/quda/commit/d04911d1db816113faa5d9a23f9e07d233cb62dd

lib/staggered_kd_reorder_xinv.cu
weinbe2: I'm not sure if I understand what you're referring to---the location of the `Arg` instantiation roughly matches between this file and `staggered_kd_build_xinv.cu`

maddyscientist: You are correct, I mixed up my filenames - I meant staggered_kd_apply_xinv.cu .  All the same comments on this file can be leveled at staggered_kd_build_xinv.cu  file as well (including my comment about double instantiation of host and device, and the `strcat`.)

What I meant here is that ideally we should be instantiating the kernel argument struct as late as possible, ideally its construction would be in the call to `launch` itself (but this isn't always possible).  It is possible in this file and in staggered_kd_build_xinv.cu though.  This helps reduce template bloat.

lib/staggered_kd_reorder_xinv.cu
weinbe2: https://github.com/lattice/quda/commit/28e91cb86b4f40d405a75eefdacb4348d03437eb

include/kernels/staggered_kd_apply_xinv_kernel.cuh
weinbe2: https://github.com/lattice/quda/commit/28e91cb86b4f40d405a75eefdacb4348d03437eb

lib/staggered_kd_reorder_xinv.cu
weinbe2: Let me know what you think of https://github.com/lattice/quda/commit/28e91cb86b4f40d405a75eefdacb4348d03437eb

include/dirac_quda.h
weinbe2: Agreed! https://github.com/lattice/quda/commit/0cb94afc8fe13b15ce3eb30a24dcdd89d2165815

lib/staggered_kd_reorder_xinv.cu
maddyscientist: This commit does address my comments, thanks.  I've made a few comments on the commit which might need some attention.

include/dirac_quda.h
maddyscientist: Great, that looks good.

lib/staggered_kd_reorder_xinv.cu
weinbe2: Mostly addressed, will be fully addressed in a subsequent commit where I refactor `staggered_kd_build_xinv.cu`

weinbe2:
@maddyscientist is this ready for a `clang-format`?

include/quda.h
```diff
@@ -732,6 +732,12 @@ extern "C" {
     /** Boolean for aggregation type, implies staggered or not */
     QudaTransferType transfer_type[QUDA_MAX_MG_LEVEL];
 
+    /** Whether or not to let MG coarsening drop improvements, for ex dropping long links in small aggregation dimensions */
```
maddyscientist, line 735: These variables need to propagated into quda_fortran.F90

include/kernels/coarse_op_kernel.cuh
```diff
@@ -199,7 +204,7 @@ namespace quda {
      @param[in] nFace Depth of the halo
      @param[in] arg Kernel argument
   */
-  template <typename Arg> constexpr bool isHalo(const int coord[], int dim, int nFace, const Arg &arg)
+  template <typename Arg> constexpr bool isHalo(const int coord[], int dim, const int nFace, const Arg &arg)
```
maddyscientist: Not sure what the point is for this `const` addition for passing an integer by value.

include/kernels/coarse_op_kernel.cuh
```diff
@@ -220,36 +225,36 @@ namespace quda {
      @param[in] dim Dimension
      @param[in] arg Kernel argument
   */
-  template <typename Arg> constexpr bool isCoarseDiagonal(const int coord[], const int coord_coarse[], int dim, const Arg &arg)
+  template <typename Arg> constexpr bool isCoarseDiagonal(const int coord[], const int coord_coarse[], int dim, const int nFace, const Arg &arg)
```
maddyscientist: Not sure what the point is for this `const` qualifier for passing an integer (`nFace`) by value.

include/kernels/coarse_op_kernel.cuh
```diff
@@ -687,6 +918,110 @@ namespace quda {
...
+            }
+
+            // Load spinor
+            Vector in;
+#pragma unroll
+            for (int jc_f = 0; jc_f < Arg::fineColor; jc_f++) {
+              in(0, jc_f) = arg.V(nbr_parity, nbr_x_cb, 0, jc_f, ic_c);
+            }
+
+            out = mv_add(arg.kd_dagger ? conj(Xinv) : Xinv, in, out);
```
maddyscientist, line 999: This expression to my eye looks like it could cause the compiler to spill.  Am I being paranoid?

include/kernels/coarse_op_kernel.cuh
```diff
@@ -876,6 +1209,93 @@ namespace quda {
...
+          auto UV = make_tile_B<complex, false>(tile);
+          UV.loadCS(arg.UV, 0, 0, parity, x_cb, s_c_col, k, j0);
+
+          vuv[s_c_row*Arg::coarseSpin+s_c_col].mma_tn(V, UV);
+
+        } // fine spin
+      } // coarse spin (fine source parity)
+      // end backwards direction
+    } else if (arg.dir == QUDA_FORWARDS) {
+      // for forwards, we're tieing together <V^\dag A | U V>, so we only need
```
maddyscientist: tieing -> tying

include/kernels/coarse_op_kernel.cuh
```diff
@@ -876,6 +1209,93 @@ namespace quda {
...
+      // one component of UV
+      int s_c_col = arg.spin_map(0, (parity+1)&1);
+#pragma unroll
+      for (int s_c_row = 0; s_c_row < Arg::coarseSpin; s_c_row++) {
+#pragma unroll
+        for (int k = 0; k < TileType::k; k += TileType::K) { // Sum over fine color 
+
+          // for forwards, we're tieing together <V^\dag A | U V>
```
maddyscientist: tying

lib/staggered_kd_reorder_xinv.cu
```diff
@@ -0,0 +1,124 @@
...
+      if (dagger_approximation)  {
+        // approximate the inverse with the dagger: the free field for staggered, 
+        // B^-1 = 1 / (4 * (d + mass^2)), where the 4 is due to the factor of 2 convention
+        // Point of future investigation: it seems like we don't need the scaling factor to get
+        // a consistent norm?
+        double scale = 1.; // / (4. * (fineXinv.Ndim() + mass * mass));
+
+        // reset scales as appropriate
+        if constexpr (sizeof(Float) < QUDA_SINGLE_PRECISION) {
+          double max_scale = coarseXinv.abs_max() * abs(scale) * 1.01;
```
maddyscientist: What's the motivation for the 1% margin here?

maddyscientist (commented):
I've now tested the code, and reviewed the latest additions.  This is looking great.  Once my latest final comments have been addressed this PR will get my approval.

lib/staggered_kd_reorder_xinv.cu
weinbe2: I remembered seeing bonus paddings elsewhere, but I can't seem to find them anymore. I'll zap it!

include/quda.h
weinbe2: Not to make excuses, but the multigrid structure doesn't seem to be included in `quda_fortran.F90` anyway?

include/kernels/coarse_op_kernel.cuh
weinbe2: I haven't seen any compiler complaints on spilling for this kernel or for the kernel it's based on in `staggered_kd_apply_xinv.cu`, so I think (perhaps by good graces) it's safe.

lib/staggered_kd_reorder_xinv.cu
weinbe2: https://github.com/lattice/quda/pull/1221/commits/494c0be0466700edb2329aeac14ced9a5ea7e9dd

include/kernels/coarse_op_kernel.cuh
weinbe2: https://github.com/lattice/quda/pull/1221/commits/494c0be0466700edb2329aeac14ced9a5ea7e9dd

include/kernels/coarse_op_kernel.cuh
weinbe2: https://github.com/lattice/quda/pull/1221/commits/494c0be0466700edb2329aeac14ced9a5ea7e9dd

include/kernels/coarse_op_kernel.cuh
weinbe2: https://github.com/lattice/quda/pull/1221/commits/494c0be0466700edb2329aeac14ced9a5ea7e9dd

include/kernels/coarse_op_kernel.cuh
weinbe2: Was probably just in a `const` happy mood

https://github.com/lattice/quda/pull/1221/commits/494c0be0466700edb2329aeac14ced9a5ea7e9dd

include/quda.h
weinbe2: Resolved offline

maddyscientist (approved):

mathiaswagner:
Chroma HMC passed without issues.

weinbe2:
Can confirm jitify goes through with the latest fix, thanks @maddyscientist

PR #1225, opened 2021-12-15
fix some issues with tests

jcosborn:


mathiaswagner:

mathiaswagner:
@Jenkins please test this.

mathiaswagner:
@Jenkins ok to test.

tests/staggered_dslash_test_utils.h
```diff
@@ -503,7 +503,7 @@ struct StaggeredDslashTestWrapper {
       if (std::isnan(spinor_out_norm2)) { failed = true; }
 
       double cuda_spinor_out_norm2 = blas::norm2(*cudaSpinorOut);
-      printfQuda("Results: CPU=%f, CUDA=%f, CPU-CUDA=%f\n", spinor_ref_norm2, cuda_spinor_out_norm2, spinor_out_norm2);
+      printfQuda("Results: CPU=%f, CUDA=%f, (CPU-CUDA)/CPU=%e\n", spinor_ref_norm2, cuda_spinor_out_norm2, 1.0-cuda_spinor_out_norm2/spinor_ref_norm2);
```
maddyscientist: @jcosborn can you apply this change to dslash_test_utils.h.  Despite the fact it arose due to a misunderstanding, it's a good change to have.  Though I think it probably would be better to have `1.0 - sqrt(....)`

tests/invert_test.cpp
```diff
@@ -170,7 +169,7 @@ int main(int argc, char **argv)
     setQudaMgSolveTypes();
     setMultigridInvertParam(inv_param);
     // Set sub structures
-    mg_param.invert_param = &mg_inv_param;
+    mg_param.invert_param = &inv_param;
```
maddyscientist: This changes breaks the MG solver when run through invert_test.  This is primarily due to the poor interface of the multigrid solver, but it is what it is.  The reasoning is that the `QudaInvertParam` that is embedded in the `QudaMultigridParam` is used to drive the null space generation, and this is necessarily distinct from a solver that is preconditioned by MG.

Can you revert this change?

maddyscientist (changes requested):
Thanks for this PR @jcosborn.  This is mostly ok, but the removal of `mg_inv_param` breaks the multigrid-preconditioned solver.

maddyscientist (approved):
Thanks for the reversion of the MG param @jcosborn.  I've confirmed this fixes the MG test again.  I fully intend to address the MG interface in the future.  I've pushed a slight change to make the deviation printing more consistent with other parts of QUDA.  Once Jenkins is finished, we can merge.

PR #1230, opened 2022-01-04
Feature/multi blas misc

maddyscientist:
Primarily a quality-of-life and bug-fix PR aimed at the blas, as well as some foundational building blocks for future additions:

### Features
* Reduction buffers are now resized on demand.  This removes any artificial limit on the number of concurrent reductions.
* Multi-reductions now perform batching over the z dimension, e.g., a swizzling of the y-z thread indices
* Add support for batched reductions within the thread block (over the z thread dimension).  Utilized by multi_reduce_quda.cu (but only enabled by default with fast-compilation enabled)

### Bug fixes
* Fix bug with multi-1d multi-blas kernel template instantiation (this fixes BiCGStab(l) for example)
* Improved robustness of multi-blas NYW calculation
* Fix bug in multi `caxpyz` and add this kernel to blas_test (fixes CA-CG)
* Fixed-point (half/quarter) multi-blas kernels now have the same size limit as single/double
* Fix bug in `Tunable::advanceBlockDim` to take into account 3-d block size
* Fix issue with multi-blas when using warp splitting (thread count must be divisible by warp size)

### DIFF
--- a/include/kernels/multi_reduce_core.cuh
+++ b/include/kernels/multi_reduce_core.cuh
@@ -46,7 +46,7 @@ namespace quda
                      std::vector<ColorSpinorField *> &z, std::vector<ColorSpinorField *> &w,
                      Reducer f, int NYW, int length, int nParity) :
         // we have NYW * nParity reductions each of length NXZ
-        ReduceArg<reduce_t>(dim3(length, NYW, 1), NYW),
+        ReduceArg<reduce_t>(dim3(length, 1, NYW), NYW),
         NYW(NYW),
         f(f),
         length_cb(length / nParity),
@@ -86,7 +86,7 @@ namespace quda
       constexpr MultiReduce_(const Arg &arg) : arg(arg) {}
       static constexpr const char *filename() { return KERNEL_FILE; }
 
-      __device__ __host__ inline reduce_t operator()(reduce_t &sum, int tid, int k, int) const
+      __device__ __host__ inline reduce_t operator()(reduce_t &sum, int tid, int, int k) const
       {
         unsigned int parity = tid >= arg.length_cb ? 1 : 0;
         unsigned int i = tid - parity * arg.length_cb;
--- a/include/kernels/multi_reduce_core.cuh
+++ b/include/kernels/multi_reduce_core.cuh
@@ -14,6 +14,19 @@ namespace quda
   {
 
     /**
+       @brief Return the batch block size used for multi reductions.
+       For now, we leave this at 1 unless fast-compilation is enabled,
+       since it provides negligible benefit (but substantially longer
+       tuning).  When fast-compilation is enabled, the increased batch
+       count can dramatically improve performance.
+     */
+#ifndef QUDA_FAST_COMPILE_REDUCE
+    constexpr unsigned int max_n_batch_block_multi_reduce() { return 1; }
+#else
+    constexpr unsigned int max_n_batch_block_multi_reduce() { return 8; }
+#endif
+
+    /**
        @brief Parameter struct for generic multi-reduce blas kernel.
        @tparam real_ The precision of the calculation
        @tparam n_ The number of real elements per thread
@@ -36,6 +49,7 @@ namespace quda
       static constexpr int n = n_;
       static constexpr int NXZ = NXZ_;
       static constexpr int NYW_max = max_YW_size<NXZ, store_t, y_store_t, Reducer>();
+      static constexpr unsigned int max_n_batch_block = max_n_batch_block_multi_reduce();
       const int NYW;
       Reducer f;
 
--- a/include/multi_blas_helper.cuh
+++ b/include/multi_blas_helper.cuh
@@ -125,6 +125,19 @@ namespace quda
     */
     constexpr int max_N_multi_1d() { return 24; }
 
+    constexpr int max_N_multi_1d_pow2()
+    {
+      unsigned int v = max_N_multi_1d();
+      v--;
+      v |= v >> 1;
+      v |= v >> 2;
+      v |= v >> 4;
+      v |= v >> 8;
+      v |= v >> 16;
+      v++;
+      return v >> 1;
+    }
+
     /**
        @brief Return the maximum power of two enabled by default for
        multi-blas.  We set a lower limit for multi-reductions, since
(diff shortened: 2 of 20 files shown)


include/multi_blas_helper.cuh
```diff
@@ -125,23 +125,34 @@ namespace quda
     */
     constexpr int max_N_multi_1d() { return 24; }
 
+    /**
+       @brief Return the maximum size supported by multi-blas kernels
+       (max_N_multi_1d()) rounded down to the largest power of two.
+       This is used for the NXZ power-of-two instantiation.
+    */
+    constexpr int max_N_multi_1d_pow2()
+    {
```
weinbe2, line 134: What's the reason for going up to `v >> 16` and not, say `v >> 32`?

maddyscientist: v >> 32 doesn't tell you much with a 32-bit number 😉

maddyscientist:
@bjoo @jcosborn pinging you that this PR, once merged, will require minor updates at your end.  In particular there is a swizzling of the y-z thread mapping in the multi-reduction kernel.  For HIP this likely just means a straight copy of the include/targets/cuda/reduction_kernel.h file.

include/multi_blas_helper.cuh
weinbe2: ...yep, fair. Need more coffee.

weinbe2:
Visual review looks good. I'll test the fixes to CA-CG and BiCGStab-L in a real workflow, then I expect I'll be good to merge.

include/kernels/multi_reduce_core.cuh
```diff
@@ -13,6 +13,19 @@ namespace quda
...
   {
 
+    /**
+       @brief Return the batch block size used for multi reductions.
+       For now, we leave this at 1 unless fast-compilation is enabled,
+       since it provides negligible benefit (but substantially longer
+       tuning).  When fast-compilation is enabled, the increased batch
+       count can dramatically improve performance.
+     */
+#ifndef QUDA_FAST_COMPILE_REDUCE
```
jcosborn, line 23: should this be #ifdef (set to 1 if fast compile, 8 otherwise)?

maddyscientist: No.  😄   As the comment notes, the perf benefit from block batching is non-existent when fast-compile is disabled.  However, when enabled, block batching allows us to increase the occupancy, and helps to close the gap relative to fast-compile off.

maddyscientist: One could argue I suppose that we should just leave it on always then, but enabling it for fast compile off has an adverse impact on auto tuning time.  Hence the above code.

Long-term plan (post PR merge) is to rewrite this routine to utilize 2-d hierarchical tiling as opposed to the approach taken presently (which uses register tiling in the x dimension, and cache tiling in the y dimension).  Once I have done that, the block batch tuning will be have a significant benefit then.

lib/multi_reduce_quda.cu
```diff
@@ -31,6 +31,25 @@ namespace quda {
...
+              param.block.z < max_n_batch_block_multi_reduce()) {
+            param.block.z++;
+            param.grid.z = (n_batch + param.block.z - 1) / param.block.z;
+            return true;
+          } else { // we have run off the end so let's reset
+            param.block.z = 1;
+            param.grid.z = (n_batch + param.block.z - 1) / param.block.z;
+            return false;
+          }
+        }
```
hummingtree: Maybe this part can be abstracted in a function so it can be used else where?

maddyscientist: I'll see if I can move to the parent `TunableMultiReduction`.

jcosborn:
It looks like this would be relatively straightforward to implement in SYCL (similar to the other block operations).  Thanks for the heads-up.

lib/multi_reduce_quda.cu
maddyscientist: 62a41b7

include/tunable_reduction.h
```diff
@@ -357,14 +358,19 @@ namespace quda
...
   public:
-    TunableMultiReduction(const LatticeField &field, int n_batch,
+    TunableMultiReduction(const LatticeField &field, unsigned int n_batch, unsigned int n_batch_block_max = 1u,
                           QudaFieldLocation location = QUDA_INVALID_FIELD_LOCATION) :
-      TunableReduction2D<block_size_y>(field, location), n_batch(n_batch)
+      TunableReduction2D<block_size_y>(field, location),
+      n_batch(n_batch),
+      n_batch_block_max(n_batch_block_max)
     {
     }
```
hummingtree: I would suggest having a check of `n_batch_block_max == Arg::max_n_batch_block` here since otherwise we might get mysterious shared memory out of bound.

maddyscientist: I don't think that's possible at this point, since the `Arg` isn't visible here, and may not even be instantiated yet.

maddyscientist: Ah, but I can add that check to the launch function.  Will do that now.

hummingtree: I put it in the wrong place but I think you got my point :)

include/tunable_reduction.h
```diff
@@ -380,21 +386,36 @@ namespace quda
...
+        }
+      }
     }
 
     void initTuneParam(TuneParam &param) const
     {
       Tunable::initTuneParam(param);
-      param.block.y = 1;
-      param.grid.y = n_batch;
+      param.block = {param.block.x, 1, 1};
```
hummingtree, line 410: One way to decrease the tuning time is to start not from `block.z == 1` but from some other larger user-defined number. But this is definitely for a future PR.

hummingtree (approved):
I have been able to merge this into my MSPCG PR and make the necessary `y <-> z` changes to make things work.

include/tunable_reduction.h
maddyscientist: 1650dc622e8d9ce45dc5c2eef6bba0f5a7b601be applies this check.  Instead of checking for equality, I instead check that the max used by the tuner does not exceed the max supported by the kernel.  There may be some run-time heuristic for example for pruning the max size, so we shouldn't insist on equality.

hummingtree: Yep, I was about to suggest `<` instead of `=`.

weinbe2 (approved):
With that last fix, I can confirm this PR looks great and fixes the issues with CA-CG and BiCGstab-L. Pending Jenkins I think this is good to merge.

PR #1232, opened 2022-01-07
Fix bug where reduceEnd was being destroyed when not created

maddyscientist:
Quick fix for silly bug I introduced.  Causes badness with `endQuda` if reductions are never initialized.


weinbe2 (approved):
Nothing controversial!

PR #1233, opened 2022-01-08
Fix UB in host multi-reduction kernel

maddyscientist:
Fixes issue with host multi reduction kernel


maddyscientist:
@weinbe2 @hummingtree @mathiaswagner please approve when one of you gets the chance.  This fixes a bug in #1230 that sneaked in. Tests passing, but ASAN showed badness when running `transform_reduce` on the host, and this trivial patch fixes it.

mathiaswagner (approved):
visual review looks ok.

Issue #959, opened 2020-01-22
MILC and QUDA Q charge discrepancy


cpviolator:
As observed in #942, QUDA and MILC computed values of the topological charge Q are inconsistent.

MILC's F_munu code is here:
Make Field Strength Tensor,
https://github.com/milc-qcd/milc_qcd/blob/ac033d6faf9edb86ecb89acabb5ae3108fd618d3/generic/field_strength.c
Compute Q,
https://github.com/milc-qcd/milc_qcd/blob/ac033d6faf9edb86ecb89acabb5ae3108fd618d3/wilson_flow/fmunu.c#L41

QUDA's F_munu code is here:
Make Field Strength Tensor,
https://github.com/lattice/quda/blob/develop/include/kernels/field_strength_tensor.cuh
Compute Q,
https://github.com/lattice/quda/blob/develop/include/kernels/gauge_qcharge.cuh#L36

From what I can see, they use identical methods: compute the clover F_{mu,nu} terms, form the six independent 1/8 * ( F_{mu,nu} - F_{nu,mu} ), for nu < mu < 4, and then take the real trace of products with Levi-Civita.

If someone is willing to fine tooth comb MILC code for errors, I'll do the same for QUDA.

weinbe2:
This line looks a little weird: https://github.com/milc-qcd/milc_qcd/blob/ac033d6faf9edb86ecb89acabb5ae3108fd618d3/wilson_flow/fmunu.c#L56

But otherwise they look consistent.

cpviolator:
@weinbe2 @walkloud and I had a chat about this. It does look slightly off! However, given that the `fs` argument is antisymmetric it works out OK on paper. I checked QUDA by switching the four lines at
https://github.com/lattice/quda/blob/develop/include/kernels/gauge_qcharge.cuh#L48

with

```C++
double Q1 = getTrace(F[0] * F[5]).real();
double Q2 = getTrace(conj(F[1]) * F[4]).real();
double Q3 = getTrace(F[3] * F[2]).real();
double Q_idx = (Q1 + Q3 + Q2);
```
and QUDA's results were unchanged, so it must be something deeper.

kostrzewa:
If it's any help, here's my implementation in tmLQCD, not sure how useful this might be as one would need to check the definitions of the used macros:

https://github.com/etmc/tmLQCD/blob/master/meas/measure_clover_field_strength_observables.c

cpviolator:
[...]

PR #1234, opened 2022-01-11
Wflow interface

Marcogarofalo:
We would like to use the Wflow rotines of quda from `tmLQCD`.  We would like to have the `QudaGaugeObservableParam`  structure computed in `performWFlownStep` returned.
To do that we add the function 
```
 void performWFlownStep_param(unsigned int n_steps, double step_size, int meas_interval, QudaWFlowType wflow_type,
  QudaGaugeObservableParam *param);
```
that need as input  the pointer `QudaGaugeObservableParam *param` which must point to an array of n_steps+1 QudaGaugeObservableParam where to store the parameters computed at each time step.

The function 
```
  void performWFlownStep(unsigned int n_steps, double step_size, int meas_interval, QudaWFlowType wflow_type);
```
works as before, but now internally allocate the memory for a temporary array of `QudaGaugeObservableParam` and  calls performWFlownStep_param.

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -1493,6 +1493,15 @@ extern "C" {
    * @param step_size Size of Wilson Flow step
    * @param meas_interval Measure the Q charge and field energy every Nth step
    * @param wflow_type 1x1 Wilson or 2x1 Symanzik flow type
+   * @param[in,out] param Parameter struct that defines which
+   * observables we are making and the resulting observables.
+   */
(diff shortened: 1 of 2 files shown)


mathiaswagner:

Marcogarofalo:
Could you please tell me if there are any updates regarding
#959

maddyscientist:
ok to test

Marcogarofalo:
Moved to #1249

PR #1235, opened 2022-01-11
Improvements to fixed-iteration CA-CG

weinbe2:
This PR introduces a set of performance improvements to CA-CG when the number of iterations is fixed in advance (such as when it's a preconditioner). This includes:

* Eliminating copies that are unnecessary when there's only one "pass"
* Replacing all `Complex` data types with real types, since all coefficients in CA-CG are real (@maddyscientist)
* Removing unnecessary `blas` routines, replacing them with versions with `real` coefficients as appropriate (which also led to a small optimization in polynomial-accelerated eigensolves)
* Creating an `real` valued block `axpyz`, which gives a performance boost to CA-CG without fixed iterations --- it's relevant for the calculation of the `alpha` scalars, which isn't needed in the first "pass" (@maddyscientist)
* Exposing specifying the basis for CA-CG as a preconditioner on the command line, and persisting it across subsequent calls to `invertQuda`. This required additions to `QudaInvertParam` and `SolverParam`. Additions to `QudaInvertParam` were propagated to the fortran definitions.
* Removal of unnecessary norm printing of host fields in `invertQuda`: now only done if debug verbosity enabled (@maddyscientist)

With these improvements, single-pass `CA-CG(n)` is generally faster than `n` iterations of CG in both the power and the Chebyshev basis. The algorithmic itself is now arguably SoL.

These improvements are best tested via additive Schwarz preconditioning to Mobius DWF. A reference command is:

```
./invert_test   --matpc even-even   --dim 16 16 16 16   --Lsdim 8   --gridsize 1 1 1 1   \
  --dslash-type mobius   --b5 2.5 --c5 1.5   --load-gauge l16t16b7p0  --mass 0.01 \
  --inv-type pcg   --precon-type cg   --precon-schwarz-type additive   \
  --tol 1e-6   --tol-precondition 1e-6   --niter 860   --maxiter-precondition 4   --ca-basis-type-precondition chebyshev  \
  --prec single   --prec-sloppy half   --prec-precondition half   --verbosity verbose
```

For expedient testing, QUDA can be compiled with dwf only, fast dslash/fast reduce, reconstruct-18 only, and single and half precision only.

Outstanding actions:

- [x] `clang-format` (after review, before merge)


weinbe2:
Scope for future work noted in #1236

hummingtree (approved):
Thanks for the nice and quick PR! I have verified that it works with #1061.

maddyscientist:
Hold off merging: I just spotted a bug I’ll fix when at laptop

maddyscientist:
Fix pushed, we're good to merge once Jenkins has finished

PR #1238, opened 2022-01-18
Enable random momentum field

sbacchio:
This PR
- enables `gauge_random` for momentum field (adds `RECONSTRUCT_10` to `ReconstructFull`)
- makes usage of `sigma` in creating a random algebra (original behavior is maintained for sigma=1)


maddyscientist (approved):
Thanks for this contribution @sbacchio 👍

PR #1241, opened 2022-01-24
Add copy from reconstruct_10 to NO and viceversa for GaugeField

sbacchio:


maddyscientist (changes requested):
Thanks for the contribution @sbacchio.  This looks mostly good, just needs to be reformatted (looks like you're using four space indents in your additions for copy_gauge_inc.cu).

I presume this ordering support is required for interfacing with your host HMC code?

sbacchio:
Thanks for the feedback @maddyscientist 
Yes for interfacing and testing purposes I needed to convert from yours to full matrix format.
I could implement it outside but why not supporting it internally :)

maddyscientist (approved):

PR #1242, opened 2022-01-25
Renamed QUDAVersion.cmake to QUDAConfigVersion.cmake. It is the conve…

bjoo:
…ntion

This is a minor conventional change. I have found that if I actually want to match on a version, the version file CMake looks for is actually 'QUDAConfigVersion.cmake' rather than `QUDAVersion.cmake`


mathiaswagner (approved):
Apart from the branch name not matching our `hotfix/something` `feature/new_cool_stuff` there is nothing I could find to complain here ;-) (So I had to got with that pointless comment ;-) ) . 
Thanks for the fix.

Issue #1244, opened 2022-02-05
multishift-CG with refinement segfaults (develop regression)


kostrzewa:
The "last few" commits of `develop` seem to have introduced a regression not caught by the tests. We (@Marcogarofalo) first saw this in combination with tmLQCD in the HMC and were able to confirm using `invert_test`.

Running:

```
QUDA_RESOURCE_PATH=$(pwd) \
./invert_test \
  --prec double \
  --prec-refine single \
  --prec-sloppy single \
  --multishift 3
```

leads to

```
Rank order is column major (t running fastest)
Kappa = 0.12195122 Mass = 0.10000000
running the following test:
prec    prec_sloppy   multishift  matpc_type  recon  recon_sloppy solve_type S_dimension T_dimension Ls_dimension   dslash_type  normalization
double   single          3        even-even     18     18          normop-pc  24/ 24/ 24      24         16               wilson     kappa
Grid partition info:     X  Y  Z  T
                         0  0  0  0
QUDA 1.1.0 (git 1.1.0-3816b96d1-sm_61)
CUDA Driver version = 11050
CUDA Runtime version = 11050
Found device 0: NVIDIA GeForce GTX 1060 6GB
Found device 1: NVIDIA GeForce GTX 1060 6GB
Using device 0: NVIDIA GeForce GTX 1060 6GB
WARNING: Data reordering done on GPU (set with QUDA_REORDER_LOCATION=GPU/CPU)
Loaded 42 sets of cached parameters from /home/bartek/build/quda-develop-with_tests/tests/tunecache.tsv
Loaded 42 sets of cached parameters from /home/bartek/build/quda-develop-with_tests/tests/tunecache.tsv
WARNING: Using device memory pool allocator
WARNING: Using pinned memory pool allocator
cublasCreated successfully
Computed plaquette is 1.229479e-01 (spatial = 1.226621e-01, temporal = 1.232337e-01)
Saving 44 sets of cached parameters to /home/bartek/build/quda-develop-with_tests/tests/tunecache.tsv
MultiShift CG: Converged after 23 iterations
 shift=0, 23 iterations, relative residual: iterated = 1.198756e-07
 shift=1, 23 iterations, relative residual: iterated = 9.924532e-08
 shift=2, 22 iterations, relative residual: iterated = 9.369405e-08
Refining shift 0: L2 residual inf / 1.000000e-07, heavy quark 0.000000e+00 / 0.000000e+00 (actual / requested)
[...]

PR #1245, opened 2022-02-05
Hotfix for injecting `p` vector from multi-shift into 0th refinement solve

weinbe2:
This hotfix addresses #1244 .

See title. The issue was the `std::vector<ColorSpinorField> p;` used for pipelining was moved into `XUpdateBatch`, as did its allocation and _almost_ all references from it in `CG::operator(...)` --- the one exception being in the setup when an initial `p` is injected from a multi-shift solve.

This has now been addressed by replacing `p[0]` with `*p_init` during solver start-up (as opposed to `x_update_batch.get_current_field()` for clarity). I also removed the `std::vector<ColorSpinorField*> p;` from the declaration of `CG` as well, and cleaned up a few (harmess) compiler errors downstream from that.

@kostrzewa this passes your minimal repro, and I'm in the process of testing a few other workflows now. That said, I don't expect surprises. I'm sorry about this, again.

### DIFF
--- a/include/invert_quda.h
+++ b/include/invert_quda.h
@@ -709,7 +709,6 @@ namespace quda {
   private:
     // pointers to fields to avoid multiple creation overhead
     ColorSpinorField *yp, *rp, *rnewp, *pp, *App, *tmpp, *tmp2p, *tmp3p, *rSloppyp, *xSloppyp;
-    std::vector<ColorSpinorField*> p;
     bool init;
 
   public:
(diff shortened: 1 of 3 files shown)


maddyscientist (approved):
Thanks for the quick fix @weinbe2

mathiaswagner (approved):

weinbe2:
Do not merge yet, I found a subtle bug with double/single/half...

kostrzewa:
@weinbe2 Thanks also for https://github.com/lattice/quda/pull/1245/commits/3dedbe3176f2138a550c5c7108535b2300382f71! I had actually produced this problem in the past but thought it was a PEBKAC issue and left exploring it as a future optimization for our runs!

PR #1247, opened 2022-02-08
Hotfix/reduction precision

jcosborn:
Refactor reductions to avoid unnecessary precision conversion and include static init and apply members for reducers.


include/array.h
```diff
@@ -8,6 +8,7 @@ namespace quda
    */
   template <typename T, int n> struct array {
     using value_type = T;
+    static constexpr int N = n;
```
maddyscientist, line 11: What purpose does this serve that `array::size()` doesn't

include/float_vector.h
```diff
@@ -42,19 +44,26 @@ namespace quda {
     return c;
   }
 
-  template <typename T> constexpr T zero() { return static_cast<T>(0); }
-  template <> constexpr double2 zero() { return double2 {0.0, 0.0}; }
-  template <> constexpr double3 zero() { return double3 {0.0, 0.0, 0.0}; }
-  template <> constexpr double4 zero() { return double4 {0.0, 0.0, 0.0, 0.0}; }
+  template <typename T> constexpr std::enable_if_t<std::is_arithmetic_v<T>, T> zero() { return static_cast<T>(0); }
```
maddyscientist, line 47: Can you explain what the motivation is for this?  At present this will break run-time compilation of QUDA since `std::is_arithmetic_v` isn't supported in Jitify (I can add it easily enough though if there's a good reason).

include/targets/cuda/reduce_helper.h
```diff
@@ -59,6 +59,7 @@ namespace quda
      memory
    */
   template <typename T, bool use_kernel_arg = true> struct ReduceArg : kernel_param<use_kernel_arg> {
+    using reduce_t = T;
```
maddyscientist, line 62: The changes in this file will need to be propagated unto `include/targets/generic/reduce_helper.h`

include/array.h
jcosborn: It's static, so instantiation isn't necessary.

include/float_vector.h
jcosborn: It's needed to allow non-overlapping versions of the templated function.  This was done to allow the addition of a version for array<T,N>.  There are other ways to do it, but this was the simplest.  We could also just overload for float and double explicitly.

maddyscientist: That's motivation enough.  I can just added the needed change to Jitify and put this directly into your PR while I'm testing.  Sound ok?

maddyscientist (commented):
Thanks for this PR @jcosborn.  Once you've added the changes for the generic path (can be tested on CUDA with `QUDA_HETEROGENEOUS_ATOMIC=OFF`), then I will test this locally, and make any fixes needed for the Jitify path.

include/targets/cuda/reduce_helper.h
jcosborn: 2604207eb931edd2da6efb1768323aa2158e905b

jcosborn: and this: 960573f1a94b93e87bc4282c0e56aef986eb18ce

maddyscientist (approved):
Ok, this is all looking good @jcosborn.  I've confirmed that performance is good with your changes, and I have pushed the changes required to for Jitify to with with `std::is_arithmetic_v`.  We can merge this once Jenkins has completed.

PR #1246, opened 2022-02-08
Initialization of madwf_param_infile and madwf_param_outfile

Marcogarofalo:
co-author: [sunpho84](https://github.com/sunpho84)
this should fix the issue https://github.com/sunpho84/quda/issues/1

PR #1248, opened 2022-02-08
Hotfix/madwf param cstring

hummingtree:
This PR replaces the internal C-string `madwf_param_<in/out>file` with `std::string`, and adds additional checks to make sure there is no out of bound memory access. Also zero initializes the interface C-string `madwf_param_<in/out>file`.

This should be able to address https://github.com/sunpho84/quda/issues/1.


maddyscientist:
@hummingtree how does this relate to the fix in #1246?

hummingtree:
> @hummingtree how does this relate to the fix in #1246?

This is a more comprehensive one.

maddyscientist (approved):

weinbe2:
This is great, thanks for doing this. Can you extend it to `--load-gauge`, `--save-gauge` in the interface in an additional commit?

On top of it, can you add it to null vector in/out in the MG interface as well (intentionally in a separate commit from load/save gauge)? My ask for that being a separate commit is because there's a chance for subtle issues, for ex in the MILC bindings.

hummingtree:
> This is great, thanks for doing this. Can you extend it to `--load-gauge`, `--save-gauge` in the interface in an additional commit?
> 
> On top of it, can you add it to null vector in/out in the MG interface as well (intentionally in a separate commit from load/save gauge)? My ask for that being a separate commit is because there's a chance for subtle issues, for ex in the MILC bindings.

Can you check if the changes are good (for my ignorance of the MG-related stuff)?

tests/utils/set_params.cpp
```diff
@@ -597,8 +597,8 @@ void setMultigridParam(QudaMultigridParam &mg_param)
 
   // set file i/o parameters
   for (int i = 0; i < mg_param.n_level; i++) {
-    strcpy(mg_param.vec_infile[i], mg_vec_infile[i]);
-    strcpy(mg_param.vec_outfile[i], mg_vec_outfile[i]);
+    strcpy(mg_param.vec_infile[i], mg_vec_infile[i].c_str());
```
maddyscientist: No harm in adding a bounds check here

tests/utils/set_params.cpp
```diff
@@ -597,8 +597,8 @@ void setMultigridParam(QudaMultigridParam &mg_param)
...
+    strcpy(mg_param.vec_outfile[i], mg_vec_outfile[i].c_str());
     if (strcmp(mg_param.vec_infile[i], "") != 0) mg_param.vec_load[i] = QUDA_BOOLEAN_TRUE;
```
maddyscientist: This could be updated like you did elsewhere in this PR to use the size of the string

tests/utils/set_params.cpp
```diff
@@ -1186,8 +1186,8 @@ void setStaggeredMultigridParam(QudaMultigridParam &mg_param)
 
   // set file i/o parameters
   for (int i = 0; i < mg_param.n_level; i++) {
-    strcpy(mg_param.vec_infile[i], mg_vec_infile[i]);
-    strcpy(mg_param.vec_outfile[i], mg_vec_outfile[i]);
+    strcpy(mg_param.vec_infile[i], mg_vec_infile[i].c_str());
+    strcpy(mg_param.vec_outfile[i], mg_vec_outfile[i].c_str());
```
maddyscientist: Add bounds check here and use `size()` on the two lines below

lib/check_params.h
```diff
@@ -573,6 +573,11 @@ void printQudaInvertParam(QudaInvertParam *param) {
   }
 #endif
 
+#ifdef INIT_PARAM
+  P(madwf_param_infile[0], '\0');
```
weinbe2, line 577: Out of curiosity, is this necessary? Or shouldn't the default initialization for a `std::string` always be the empty string?

weinbe2:
> > This is great, thanks for doing this. Can you extend it to `--load-gauge`, `--save-gauge` in the interface in an additional commit?
> > On top of it, can you add it to null vector in/out in the MG interface as well (intentionally in a separate commit from load/save gauge)? My ask for that being a separate commit is because there's a chance for subtle issues, for ex in the MILC bindings.
> 
> Can you check if the changes are good (for my ignorance of the MG-related stuff)?

Looks good---I'll formally spin up a test tomorrow.

I remembered a few more `char[]` that need to be updated:
```
char eig_arpack_logfile[256] = "arpack_logfile.log";
char eig_vec_infile[256] = "";
char eig_vec_outfile[256] = "";
```

lib/check_params.h
hummingtree: This is the C-string `madwf_param_infile` in the C interface.

tests/utils/set_params.cpp
hummingtree: See the latest changes.

weinbe2 (approved):
I did some tests and everything looks good. I noticed `std::string` updates didn't go into `tests/multigrid_evolve_test.cpp` (look for the comment `// restore output file name`). I'll still give this an approve in the interim.

The ideas in this PR _should_ go into the HISQ MG MILC interface that I created, but that has such a train wreck design that I won't waste your (or my, for now) time with it.

Once the fix for `tests/multigrid_evolve_test.cpp` goes in, this is good to `clang-format` and (once Jenkins completes) merge.

PR #1249, opened 2022-02-16
Feature/wflow interface

Marcogarofalo:
I closed #1234  and reopened here.


lib/interface_quda.cpp
```diff
@@ -5459,16 +5473,13 @@ void performWFlownStep(unsigned int n_steps, double step_size, int meas_interval
 
   GaugeField *in = gaugeSmeared;
   GaugeField *out = gaugeAux;
-
-  QudaGaugeObservableParam param = newQudaGaugeObservableParam();
-  param.compute_plaquette = QUDA_BOOLEAN_TRUE;
-  param.compute_qcharge = QUDA_BOOLEAN_TRUE;
+  
+  gaugeObservables(*in, param[0], profileWFlow);
```
kostrzewa: @maddyscientist @cpviolator This change makes the flow itself a little more expensive compared to what was previously there for verbosity levels below `QUDA_SUMMARIZE`. Is this acceptable or do we need to add some more logic?

One could, for example, change lines 5449 and 5450 above to be `getVerbosity() >= QUDA_SUMMARIZE ? QUDA_BOOLEAN_TRUE : QUDA_BOOLEAN_FALSE` to keep the original behaviour of `performWFlownStep`.

Marcogarofalo: An other possibility will be  setting `meas_interval>n_steps` if `getVerbosity() < QUDA_SUMMARIZE`

cpviolator: I think I prefer @kostrzewa's suggestion of setting the booleans according to verbosity level. That will ~~render the call to gaugeObservables a NOP and~~ not waste any FLOPS.

I'm was unsure as to why it is necessary to move the call outside the verbosity boolean guard, but now I see that there an SU(N) projection step in the call to gaugeObservables. Is this the motivation for moving the call?

Marcogarofalo: In general, I thought it could be possible that we want to call quda from tmLQCD with  `QUDA_SILENT` but we would like to have the `QudaGaugeObservableParam *param` computed.
I can remove the change if it is in conflict with the logic in quda.

kostrzewa: @cpviolator In the original implementation, the only way to get QUDA to compute the gauge observables during the Wilson flow was to switch to a higher verbosity level. Instead, we would like to have QUDA compute the observables for us for a whole Wilson flow integration trajectory via the new `performWFlownStep_param` (when called from tmLQCD) without having lots of console output at the same time *and* without changing the interface of the original `performWFlownStep` function. This seemed like a reasonable way to achieve what we want to do (modulo the change suggested).

If you now say, however, that there is an SU(N) projection hidden in `gaugeObservables`, then we are of course adding operations even if we don't compute the plaquette and topological charge...

cpviolator: @Marcogarofalo Ahh, so tmLQCD can pass in the array of param structs, have QUDA be silent, then have tmLQCD dump the information. I think the extra expense of a single call to gaugeObservables is OK.

cpviolator: So moving the call outside the guard is actually solving a hidden bug, because verbosity should have no relevance at all to an suN_project policy. The default behaviour in `develop` is to not perform suN projection. I think the addition of `getVerbosity() >= QUDA_SUMMARIZE ? QUDA_BOOLEAN_TRUE : QUDA_BOOLEAN_FALSE` is good.

@maddyscientist What are your thoughts on default behaviour WRT reunitarization during flow computation? We do not have a CL option for it in develop. I would prefer to keep it as (no default reunitarization), add a CL option, and a warning in the flow function giving the state of the reunitarization. This may be why MILC data and QUDA data differ slightly as MILC reunitarises all the time.

maddyscientist: The cost of computing the plaquette and the topological charge is negligible overall here, so I don't see any reason to just do them anyway.  It's not like this will be called after every solve.

@cpviolator Interesting thought regarding the potential source of the deviation between MILC and QUDA.  Is this something you can look into?

cpviolator: @maddyscientist I'll hotfix `develop` with a CL addition for su projection and re run against MILC data to see if there is any effect.

cpviolator: I think https://github.com/lattice/quda/pull/1249/commits/2b66ed620e5a6f2a8a4e75a7f9807a055eb02963 should be propagated to all the gauge smearing routines as it is the correct way to call `gaugeObservables` which I will include in the hotfix. This does, however, induce a problem with respect to how reunitarisation is enabled/disabled. Each smearing interface function call creates its own `gaugeObservablesParam` struct. At present there is no way for the user to interact with it. Updating the function signature in `interface_quda.cpp` comes with the usual pitfall of breaking code downstream.

One way to avoid that is to introduce `smearingFunction(Args..., gaugeObservablesParam param = nullptr)` signatures in `quda.h` so that any existing downstream code that uses `smearingFunction(Args...)` remains as is, but users are free to change their code to `smearingFunction(Args..., user_defined_gauge_obs_param)`. Naturally the smearing functions will detect when the `gaugeObservablesParam` arg is a nullptr and dictate its own behaviour.

Any thoughts on this before I go ahead?

cpviolator: @kostrzewa @Marcogarofalo Just had an offline conversation with maddyscientist and it's probably better for all if we switch all the smearing routines to behave in this fashion (with arrays of QUDA param structs) and ditch the old interfaces. This will allow for more fine grained control over verbosity and reunitarisation.

If you don't mind I can add the hotfix additions to this branch, and update the testing routines so that arrays of param structs are passed. I'll get this done sooner rather than later so your PR is not held up.

Marcogarofalo: thank you, that ok for me.

maddyscientist:
@Marcogarofalo Jenkins failures are caused by the following warning (if you click on the red cross / green tick mark, and then click on the "Console Output" on the Jenkins panel, you'll be able to see the test logs.
```
../lib/interface_quda.cpp:5453:5: error: 'delete' applied to a pointer that was allocated with 'new[]'; did you mean 'delete[]'? [-Werror,-Wmismatched-new-delete]
    delete param;
    ^
          []
../lib/interface_quda.cpp:5446:39: note: allocated with 'new[]' here
    QudaGaugeObservableParam* param = new QudaGaugeObservableParam[n_steps/meas_interval+1];
                                      ^
1 error generated.
```

Marcogarofalo:
Sorry for that, I think I correct it now.

cpviolator:
@Marcogarofalo @kostrzewa I've updated all the smearing interface function to perform suN projection, and each function accepts an array of GaugeObservalesParam to perform measurements. Please be sure to change your code so that only `n_smear/measurement_interval+1` structs are declared, rather than `n_smear+1`

The suN projection has a non-trivial effect on the data, so this PR has solved a hidden bug. I'll be cross checking QUDA data with MILC data in the near future, but that should not hold up this PR any longer. Please take a look over my last ~~two~~ commits and please don't hesitate to give feedback.

lib/interface_quda.cpp
```diff
@@ -5350,29 +5351,33 @@ void performAPEnStep(unsigned int n_steps, double alpha, int meas_interval)
...
-
+  // The nth measurement to take
+  int measurement_n = 0;
+  
   if (getVerbosity() >= QUDA_SUMMARIZE) {
-    gaugeObservablesQuda(&param);
-    printfQuda("Q charge at step %03d = %+.16e\n", 0, param.qcharge);
+    gaugeObservablesQuda(&param[measurement_n]);
+    printfQuda("Q charge at step %03d = %+.16e\n", 0, param[measurement_n].qcharge);
   }
```
Marcogarofalo, line 5350: Thank you for the changes. Is there a reason why the call to `gaugeObservablesQuda` is guarded by `if (getVerbosity() >= QUDA_SUMMARIZE)` here?

cpviolator: Good catch, there was another misplaced `gaugeObservablesQuda` too.

Are you happy with this version works with TMLQCD?

kostrzewa: yes, thank you

cpviolator:
@Marcogarofalo @kostrzewa A good few PRs have made their way into develop recently, I'm happy to merge them into this branch while it awaits PR.

include/quda.h
```diff
@@ -1524,37 +1524,47 @@ extern "C" {
   /**
    * Performs APE smearing on gaugePrecise and stores it in gaugeSmeared
    * @param n_steps Number of steps to apply.
-   * @param alpha  Alpha coefficient for APE smearing.
+   * @param alpha   Alpha coefficient for APE smearing.
    * @param meas_interval Measure the Q charge every Nth step
    */
-  void performAPEnStep(unsigned int n_steps, double alpha, int meas_interval);
+  void performAPEnStep(unsigned int n_steps, double alpha, int meas_interval,
+		       QudaGaugeObservableParam *param);
```
maddyscientist: While we're updating these interface functions, I wonder if we should take the opportunity to fix their names.  QUDA interface functions should end in Quda, e.g., `performanceAPEnStepQuda`.

maddyscientist: Also, do any of these parameters (`alpha`, `n_steps`, `meas_interval`) belong in the `QudaGaugeObervableParam` struct)?

include/quda.h
```diff
@@ -1524,37 +1524,47 @@ extern "C" {
   /**
    * Performs APE smearing on gaugePrecise and stores it in gaugeSmeared
    * @param n_steps Number of steps to apply.
-   * @param alpha  Alpha coefficient for APE smearing.
+   * @param alpha   Alpha coefficient for APE smearing.
    * @param meas_interval Measure the Q charge every Nth step
```
maddyscientist: Missing doxygen for `param` here

include/quda.h
```diff
@@ -1524,37 +1524,47 @@ extern "C" {
   /**
    * Performs APE smearing on gaugePrecise and stores it in gaugeSmeared
    * @param n_steps Number of steps to apply.
-   * @param alpha  Alpha coefficient for APE smearing.
+   * @param alpha   Alpha coefficient for APE smearing.
```
maddyscientist: Apologies for being OCD: extra space added here

include/quda.h
```diff
@@ -1524,37 +1524,47 @@ extern "C" {
...
    */
-  void performAPEnStep(unsigned int n_steps, double alpha, int meas_interval);
+  void performAPEnStep(unsigned int n_steps, double alpha, int meas_interval,
+		       QudaGaugeObservableParam *param);
 
   /**
    * Performs STOUT smearing on gaugePrecise and stores it in gaugeSmeared
    * @param n_steps Number of steps to apply.
-   * @param rho    Rho coefficient for STOUT smearing.
+   * @param rho     Rho coefficient for STOUT smearing.
```
maddyscientist: Extra space added before Rho

include/quda.h
cpviolator: > While we're updating these interface functions, I wonder if we should take the opportunity to fix their names. QUDA interface functions should end in Quda, e.g., `performanceAPEnStepQuda`.

That's an easy fix and certainly should be applied.

> Also, do any of these parameters (`alpha`, `n_steps`, `meas_interval`) belong in the `QudaGaugeObervableParam` struct)?

Not at the moment, but we could add generic "meas_n_steps", "alpha", "rho", "meas_interval" which could serve for all three smearing types and flow types with just a bit of documentation. Wilson flow has fixed coefficients, stout and APE use rho and alpha respectively, and over improved stout uses two coeffs, so as long as the user knows which is which there will be no ability and a `QudaGaugeObervableParam` struct can be used.

maddyscientist (commented):
Apologies taking so long to review this.  Nothing controversial here.  Left a few comments: some doxygen fixes needed in `quda.h`, and it would be nice to make the function signatures consistent with the rest of QUDA.

Once these are addressed we can merge this.

include/quda.h
cpviolator: I'll fix all the whitespace stuff with the next round.

tests/su3_test.cpp
```diff
@@ -72,41 +59,40 @@ void display_test_info()
...
-  opgroup->add_option("--su3-ape-rho", ape_smear_rho, "rho coefficient for APE smearing (default 0.6)");
-
-  opgroup->add_option("--su3-stout-rho", stout_smear_rho,
-                      "rho coefficient for Stout and Over-Improved Stout smearing (default 0.08)");
 
-  opgroup->add_option("--su3-stout-epsilon", stout_smear_epsilon,
-                      "epsilon coefficient for Over-Improved Stout smearing (default -0.25)");
-
-  opgroup->add_option("--su3-smear-steps", smear_steps, "The number of smearing steps to perform (default 50)");
+  opgroup->add_option("--su3-smear-type", gauge_smear_type, "The type of action to use in the smearing (default stout)")
```
maddyscientist: probably should display the different smearing types supported here in brackets

tests/su3_test.cpp
```diff
@@ -72,41 +59,40 @@ void display_test_info()
 
 void add_su3_option_group(std::shared_ptr<QUDAApp> quda_app)
 {
-  CLI::TransformPairs<QudaWFlowType> wflow_type_map {{"wilson", QUDA_WFLOW_TYPE_WILSON},
-                                                     {"symanzik", QUDA_WFLOW_TYPE_SYMANZIK}};
+  CLI::TransformPairs<QudaGaugeSmearType> gauge_smear_type_map {{"ape", QUDA_GAUGE_SMEAR_APE},
+                                                                {"stout", QUDA_GAUGE_SMEAR_STOUT},
+                                                                {"ovrimp_stout", QUDA_GAUGE_SMEAR_OVRIMP_STOUT},
```
maddyscientist: for consistency with the rest of the CLI commands, change the underscore to a hyphen

tests/su3_test.cpp
```diff
@@ -191,57 +177,66 @@ int main(int argc, char **argv)
...
+      obs_param[i].compute_plaquette = QUDA_BOOLEAN_TRUE;
+    }
+    performWFlowQuda(&smear_param, obs_param);
     break;
-  default: errorQuda("Undefined test type %d given", test_type);
   }
+  default: errorQuda("Undefined gauge smear type %d given", smear_param.smear_type);
+  }
+
+  time0 += clock(); // stop the timer
```
maddyscientist: Use the `host_timer_t` class for doing this kind of timing.   Look in dslash_test_util.h for an example.

tests/su3_test.cpp
cpviolator: 8c6db1f5c705446dfb5fed55428c694edc41e73e

tests/su3_test.cpp
cpviolator: 8c6db1f5c705446dfb5fed55428c694edc41e73e

tests/su3_test.cpp
cpviolator: 8c6db1f5c705446dfb5fed55428c694edc41e73e

maddyscientist (approved):
Great work @cpviolator with the consolidation and cleanup.  Approving this.

@Marcogarofalo @kostrzewa hopefully you like the changes @cpviolator has made here.  Before we merge this, can you sign off on these changes?

cpviolator:
@Marcogarofalo @kostrzewa After @maddyscientist's suggestions I consolidated the smearing functions and revamped the test suite. Most of this is orthogonal to your work, except for four important changes:

1. The name and function signature of the flow interface function has changed.
2. The signature now takes a new param `QudaGaugeSmearParam` which contains all the necessary variables for any/all of the smearing and flow functions.
3. `QudaWFlowType` is now consolidated into a new type `QudaGaugeSmearType` and Wilson/Symanzik flow have new enums to be consistent: `QUDA_WFLOW_TYPE_WILSON` is now `QUDA_GAUGE_SMEAR_WILSON_FLOW` and similar for symanzik.
4. Other variables such as `epsilon` and `smear_steps` are also included in this param and have different names for consistency.

Sorry to churn this back one last time, but your flow additions were a perfect opportunity to clean the smearing code up too. @maddyscientist has signed off on the present state of the code for PR. If you are happy with the changes going forward we can merge, else we can do more changes for TMLQCD.

kostrzewa:
@cpviolator @maddyscientist Awesome work! @Marcogarofalo tested this with our interface and things seem to work as expected.

kostrzewa (approved):

PR #1250, opened 2022-02-17
Chebyshev-basis CA-GCR and some boilerplate cleanup

weinbe2:
This PR adds support for the Chebyshev basis in CA-GCR, does various CA boilerplate clean-up and (finally) exposes CA-related parameters for MG smoothers on the command line. More details:

- Adds support for choosing the Power or Chebyshev basis in CA-GCR
- Abstracts the computation of search vectors in different bases (power/Chebyshev) into a new function `computeCAKrylovSpace`
- Abstracts performing power iterations, used in the context of Chebyshev-basis CA algorithms to approximate lambda_max, to a new function `performPowerIterations`
- Exposes choosing the CA basis, and associated other parameters, for smoothers on the command line.
- Pushes some heuristics for which CA basis to use into the MILC HISQ MG interface

Outstanding:

- [x] `clang-format` (waiting on full review)


include/invert_quda.h
```diff
@@ -694,6 +694,98 @@ namespace quda {
     */
     void setRecomputeEvals(bool flag) { recompute_evals = flag; };
 
+    /**
```
maddyscientist, line 697: I'm not a huge fan of these extensive functions being in headers.  I know you're doing this because of the template, but the simple approach might be better here.  E.g.,
```cpp
static double performPowerIterations(const DiracMatrix& diracm, ColorSpinorField& start, 
ColorSpinorField* tempvec1, ColorSpinorField* tempvec2,int niter, int normalize_freq, ColorSpinorField *tmp1 = nullptr, ColorSpinorfield *tmp2 = nullptr);
```
Moreover, I can easily make these last arguments references with my default constructor branch, so the use of pointers would be temporary.

maddyscientist: Or if you you really want to keep the template interface, I would move the definition implementation to, say, `solver.hpp`, leaving the declaration in `invert_quda.h`.  Then `solver.hpp` would be included in the CA solver cpp files.

lib/multigrid.cpp
```diff
@@ -347,6 +347,14 @@ namespace quda
 
     param_presmooth->Nkrylov = param_presmooth->maxiter;
     param_presmooth->pipeline = param_presmooth->maxiter;
+
+    if (param_presmooth->inv_type == QUDA_CA_GCR_INVERTER || param_presmooth->inv_type == QUDA_CA_CG_INVERTER ||
```
maddyscientist: Would an `is_ca_solver` method be helpful?
```c++
bool is_ca_solver(QudaInvertType type)
{
  switch (type) {
  case QUDA_CA_GCR_INVERTER:
  case QUDA_CA_CG_INVERTER:
  case QUDA_CA_CGNR_INVERTER:
  case QUDA_CA_CGNE_INVERTER:
    return true;   
  default: return false;
}
```

include/invert_quda.h
weinbe2: As much as I love using fancy templates for no reason, this was because there are three flavors of `DiracMatrix::operator()`, each of which take a different number of temporaries at the end (0, 1, and 2). Variadic templates were the simplest way to handle this, so I can just pass `args...` to the end of the `DiracMatrix::operator()` calls and not care about the details.

To that end---I do like the idea of moving it to a `solver.hpp` file that only gets included in the CA files, so I'll do that.

lib/multigrid.cpp
weinbe2: Haha, yes, it would, I'll slot that in somewhere. Sometimes I need to do something extra for CA-GCR but that becomes a neater additional conditional.

include/invert_quda.h
weinbe2: https://github.com/lattice/quda/commit/08f02ac23a0b89da008ccf9dcf5d345c11a0885e

lib/multigrid.cpp
weinbe2: https://github.com/lattice/quda/commit/06c45e4fa09f664cfc83aca932d79719bf6f65c5

maddyscientist (approved):
This is a short and sweet PR.  In the tests I've done, CA-GCR is seemingly very resilient, and I'm loving the improved stability.  Great addition @weinbe2

PR #1252, opened 2022-03-01
Fix float max/min reductions

maddyscientist:
Quick fix for max/min reductions for fixed-point fields with `transform_reduce` which was broken by #1247.  A more expansive PR will be forthcoming which will go beyond this making reductions on non-double types robust, but for now this is sufficient to fix the regression.

### DIFF
--- a/include/tunable_reduction.h
+++ b/include/tunable_reduction.h
@@ -38,6 +38,16 @@ namespace quda
     }
   };
 
+  template <> struct comm_reduce_max<float> {
+    void operator()(std::vector<float> &v)
+    {
+      std::vector<double> v_(v.size());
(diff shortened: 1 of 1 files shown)


mathiaswagner (approved):
Looks good and passed Chroma test

kostrzewa:
Awesome, this fixes a performance regression that we were seeing with QUDA-MG in the `develop` branch and saves me from having to try to reproduce it with QUDA's tests. Thanks a lot!

@marcuspetschlies @Marcogarofalo @simone-romiti This fixes the performance problems that we were seeing (and perhaps also the stability issues that you observed in the HMC @Marcogarofalo ).

PR #1253, opened 2022-03-03
Feature/reduction comm cleanup

maddyscientist:
* Cleanup of reduction kernels to avoid any blind casting, which could lead to undefined behaviour
* The appropriate comm_reducer function to call is now built into the reducer functor. 
* comm_quda.h, communicator_quda.h, invert_x_update.h and reliable_updates.h are now correctly in the `quda` namespace.  Fix the resulting cascade of issues.
* Some `std::enable_if_t` -> `if constexpr` cleanup


tests/unitarize_link_test.cpp
```diff
@@ -47,7 +47,7 @@ TEST(unitarization, verify) {
                            unittol, cpu_prec);
 
 #ifdef MULTI_GPU
-  comm_allreduce_int(&res);
```
mathiaswagner, line 50: this file still has a `using namespace` but to get things moving away from that we should use quda::comm_allreduce_inte here.

tests/hisq_unitarize_force_test.cpp
```diff
@@ -144,7 +144,7 @@ TEST(hisq_force_unitarize, verify)
     res[dir] = compare_floats(((char **)cpuReference->Gauge_p())[dir], ((char **)cpuResult->Gauge_p())[dir],
                               cpuReference->Volume() * gauge_site_size, accuracy, gaugeParam.cpu_prec);
 
-    comm_allreduce_int(&res[dir]);
```
mathiaswagner, line 147: add `quda::` ?

tests/staggered_invert_test.cpp
```diff
@@ -353,11 +353,11 @@ int main(int argc, char **argv)
       inv_param.num_src_per_sub_partition = Nsrc / num_sub_partition;
       invertMultiSrcStaggeredQuda(_hp_x.data(), _hp_b.data(), &inv_param, (void *)milc_fatlink, (void *)milc_longlink,
                                   &gauge_param);
-      comm_allreduce_int(&inv_param.iter);
```
mathiaswagner, line 356: add `quda::` here?

tests/utils/host_utils.cpp
```diff
@@ -1612,7 +1612,7 @@ double mom_action(void *mom, QudaPrecision prec, int len)
   } else if (prec == QUDA_SINGLE_PRECISION) {
     action = mom_action<float>((float *)mom, len);
   }
-  comm_allreduce(&action);
```
mathiaswagner, line 1615: `quda::` ?

mathiaswagner (commented):
Any thoughts on clang-format for the files that got a `namespace quda { ...}` added?

lib/gauge_qcharge.cu
```diff
@@ -31,7 +31,7 @@ namespace quda
     {
       TuneParam tp = tuneLaunch(*this, getTuning(), getVerbosity());
 
-      std::vector<double> result(3);
+      array<double, 3> result;
```
mathiaswagner: just wondering: Is this std::array or quda::array and since I am wondering would it be better to be more verbose here?

maddyscientist: This is `quda::array` type, we certainly don't have any `using namespace std;` statements in QUDA.

maddyscientist: While this isn't appropriate for all reductions, I've updated gauge_qcharge.cu to use the `reduce_t` as opposed to needing to know the reduction type.  96ee997e1118bd7d8cb8888956a4a469a6b35a95

tests/unitarize_link_test.cpp
maddyscientist: 57a14f2df68d0e75121fa85d92e862a1935c6d27

tests/staggered_invert_test.cpp
maddyscientist: 57a14f2df68d0e75121fa85d92e862a1935c6d27

tests/utils/host_utils.cpp
maddyscientist: 57a14f2df68d0e75121fa85d92e862a1935c6d27

tests/hisq_unitarize_force_test.cpp
maddyscientist: 57a14f2df68d0e75121fa85d92e862a1935c6d27

maddyscientist:
> Any thoughts on clang-format for the files that got a `namespace quda { ...}` added?

As you may have expected, the files that have been newly namespaced all end up with significant reformatting.  I haven't yet apply clang format to those files yet as a result.  What are your thoughts @mathiaswagner?  I'm ok with formatting them regardless.

maddyscientist:
@bjoo, @jcosborn, @jxy, @alexstrel knock on changes that you should be aware of that may require updates in your target branches:
* Contents of lib/targets/XXX/comm_target.cpp should now be in the `quda` namespace
* Contents of lib/targets/XXX/FFT_Plans.h should now be in the `quda` namespace
* For consistency with the multi-reductions, the block reduction kernel batching thread dimension is now the Z dimension, and not the Y dimension.  The only change needed is to make sure that both the y and z dimensions are filled out in the `block_idx` and `thread_idx` objects that are passed to the functors (include/targets/XXX/block_reduction_kernel.h).  (Previously on the CUDA target, the z dimensions were zero-ed out)

@bjoo Not sure if my PR or yours will be merged first.  Regardless, the updates needed will be fairly trivial.

tests/unitarize_link_test.cpp
maddyscientist: I've made the changes suggested here, but I see that the tests are littered with `using namespace quda`, so cleaning that up goes way beyond the scope of this PR.

mathiaswagner:
I would be happy to format the newly-namespaces files. Some conflicts are to be expected but I don't think later will be a better time than now.

lib/dslash_pack2.cu
```diff
@@ -16,14 +16,14 @@
 namespace quda
 {
 
-  static int commDim[QUDA_MAX_DIM];
+  static int comm_dim_pack[QUDA_MAX_DIM];
```
hummingtree, line 19: What is the purpose of this name change here?

maddyscientist: Because of the namespace cleanup, this conflicted with `int commDim(int)` in `comm_quda.h`.

include/kernels/block_orthogonalize.cuh
```diff
@@ -63,7 +63,7 @@ namespace quda {
     template <typename... T>
     BlockOrthoArg(ColorSpinorField &V, const int *fine_to_coarse, const int *coarse_to_fine, int parity,
                   const int *geo_bs, const int n_block_ortho, const ColorSpinorField &meta, T... B) :
-      kernel_param(dim3(meta.VolumeCB() * (fineSpin > 1 ? meta.SiteSubset() : 1), chiral_blocks, 1)),
+      kernel_param(dim3(meta.VolumeCB() * (fineSpin > 1 ? meta.SiteSubset() : 1), 1, chiral_blocks)),
```
weinbe2, line 66: Out of curiosity, what type of performance difference does this make?

include/kernels/madwf_tensor.cuh
```diff
@@ -68,6 +68,9 @@ namespace quda
       constexpr Tensor5DReduce(const Arg &arg) : arg(arg) { }
       static constexpr const char *filename() { return KERNEL_FILE; }
 
+      // overload comm_reduce to prevent any global reduction
+      static inline void comm_reduce(std::vector<reduce_t> &) { }
```
weinbe2, line 72: This approach seems prone to "gotchas". Can we do something along the lines of a `push`/`pop` disabling global reductions in stretches of code instead? (Maybe we already have something like this?)

include/kernels/block_orthogonalize.cuh
maddyscientist: None.  It's a cleanup really for consistency with the multi-reductions where the batched index is in the z dimension.

It's a bit more consequential in the restrictor, where we actually have a batch size greater than 1 in the block.  There we were mapping the y thread index to the batch index, which is the z index in the block reducer.  It didn't break anything, or rely on UB, but it gave me nightmares, e.g., if CUB changed its internals somehow, it could break.  So making this all consistent was a good thing to do.

include/tunable_block_reduction.h
```diff
@@ -22,115 +22,127 @@ namespace quda
...
+       @param[in] tp The launch parameters
+       @param[in] stream The stream on which the execution is done
      */
     template <template <typename> class Functor, typename Block, bool enable_host = false, typename Arg>
-    std::enable_if_t<!enable_host, void> launch(const TuneParam &tp, const qudaStream_t &stream, const Arg &arg)
+    void launch(const TuneParam &tp, const qudaStream_t &stream, const Arg &arg)
     {
       if (location == QUDA_CUDA_FIELD_LOCATION) {
         launch_device<Functor, Block>(tp, stream, arg);
+      } else if constexpr (enable_host) {
```
weinbe2, line 124: I don't know if I like this being popped into this PR --- is there a workflow that tests this?

maddyscientist: not sure I follow.  This isn't a few feature.  It's just a cleanup of the prior code to use `if constexpr` instead of SFINAE.  And yes, this is used all over the place, e.g., the copy kernels.

include/kernels/madwf_tensor.cuh
maddyscientist: push/pop is what we want to be avoiding here.  A kernel functor should know what type of reduction it's doing, and if it's doing a global or local reduction.

include/tunable_reduction.h
```diff
@@ -289,39 +232,41 @@ namespace quda
...
     template <int block_size_x, template <typename> class Functor, typename FunctorArg>
-    std::enable_if_t<block_size_x != device::warp_size(), qudaError_t> launch(FunctorArg &arg, const TuneParam &tp,
-                                                                              const qudaStream_t &stream)
+    qudaError_t launch(FunctorArg &arg, const TuneParam &tp, const qudaStream_t &stream)
     {
       if (tp.block.x == block_size_x) {
         using Arg = ReduceKernelArg<block_size_x, block_size_y, FunctorArg>;
         return TunableKernel::launch_device<Functor, grid_stride>(KERNEL(MultiReduction), tp, stream, Arg(arg));
-      } else {
+      } else if constexpr (block_size_x != device::warp_size()) {
```
weinbe2, line 249: ~Should there be an extra `else` after this that throws an error?~ Nevermind, it's a handful of lines down, the way `git` handled the diff threw me off.

include/kernels/madwf_tensor.cuh
weinbe2: I'm reminded that this type of mechanism does exist and can be queried by `commGlobalReduction();`

maddyscientist: This abstraction also allows us, for example, to implement custom global reductions for a given functor.  E.g., a time-slice reduction for a contraction could be implemented here (or we could define it in a parent class) and derive from it.

maddyscientist: Yes, we have the push/pop functionality.  That's more for the case where we want to enable/disable global reductions at run time, e.g., switching between a Schwarz preconditioner and the outer solver.  This functionality here is for hard disablement at compile time, where know we never want to have a global reduction, and / or provide a custom one.

weinbe2: Okay, I now understand and appreciate the compile time vs runtime motivation, thank you for explaining that. I think the way it's implemented is still a bit prone to confusion---overriding a global-scope function with a `static inline` class routine, if I'm gathering it correctly.

While it's not something I'm worried about here (we're in a *.cuh), I could fathom a situation where:
 * I override `comm_reduce` in my class as such
 * Within my class, I call some function that includes reductions, but is defined in another `*.o` in a context where `comm_reduce` has _not_ been overriden
 
In this contrived example, I think I'd get "unexpected" (though strictly correct) behavior, where some functions have global reductions enabled, while others don't.

This might be too contrived to be worth worrying about in practice, but I think it's technically still a risk?

maddyscientist: Not following the "global-scope" aspect of this.  It's a static class function, and not global.

weinbe2: Resolved offline---I missed (aka overlooked reading) how `comm_reduce` is a member of the parent reduction class, so I ended up running myself into a bit of a confused circle.

include/targets/hip/block_reduction_kernel.h
```diff
@@ -64,10 +63,12 @@ namespace quda
   template <template <typename> class Functor, typename Arg>
   __forceinline__ __device__ void BlockKernel2D_impl(const Arg &arg)
   {
-    const dim3 block_idx(virtual_block_idx(arg), blockIdx.y, 0);
-    const dim3 thread_idx(threadIdx.x, threadIdx.y, 0);
+    const dim3 block_idx(virtual_block_idx(arg), blockIdx.y, blockIdx.z);
+    const dim3 thread_idx(threadIdx.x, threadIdx.y, threadIx.z);
```
bjoo: should be threadIdx  instead of threadIx ?

maddyscientist: Thank you

bjoo:
This is a pretty big PR, so let me check it with a build. Be back ASAP...

weinbe2 (approved):
Looks good from a staggered standpoint!

mathiaswagner (approved):
looks good to me and passed a quick test in Chroma

bjoo (approved):
With the couple of typos (fixed already) this branch merged into develop compiles fine on HIP. I can give it some runtime testing, but I will trust the other approvers and approve it now to not be the one person holding it back. If I find anything at runtime it will need to be tracked anyhow, and then I can raise an issue.

bjoo:
I  am seeing a funny, that even when I turn off QMP, somehow the tests seem to have some form of QMP linkage... This may just be my fault for not doing a clean build tho, so don't let this hold you up.

mathiaswagner:
@bjoo please file an issue if that persits with a clean build. Since tests and build system have not been touched here I don't think this is related.

bjoo:
@mathiaswagner  I suspect this is my preset `LD_LIBRARY_PATH` pointing to `libquda.so` from a different build but I am checking to make sure.

bjoo:
The link path turned out to be the issue. So my own fault. Please carry on....

PR #1259, opened 2022-04-01
Hotfix/overflow test fixes

maddyscientist:
Quick fix PR:
* Check dslash policy is valid when enabling it
* Use 64-bit variables for `*_site_size` variables in the host reference code: this fixes a variety of 32-bit overflow in the host reference code, and for example enables use to run clover fermions at a volume of 72^4, where previously the limit was < 70^4

### DIFF
--- a/lib/dslash_policy.cuh
+++ b/lib/dslash_policy.cuh
@@ -1736,7 +1736,7 @@ namespace quda
   inline void enable_policy(QudaDslashPolicy p)
   {
     size_t p_idx = static_cast<std::size_t>(p);
-    if (p > QudaDslashPolicy::QUDA_DSLASH_POLICY_DISABLED) errorQuda("Invalid policy %lu", p_idx);
+    if (p >= QudaDslashPolicy::QUDA_DSLASH_POLICY_DISABLED) errorQuda("Invalid policy %lu", p_idx);
     policies[p_idx] = p;
   }
 
--- a/lib/dslash_policy.cuh
+++ b/lib/dslash_policy.cuh
@@ -1733,7 +1733,12 @@ namespace quda
     }
   };
 
-  inline void enable_policy(QudaDslashPolicy p) { policies[static_cast<std::size_t>(p)] = p; }
+  inline void enable_policy(QudaDslashPolicy p)
+  {
+    size_t p_idx = static_cast<std::size_t>(p);
+    if (p > QudaDslashPolicy::QUDA_DSLASH_POLICY_DISABLED) errorQuda("Invalid policy %lu", p_idx);
+    policies[p_idx] = p;
+  }
 
   inline void disable_policy(QudaDslashPolicy p)
   {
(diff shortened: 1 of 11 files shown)


tests/utils/staggered_host_utils.cpp
```diff
@@ -166,7 +166,7 @@ void constructFatLongGaugeField(void **fatlink, void **longlink, int type, QudaP
       const complex<double> z = std::polar(1.0, phase);
       for (int dir = 0; dir < 4; ++dir) {
         for (int i = 0; i < V; ++i) {
-          for (int j = 0; j < gauge_site_size; j += 2) {
+          for (unsigned int j = 0; j < gauge_site_size; j += 2) {
```
mathiaswagner: any particular reason for `unsigned int` here when `gauge_site_size` is `size_t`? Anyway, how about `auto` ?

mathiaswagner (changes requested):
I think a lot of the for loops changed here could just use `auto` ?

lib/dslash_policy.cuh
```diff
@@ -1733,7 +1733,12 @@ namespace quda
     }
   };
 
-  inline void enable_policy(QudaDslashPolicy p) { policies[static_cast<std::size_t>(p)] = p; }
+  inline void enable_policy(QudaDslashPolicy p)
+  {
+    size_t p_idx = static_cast<std::size_t>(p);
+    if (p > QudaDslashPolicy::QUDA_DSLASH_POLICY_DISABLED) errorQuda("Invalid policy %lu", p_idx);
```
hummingtree: The policy index should be strictly less than `QUDA_DSLASH_POLICY_DISABLED`?
```suggestion
    if (p >= QudaDslashPolicy::QUDA_DSLASH_POLICY_DISABLED) errorQuda("Invalid policy %lu", p_idx);
```

tests/utils/host_utils.cpp
```diff
@@ -1203,7 +1203,7 @@ template <typename Float> void constructCloverField(Float *res, double norm, dou
 
   Float c = 2.0 * norm / RAND_MAX;
 
-  for (int i = 0; i < V; i++) {
+  for (size_t i = 0; i < (size_t)V; i++) {
```
hummingtree:
```suggestion
  for (size_t i = 0; i < static_cast<size_t>(V); i++) {
```

tests/utils/staggered_host_utils.cpp
hummingtree: I am also wondering about the same thing here ...

tests/utils/staggered_host_utils.cpp
```diff
@@ -187,7 +187,7 @@ void constructFatLongGaugeField(void **fatlink, void **longlink, int type, QudaP
   if (dslash_type == QUDA_STAGGERED_DSLASH) {
     for (int dir = 0; dir < 4; ++dir) {
       for (int i = 0; i < V; ++i) {
-        for (int j = 0; j < gauge_site_size; j += 2) {
+        for (unsigned int j = 0; j < gauge_site_size; j += 2) {
```
hummingtree:
```suggestion
        for (auto j = 0lu; j < gauge_site_size; j += 2) {
```

hummingtree (approved):
Thanks @maddyscientist for the quick fix!

mathiaswagner (approved):

PR #1260, opened 2022-04-06
prototype for a 2-step tuning process.

mathiaswagner:
First iterate over full tune space with 2 iterations and keep 5 best candidates.
Do a longer measurement (at least 1e-5s or 3 iterations) with these 5 candidates to find the best launch parameters.

### DIFF
--- a/include/tune_quda.h
+++ b/include/tune_quda.h
@@ -245,31 +245,33 @@ namespace quda {
     virtual void postTune() { }
 
     /**
-     * @brief Parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
-     *
-     * @return int minimum number of iterations
-     */
-    virtual int min_tune_iter() const { return 3; }
-    /**
-     * @brief Parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
-     *
-     * @return float
-     */
-    virtual float min_tune_time() const { return 1e-3; }
-
-    /**
      * @brief Number of iterations used in the 1st phase of tuning, i.e. finding the candidates for the 2nd phase/
      *
-     * @return int number of iterations
+     * @return number of iterations
      */
     virtual int candiate_iter() const { return 2; }
 
     /**
-     * @brief Number of candidates to use for the 2dn tuning phase
+     * @brief Number of candidates to be identified in the 1st phase for the 2nd tuning phase
      *
-     * @return size_t number of candidates
+     * @return number of candidates
      */
     virtual size_t num_candidates() const { return 10; }
+
+    /**
+     * @brief Parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
+     *
+     * @return minimum number of iterations
+     */
+    virtual int min_tune_iter() const { return 3; }
+
+    /**
+     * @brief Time parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
+     * This controls that the measured time for the number of iterations is at least this long based on the time from phase 1.
+     *
+     * @return minimum time that should be measured
+     */
+    virtual float min_tune_time() const { return 1e-3; }
 
     virtual std::string paramString(const TuneParam &param) const
     {
--- a/include/tune_quda.h
+++ b/include/tune_quda.h
@@ -243,7 +243,6 @@ namespace quda {
     virtual void apply(const qudaStream_t &stream) = 0;
     virtual void preTune() { }
     virtual void postTune() { }
-    virtual int tuningIter() const { return 1; }
 
     /**
      * @brief Parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
--- a/include/tune_quda.h
+++ b/include/tune_quda.h
@@ -249,7 +249,7 @@ namespace quda {
      *
      * @return number of iterations
      */
-    virtual int candiate_iter() const { return 2; }
+    virtual int candidate_iter() const { return 2; }
 
     /**
      * @brief Number of candidates to be identified in the 1st phase for the 2nd tuning phase
--- a/include/tune_quda.h
+++ b/include/tune_quda.h
@@ -245,6 +245,33 @@ namespace quda {
     virtual void postTune() { }
     virtual int tuningIter() const { return 1; }
 
+    /**
+     * @brief Parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
+     *
+     * @return int minimum number of iterations
+     */
+    virtual int min_tune_iter() const { return 3; }
+    /**
+     * @brief Parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
+     *
+     * @return float
+     */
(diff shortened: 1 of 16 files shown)


lib/tune.cpp
```diff
@@ -787,7 +819,8 @@ namespace quda
         param.aux = make_int4(-1, -1, -1, -1);
         tunable.initTuneParam(param);
 
-        while (tuning) {
+        constexpr int candiate_iterations = 2;
```
maddyscientist: probably want to leave this as a variable that can be changed by the user

lib/tune.cpp
```diff
@@ -678,6 +680,35 @@ namespace quda
...
 
+  using tunepair = std::pair<float, TuneParam>;
+  struct tunepaircomp {
+    bool operator()(const tunepair &left, const tunepair &right) const { return (left.first) < (right.first); }
+  };
+
+  class TuneCandidates : public std::priority_queue<tunepair, std::vector<tunepair>, tunepaircomp>
+  {
+  private:
+    static constexpr size_t maxSize = 5;
```
maddyscientist: style comment: prefer `max_size`, I think we should be moving away from camel case, especially for member variables (I'll _tolerate_ it for member functions).

lib/tune.cpp
```diff
@@ -678,6 +680,35 @@ namespace quda
 
   static TimeProfile launchTimer("tuneLaunch");
 
+  using tunepair = std::pair<float, TuneParam>;
```
maddyscientist: Is this new type needed: can't we use the `time` variable already in the `TuneParam`, and just use a queue of `TuneParam`s?

lib/tune.cpp
mathiaswagner: good point.

lib/tune.cpp
```diff
@@ -801,7 +834,59 @@ namespace quda
...
+            }
+          }
+          candidatetuning = tunable.advanceTuneParam(param);
+          tunable.launchError() = QUDA_SUCCESS;
+        }
+
+        if (tc.empty()) { errorQuda("Auto-tuning failed for %s with %s at vol=%s", key.name, key.aux, key.volume); }
+
+        constexpr float min_tune_time = 1e-5f;
+        const int tuneiterations = std::max(static_cast<int>(std::ceil(min_tune_time / tc.getBestTime())), 3);
```
maddyscientist: Like the candidate iterations, we should leave the `3` more exposed and not buried in this equation.

maddyscientist: Also, I think the default here should 10.  Given you have the timer cap, there's no reason not to use this.

mathiaswagner: This is to make sure we *at least* do 3 iterations here. The timer cap is to make sure we do more iterations on fast running kernels. And `min_time_tune` should get a sensible default to make sure we don't pick up too much noise / timing inaccuracy.

maddyscientist: Ha, I read it backwards 😆

lib/tune.cpp
weinbe2: I'm fine with baking this in at 2, combined with some minimum time requirement.

lib/tune.cpp
weinbe2: I think `5` or `10` would be a good default (actually, some preference for `10`, given how much we're thinning the space already). I think this should also be overridable via a `virtual` class function, as well as a global override via an environment variable, where by "global" I mean ignore this value here and all `virtual` values.

The motivation for the "global" override is for development purposes, sometimes it's worthwhile to have a sketchily tuned kernel as opposed to the current alternative of outright disabling tuning.

mathiaswagner: 86b46b64e4871473715e683dfc1ab46d5de8d18c

lib/tune.cpp
mathiaswagner: 91c29f86284deea69ca9b6c87fdadcb3525a6cb9

lib/tune.cpp
mathiaswagner: we have this as a virtual function now

lib/tune.cpp
```diff
@@ -678,6 +680,54 @@ namespace quda
...
+   */
+  class TuneCandidates : public std::priority_queue<TuneParam, std::vector<TuneParam>, TuneParamComp>
+  {
+  private:
+    const size_t max_size = 5;
+    float besttime = FLT_MAX;
+
+  public:
+    /**
+     * @brief Construct a new Tune Candidates object with specifiwd size
```
hummingtree:
```suggestion
     * @brief Construct a new Tune Candidates object with specified size
```

lib/tune.cpp
```diff
@@ -678,6 +680,54 @@ namespace quda
 
   static TimeProfile launchTimer("tuneLaunch");
 
+  using tunepair = std::pair<float, TuneParam>;
+  struct tunepaircomp {
+    bool operator()(const tunepair &left, const tunepair &right) const { return (left.first) < (right.first); }
+  };
```
hummingtree: This struct is no longer used anymore?

mathiaswagner: Yes, forgot to remove that. Thanks!

include/tune_quda.h
```diff
@@ -243,7 +243,33 @@ namespace quda {
...
+
+    /**
+     * @brief Number of iterations used in the 1st phase of tuning, i.e. finding the candidates for the 2nd phase/
+     *
+     * @return int number of iterations
+     */
+    virtual int candiate_iter() const { return 2; }
+
+    /**
+     * @brief Number of candidates to use for the 2dn tuning phase
```
maddyscientist: typo: 2dn -> 2nd

include/tune_quda.h
```diff
@@ -243,7 +243,33 @@ namespace quda {
     virtual void apply(const qudaStream_t &stream) = 0;
     virtual void preTune() { }
     virtual void postTune() { }
-    virtual int tuningIter() const { return 1; }
+
+    /**
+     * @brief Parameter to control the number of iteration used in the 2nd phase of tuning, i.e. for the candidates.
+     *
+     * @return int minimum number of iterations
```
maddyscientist: The `return` statement shouldn't include the type being returned (same applies for the next few functions in this file).

include/tune_quda.h
```diff
@@ -243,7 +243,33 @@ namespace quda {
...
+     */
+    virtual int min_tune_iter() const { return 3; }
```
maddyscientist: I think the order of these declarations should be reversed, such that the parameters for the first stage of tuning should come first, and the candidate (2nd) tuning should come second.

maddyscientist (commented):
Great addition.  Beyond the quick-to-fix comments I've left, this PR is good to go.  I'll approve once those fixes have been applied.

maddyscientist (approved):

mathiaswagner:
Before merging that I need to run some more test with Chroma + NVSHMEM.

mathiaswagner:
Did some more tests and things are looking good now.

weinbe2 (approved):
This has passed all my various correctness and perf tests, so this is good to go. Great contribution @mathiaswagner !

PR #1261, opened 2022-04-08
Feature/multiblas adapt

jcosborn:
Make multi_blas arg use constant mem if too big for kernel parameter.
Fix 2D FFT.
Some other small fixes.


include/multi_blas_helper.cuh
```diff
@@ -206,21 +211,25 @@ namespace quda
...
       constexpr auto arg_known_size = (sizeof(kernel_param<>)                                        // kernel_param parent
-                                       + sizeof(int)                                                 // NYW parameter
-                                       + sizeof(SpinorX[NXZ])                                        // SpinorX array
-                                       + (Functor::use_z ? sizeof(SpinorZ[NXZ]) : sizeof(SpinorZ *)) // SpinorZ array
-                                       + sizeof(Functor)                                             // functor
-                                       + sizeof(dim3)                                                // threads parameter
-                                       + (!Functor::use_w ? sizeof(SpinorW *) : 0)                   // subtract pointer if not using W
-                                       + (Functor::reducer ? sizeof(ReduceArg<device_reduce_t>) : 0) // reduction buffers
-                                       );
+				       + sizeof(int)                                                 // NYW parameter
```
maddyscientist, line 217: clang-format will catch this, but noting that this diff here mostly arises due to using tab indents

maddyscientist (approved):

maddyscientist:
Thanks for this @jcosborn.  I have confirmed the issue you fix with the FFT plans: https://docs.nvidia.com/cuda/cufft/index.html#function-cufftplanmany
>  Array of size rank, describing the size of each dimension, n[0] being the size of the outermost and n[rank-1] innermost (contiguous) dimension of a transform.

The point being that the argument is an array specifying the dimensions with slowest running to fastest order.  E.g., exactly as you have fixed.  I have added a comment to the source to make this clear.

Once Jenkins is finished we can merge this.

PR #1263, opened 2022-04-12
remove clear from LatticeField/ColorSpinirField 

mathiaswagner:
and fix blas and dslash test for that


maddyscientist (approved):

maddyscientist:
This PR is a good cleanup.  I have pushed a further cleanup of blas_test.cpp to remove the use of `std::vector<ColorSpinorField*>` replacing it with `std::vector<ColorSpinorField>`.

PR #1267, opened 2022-04-29
Feature/dilution

maddyscientist:
This PR adds dilution functionality to QUDA
* Interface function is `spinorDilute`, which dilutes a fermion field into a vector of fields
* Initial dilution types supported are:
  * spin
  * color
  * spin-color
  * spin-color-even-odd (only applicable to full fields)
* An accompanying regression test is included `dilution_test` which tests that the various dilution patterns are all lossless (e.g., the original vector can be exactly reconstructed), testing both full fields and single parity fields

I also snuck in a compilation fix for nvc++ (which doesn't yet support `__builtin_assume`)


tests/CMakeLists.txt
```diff
@@ -1049,4 +1055,11 @@ foreach(prec IN LISTS TEST_PRECS)
                      --gtest_output=xml:gauge_arg_test_${prec}.xml)
   endif()
 
+  if (QUDA_NSPIN4)
```
mathiaswagner: Better use 
```suggestion
  if (TARGET dilution_test)
```

tests/dilution_test.cpp
```diff
@@ -0,0 +1,148 @@
+#include <stdlib.h>
```
mathiaswagner: Use C++ version of headers here?

weinbe2:
This is a great, narrow PR, thanks @maddyscientist ! My only ask is if you can extend the test to cover Nspin = 1 as well? It looks like that'd require a new `constructStaggeredTestSpinorParam` function, but that should be relatively straightforward.

tests/dilution_test.cpp
maddyscientist: These headers weren't even needed, so I just deleted them 5ca61ad2adaa22da1d40d28ab496bc1dd018ea2b

tests/CMakeLists.txt
maddyscientist: Done 5ca61ad2adaa22da1d40d28ab496bc1dd018ea2b

maddyscientist:
Good suggestion @weinbe2 for adding a staggered test: done with 5ca61ad2adaa22da1d40d28ab496bc1dd018ea2b.

Jenkins should now pass as well.

mathiaswagner (approved):
LGTM. Clean addition.

weinbe2 (approved):
This looks great to me. Thanks @maddyscientist !

hummingtree (approved):
This looks nice and clean to me!

PR #1271, opened 2022-05-04
Alternative `i2f`

### LINKED ISSUE #1270: The alternative `i2f` no longer works
The alternative `i2f` conversion routine [here](https://github.com/lattice/quda/blob/466e5dfe9ec38fe8b217cc784a445169022c8842/include/convert.h#L79) (the `#if 0` version) no longer works with the latest CUDA compiler, since in principle writing to one member of a union and then accessing another member is an undefined behavior. Per https://en.cppreference.com/w/cpp/language/union,

> It's undefined behavior to read from the member of the union that wasn't most recently written. Many compilers implement, as a non-standard language extension, the ability to read inactive members of a union.

hummingtree:
- Add `QUDA_ALTERNATIVE_I_TO_F` as a cmake option.
- Fix the alternative `i2f`: the original code no longer work as described in #1270. In principle the fix is still undefined behavior, but it works for now.

Closes #1270.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,6 +233,8 @@ option(QUDA_CTEST_DISABLE_BENCHMARKS "Disable benchmark test" ON)
 option(QUDA_FAST_COMPILE_REDUCE "enable fast compilation in blas and reduction kernels (single warp per reduction)" OFF)
 option(QUDA_FAST_COMPILE_DSLASH "enable fast compilation in dslash kernels (~20% perf impact)" OFF)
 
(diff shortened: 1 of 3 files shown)


maddyscientist (approved):
Curious: have you seen any benefit for this path at all?

hummingtree:
> Curious: have you seen any benefit for this path at all?

I have tried with some of the new DWF kernels I was experimenting with. It does what it does, i.e. reducing the XU pipe utilization at the expense of more FMA, but I have not seen any cases where it leads to visible speed ups (yet).

PR #1272, opened 2022-05-06
Staggered MG hotfix: field location and zeroing for coarse KD build.

weinbe2:
This quick hotfix PR fixes some holes in appropriately zeroing fields (we lucked out on UB before) and setting field locations for staggered MG when not using the optimized KD pathway.


maddyscientist (approved):
Looks good.

Issue #1273, opened 2022-05-06
2.8 * M_PI, is it a typo?


jxy:
https://github.com/lattice/quda/blob/6d16d9e06ac1aca084638db116b0dd1a0ab035b6/include/targets/cuda/math_helper.cuh#L96

maddyscientist:
That is a typo.  Thanks for the catch.  It's a harmless one, since `a < 2 π`, but good to fix

maddyscientist:
Bug fix pushed, will close then when the branch in question is merged in.

PR #1277, opened 2022-05-17
Feature/reduce runtime blocksize

maddyscientist:
This PR's main addition is the change of QUDA utilizing a compile-time definition of the block size in the (block) reduction to one deployed using a run-time block size
* All kernels that utilize either the `reduce` function or the `BlockReduce` helper now do so using the run-time block size only.
* Major reduction in compilation time since we no longer instantiate a kernel template for each block size
* Major reduction in library binary size as well
* In the CUDA target, we can now use either cooperative groups or CUB for the warp reduction
  * Present default is cooperative groups (`#define USE_CG` at the top of block_reduction_helper.h)  
* No measurable performance regression from this change, and indeed, many reduction kernels are actually faster
* Added heuristics to reduce autotuning time for (multi-) reduction kernels (pruning the grid space)
* This PR affects the target-specific code.  I have already applied the necessary changes for HIP, @jcosborn you'll likely need to make some tweaks at your end for your SYCL branch
* Improved compilation time for multi-blas through avoiding unnecessarily instantiated templates (`NXZ = 128` is not ever needed).  This also avoids any local memory in these kernels.
* `atomic_type` helper for splitting up objects into atomic safe constituents is now done through explicit instantiation to prevent silent errors which reared its head during development of this branch
* Improved robustness of multigrid verification to prevent erroneous test passing if `Nan` or `inf` permeates the result
* Fixed a bug in the plaquette kernel which reared its head with the changes in this branch
* Fixed a bug in include/targets/cuda/math_helper.cuh (#1273)
* Some compiler warnings fixed from when using clang 14


maddyscientist:
Regarding portability:
* All changes required for HIP should have been done already.  I have confirmed that the HIP target compiles at least, @bjoo I'll leave to you to test how well this works on red team
* @jcosborn Some changes for your SYCL branch will be needed here, but I imagine nothing to onerous.

include/targets/cuda/reduce_helper.h
```diff
@@ -180,63 +237,46 @@ namespace quda
...
-
-    // finish the reduction if last block
-    if (isLastBlockDone[target::thread_idx().z]) {
-      auto i = target::thread_idx().y * block_size_x + target::thread_idx().x;
-      T sum = r.init();
-      while (i < target::grid_dim().x) {
-        sum = r(sum, arg.partial[idx * target::grid_dim().x + i].load(cuda::std::memory_order_relaxed));
-        i += block_size_x * block_size_y;
+        // determine if last block
+        isLastBlockDone[target::thread_idx().z] = (value == (target::grid_dim().x - 1));
```
hummingtree, line 265: Maybe I am missing something, but shouldn't here be `value == target::grid_dim().x * target::grid_dim().y - 1`?

hummingtree: And the above short circuit should take `target::grid_dim().y` into account as well?

maddyscientist: No, this is intentional.  As per the doxygen for the reduction kernels:
> The y thread dimension is constrained to remain inside the thread block and this dimension is contracted in the reduction.

maddyscientist: In other words the global reduction is only done over the x dimension.  So in summary:
* x: global reduction
* y: block reduction
* z: batch dimension, no reduction done

hummingtree: Got it. Thanks.

hummingtree (approved):
Getting rid of the block size template is just too good to be true. :) Thanks for this great clean up.

mathiaswagner:
Chroma looks good now. @weinbe2 we probably should check with MILC.

include/color_spinor_field_order.h
```diff
@@ -876,6 +876,20 @@ namespace quda
 
 #endif
 
+      template <typename reducer, typename helper>
```
weinbe2, line 889: doxygen

include/kernels/multi_reduce_core.cuh
```diff
@@ -15,16 +15,10 @@ namespace quda
 
     /**
        @brief Return the batch block size used for multi reductions.
-       For now, we leave this at 1 unless fast-compilation is enabled,
-       since it provides negligible benefit (but substantially longer
-       tuning).  When fast-compilation is enabled, the increased batch
-       count can dramatically improve performance.
+       For now, we leave this at 1, but this is something that will be
```
weinbe2: Typo, should be "8"

maddyscientist: Actually that comment is completely wrong now since fast reduce compilation is gone now.  Will update in next push

include/targets/cuda/reduce_helper.h
```diff
@@ -167,76 +151,137 @@ namespace quda
...
+      if constexpr (coalesced_write) {
+        static_assert(n <= device::warp_size(), "reduction array is greater than warp size");
+        auto mask = __ballot_sync(0xffffffff, target::thread_idx().x < n);
+        if (target::thread_idx().x < n && target::thread_idx().y == 0) {
+          atomic_t sum_tmp[n];
+          memcpy(sum_tmp, &sum, sizeof(sum));
+
+          atomic_t s = sum_tmp[0];
+#pragma unroll
+          for (int i = 1; i < n; i++) {
```
weinbe2, line 185: I believe there's a corner case where this could break if `blockDim.x < n <= device::warp_size()` --- is this a situation worth worrying about?

include/targets/generic/reduce_helper.h
```diff
@@ -143,17 +141,17 @@ namespace quda
 
     // finish the reduction if last block
     if (isLastBlockDone[target::thread_idx().z]) {
-      auto i = target::thread_idx().y * block_size_x + target::thread_idx().x;
+      auto i = target::thread_idx().y * target::block_dim().x + target::thread_idx().x;
```
weinbe2: `thread_idx_linear`

include/targets/generic/reduce_helper.h
```diff
@@ -143,17 +141,17 @@ namespace quda
...
         sum = r(sum, const_cast<T &>(static_cast<volatile T *>(arg.partial)[idx * target::grid_dim().x + i]));
-        i += block_size_x * block_size_y;
+        i += target::block_dim().x * target::block_dim().y;
       }
 
       sum = BlockReduce(target::thread_idx().z).Reduce(sum, r);
 
       // write out the final reduced value
-      if (target::thread_idx().y * block_size_x + target::thread_idx().x == 0) {
+      if (target::thread_idx().y * target::block_dim().x + target::thread_idx().x == 0) {
```
weinbe2: `thread_idx_linear`

maddyscientist: 7eba3b448a0f8cc73db93469991e3d6ae0b911d0

include/targets/generic/reduce_helper.h
maddyscientist: 7eba3b448a0f8cc73db93469991e3d6ae0b911d0

include/targets/cuda/reduce_helper.h
maddyscientist: 7eba3b448a0f8cc73db93469991e3d6ae0b911d0

include/kernels/multi_reduce_core.cuh
maddyscientist: aa3f7b9db9b23ed8a60628df9b078e1d9076d216

include/color_spinor_field_order.h
maddyscientist: c78f27c0a8e1af7736657bdca3c8b710254317f7

tests/utils/host_utils.cpp
```diff
@@ -98,7 +98,7 @@ void setQudaDefaultMgTestParams()
     smoother_solve_type[i] = QUDA_INVALID_SOLVE;
     mg_schwarz_type[i] = QUDA_INVALID_SCHWARZ;
     mg_schwarz_cycle[i] = 1;
-    smoother_type[i] = QUDA_GCR_INVERTER;
+    smoother_type[i] = QUDA_MR_INVERTER;
```
weinbe2, line 101: What was this change for?

weinbe2 (approved):
Minus a spurious question (and mayhaps the need for a `clang-format`?) I approve of this PR. Great clean-up/optimization work!

tests/utils/host_utils.cpp
maddyscientist: Because this was the documented default.  It seemingly got changed seemingly, I assume when the test clean up was done.  Good to have docs match actual behavior.

mathiaswagner (approved):

bjoo (approved):
Finally. I figured out how to review this. - after accidentally taking myself off the reviewer list and adding Carleton on and taking him off again.... 

In any case I built the code on Crusher, and ran it. It seems fine functionally in terms of HMC and perf on 1 node is in line with expectations.

PR #1257, opened 2022-03-17
Feature/colorspinor default

maddyscientist:
This is a significant PR which consists of foundational changes, cleanup and improved testing:
* Introduce null constructor, move constructor and move assignment operator to `ColorSpinorField` class.   The long-term goal here is to enable the complete removal of heap storage of `ColorSpinorField`.
* Adapted both `LatticeField` and `ColorSpinorField` to be more amenable to move assignment, replacing C-style arrays with `array` objects, and `char *` with `std::string`.
* Added new `lat_dim_t` type, to be used for specifying n-dimensional quantities (replaces C-style array with a moveable object)
* Introduce `array_2d` and `array_3d` objects, which are nested `array` types
* Introduced a new interface for the multi-blas kernels, instead of `std::vector<ColorSpinorField*>` we now use `std::vector<std::reference_wrapper<ColorSpinorField>>`.  A compatibility layer using `std::vector<ColorSpinorField*>` remains for now, until the heap -> stack solver cleanup is universal.
* Significant cleanup of many solvers, and other functions, enabled by both null constructor and move assignment (I have intentionally not rewritten all solvers in an attempt to narrow the scope of this PR).  Clean up can be done on a more narrow basis in follow up work.
* Introduced new method `Solver::get_residual()` for obtaining the residual from a solve, and deployed this in multigrid, removing the need for the hack that overwrote the source vector.
* Introduced `logQuda` which allows for more concise verbosity guarded printing.
* Introduced automated testing of the solvers for all Wilson-type fermions, for both symmetric and asymmetric preconditioning if appropriate.  Focus here is on traditional Krylov solvers (CG, CGNR, CCGNE, CA-CG, CA-CGNR, CA-CGNE, GCR, CA-GCR, MR multi-shift CG).  Testing for Schwarz and multigrid is out of scope of this PR.
* Multiple bug fixes uncovered from a results of this much broader testing coverage.

PR #1278, opened 2022-05-19
Fixing location of fields

sbacchio:
In the current version of develop the `multigrid_benchmark_tests` is not running.
These changes fix this.

### DIFF
--- a/tests/multigrid_benchmark_test.cpp
+++ b/tests/multigrid_benchmark_test.cpp
@@ -74,6 +74,7 @@ void initFields(QudaPrecision prec)
   param.create = QUDA_ZERO_FIELD_CREATE;
   param.setPrecision(prec);
   param.fieldOrder = QUDA_FLOAT2_FIELD_ORDER;
+  param.location = QUDA_CUDA_FIELD_LOCATION;
 
   xD = std::make_unique<ColorSpinorField>(param);
   yD = std::make_unique<ColorSpinorField>(param);
@@ -99,6 +100,7 @@ void initFields(QudaPrecision prec)
   gParam.nFace = 1;
 
   gParam.geometry = QUDA_COARSE_GEOMETRY;
+  gParam.location = QUDA_CPU_FIELD_LOCATION;
   Y_h = new cpuGaugeField(gParam);
   Yhat_h = new cpuGaugeField(gParam);
 
@@ -121,6 +123,7 @@ void initFields(QudaPrecision prec)
   gParam.pad = gParam.nFace * pad * 2;
 
   gParam.setPrecision(prec_sloppy);
+  gParam.location = QUDA_CUDA_FIELD_LOCATION;
 
   Y_d = new cudaGaugeField(gParam);
   Yhat_d = new cudaGaugeField(gParam);


maddyscientist (approved):
Thanks for the fix @sbacchio

kostrzewa:
I haven't gotten around to submitting an issue yet but I've observed issues with field locations when calling the gradient flow routines from tmLQCD. Around the time that the GF stuff was merged, there was a lot of movement in `develop` and it's very hard to cherry-pick. I'll try to submit an issue as I think the situation is similar to what was observed here.

maddyscientist:
@kostrzewa  Issues you are seeing are likely stemming from #1257, which made a change requiring that we specify the location explicitly when creating gauge fields, as opposed to inferring the location from it being a `cpuGaugefield` or `cudaGaugeField`.  This was a precursor to a soon-to-be-written PR that will unify these fields into just being `GaugeField` as has already been done for the quark and clover fields.  I thought I caught all of the potential issues, but as evidenced by this PR and your continued issues, clearly I did not. 😞 

If you tell me the algorithms and/or kernels you are running, that may be enough for me to isolate the issue.

kostrzewa:
That would be extremely helpful!

We call the Wflow and (impliclty) a couple of gauge observables:

```
  QudaGaugeSmearParam wflow_params = newQudaGaugeSmearParam();
  wflow_params.smear_type = QUDA_GAUGE_SMEAR_WILSON_FLOW; 
  wflow_params.n_steps = (int)(tmax / eps) + 3; 
  wflow_params.epsilon = eps;
  wflow_params.meas_interval = 1;

  int n_meas= wflow_params.n_steps / wflow_params.meas_interval + 1 ;
  QudaGaugeObservableParam *obs_param;
  obs_param = (QudaGaugeObservableParam*) malloc(sizeof(QudaGaugeObservableParam) * n_meas);
  for (int i=0; i<n_meas; i++){
    obs_param[i] = newQudaGaugeObservableParam();   
    obs_param[i].compute_plaquette = QUDA_BOOLEAN_TRUE;
    obs_param[i].compute_qcharge = QUDA_BOOLEAN_TRUE; 
    obs_param[i].su_project = QUDA_BOOLEAN_TRUE; 
  }

  setVerbosityQuda(QUDA_SILENT, "# QUDA: ", stdout);  // suppress QUDA's Wflow stdout output
  performWFlowQuda(&wflow_params, obs_param);
  _setVerbosityQuda();  // this resets verbosity
```

maddyscientist:
@kostrzewa this patch may be what you need:
```$ git diff lib/
diff --git a/lib/gauge_observable.cpp b/lib/gauge_observable.cpp
index 063c381dc..6ab8467cd 100644
--- a/lib/gauge_observable.cpp
+++ b/lib/gauge_observable.cpp
@@ -34,6 +34,7 @@ namespace quda
     lat_dim_t x;
     for (int i = 0; i < 4; i++) x[i] = u.X()[i] - 2 * u.R()[i];
     GaugeFieldParam tensorParam(x, u.Precision(), QUDA_RECONSTRUCT_NO, 0, QUDA_TENSOR_GEOMETRY);
+    tensorParam.location = QUDA_CUDA_FIELD_LOCATION;
     tensorParam.siteSubset = QUDA_FULL_SITE_SUBSET;
     tensorParam.order = QUDA_FLOAT2_GAUGE_ORDER;
     tensorParam.ghostExchange = QUDA_GHOST_EXCHANGE_NO;
```
I've added this to #1277 which should be merged shortly

maddyscientist:
Actually, I see Evan has already fixed this, and the change is in develop already for the last 3 weeks.  Can you confirm the issue persists still?

maddyscientist:
And if it is still an issue, can you run with `QUDA_BACKWARDS=ON` enabled.  This will give a stack trace which should allow us to pin point the failure.

kostrzewa:
@maddyscientist Indeed it seems that when we first encountered the issue, dbc1e288315201f35f0069c60e4f936daad54d22 had not yet landed. Then I started a bisection which took a long while, at which point I left for two weeks! Thanks, it works now!

PR #1279, opened 2022-05-20
Reduce ctest time by not testing all partitions for all Dirac operators

mathiaswagner:
modify (staggered)_dslash_ctest to only run partitions 0, 1, 2, 3, 4, 7, 8, 15 by default.
New option `--all-partitions' add to these tests to enable all partitions.

ctest will test all-partitions for  wilson and asqtad.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -331,9 +331,10 @@ foreach(pol IN LISTS DSLASH_POLICIES)
 
   if(QUDA_DIRAC_WILSON)
     set(DIRAC_NAME wilson)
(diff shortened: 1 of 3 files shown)


maddyscientist:
This looks good, but the default partitions to test here should be:
* no partitioning (0)
* each dimension separately (1, 2, 4, 8)
* splitting multiple dimensions starting from T (12, 14, 15)

The rationale is that these are the most commonly used.

mathiaswagner:
Obviously you are right for partition choices and I reversed order accidentally. Thanks for catching this.

maddyscientist (approved):

PR #1281, opened 2022-05-25
remove debug information from STRICT build

mathiaswagner:
This is primarily used for CI, not development. Also no debug info in CXX already.


maddyscientist (approved):
How much disk space does this save?

mathiaswagner:
For a jitify build I have seen 200MB vs 600MB.

PR #1228, opened 2021-12-22
Feature/shmem exchangeghost

mathiaswagner:
This enables nvshmem for exchange ghost and a first initial version for dslash_coarse.

PR #1282, opened 2022-05-26
Hotfix/nvshmem 11.5

maddyscientist:
This PR fixes some regressions introduced with #1228 
* Move shmem dslash bits from `dslash_quda.h` to its own header  `dslash_shmem.h` which is only included in device target files (fixes issues with CUDA 11.5 `cuda/atomic` when compiled by gcc)
* Fix compilation of color_spinor_pack.cu which was broken if no fermion types are enabled

### DIFF
--- a/include/dslash_helper.cuh
+++ b/include/dslash_helper.cuh
@@ -7,6 +7,7 @@
 #include <shmem_helper.cuh>
 #include <fast_intdiv.h>
 #include <dslash_quda.h>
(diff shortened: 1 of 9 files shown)


weinbe2 (approved):
Works as advertised, thanks for the quick fix @maddyscientist ! Good to merge once Jenkins has completed

mathiaswagner (approved):

PR #1283, opened 2022-05-27
Refactor/expansion of MG setup methods

weinbe2:
This PR refactors and expands the number of methods by which near-null vectors can be generated in QUDA. Due to the nature of the refactor _and cleanup_, this PR is **interface breaking**, but in principle in a future-proof way -- near-null vector generation methods are now specified via an `enum`, `QudaNullVectorSetupType`, so it is straightforward to add more options in a non-breaking fashion. This PR also codifies the existing behavior that, if an input near-null vector is specified (via `--mg-load-vec` from the command line, for ex), it is loaded and all other options are ignored.

The full list of methods now includes:

- Inverse iterations (still the default)
- Chebyshev filter a la `arXiv:2103.05034, P. Boyle and A. Yamaguchi`
- Eigenvectors (as before)
- Free field vectors (as before, and only for sanity tests)
- Restricting as many near null vectors as possible from the finer level, then generating more from another method (that is --- if level 1 is Nc = 24, and level 2 is Nc = 32, then for level 2 all 24 near-null vectors from level 1 are restricted then 8 more are generated)

This PR also supports "polishing" near-null vectors generated by other methods with more iterations of inverse iterations.

The incomplete test vector support in QUDA has been mostly removed as it requires a fuller refactor that is outside of the scope of this PR, though it is on the to-do list in the future as it is a demonstrably successful approach.

### Command line arguments

The core command line argument, which has been repurposed, is

- `--mg-setup-type [level] [method]`, where `[method]` can be `inverse-iterations` (default), `chebyshev-filter`, `eigenvectors`, `test-vectors`, `restrict-fine`, and `free-field` (where `test-vectors` gracefully `errorQuda`s out)

**Inverse iterations**

No options for inverse iterations have been changed; `--mg-setup-tol`, `--mg-setup-maxiter`, etc, all behave as expected.

**Eigenvectors**

No options for eigenvectors have been changed.

**Chebyshev Filter**

The Chebyshev filter has a flexible set of parameters describing generating a set of near-null vectors related to the initial low-pass filter, the number of starting vectors, and subsequent generation from a low-passed starting vector.

 - `--mg-setup-filter-startup-vectors` - the number of random starting vectors, default 1. As some examples, if the number of near-null vectors for level 1 is 24, `--mg-setup-filter-startup-vectors 1 1` corresponds to one starting vector with 24 near-nulls generated; `[...] 1 3` corresponds to three starting vectors with 3 near-nulls generated from each, etc. In cases like `[...] 1 5`, where 5 doesn't divide into 24, 5 near-null vectors are generated from the first four starting vectors, and 4 from the last --- 4 * 4 + 4 = 24.
 - `--mg-setup-filter-startup-iterations` - number of iterations for the initial low-pass filter, default 1000.
 - `--mg-setup-filter-startup-rescale-frequency` - an empirical feature; since the norm of a vector could overflow (or individual values thereof), the vector can be renormalized with some frequency in a way that preserves the Chebyshev recursion; empirical default 50.
 - `--mg-setup-filter-lambda-max` - upper bound for the Chebyshev filters, default power iterations to guess an upper bound
 - `--mg-setup-filter-lambda-min` - lower bound to use for the initial low pass filter, modes smaller than that value are enhanced. Default 1.
 - `--mg-setup-filter-iterations-between-vectors` - number of iterations between subsequent near-null vectors after the initial low-pass filter. As an example, if startup iterations is 1000, and the number of iterations between vectors is 150, a near-null vector is generated after 1000 (initial) matrix applications, then at 1150, 1300, 1450... . Default 150.

**Restriction**

There are no special flags for restriction in and of itself, but:
 - `--mg-setup-restrict-remaining-type [level] [method]` can be used to specify which method to use to generate the "remaining" near null vectors if fine nvec != coarse nvec. Parameters for the remainder method are taken from the flags for each method.

**"Polishing" near-null vectors**

"Polishing" near-null vectors with inverse iterations is enabled by specifying a non-zero number of polish iterations via
 - `--mg-setup-maxiter-inverse-iterations-polish [level] [numbers]` , where the default `[numbers]` is 0, corresponding to no polishing. Parameters for polishing are taken from the flags for inverse iteration setup.

### Reference commands

A base command where we use inverse iterations for level 1 and a custom method for level 2 (specified by `SETUP_FLAGS_LEVEL2`) is:

```
CONFIG=[...]
SETUP_FLAGS_LEVEL2=[...]
./invert_test \
  --prec double --prec-sloppy single --prec-null half --prec-precondition half \
  --recon 12 --recon-sloppy 8 --recon-precondition 8 \
  --dim 16 16 16 16 --gridsize 1 1 1 1 --load-gauge $CONFIG \
  --dslash-type wilson --kappa 0.1394265 --tol 1e-10 --nsrc 1 --niter 20 \
  --verbosity verbose --solve-type direct-pc --solution-type mat --inv-type gcr \
  --inv-multigrid true --mg-levels 3 \
  --mg-block-size 0 4 4 4 4 --mg-nvec 0 24 \
  --mg-block-size 1 2 2 2 2 --mg-nvec 1 32 \
  --mg-setup-tol 0 5e-7 --mg-setup-inv 0 cgnr $SETUP_FLAGS_LEVEL2 \
  --mg-use-mma true \
  --mg-smoother 0 ca-gcr --mg-smoother-solve-type 0 direct-pc --mg-nu-pre 0 0 --mg-nu-post 0 4 \
  --mg-smoother 1 ca-gcr --mg-smoother-solve-type 1 direct-pc --mg-nu-pre 1 0 --mg-nu-post 1 4 \
  --mg-coarse-solver 1 gcr --mg-coarse-solve-type 1 direct-pc --mg-coarse-solver-tol 1 0.35 --mg-coarse-solver-maxiter 1 10 \
  --mg-coarse-solver 2 gcr --mg-coarse-solve-type 2 direct-pc --mg-coarse-solver-tol 2 0.01 --mg-coarse-solver-maxiter 2 20 \
  --mg-verbosity 0 verbose --mg-verbosity 1 verbose --mg-verbosity 2 verbose
```

Where we will fill in `SETUP_FLAGS_LEVEL2` for different options.

**Inverse Iterations**

A standard setup is

```
SETUP_FLAGS_LEVEL2="--mg-setup-type 1 inverse-iterations --mg-setup-tol 1 5e-7 --mg-setup-inv 1 cgnr"
```

Where the `--mg-setup-type` flag is optional as inverse iterations are the default

**Eigenvectors**

A reference setup without polynomial acceleration is

```
SETUP_FLAGS_LEVEL2="--mg-setup-type 1 eigenvectors --mg-eig-type 1 trlm --mg-eig-use-dagger 1 false --mg-eig-use-normop 1 true --mg-eig-use-poly-acc 1 false --mg-eig-n-ev 1 36 --mg-eig-n-kr 1 64 --mg-eig-tol 1 1e-5"
```

**Chebyshev filter**

A reference setup where 4 base vectors are used -> 8 near-null vectors are generated from each base vector, the minimum of the low pass filter is `1.0`, a 500 iteration low pass filter with rescaling every 50 iterations is used, and there are 100 iterations between subsequent near-nulls is:

```
SETUP_FLAGS_LEVEL2="--mg-setup-type 1 chebyshev-filter --mg-setup-filter-startup-vectors 1 4 --mg-setup-filter-startup-iterations 1 500 --mg-setup-filter-startup-rescale-frequency 1 50 --mg-setup-filter-lambda-min 1 1.0 --mg-setup-filter-iterations-between-vectors 1 50"
```

**Restriction**

A reference setup with restriction, then using inverse iterations for the remaining 8 vectors (32 on level 2 minus 24 on level 1) is:

```
SETUP_FLAGS_LEVEL2="--mg-setup-type 1 restrict-fine --mg-setup-restrict-remaining-type 1 inverse-iterations --mg-setup-tol 1 5e-7 --mg-setup-inv 1 cgnr"
```

`--mg-setup-restrict-remaining-type` can be changed appropriately, grabbing other reference flags as appropriate.

**Polishing with inverse iterations**

As an example, the parameters for a Chebyshev filter can be included, and then they can be polished for 50 iterations via adding:

```
SETUP_FLAGS_LEVEL2="${SETUP_FLAGS_LEVEL2} --mg-setup-maxiter-inverse-iterations-refinement 1 50 --mg-setup-tol 1 5e-7 --mg-setup-inv 1 cgnr"
```

### Outstanding work

- [ ] Exposing these options in the MILC interface
- [x] Copy plus pasting all of this to a wiki page: https://github.com/lattice/quda/wiki/Multigrid-Solver
- [ ] Interface testing/updating with Chroma, ...?
- [ ] `clang-format`


tests/utils/host_utils.h
```diff
@@ -260,7 +260,22 @@ inline void safe_strcpy(char *cstr, const std::string &str, size_t limit, const
 }
 
 // MG param types
+/**
+  @brief Set multigrid parameters that are universal across Wilson-type
+         and staggered-type
+*/
 void setMultigridParam(QudaMultigridParam &mg_param);
```
maddyscientist: Does this declaration still need to be there.  E.g., should the Wilson and Staggered functions be the only functions exposed outside of set_params.cpp?

tests/utils/host_utils.cpp
```diff
@@ -128,6 +129,15 @@ void setQudaDefaultMgTestParams()
     mg_eig_amax[i] = -1.0; // use power iterations
     mg_eig_save_prec[i] = QUDA_DOUBLE_PRECISION;
 
+    setup_type[i] = QUDA_SETUP_NULL_VECTOR_INVERSE_ITERATIONS;
+    setup_restrict_remaining_type[i] = QUDA_SETUP_NULL_VECTOR_INVERSE_ITERATIONS;
+    filter_startup_vectors[i] = 1;
+    filter_startup_iterations[i] = 1000;
+    filter_startup_rescale_frequency[i] = 50;
+    filter_iterations_between_vectors[i] = 150;
+    filter_lambda_min[i] = 1.0;
```
maddyscientist, line 138: Is 1.0 a good default for the min value?

lib/transfer.cpp
```diff
@@ -455,7 +455,7 @@ namespace quda {
 
       if (use_gpu) {
         if (out.Location() == QUDA_CPU_FIELD_LOCATION) output = coarse_tmp_d;
-        if (in.Location() == QUDA_CPU_FIELD_LOCATION || in.GammaBasis() != V->GammaBasis())
+        if (in.Location() == QUDA_CPU_FIELD_LOCATION || in.GammaBasis() != V->GammaBasis() || in.FieldOrder() != QUDA_FLOAT2_FIELD_ORDER)
```
maddyscientist, line 458: Remind me: why did you need this extra check for FLOAT2-ordered fields?

lib/inv_cg_quda.cpp
```diff
@@ -567,7 +567,7 @@ namespace quda {
         }
 
         // if L2 broke down already we turn off reliable updates and restart the CG
-        if (ru.reliable_heavy_quark_break(L2breakdown, heavy_quark_res, heavy_quark_res_old, heavy_quark_restart)) {
+        if (use_heavy_quark_res && ru.reliable_heavy_quark_break(L2breakdown, heavy_quark_res, heavy_quark_res_old, heavy_quark_restart)) {
```
maddyscientist, line 570: Did this additional check for `use_heavy_quark_res` fix something?

lib/eigensolve_quda.cpp
```diff
@@ -337,8 +337,8 @@ namespace quda
     // C_1 is the current 'out' vector.
 
     // Clone 'in' to two temporary vectors.
-    auto tmp1 = std::make_unique<ColorSpinorField>(in);
-    auto tmp2 = std::make_unique<ColorSpinorField>(out);
+    auto tmp_1 = std::make_unique<ColorSpinorField>(in);
```
maddyscientist: Not sure what the point of the rename was here, and I don't care 😄 

But while you're there, you might as well use objects here and not unique pointers now we have a move constructor.

include/multigrid.h
```diff
@@ -394,9 +394,8 @@ namespace quda {
 
     /**
        @brief Load the null space vectors in from file
-       @param B Loaded null-space vectors (pre-allocated)
     */
-    void loadVectors(std::vector<ColorSpinorField *> &B);
+    void loadVectors();
```
maddyscientist: Just wondering what the rationale was for removing the argument to this function, but retaining it for `saveVectors` and `initializeRandomVectors`?

include/multigrid.h
```diff
@@ -405,22 +404,59 @@ namespace quda {
...
+    void generateFreeVectors(std::vector<ColorSpinorField*> &B);
+
+    /**
+      @brief Orthonormalize a vector of ColorSpinorField, erroring out if
+             the fields are not sufficiently linearly independent
+      @param vecs vector of ColorSpinorFields to normalize
+      @param count number of near-null vectors to orthonormalize, default all
     */
-    void buildFreeVectors(std::vector<ColorSpinorField*> &B);
+    void orthonormalize_vectors(const std::vector<ColorSpinorField*>& vecs, int count = -1) const;
```
maddyscientist: Probably should keep consistency with the functions declared above here and make this `orthonormalizeVectors` ?

include/quda.h
```diff
@@ -638,6 +638,9 @@ extern "C" {
     /** Maximum number of iterations for refreshing the null-space vectors */
     int setup_maxiter_refresh[QUDA_MAX_MG_LEVEL];
 
+    /** Maximum number of iterations for inverse iteration polishing of null-space vectors */
+    int setup_maxiter_inverse_iterations_polish[QUDA_MAX_MG_LEVEL];
```
maddyscientist: For interface self-consistency, this should be _refinement_ rather than _polish_.

lib/check_params.h
```diff
@@ -821,6 +825,23 @@ void printQudaMultigridParam(QudaMultigridParam *param) {
     P(setup_ca_lambda_max[i], INVALID_DOUBLE);
 #endif
 
+#ifdef INIT_PARAM
+    P(filter_startup_vectors[i], 1);
+    P(filter_startup_iterations[i], 1000);
+    P(filter_startup_rescale_frequency[i], 50);
+    P(filter_iterations_between_vectors[i], 150);
+    P(filter_lambda_min[i], 1.0);
```
maddyscientist, line 833: Again, just wondering why 1.0 is a good default for a min here?

lib/multigrid.cpp
```diff
@@ -140,13 +130,14 @@ namespace quda
     // if we aren't doing a staggered KD solve
     if (param.level != 0 || param.transfer_type == QUDA_TRANSFER_AGGREGATE) {
       // Refresh the null-space vectors if we need to
-      if (refresh && param.level < param.Nlevel - 1) {
-        if (param.mg_global.setup_maxiter_refresh[param.level]) generateNullVectors(param.B, refresh);
+      // FIXME: can we meaningfully refresh Chebyshev filter null vecs?
```
maddyscientist, line 134: I think the answer is yes here.  I don't think this case should be any difference from having a null space set generated from inverse iteration.  Or am I overlooking something?

lib/multigrid.cpp
```diff
@@ -824,6 +779,23 @@ namespace quda
     popLevel();
   }
 
+  void MG::orthonormalize_vectors(const std::vector<ColorSpinorField*>& vecs, int count) const {
+    int num_vec = (count == -1) ? static_cast<int>(vecs.size()) : count;
+
+    if (num_vec > (int)vecs.size())
+      errorQuda("Trying to orthonormalize %d vectors which is larger than std::vector size %ld", num_vec, vecs.size());
+
+    for (int i = 0; i < num_vec; i++) {
```
maddyscientist: Why not do this using a CA-approach?  Even a two-pass algorithm would likely be much faster.

I'd suggest at least making it an option, anyway.

lib/multigrid.cpp
```diff
@@ -1591,22 +1602,156 @@ namespace quda
...
       for (int i=0; i<QUDA_MAX_DIM; i++) commDim[i] = 0;
       diracSmootherSloppy->setCommDim(commDim);
     }
 
-    if (param.mg_global.vec_store[param.level] == QUDA_BOOLEAN_TRUE) { // conditional store of null vectors
-      saveVectors(B);
+    popLevel();
+  }
+
+  void MG::generateChebyshevFilter(std::vector<ColorSpinorField *> &B)
```
maddyscientist, line 1655: Is there any reason one might think that the Chebyshev filter would be better placed in a function or method outside of the MG class?  Just wondering if it could be helpful elsewhere, and / or to reduce redundant code elsewhere.

lib/multigrid.cpp
```diff
@@ -1591,22 +1602,156 @@ namespace quda
...
+        blas::copy(*B[i], pk);
+
+        if (first) first = false;
+
+      }
+
+    }
+
+    // print norms
+    if (getVerbosity() >= QUDA_VERBOSE) {
```
maddyscientist: Could make this a `logQuda`

lib/multigrid.cpp
```diff
@@ -1663,55 +1803,15 @@ namespace quda
...
-
-          // 1110, 1111
-          tmp.Source(QUDA_CORNER_SOURCE, 1, 0xE, c);
-          xpy(tmp, *B[8 * c + 7]);
-          tmp.Source(QUDA_CORNER_SOURCE, 1, 0xF, c);
-          xpy(tmp, *B[8 * c + 7]);
+          // Need to pair an even+odd corner together since they'll get split up
+          // during chiral doubling. Vector 0 is `0000`,`0001`;
+          // vector 1 is `0010`,`0011`, etc.
+          for (int corner = 0; corner < 8; corner++) {
```
maddyscientist, line 1820: nice clean up 👍

lib/multigrid.cpp
```diff
@@ -1765,81 +1859,165 @@ namespace quda
...
     csParam.create = QUDA_ZERO_FIELD_CREATE;
     // This is the vector precision used by matResidual
     csParam.setPrecision(param.mg_global.invert_param->cuda_prec_sloppy, QUDA_INVALID_PRECISION, true);
 
     for (int i = 0; i < n_conv; i++) B_evecs.push_back(new ColorSpinorField(csParam));
 
     // before entering the eigen solver, let's free the B vectors to save some memory
-    ColorSpinorParam bParam(*param.B[0]);
-    for (int i = 0; i < (int)param.B.size(); i++) delete param.B[i];
+    //ColorSpinorParam bParam(*B[0]);
```
maddyscientist: delete this commented code

lib/multigrid.cpp
```diff
@@ -1765,81 +1859,165 @@ namespace quda
...
+
+    if (is_fine_grid()) errorQuda("There are no fine near-null vectors to restrict on level 0");
+    if (param.Nvec != param.fine->param.Nvec)
+      logQuda(QUDA_VERBOSE, "The number of near-null vectors on the fine level (%d) and coarse level (%d) do not match, restricting as many as possible\n", param.fine->param.Nvec, param.Nvec);
+
+    logQuda(QUDA_VERBOSE, "Restricting null space vectors\n");
+
+    // B vectors can be in half precession on the fine level, need to convert to single
+    ColorSpinorField Btmp;
+    bool need_tmp_single = param.fine->param.B[0]->Precision() != QUDA_SINGLE_PRECISION;
```
maddyscientist: Just in case we ever have the B vectors in double precision, I would probably rewrite this as `param.fine->param.B[0]->Precision() < QUDA_SINGLE_PRECISION`

maddyscientist:
Completed an initial visual review of this PR.  Looks like a good contribution.  Left a few comments, and I'll test it on some clover multigrid shortly.

include/multigrid.h
weinbe2: No idea! I'll add it back.

include/multigrid.h
weinbe2: Can do, or I can rip the band-aid off and switch more/all functions over to underscores. Thoughts?

include/quda.h
weinbe2: Will do.

lib/check_params.h
weinbe2: Semi-empirical, in some cases the gapping relative to, say, 1e-2, was necessary to get good enhancement of very small eigenvalues; in other cases it was necessary to get enhancement when I was testing far from the critical point (aka when I didn't bother finding the critical point in the Wilson case).

lib/eigensolve_quda.cpp
weinbe2: I think at some point I was wrestling with a bug and thought that `tmp1` and `tmp2` aliasing other variables in the class was causing an issue, and never bothered to revert it after I hunted down the issue. I imagine I could change it back. In any case, I can switch to objects here! (assuming with the objects I can still use `std::swap`?)

lib/inv_cg_quda.cpp
weinbe2: It made a difference _specifically_ during near-null generation when I was lazily using well-conditioned operators and I was triggering heavy quark convergence nearly immediately, I think a quirk of the zero rhs. I didn't dig too deeply into it once this fix "just worked".

lib/multigrid.cpp
weinbe2: Will discuss offline

lib/multigrid.cpp
weinbe2: One answer is "no good reason". My motivation at the time was it interferes with the `if (nrm2 > 1e-16)` check (since that's done before normalization), but we might be fine just throwing that out---I don't think I've ever hit it in practice, regardless of how good, bad, or stupid my setup was.

lib/eigensolve_quda.cpp
maddyscientist: Swap works as long as objects are movable, so this all should just work.

lib/inv_cg_quda.cpp
maddyscientist: Good enough for me

include/multigrid.h
maddyscientist: 👍

include/multigrid.h
maddyscientist: Let’s keep the bandaid here for now to keep things simple.

lib/multigrid.cpp
weinbe2: I think I would need to make the existing Chebyshev polynomial routine (which gets reused across `CA-CG` and `CA-GCR`) more modular/featureful. More specifically, I'd need to expose the periodic rescaling (line 1687-ish), plus saving vectors along the way (loop on line 1719-ish). I can think that through.

lib/multigrid.cpp
weinbe2: The main reason I didn't was because of line 1740; the `blas::norm2` would get evaluated if I used `logQuda` (I believe)

lib/transfer.cpp
weinbe2: The context here was restricting Wilson-type near-null vectors from level 0 -> level 1; the fine `B` vectors were in `FLOAT4` order. (I'm sure this was an issue before, but I don't know if the restriction workflow had been tested in a long while.)

lib/multigrid.cpp
maddyscientist: It won’t be evaluated: 
https://github.com/lattice/quda/blob/9bae409b742765563b28939e93fa5c43b03da20b/include/util_quda.h#L130

tests/utils/host_utils.cpp
weinbe2: (answered elsewhere)

tests/utils/host_utils.h
weinbe2: Not sure at this point, I'll look back through it.

lib/multigrid.cpp
maddyscientist: Yeah I think that nrm check is a holdover from my original qcdmg code I wrote when I was a postdoc.  😜

lib/transfer.cpp
maddyscientist: Ah gotcha.  Yes, this makes sense.

it’s an issue for the B vectors because they are in a relativistic basis so don’t trigger the copy unlike a general residual vector which will be in the non relativistic basis.

lib/multigrid.cpp
weinbe2: Punted because we need an in-place `cax_U`. Recorded for future reference in #1286

lib/multigrid.cpp
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

lib/multigrid.cpp
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

lib/multigrid.cpp
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

include/quda.h
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

include/multigrid.h
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

include/multigrid.h
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

lib/eigensolve_quda.cpp
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

tests/utils/host_utils.h
weinbe2: https://github.com/lattice/quda/commit/5ac5288cf79843ba7c64c7cd48859130c48fc6e6

lib/multigrid.cpp
weinbe2: https://github.com/lattice/quda/pull/1283/commits/b4cdb6f76184f592017bd8f43428dd875d45b02d

kostrzewa:
@weinbe2 sorry that I haven't gotten back to this yet. I started taking a look to see what I need to adjust in tmLQCD to support the changes here when #1287 got in the way... I'll try to spin this up in the next few days and provide feedback, if any.

weinbe2:
> @weinbe2 sorry that I haven't gotten back to this yet. I started taking a look to see what I need to adjust in tmLQCD to support the changes here when #1287 got in the way... I'll try to spin this up in the next few days and provide feedback, if any.

No problem! There are certainly plenty of higher-priority things bouncing around, on all of our plates. Whenever you can get to it is fine and appreciated.

weinbe2:
FYI, one of the recent merges of `develop` slightly broke this branch, will fix it this week.

include/multigrid.h
```diff
@@ -495,6 +532,14 @@ namespace quda {
 
       return (param.level == 0 || kd_nearnull_gen);
     }
+
+    /**
+      @brief Return if we're on the coarsest grid right now
```
maddyscientist, line 537: Use`@return` here

include/multigrid.h
```diff
@@ -449,7 +449,7 @@ namespace quda {
 
     /**
        @brief Load the null space vectors in from file
-       @param B Loaded null-space vectors (pre-allocated)
+       @param B Load null-space vectors to here
```
maddyscientist, line 452: Can you add `[in]`/`[out]` tags to all the doxgyen you've touched in this file?

include/quda.h
```diff
@@ -652,8 +655,29 @@ extern "C" {
...
+    /** Number of iterations for initial Chebyshev filter */
+    int filter_startup_iterations[QUDA_MAX_MG_LEVEL];
+
+    /** Frequency of rescales during initial filtering which helps avoid overflow */
+    int filter_startup_rescale_frequency[QUDA_MAX_MG_LEVEL];
+
+    /** Number of iterations between null vectors generated from each starting vector */
+    int filter_iterations_between_vectors[QUDA_MAX_MG_LEVEL];
+
+    /** Conservative estimate of largest eigenvalue of operator used for Chebyshev filter setup */
```
maddyscientist, line 670: Is the doxygen correct here: `min` is largest e-value and `max` is lower bound?

lib/eigensolve_quda.cpp
```diff
@@ -372,14 +372,14 @@ namespace quda
...
 
-      blas::axpbypczw(d3, *tmp1, d2, *tmp2, d1, out, *tmp1);
-      std::swap(tmp1, tmp2);
+      blas::axpbypczw(d3, tmp_1, d2, tmp_2, d1, out, tmp_1);
+      std::swap(tmp_1, tmp_2);
 
       sigma_old = sigma;
     }
-    blas::copy(out, *tmp2);
+    blas::copy(out, tmp_2);
```
maddyscientist, line 382: Just to note that this `copy` can be replaced with a `swap`.  I've already applied this optimization to the `feature/multi-rhs`, so it's perhaps moot.

tests/utils/command_line_params.h
```diff
@@ -267,33 +267,46 @@ extern quda::mgarray<QudaVerbosity> mg_verbosity;
...
+
+// Parameters for Chebyshev filter setup
+extern quda::mgarray<int> filter_startup_vectors;
+extern quda::mgarray<int> filter_startup_iterations;
+extern quda::mgarray<int> filter_startup_rescale_frequency;
+extern quda::mgarray<int> filter_iterations_between_vectors;
+extern quda::mgarray<double> filter_lambda_min;
+extern quda::mgarray<double> filter_lambda_max;
+
+
```
maddyscientist, line 293: delete extra space

tests/utils/set_params.cpp
```diff
@@ -495,7 +468,7 @@ void setMultigridParam(QudaMultigridParam &mg_param)
     mg_param.smoother_solver_ca_lambda_min[i] = smoother_solver_ca_lambda_min[i];
     mg_param.smoother_solver_ca_lambda_max[i] = smoother_solver_ca_lambda_max[i];
 
-    // Set set coarse_grid_solution_type: this defines which linear
+        // Set set coarse_grid_solution_type: this defines which linear
```
maddyscientist, line 471: remove random indent

lib/polynomial.hpp
```diff
@@ -0,0 +1,260 @@
...
+	      blas::ax(tmp_inv_nrm, pkm1);
+	      blas::ax(tmp_inv_nrm, pkm2);
+	      blas::ax(tmp_inv_nrm, Apkm1);
+            }
+          }
+        }
+      }
+
+      // if save_freq != 0, the last vector has already been saved into output_vecs
+      if (save_freq == 0) blas::copy(output_vecs[0], pk);
```
maddyscientist, line 159: Like the optimization I noted above in `eigensolve_quda.cpp`, could the `blas::copy` be replaced with a `std::swap` here?

lib/polynomial.hpp
```diff
@@ -0,0 +1,260 @@
...
+    PolynomialBasisParams::check_params(poly_params);
+    if (poly_params.basis != QUDA_POWER_BASIS) errorQuda("Invalid basis %d", poly_params.basis);
+
+    auto tempvec1 = poly_params.tmp_vectors[0];
+    auto tempvec2 = poly_params.tmp_vectors[1];
+
+    applyMatrixPolynomial(diracm, tempvec1, start, poly_params, args...);
+
+    // Get Rayleigh quotient
+    double tmpnrm = sqrt(blas::norm2(tempvec1));
```
maddyscientist, line 204: could replace this with `quda::rsqrt`

lib/inv_ca_cg.cpp
```diff
@@ -488,7 +488,14 @@ namespace quda
       // Perform 100 power iterations, normalizing every 10 mat-vecs, using r as an initial seed
       // and Q[0]/AQ[0] as temporaries for the power iterations. tmp_sloppy/tmp2_sloppy get passed in as temporaries
       // for matSloppy.
-      lambda_max = 1.1 * Solver::performPowerIterations(matSloppy, r, Q[0], AQ[0], 100, 10, tmp_sloppy, tmp2_sloppy);
+      PolynomialBasisParams basis_params;
+      basis_params.basis = QUDA_POWER_BASIS;
+      basis_params.n_order = 100;
+      basis_params.normalize_freq = 10;
+      basis_params.tmp_vectors.emplace_back(std::ref(Q[0]));
```
maddyscientist, line 495: is the `std::ref` wrapper needed here?

lib/multigrid.cpp
```diff
@@ -824,6 +780,63 @@ namespace quda
     popLevel();
   }
 
+  void MG::orthonormalizeVectors(const std::vector<ColorSpinorField*>& vecs, int count) const {
```
maddyscientist, line 783: Just make `count` and `num_vec` unsigned to cleanup all this casting?

lib/multigrid.cpp
```diff
@@ -824,6 +780,63 @@ namespace quda
...
+
+    int num_vec = (count == -1) ? static_cast<int>(vecs.size()) : count;
+
+    if (num_vec > (int)vecs.size())
+      errorQuda("Trying to orthonormalize %d vectors which is larger than std::vector size %ld", num_vec, vecs.size());
+
+    typedef Matrix<Complex, Dynamic, Dynamic> matrix;
```
maddyscientist, line 790: `using`

lib/multigrid.cpp
```diff
@@ -824,6 +780,63 @@ namespace quda
...
+
+    // Create std::vector of references to ColorSpinorFields we're orthonormalizing
+    std::vector<ColorSpinorField_ref> ortho_vecs;
+    for (int i = 0; i < num_vec; i++)
+      ortho_vecs.emplace_back(std::ref(*vecs[i]));
```
maddyscientist, line 795: is `std::ref` needed here?

lib/multigrid.cpp
```diff
@@ -824,6 +780,63 @@ namespace quda
...
+
+    // Prepare to do the Cholesky decomposition for a thin-QR
+    std::vector<Complex> Vdagv_(num_vec * num_vec);
+
+    // outstanding bugfix
```
maddyscientist, line 800: what's the issue here?

lib/multigrid.cpp
```diff
@@ -1406,22 +1396,128 @@ namespace quda
...
+          generateRestrictedVectors(refresh);
+        } else if (setup_type == QUDA_SETUP_NULL_VECTOR_EIGENVECTORS) {
+
+          // Run the eigensolver
+          generateEigenvectors(param.B);
+        } else if (setup_type == QUDA_SETUP_NULL_VECTOR_CHEBYSHEV_FILTER) {
+
+          // Initializing to random vectors 
+          if (!refresh) {
+            int num_initialize = param.mg_global.filter_startup_vectors[param.level];
```
maddyscientist, line 1466: make this `unsigned`?

lib/multigrid.cpp
```diff
@@ -1505,82 +1601,38 @@ namespace quda
...
-        for(int i=0; i<(int)B.size(); i++) {
-          for (int j=0; j<i; j++) {
-            Complex alpha = cDotProduct(*B[j], *B[i]);// <j,i>
-            caxpy(-alpha, *B[j], *B[i]); // i-<j,i>j
-          }
-          double nrm2 = norm2(*B[i]);
-          if (sqrt(nrm2) > 1e-16) ax(1.0/sqrt(nrm2), *B[i]);// i/<i,i>
-          else errorQuda("\nCannot normalize %u vector (nrm=%e)\n", i, sqrt(nrm2));
-        }
+      if (getVerbosity() >= QUDA_VERBOSE) {
```
maddyscientist, line 1619: replace these four lines with two lines of `logQuda`

lib/multigrid.cpp
```diff
@@ -1505,82 +1601,38 @@ namespace quda
...
-        } else {
-          reset();
-        }
-      }
+      ColorSpinorField *out=nullptr, *in=nullptr;
+      diracSmoother->prepare(in, out, x, b, QUDA_MAT_SOLUTION);
+      (*solve)(*out, *in);
+      diracSmoother->reconstruct(x, b, QUDA_MAT_SOLUTION);
+
+      if (getVerbosity() >= QUDA_VERBOSE) printfQuda("Solution = %g\n", norm2(x));
```
maddyscientist, line 1629: `logQuda`

lib/multigrid.cpp
```diff
@@ -1765,81 +1870,161 @@ namespace quda
...
+    if (B[0]->Nspin() == 4) csParam.gammaBasis = QUDA_UKQCD_GAMMA_BASIS;
     csParam.create = QUDA_ZERO_FIELD_CREATE;
     // This is the vector precision used by matResidual
     csParam.setPrecision(param.mg_global.invert_param->cuda_prec_sloppy, QUDA_INVALID_PRECISION, true);
 
     for (int i = 0; i < n_conv; i++) B_evecs.push_back(new ColorSpinorField(csParam));
 
-    // before entering the eigen solver, let's free the B vectors to save some memory
-    ColorSpinorParam bParam(*param.B[0]);
-    for (int i = 0; i < (int)param.B.size(); i++) delete param.B[i];
```
maddyscientist, line 1803: Did this optimization to reduce memory get deleted intentionally?

maddyscientist (changes requested):
Finished visual review of this PR.  Looks mostly good to my eye with some relatively small tweaks needed.

include/quda.h
```diff
@@ -742,12 +766,6 @@ extern "C" {
         memory */
     QudaBoolean setup_minimize_memory;
 
-    /** Whether to compute the null vectors or reload them */
-    QudaComputeNullVector compute_null_vector;
```
kostrzewa, line 746: @weinbe2 with this gone I seem to hit an issue in tmLQCD (https://github.com/etmc/tmLQCD/pull/548) where QUDA attempts to load the null vectors from file by default. I do set the new per level `setup_type` to `QUDA_SETUP_NULL_VECTOR_INVERSE_ITERATIONS` and don't really know if I'm missing something else. Any pointer would be very helpful!

kostrzewa: Ah, I think I might have missed a major change in this PR: `setMultigridParam -> setWilsonMultigridParam`. I'll try that to see if it helps.

kostrzewa: Actually, we never use `setMultigridParam`, instead we have our own.

weinbe2: For better or for worse, this is consistent with the behavior I put into `multigrid.cpp`:

```
if (strcmp(param.mg_global.vec_infile[param.level], "") != 0 && !refresh) {
  load[...]
} else {
  generate[...]
}
```

The reasoning behind it was if `vec_infile` is set, the user must want us to load near nulls, so you should check to make sure that value is the empty string.

That said... this behavior is worth revisiting. We could add a flag `override_load_by_default` or something like that, so it's not _only_ based on not setting `vec_infile`. I also didn't think about how that could be a pain in workflows that generate once, recreate the MG setup for one reason or another, then want to reload from there.

Let me know if you think keeping this default behavior (querying if the string isn't empty), but also adding an override flag, would be a good solution.

kostrzewa: Ah, I missed the fact that there are now per-level `vec_infile` and `vec_outfile` parameters. We actually did something really bad in the past:

```
  // set file i/o parameters
  strcpy(mg_param->vec_infile, "");
  strcpy(mg_param->vec_outfile, "");
```

which appears to be a bug in our interface dating back all the way to 1dec1db7a57e4fe00b14e04ba2088ca5039f8760 (June 2019)... Ouch :)


I think keeping this behaviour is historically consistent (also with the eigensolver).

Thanks a lot!

lib/inv_cg_quda.cpp
kostrzewa: Interestingly I seem to have hit the same thing in a different context:

https://github.com/lattice/quda/issues/1323

kostrzewa:
@weinbe2 Seems to work fine for me with with the changes to our interface that I implemented in https://github.com/etmc/tmLQCD/pull/548 to preserve the status quo. I will keep track of this PR and make any adjustments that may become necessary due to ongoing changes.

kostrzewa (approved):
No issues from my side, haven't tested anything beyond our status quo, however.

weinbe2:
For future reference, do not merge, it appears an issue creeped in not in `develop`. For my own convenience, a command that caused an issue was

```
LOCAL_DIM="32 32 32 16"
GRIDSIZE="1 1 1 2"
OUTER_OPERATOR=direct-pc
GAUGE_PATH=./l32t32b7p0
MASS=0.00945
TADPOLE_COEFF=1.0
TOL=1e-5
FIRST_AGGREGATE="4 4 4 4"
SECOND_AGGREGATE="4 4 4 4"
COARSEST_NVEC=128
COARSEST_NEV=160
COARSEST_NKR=192
COARSEST_POLY_DEG=32
COARSEST_AMIN=0.85
COARSEST_AMAX=0.0

export QUDA_RESOURCE_PATH=.
export QUDA_TUNE_VERSION_CHECK=0

mpirun -np 2 --allow-run-as-root \
                        ./staggered_invert_test --inv-multigrid true --dim ${LOCAL_DIM} --gridsize ${GRIDSIZE} --verbosity verbose --nsrc 1 \
                        --dslash-type asqtad --solve-type ${OUTER_OPERATOR} --solution-type mat --compute-fat-long true \
                        --load-gauge ${GAUGE_PATH} --mass ${MASS} --tadpole-coeff ${TADPOLE_COEFF} --tol ${TOL} \
                        --prec single --prec-sloppy single --prec-precondition single --prec-null single --mg-use-mma false \
                        --recon 18 --recon-sloppy 18 --recon-precondition 18 \
                        --mg-verbosity 2 verbose --mg-verbosity 3 verbose \
                        --mg-save-vec 1 blah --mg-save-vec 2 blah \
                        --mg-levels 4 --mg-coarse-solve-type 0 ${OUTER_OPERATOR} --mg-staggered-coarsen-type kd-optimized \
                        --mg-block-size 0 1 1 1 1 --mg-nvec 0 3 \
                        --mg-smoother-solve-type 0 ${OUTER_OPERATOR} --mg-smoother 0 ca-gcr --mg-nu-pre 0 0 --mg-nu-post 0 8 --mg-smoother-tol 0 1e-10  \
                        --mg-coarse-solve-type 1 direct --mg-coarse-solver-tol 1 5e-2 \
                        --mg-coarse-solver-maxiter 1 16 --mg-coarse-solver 1 gcr \
                        --mg-setup-inv 1 cgnr --mg-setup-maxiter 1 1000 --mg-setup-tol 1 1e-5 \
                        --mg-block-size 1 ${FIRST_AGGREGATE} --mg-nvec 1 64 --mg-n-block-ortho 1 2 \
                        --mg-smoother-solve-type 1 direct --mg-smoother 1 ca-gcr --mg-nu-pre 1 0 --mg-nu-post 1 2 --mg-smoother-tol 1 1e-10 \
                        --mg-coarse-solve-type 2 direct-pc --mg-coarse-solver-tol 2 0.25 \
                        --mg-coarse-solver-maxiter 2 16   --mg-coarse-solver 2 gcr \
                        --mg-setup-inv 2 cgnr --mg-setup-maxiter 2 1000 --mg-setup-tol 2 1e-5 \
                        --mg-block-size 2 ${SECOND_AGGREGATE} --mg-nvec 2 96 --mg-n-block-ortho 2 2 \
                        --mg-smoother-solve-type 2 direct-pc --mg-smoother 2 ca-gcr --mg-nu-pre 2 0 --mg-nu-post 2 2 --mg-smoother-tol 2 1e-10 \
                        --mg-coarse-solver 3 ca-gcr --mg-coarse-solver-ca-basis-size 3 16 --mg-coarse-solver-maxiter 3 16 \
                        --mg-nvec 3 ${COARSEST_NVEC} --mg-eig-n-ev 3 ${COARSEST_NEV} --mg-eig-n-kr 3 ${COARSEST_NKR} --mg-eig-tol 3 1e-4 \
                        --mg-eig 3 true --mg-eig-type 3 trlm --mg-eig-use-dagger 3 false --mg-eig-use-normop 3 true \
                        --mg-eig-use-poly-acc 3 true --mg-eig-poly-deg 3 ${COARSEST_POLY_DEG} --mg-eig-amin 3 ${COARSEST_AMIN} --mg-eig-amax 3 ${COARSEST_AMAX} \
                        --mg-eig-max-restarts 3 1000
```

and the error that popped up is

```
MG level 2 (GPU): Xinv = 1.574662e+04
MG level 2 (GPU): Yhat[0] = 6.719429e+01 (1.147926e-01 < 4.053768e-02 x 3.218868e+00)
MG level 2 (GPU): Yhat[1] = 6.493005e+01 (1.500221e-01 < 5.319621e-02 x 3.218868e+00)
MG level 2 (GPU): Yhat[2] = 6.315730e+01 (1.395274e-01 < 5.026488e-02 x 3.218868e+00)
MG level 2 (GPU): Yhat[3] = 2.209952e+02 (3.106113e-01 < 8.176686e-02 x 3.218868e+00)
MG level 2 (GPU): Yhat[4] = 4.554047e+01 (1.680291e-01 < 6.054319e-02 x 3.218868e+00)
MG level 2 (GPU): Yhat[5] = 4.207957e+01 (1.500221e-01 < 5.088072e-02 x 3.218868e+00)
MG level 2 (GPU): Yhat[6] = 4.217662e+01 (1.335199e-01 < 4.538731e-02 x 3.218868e+00)
MG level 2 (GPU): Yhat[7] = 1.093643e+02 (3.106113e-01 < 7.678801e-02 x 3.218868e+00)
MG level 2 (GPU): ....done computing Yhat field
MG level 2 (GPU): Finished creating the preconditioned coarse op
MG level 2 (GPU): Coarse Dirac operator done
MG level 2 (GPU): Creating smoother
MG level 2 (GPU): Creating a CA-GCR solver
MG level 2 (GPU): Smoother done
MG level 2 (GPU): Checking 0 = (1 - P P^\dagger) v_k for 96 vectors
MG level 2 (GPU): Vector 0: norms v_k = 1.000000e+00 P^\dagger v_k = 6.917310e-01 (1 - P P^\dagger) v_k = 1.202923e+00, L2 relative deviation = 1.096779e+00
MG level 2 (GPU): ERROR: L2 relative deviation for k=0 failed, 1.096779e+00 > 1.000000e-03 (rank 0, host luna-0196, multigrid.cpp:920 in void quda::MG::verify(bool)())
MG level 2 (GPU): ERROR: L2 relative deviation for k=0 failed, 1.096779e+00 > 1.000000e-03 (rank 1, host luna-0196, multigrid.cpp:920 in void quda::MG::verify(bool)())
MG level 2 (GPU):        last kernel called was (name=N4quda4blas5Norm2IdfEE,volume=8x8x8x4x1,aux=GPU-offline,nParity=2,vol=2048,parity=2,precision=4,order=2,Ns=2,Nc=64)
MG level 2 (GPU):        last kernel called was (name=N4quda4blas5Norm2IdfEE,volume=8x8x8x4x1,aux=GPU-offline,nParity=2,vol=2048,parity=2,precision=4,order=2,Ns=2,Nc=64)
QMP m0,n2@luna-0196 error: abort: 1
QMP m1,n2@luna-0196 error: abort: 1
```

sm_80, fast build, QUDA_PRECISION=12

PR #1284, opened 2022-06-01
CTestTestfile install helper

mathiaswagner:
This generates a little helper script that copies a modified CTestTestfile to the installation directory
in order to run ctest after installation. Note that this is not called by default. and controlled by
`QUDA_CTEST_INSTALL`.

Background:
CMake does not support installing the tests and this is a workaround so we can compile on machines without GPUs, install QUDA and then run CI on a machine with GPUs after install.


maddyscientist:
Closing as no longer relevant.

PR #1285, opened 2022-06-01
Feature/sincospi

maddyscientist:
This PR does the following:
* Use `sincospi`, `sinpi` and `cospi` instead of the regular trig functions, where appropriate.
* Remove legacy `Trig` namespace
* Kernel optimization: remove stack frame from many kernels, e.g., using shared memory for per-thread arrays, loop unrolling
* Compile-time optimization: default reconstruct type use for `instantiate` function is now `ReconstructNo12`, e.g., reconstruction-no and reconstruct-12.  This avoids a lot of unnecessary template instantiation.
* Removed the `QUDA_GAUGE_ALG`, `QUDA_GAUGE_TOOLS` and `QUDA_FORCE_GAUGE` CMake options: the corresponding kernels are now always built
* Removed the `QUDA_FORCE_HISQ` CMake option: the HISQ force is now enabled if staggered fermions are enabled 
* Apply some cleanup to the multigrid prolongator instantiation
* Update QIO to version 3.0


mathiaswagner (approved):

include/instantiate.h
```diff
@@ -164,7 +150,7 @@ namespace quda
 
   /**
      @brief This instantiate function is used to instantiate the colors
-     @param[in] U Gauge field
+     @param[in] U Gauge fieldOB
```
weinbe2: I think `fieldOB` might be an accident :)

maddyscientist: Ha, yes.  Probably some failed attempt at some obscure emacs key combination.

Will fix when I do a clang-format.

lib/prolongator.cu
```diff
@@ -49,165 +50,152 @@ namespace quda {
...
+  template <> struct enabled<2,  6, 2,  6> : std::true_type { };
+#endif
+#ifdef NSPIN1
+  template <> struct enabled<1,  3, 2, 24> : std::true_type { };
+  template <> struct enabled<1,  3, 2, 64> : std::true_type { };
+  template <> struct enabled<1,  3, 2, 96> : std::true_type { };
+  template <> struct enabled<2, 24, 2, 64> : std::true_type { };
+  template <> struct enabled<2, 24, 2, 96> : std::true_type { };
+  template <> struct enabled<2, 64, 2, 64> : std::true_type { };
+  template <> struct enabled<2, 64, 2, 96> : std::true_type { };
```
weinbe2, line 71: This is missing 96 -> 96

maddyscientist: I don't think that's necessarily a functional regression, since that's copied directly from the restrictor which doesn't have the 96->96 path either.  It could well have been a regression though when the matching refactoring was applied to the restrictor (in the reduce run-time PR).

Is the 96->96 path one that should be added (back) in?

weinbe2: Ah, I missed it during the restrictor then! Please add 96 -> 96 back in, yes.

maddyscientist: Ok, will do.

weinbe2 (approved):
I've checked MILC rhmc and a propagator workflow, correctness and performance all looks good. Pending the few outstanding tweaks and a final `clang-format` this has my approval.

include/instantiate.h
maddyscientist: e4375a12dfa0144a9c2ece9332a61889c981f964

lib/prolongator.cu
maddyscientist: 2b195025a8fccffb71846155ac7c79de9816acf3

PR #1288, opened 2022-06-06
Hotfix/hdot

maddyscientist:
* Fix for hDotProduct function: recursion that identified if the present tile is exclusively located in the lower triangular part of the matrix was broken for non-powers of two matrix dimensions
* Add a unit test for `hDotProduct`
* Some minor cleanup of multi-blas


weinbe2 (approved):
Verified via pulling it into my MG branch and testing that it worked correctly for thin-QR-based orthonormalization. Thanks for the quick hotfix @maddyscientist !

PR #1289, opened 2022-06-07
Feature/stag 128bit

maddyscientist:
* Half-precision staggered quark fields now use a single 128-bit struct per site instead of SoAoS (nColor x volume x complex).  This optimizes memory access patterns and gives a small boost to the staggered fermion dslash.
* Bug fix in hDotProduct unit test


weinbe2 (approved):
Just went through and did some performance and correctness tests, everything looks great! I'm seeing a few percent speed-up across different volumes, also always welcome. Awesome PR @maddyscientist !

Issue #1287, opened 2022-06-05
updateMultigridQuda performance regression sometime after be1766a519995f8c16b9541c300f883efabc3df8


kostrzewa:
I've observed a (rather substantial) performance regression in `updateMultigridQuda` sometime after commit be1766a519995f8c16b9541c300f883efabc3df8. The earliest  'bad' commit (which I could compile and which ran without erroring out) that I could identify is b50f16a55d2eaae816c00812b3996138c5c4dfa9. 

Running on four nodes of Juwels Booster on a 64c128 lattice at the physical point, calling `updateMultigridQuda` (without refresh iterations):

### good

```
# TM_QUDA: Time for MG_Preconditioner_Setup_Update 1.128518e+00 s level: 4 proc_id: 0 /HMC/cloverdetratio3light:cloverdetratio_derivative/solve_degenerate/invert_eo_degenerate_quda/MG_Preconditioner_Setup_Update
```

[tunecache_be1766a51.tsv.zip](https://github.com/lattice/quda/files/8839700/tunecache_be1766a51.tsv.zip)

When it performs like this, the Update scales reasonably:

| nodes | seconds |
|---------|-------------|
|    4     |     1.1     |
|    8     |     0.62   |
|   16    |     0.42   |
|   32    |     0.21   |

### bad

```
# TM_QUDA: Time for MG_Preconditioner_Setup_Update 5.131241e+00 s level: 4 proc_id: 0 /HMC/cloverdetratio3light:cloverdetratio_derivative/solve_degenerate/invert_eo_degenerate_quda/MG_Preconditioner_Setup_Update
```

[tunecache_b50f16a55.tsv.zip](https://github.com/lattice/quda/files/8839702/tunecache_b50f16a55.tsv.zip)

This poorly performing behaviour doesn't scale. With commit 9bae409b742765563b28939e93fa5c43b03da20b (which happened to be the develop head commit when I started observing this), I have:

| nodes | seconds |
|---------|-------------|
|    4      |      5.1    |
|    8      |      6.6    |
|   16     |     4.8     |

This suggest perhaps a constant factor.

Testing reproduction on a different machine now.

kostrzewa:
Just to make sure I repeated with 605e92c83a32b68cd8f889b69aee8c66172b5010 and the issue still persists.

[tunecache_605e92c83.tsv.zip](https://github.com/lattice/quda/files/8839836/tunecache_605e92c83.tsv.zip)

kostrzewa:
[...]

PR #1292, opened 2022-06-10
Hotfix/vuv

### LINKED ISSUE #1291: Potential multigrid regression due to two-stage tuning
While trying to understand the regression reported #1287 I have found a potentially related regression that I don't understand.  This was introduced as part of the two-stage tuning (#1260).  This regression concerns the VUV kernel, and it is ~33% slower with the two-stage tuning.  What is especially perplexing is that the regression isn't apparent in the first sweep to find the candidates, it seems to appear in the second stage of the tuning.

E.g., VUV kernel, half precision, 16^4 volume, 4^4 blocking on a Quadro GV100 (prior to #1260) reports performance around ~1800-1900 GFLOPS.  This performance is observed in the first stage of autotuning:
```
MG level 0 (GPU): About to call tunable.apply block=(128,1,1) grid=(256,2,4) shared_bytes=16384 aux=(2,1,1)
MG level 0 (GPU): C   block=(128,1,1), grid=(256,2,4), shared_bytes=16384, aux=(2,1,1,1) gives 1833.42 Gflop/s, 923.87 GB/s
MG level 0 (GPU): About to call tunable.apply block=(128,1,1) grid=(256,2,4) shared_bytes=16385 aux=(2,1,1)
MG level 0 (GPU): C   block=(128,1,1), grid=(256,2,4), shared_bytes=16385, aux=(2,1,1,1) gives 1835.08 Gflop/s, 924.71 GB/s
```
However, when these are re-run in the second step, the kernels are significantly slower
```
MG level 0 (GPU): About to call tunable.apply block=(128,1,1) grid=(256,2,4) shared_bytes=16384 aux=(2,1,1)
MG level 0 (GPU): T   block=(128,1,1), grid=(256,2,4), shared_bytes=16384, aux=(2,1,1,1) gives 1292.92 Gflop/s, 651.51 GB/s
MG level 0 (GPU): About to call tunable.apply block=(128,1,1) grid=(256,2,4) shared_bytes=16385 aux=(2,1,1)
MG level 0 (GPU): T   block=(128,1,1), grid=(256,2,4), shared_bytes=16385, aux=(2,1,1,1) gives 1288.45 Gflop/s, 649.26 GB/s
```

I don't understand how this can be happening.  If anything, the increased iteration count in the second step of tuning should lead to faster performance since there is less host induced latency.

@mathiaswagner thoughts?

maddyscientist:
This hotfix fixes the performance of the VUV kernel:
* When two-stage tuning was enabled, it was observed that the VUV kernel seemingly regressed (#1291)
* It turns out this wasn't a regression per se, rather it revealed a long standing issue that the launch parameters of the VUV kernel weren't being set correctly outside of the tuning loop
* Specifically, the `Arg::coarse_color_wave` parameter, which maps to `TuneParam::aux` wasn't being set in the `apply` function, it was only being set during the tuning process.
* So while the kernel autotuned correctly, post tuning, it would not be using the desired parameters, leading to reduced performance.

I have also made some minor changes to the tuning:
* Display a warning if the best second-stage tuning time regresses by more than 10% versus the first stage.  In general we should expect the second stage to be faster than the first stage, since it will generally involve more iterations.  The 10% margin is left due to latency bubbles, etc.
* Even then, this warning can still be triggered if one doesn't lock clocks for small short running kernels.
* Fix a minor bug when printing the `aux` parameter while tuning.
* Replace use of `new`/`delete` with a `std::vector`

Closes #1291

### DIFF
--- a/lib/tune.cpp
+++ b/lib/tune.cpp
@@ -314,12 +314,11 @@ namespace quda
       if (comm_rank_global() == 0) {
         comm_broadcast_global(const_cast<char *>(serialized.str().c_str()), size);
       } else {
-        char *serstr = new char[size + 1];
-        comm_broadcast_global(serstr, size);
+        std::vector<char> serstr(size + 1);
+        comm_broadcast_global(serstr.data(), size);
         serstr[size] = '\0'; // null-terminate
-        serialized.str(serstr);
+        serialized.str(serstr.data());
         deserializeTuneCache(serialized);
-        delete[] serstr;
       }
     }
 #endif
@@ -754,6 +753,7 @@ namespace quda
         push(candidate);
       }
     }
+
     /**
      * @brief Broadcast candidates among ranks to make sure policy tuning does not break.
      *
@@ -761,7 +761,6 @@ namespace quda
     void broadcast()
     {
 #ifdef MULTI_GPU
-
       size_t size;
       std::string serialized;
       if (comm_rank_global() == 0) {
@@ -774,12 +773,11 @@ namespace quda
         if (comm_rank_global() == 0) {
           comm_broadcast_global(const_cast<char *>(serialized.c_str()), size);
         } else {
-          char *serstr = new char[size + 1];
-          comm_broadcast_global(serstr, size);
+          std::vector<char> serstr(size + 1);
+          comm_broadcast_global(serstr.data(), size);
           serstr[size] = '\0'; // null-terminate
-          std::string_view deserialized(serstr);
+          std::string_view deserialized(serstr.data());
           deserialize(deserialized);
-          delete[] serstr;
         }
       }
 #endif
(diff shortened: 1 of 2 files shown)


weinbe2:
Visual review looks great, I'm always a fan of a simple fix. I've got a due-diligence build+run going now, once that's complete I'll give it the approval + merge.

weinbe2 (approved):
This passes a range of (staggered) perf and correctness checks, so I'll consider this good---thanks for the fix @maddyscientist !

lib/tune.cpp
```diff
@@ -966,13 +964,14 @@ namespace quda
...
-            printfQuda("About to call tunable.apply block=(%d,%d,%d) grid=(%d,%d,%d) shared_bytes=%d aux=(%d,%d,%d)\n",
+            printfQuda("About to call tunable.apply block=(%d,%d,%d) grid=(%d,%d,%d) shared_bytes=%d aux=(%d,%d,%d,%d)\n",
                        static_cast<int>(param.block.x), static_cast<int>(param.block.y),
                        static_cast<int>(param.block.z), static_cast<int>(param.grid.x), static_cast<int>(param.grid.y),
                        static_cast<int>(param.grid.z), static_cast<int>(param.shared_bytes),
-                       static_cast<int>(param.aux.x), static_cast<int>(param.aux.y), static_cast<int>(param.aux.z));
+                       static_cast<int>(param.aux.x), static_cast<int>(param.aux.y), static_cast<int>(param.aux.z), static_cast<int>(param.aux.w));
           }
 
+          tunable.apply(stream);
```
mathiaswagner, line 974: why is this needed?

maddyscientist: I made a comment on this in [c8a53f7](https://github.com/lattice/quda/pull/1292/commits/c8a53f7d6482cf91ad39b1a1dcf99047cccab7a1).  Without this warmup call, even with locked clocks there were some warnings about second-stage regressions.  This additional call costs nothing, but made the warnings go away.

mathiaswagner: might be worth adding a comment in the code as well.

maddyscientist: Fair point.  I'll add it in my next PR, whatever that may be....

maddyscientist: Hopefully one that fixes #1287.....

PR #1293, opened 2022-06-14
Fix a bug in `invert_x_batch.h` where `axpy` is called

hummingtree:
- Fix a bug in `invert_x_batch.h` where the coefficient size is not equal to the input vector size for `axpy`.
- Added a corresponding invert ctest parameter dimension to catch this.


maddyscientist (approved):
Thanks for the fast fix here, and `ctest` addition

PR #1294, opened 2022-06-17
Hotfix/stand alone eigensolver

cpviolator:
This PR adds gtest functionality to the Wilson type eigensolvers. The test will compute the following variations:

1. Normal operator and Direct operator
2. Preconditioned and unpreconditioned
3. With SVD if the normal operator is chosen
4. All possible spectrum types available to the eigensolver.
5. Thick Restarted Lanczos, Block Thick Restarted Lanczos, Implicitly Restarted Arnoldi.

Each test will be performed with no polynomial acceleration as that requires the user to fine tune input parameters tailored to a specific input matrix.

For the SVD testing, the routine will compute the L_2 norm of
```
M * X_left - sigma * X_right.
```
This explicitly tests the SVD data, and implicitly tests the underlying eigenvector data, as the SVD of M(Mdag) is computed via the eigendecomposition of MdagM(MMdag). For the Eigenvectors produced in Direct solves, the host will measure the L_2 norm of
```
M * X - lambda * X.
```

This gtest was tested with Wilson, Clover, Twisted-Mass, Twisted Clover, and all Domain Wall actions. For the Block solver, a block size of 4 was used, and a batched rotation size of 4 was used for all tests. The following command can be used to test this PR:

```
#!/bin/bash

export QUDA_RESOURCE_PATH=.

for ACTION in wilson clover twisted-mass twisted-clover domain-wall domain-wall-4d mobius mobius-eofa ; do
	      
    ./eigensolve_test --sdim 4 --tdim 4 --gridsize 1 1 1 1 --eig-n-conv 12 --eig-n-ev 24 --eig-n-kr 96 --eig-tol 1e-6 --mass 0.01 --eig-max-restarts 1000 --verbosity verbose --prec single --enable-testing true --verify true --dslash-type ${ACTION}
done
```


kostrzewa:
@cpviolator is the standalone eigensolver capable of computing eigenvalues also for the non-degenerate twisted mass (clover) operators? The reason that I ask is that we use RHMC to simulate the non-degenerate doublet and it is necessary to check in the simulation that our eigenvalue bounds are appropriate for the chosen parameters. It's on my TODO list to add to enable our QUDA interface to perform the necessary measurement of the minimal and maximal evals using QUDA.

cpviolator:
@kostrzewa I've yet to come across an action that the eigensolver cannot handle. Some dslash options such as laplace or clover hasenbuch have no host side verification routines, and I see also now that non-degenerate twisted (clover) have no explicit command lined dslash option, one must pass that as an extra command. I'm confident that the eigensolver can handle it. Allow me to make the extra modifications to test this path and I'll get back to you.

lib/interface_quda.cpp
```diff
@@ -2205,70 +2205,123 @@ void cloverQuda(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity
...
-  pushVerbosity(inv_param->verbosity);
-  if (getVerbosity() >= QUDA_DEBUG_VERBOSE) {
-    printQudaInvertParam(inv_param);
-    printQudaEigParam(eig_param);
+  // Specific changes to the default invert param for the eigensolver.
+  //------------------------------------------------------------------
+  // QUDA's device routines require UKQCD gamma basis. QUDA will
+  // automatically rotate from the Degrand-Rossi basis on the host, to UKQCD
+  // on the device, and back to this basis upon completion.
+  inv_param->gamma_basis = QUDA_DEGRAND_ROSSI_GAMMA_BASIS;
```
maddyscientist: If one is interfacing with an application that requires UKQCD ordering or another ordering completely, hardcoding this to DeGrand-Rossi doesn't seem like the right approach.  Shouldn't it just use what value of `gamma_basis` the user has passed in `QudaEigParam::QudaInvertParam`?

lib/interface_quda.cpp
```diff
@@ -2205,70 +2205,123 @@ void cloverQuda(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity
...
+
+  // QUDA can employ even-odd preconditioning to an operator.
+  // For the eigensolver (I think...) the solution type must match
```
maddyscientist: Probably delete the `(I think...)`

tests/eigensolve_test_gtest.hpp
```diff
@@ -0,0 +1,89 @@
...
+  if (is_chiral(dslash_type) && (::testing::get<1>(param) == QUDA_BOOLEAN_FALSE)) return true;
+  return false;
+}
+
+std::vector<double> eigensolve(test_t test_param);
+
+TEST_P(EigensolveTest, verify)
+{
+  if (skip_test(GetParam())) GTEST_SKIP();
+  auto tol = 5 * eig_param.tol;
```
maddyscientist: What's this factor of 5 for?

lib/interface_quda.cpp
```diff
@@ -2205,70 +2205,123 @@ void cloverQuda(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity
...
+  //------------------------------------------------------------------
 
+  // Ensure that the parameter structures are sound.
   checkInvertParam(inv_param);
   checkEigParam(eig_param);
-  cudaGaugeField *cudaGauge = checkGauge(inv_param);
 
-  bool pc_solve = (inv_param->solve_type == QUDA_DIRECT_PC_SOLVE) || (inv_param->solve_type == QUDA_NORMOP_PC_SOLVE)
-    || (inv_param->solve_type == QUDA_NORMERR_PC_SOLVE);
+  // Check that the gaige field is valid
```
maddyscientist: typo (gaige)

lib/interface_quda.cpp
```diff
@@ -2205,70 +2205,123 @@ void cloverQuda(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaParity
...
-  for (int i = 0; i < eig_param->n_conv; i++) {
+  // Construct vectors
+  //------------------------------------------------------
+  // Create host wrappers around application vector set
+  ColorSpinorParam cpuParam(host_evecs[0], *inv_param, cudaGauge->X(), inv_param->solution_type,
+                            inv_param->input_location);
+
+  int n_eig = eig_param->n_conv;
+  if (eig_param->compute_svd) n_eig *= 2;
+  std::vector<ColorSpinorField *> host_evecs_(n_eig);
```
maddyscientist, line 2281: While reading this, I was thinking to suggest modifying this to vector of fields, rather than vector of objects.  But on second though, I think it makes more sense to get this PR in with all the tests and only then to perform surgery of that nature.

maddyscientist (commented):
@cpviolator this is an excellent PR, thank you for adding all these tests.  Beyond the minor comments I made, the main remaining ask is to add test invocations to tests/CMakeLists.txt in a similar fashion as is already done for the solver tests.

cpviolator:
@kostrzewa I checked the non-deg twisted (clover) paths with the eigensolver and all seems well. I will caveat that by saying that when I tried to run with `QUDA_OPENMP=ON` I saw some regression, so I would avoid that option for the moment.

As soon as this PR is complete please feel free to reach out so we can discuss proper command line/param options for the eigendecompostion of non-deg operators.

lib/interface_quda.cpp
cpviolator: All works well without this, so removed. 7e005b04f

cpviolator: 7e005b04f

cpviolator: All works well without this, so removed. 7e005b04f

lib/interface_quda.cpp
cpviolator: 7e005b04f

tests/eigensolve_test_gtest.hpp
cpviolator: The IRAM eigensolver will sometimes report convergence with tolerances slightly higher than requested. The same phenomenon occurs in ARPACK. This fudge factor prevents failure when IRAM has solved to say 2e-6 when 1e-6 is requested. The solution to avoid this is to use a Krylov space (n-kr) about 3-4 times the size of the search space (n-ev), or use a well chosen Chebyshev polynomial, or use a tighter than necessary tolerance.

lib/interface_quda.cpp
cpviolator: https://github.com/lattice/quda/commit/7e005b04fb5aeeec124c87bfb099a9b4580591e9

tests/eigensolve_test_gtest.hpp
maddyscientist: Ok, could you leave a comment to this effect then and only apply the fudge factor if doing IRAM?

lib/eigensolve_quda.cpp
```diff
@@ -130,8 +135,13 @@ namespace quda
       if (!eig_solver->hermitian()) errorQuda("Polynomial acceleration not supported with non-Hermitian solver");
     }
 
+    // Cannot solve for imaginary spectrum of hermitian systems
     if (mat.hermitian() && (eig_param->spectrum == QUDA_SPECTRUM_SI_EIG || eig_param->spectrum == QUDA_SPECTRUM_LI_EIG))
       errorQuda("The imaginary spectrum of a Hermitian operator cannot be computed");
+
+    // Cannot compute SVD of non-normal operators
+    if (!eig_param->use_norm_op && eig_param->compute_svd)
```
weinbe2, line 143: I thought that a point of the SVD was that it exists for any matrix regardless of its property and shape? https://en.wikipedia.org/wiki/Singular_value_decomposition

cpviolator: @weinbe2 Yes, that's right, but I haven't coded up a routine that can compute the SVD of a non-hermitain matrix M. Instead, we compute the eigendecomposition of MdagM, which also furnishes the right singular vectors of M:
```
MdagM * X = lambda X
MdagM * X_right = Mdag * sigma * X_left = sigma^2 X_right
``` 
hence the sigma and X_right data can be computed via the (admittedly poorer conditioned) MdagM operator. The X_left vectors are easily identified via
```
M * X_right = sigma * X_left
```
I think it's OK to use the poorer MdagM as we only use the SVD in QUDA to deflate non-hermitian solvers, so the extra time required to identify the SVD via MdagM rather than M is amortised by the solver speed-up.

tests/eigensolve_test_gtest.hpp
cpviolator: Done 9b58a9770

cpviolator:
@lattice/quda_core I'll Give this one more full build with SANITIZE on and roll through all the actions one last time and dump the stdout here in a .txt file for reference, as well as adding test invocations to tests/CMakeLists.txt.

cpviolator:
@kostrzewa I added the ndeg twisted(clover) tests to the ctest suite and all is working well. If you have any difficulty obtaining a spectrum let me know and I'll help tweak the Arnoldi parameters.

weinbe2 (approved):
I just gave this a quick spin and everything looks good from the staggered side, and I didn't have to change any flags for coarsest-level deflation in MG to go through which is great.. I was a bit confused about explicitly adding a flag for the preconditioned vs non-preconditioned operator, but after thinking about it/toying with test runs a bit I came to appreciate its clarity. Thanks for the contribution @cpviolator !

This has my approval, but I think it'd be good if @kostrzewa or another applications person gave the PR a final spin before hitting merge. Happy to be overriden, though.

kostrzewa:
Please go ahead with the merge. I won't really have the time to work on the interface for our non-degenerate operator before September or so... @cpviolator many thanks for taking nd twisted mass (clover) into account: this will prove very useful in the near future

weinbe2:
Just posted in with @maddyscientist and there are still a few more tests she wants to do, so no hitting merge just yet.

maddyscientist (approved):
Approving.  The one outstanding thing I wanted to do was fix the function `display_test_info` so that each invoked test's parameters would be printed out correctly, instead of just printing out the initial values repeatedly.

(I also sneaked in a link error with quarter precision that cropped up elsewhere.)

I think we can merge this when Jenkins is done.

PR #1296, opened 2022-07-04
Hotfix/init check mg mma

maddyscientist:
This PR fixes a few long-standing issues
* Silence some false-positive warnings from `compute-sanitizer` init check when taking the norm of fields during the multigrid setup
* Introduce an optional index-mapper functor to the `TransformReduce` which can be used map the iterator index
* Use a milc-mapper functor to allow us to correctly print per-dimension norms of MILC ordered gauge field in `TransformReduce` (this fixes a known issue where the norms weren't printed correctly during an MMA-enabled multigrid setup)
* Prefer the use of `std::vector::resize` with only a size argument and no second object argument to copy from.  This removes some unnecessary copies introduced with #1257  
* (Re)enable HISQ force tests with `ctest`


lib/cuda_gauge_field.cpp
```diff
@@ -70,7 +70,18 @@ namespace quda {
...
-    if (create != QUDA_ZERO_FIELD_CREATE && isNative() && ghostExchange == QUDA_GHOST_EXCHANGE_PAD) zeroPad();
+
+    if (ghostExchange == QUDA_GHOST_EXCHANGE_PAD) {
+      if (isNative()) {
+        if (create != QUDA_ZERO_FIELD_CREATE) zeroPad();
+      } else {
+        for (int i=0; i<nDim; i++) {
+          size_t nbytes = nFace * surface[i] * nInternal * precision;
+          qudaMemset(ghost[i], 0, nbytes);
+          if (nbytes && geometry == QUDA_COARSE_GEOMETRY) qudaMemset(ghost[i], 0, nbytes);
```
weinbe2: `ghost[i+4]`

maddyscientist: Good catch

maddyscientist: ec2803168

weinbe2 (approved):
This is a great set of fixes, thanks @maddyscientist !

PR #1298, opened 2022-07-14
Feature/milc gaugefix ovr definition

detar:
Two small changes to enable access to gauge_fixingOVR via the MILC interface and to fix interval reporting.


mathiaswagner:

weinbe2:
Thank you for the contribution @detar ! I just pushed a quick `clang-format`, and I'll do my own cursory test before approving, but this should be good to go.

weinbe2:
@detar I just ran a quick `ks_spectrum` workflow, replacing `no_gauge_fix` with `coulomb_gauge_fix`, and got the following error:

```
Fixing to Coulomb gauge
Starting Coulomb gauge fixing...
        Overrelaxation boost parameter: 1.800000e+00
        Tolerance: 2.000000e-06
        Stop criterion method: Delta
        Maximum number of iterations: 500
        Reunitarize at every 20 steps
        Print convergence results at every 20 steps
ERROR: CPU not supported yet (rank 0, host luna-0199, tunable_reduction.h:130 in void quda::TunableReduction2D::launch(T&, const quda::TuneParam&, const quda::qudaStream_t&, Arg&) [with Functor = quda::FixQualityOVR; bool enable_host = false; T = quda::array<double, 2>; Arg = quda::GaugeFixQualityOVRArg<double, QUDA_RECONSTRUCT_NO, 3>]())
       last kernel called was (name=N4quda15GaugeFixQualityINS_21GaugeFixQualityOVRArgIdL21QudaReconstructType_s18ELi3EEEEE,volume=16x16x16x16,aux=CPU,nParity=2,vol=65536stride=32768precision=8geometry=4Nc=3)
```

Investigating now... I did notice that, unlike other MILC <-> QUDA bindings, your modifications to MILC went directly through the `quda.h` interface, so maybe I need to do a little work around the edges with that. I'll keep you posted.

weinbe2:
Ah, here's the issue, the function you added to `quda.h` is already declared in `quda_milc_interface.h`:

```
$ grep -r "qudaGaugeFixingOVR" *
include/quda_milc_interface.h:  void qudaGaugeFixingOVR( const int precision,
include/quda.h:  void qudaGaugeFixingOVR(int precision, unsigned int gauge_dir, int Nsteps, int verbose_interval, double relax_boost,
lib/milc_interface.cpp:void qudaGaugeFixingOVR(int precision, unsigned int gauge_dir, int Nsteps, int verbose_interval, double relax_boost,
```

So it didn't need to go into `quda.h`---it may be worth trying just including `quda_milc_interface.h` instead of `quda.h` in the MILC file `generic/gaugefix2.c`.

I can work on addressing this over the course of the day, but if you also have some cycles please let me know.

weinbe2:
@detar I just made some changes to your branch, as well as the corresponding changes in MILC in the `develop` branch of https://github.com/lattice/milc_qcd ; I made a PR here for your convenience: https://github.com/milc-qcd/milc_qcd/pull/54

Can you test this combination and make sure it works?

weinbe2:
@maddyscientist this is ready for your review, it passed various tests on my end. Since I had a heavy hand in taking the PR an extra mile I don't see it as mine to approve and merge.

maddyscientist:
@weinbe2 is there a simple test on the MILC side that can be used to verify this PR (e.g., for myself to run as part of this review)?

weinbe2:
> @weinbe2 is there a simple test on the MILC side that can be used to verify this PR (e.g., for myself to run as part of this review)?

I can provide you a spectrum workflow with Coulomb gauge fixing enabled; I noticed there was some deviation from the CPU and the GPU side but that could be from a low convergence tolerance in MILC (1e-6).

include/quda_milc_interface.h
```diff
@@ -962,16 +962,9 @@ extern "C" {
...
-    const unsigned int gauge_dir,
-    const int Nsteps,
-    const int verbose_interval,
-    const double relax_boost,
-    const double tolerance,
-    const unsigned int reunit_interval,
-    const unsigned int stopWtheta,
-    void* milc_sitelink
-    );
+  void qudaGaugeFixingOVR(const int precision, const unsigned int gauge_dir, const int Nsteps,
```
maddyscientist: I'm a bit confused here.  The function definition in milc_interface.cpp is lacking the `const` qualification on all these parameters, and yet the compiler isn't complaining.  Perhaps this is some "Cism" that I'm not aware of?

Regardless, I think the parameters should not be `const` qualified since it doesn't mean that much to `const` qualify a scalar that is passed by value.

jxy: `const` and `volatile` parameters are ignored in function declarations, but they have an effect in function definition.

maddyscientist: One learns something new every day 🙂

weinbe2: https://github.com/lattice/quda/commit/0138cf3566353710dd0e569f4060a76065a16290

maddyscientist (approved):

detar:
Hi Evan,

My mistake.  I had forgotten about quda_milc_interface.h.  I'll try that.

Best,

Carleton

PR #1303, opened 2022-07-27
Hotfix/nvcxx

maddyscientist:
This PR is primarily to fix some issues with using nvc++ and some reorganization of the shared memory cache header
* Fix copy_clover kernel with `nvc++` (`length` function variable should just be `constexpr`)
* Work around for `sinpif`, `cospif` and `sincospif` linkage with `nvc++`
* Make `shared_memory_cache_helper.h` target-specific
* Move `thread_array` into its own target-specific header, `thread_array.h`
* Workaround for `nvc++` which doesn't at present support in-class static shared memory (fallback to using a thread-local array)

### DIFF
--- a/include/gauge_fix_ovr_hit_devf.cuh
+++ b/include/gauge_fix_ovr_hit_devf.cuh
@@ -3,7 +3,7 @@
 #include <quda_internal.h>
 #include <quda_matrix.h>
 #include <atomic_helper.h>
-#include <shared_memory_cache_helper.cuh>
+#include <shared_memory_cache_helper.h>
 
 namespace quda {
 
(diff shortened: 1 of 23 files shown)


weinbe2:
A cursory visual review looks good. I'll do a due-diligence compile and run but otherwise I think this will be a non-controversial merge.

mathiaswagner (approved):
looks fine to me

weinbe2 (approved):
Various tests with routines that got poked when shared-memory-related files got moved around (Mobius local preconditioner, various gauge observables) looked good, so this also gets my approval.

maddyscientist:
i've confirmed that this builds fine with ROCm, so we can merge this.

Issue #1305, opened 2022-07-30
Compiler warning flags are not being passed to nvcc


maddyscientist:
This is a regression that I think happened when the CUDA specifics were moved to `lib/targets/cuda/target_cuda.cmake`.  For example, `-Wall -Wextra` is no long passed to the host compiler for CUDA targets:
```
nvcc -forward-unknown-to-host-compiler -ccbin=/usr/bin/c++ -DBUILD_MILC_INTERFACE -DBUILD_QDP_INTERFACE -DDEVEL -DEIGEN -DGPU_STAGGERED_DIRAC -DNATIVE_LAPACK_LIB -DQUDA_FAST_COMPILE_DSLASH -DQUDA_FAST_COMPILE_REDUCE -DQUDA_PRECISION=4 -DQUDA_RECONSTRUCT=4 -Dquda_EXPORTS -I/scratch2/kate/quda-develop/lib -I/home/kate/github/quda-develop/lib -I/home/kate/github/quda-develop/include/targets/cuda -I/scratch2/kate/quda-develop/include/targets/cuda -I/home/kate/github/quda-develop/lib/. -I/home/kate/github/quda-develop/lib/../include -I/scratch2/kate/quda-develop/include -I/home/kate/github/quda-develop/lib/../include/targets/generic -isystem=/home/kate/github/quda-develop/include/targets/cuda/externals -isystem=/home/kate/github/quda-develop/lib/../include/externals -isystem=/scratch2/kate/quda-develop/_deps/eigen-src -isystem=/usr/local/cuda-11.7/include -g -O3 --generate-code=arch=compute_70,code=[compute_70,sm_70] -Xcompiler=-fPIC -ftz=true -prec-div=false -prec-sqrt=false -Wno-deprecated-gpu-targets --expt-relaxed-constexpr -Xfatbin=-compress-all -Wreorder -Xcompiler=-Wno-unknown-pragmas -Xptxas -warn-lmem-usage,-warn-spills -lineinfo -std=c++17 -MD -MT lib/CMakeFiles/quda.dir/coarse_op.cu.o -MF CMakeFiles/quda.dir/coarse_op.cu.o.d -x cu -c /home/kate/github/quda-develop/lib/coarse_op.cu -o CMakeFiles/quda.dir/coarse_op.cu.o
```

Assigning to @mathiaswagner

PR #1306, opened 2022-08-03
move warning flags to target compile options

mathiaswagner:
This applies the warning flags only to the targets defined in QUDA's CMakeLists and should help with warning from external dependencies as well as fixing #1305.

Note that this might trigger some warning that have previously gone unnoticed.


mathiaswagner:
@Jenkins test this please

maddyscientist (approved):
Thanks @mathiaswagner.

I have fixed the warnings incurred with this PR, so we should be good to go once Jenkins completes.

include/gauge_field_order.h
```diff
@@ -1552,14 +1549,10 @@ namespace quda {
           for (int j = 0; j < N; j++) copy(tmp[i * N + j], reinterpret_cast<Float *>(&vecTmp)[j]);
         }
 
-        real phase = 0.;
```
mathiaswagner, line 1555: just wondering as I had looked at that warning before: For `!hasPhase` we had `phase=0.` in the code and now it seems that has gone. I have not looked at it in detail but did not want to forget about it.

maddyscientist: Yeah, it doesn’t affect anything.  For the non-phased reconstruct types, it doesn’t use the phase value at all so it doesn’t matter what this is set to.  I checked that the ctest staggered and Wilson tests all still pass with a full reconstruct build.

include/kernels/dslash_wilson2.cuh
```diff
@@ -0,0 +1,249 @@
+#pragma once
```
mathiaswagner: Why do we need this file?

maddyscientist: Accidental addition.  Can fix when I’m back 💻

mathiaswagner (commented):
Looks good now

include/kernels/dslash_wilson2.cuh
mathiaswagner: I just removed it.

PR #1308, opened 2022-08-03
Gauge observables: trace of arbitrary gauge loops (useful for gauge action), Polyakov loop

weinbe2:
This PR adds two new types of gauge measurements to QUDA: batch computation of the summed trace of arbitrary gauge loops (relevant for gauge action calculations), and the calculation of the (temporal) Polyakov loop. These new capabilities are exposed through both the standard QUDA interface and through the MILC interface, utilized in a complimentary PR to MILC.

### Gauge loops
Standalone code to compute products of gauge links along arbitrary paths has been moved from the gauge force routines to a helper file `include/gauge_path_helper.cuh`. This code is then used in the new gauge loop code and in a refactor of the gauge force routines.

A similar refactor was done in the host verification code, and verification of the gauge loop calculation has been added to the gauge force tests (since it's a very similar workflow).

### Polyakov loops
The Polyakov loop code is fully new, and is implemented via a routine that parallelizes over discrete (x, y, z) coordinates, where each work item computes the product of temporal gauge links across the entire local `t` dimension. This code is utilized recursively when the `t` dimension is split across MPI ranks, where:
 * A "face" of gauge link products of local size `x` x `y` x `z` is computed
 * The face is saved as a gauge field of size `x` x `y` x `z` x `1` (where the odd `t` dimension "works" since it is the slowest dimension)
 * This data is exchanged across temporal ranks at fixed `X`, `Y`, and `Z` rank, inserted as "fictitious" timeslices in a new gauge field of size `x` x `y` x `z` x `N_t`
 * The gauge link product code is then re-used to compute the product in the `t` dimension again, with a final trace + reduce

To make this code relatively robust against different partitionings of the `T` dimension, all intermediate calculations are done in full double precision, independent of the precision of the input gauge field.

For now, this is only implemented for Polyakov loops in the `t` dimension, but the plumbing is in place to generalize this---it just doesn't have a compelling use case at the moment.

This code has been verified against MILC's calculation of the Polyakov loop. The temporal Polyakov loop also gets calculated in QUDA's `su3_test` test executable. While it would be a best practice to also implement calculating the Polyakov loop in QUDA's own host verification code, implementing the full machinery to handle a distributed calculation of the Polyakov loop on the CPU felt a bit too much like reinventing the wheel... 

### Miscellaneous interface notes

Routines have been added to the QUDA interface to calculate gauge loop traces as well as the Polyakov loop. In addition, this PR adds routines that calculate the plaquette and the Polyakov loop from an input CPU field as opposed to using (and thus requiring) a resident field.

### MILC interface notes

As it stands, the MILC interface does not take advantage of keeping the gauge fields resident. This is a somewhat conscious decision because the gauge links may or may not have phases applied at different points in the RHMC workflow, leading to a book-keeping headache, and in any case the time saved by offloading gauge loop/etc calculations to the GPU more than amortizes the cost of redundant host -> device copies.

### Performance results

This optimization corresponds to a ~5% performance boost in the standard NERSC medium RHMC workflow.

### Outstanding work

- [x] Opening an analogous MILC PR: https://github.com/lattice/milc_qcd/pull/24 (updated)
- [x] ~Implementing a calculation of the Polyakov loop in spatial directions~ Punted, see #1311
- [x] Cleaning up comments, doxygen
- [x] `clang-format`


weinbe2:
Complimentary reference PR for MILC: https://github.com/milc-qcd/milc_qcd/pull/55

weinbe2:
With the exception of `clang-format`, which I can do as a last step, this is ready for review.

include/gauge_tools.h
```diff
@@ -169,4 +169,13 @@ namespace quda
   */
   void computeQChargeDensity(double energy[3], double &qcharge, void *qdensity, const GaugeField &Fmunu);
 
+  /**
+   * @brief Compute the trace of the polyakov loop in a given dimension
+   * @param[out] ploop The real adnd imaginary parts of the polyakov loop
```
maddyscientist: typo (adnd -> and)

maddyscientist: ....and I guess it should be Polyakov (not polyakov)

include/gauge_path_helper.cuh
```diff
@@ -0,0 +1,115 @@
...
+      length = buffer + dim*num_paths*max_length;
+      path_coeff = reinterpret_cast<double*>(buffer + dim * num_paths * max_length + num_paths + pad);
+    }
+
+    void free() {
+      pool_device_free(buffer);
+    }
+  };
+
+  constexpr int flipDir(int dir) { return (7-dir); }
```
maddyscientist, line 73: I would love it if there was a future where we can move away from this MILC convention of encoding direction and dimension like this in a 1-d index and have a 2-d index instead, e.g., `[dim][dir]` like in the rest of QUDA.  Not necessarily for this PR, but what's your thoughts on how hard this would be to replumb?

include/gauge_path_helper.cuh
```diff
@@ -0,0 +1,115 @@
...
+     @param[in] arg Kernel argumnt
+     @param[in] x Full index array
+     @param[in] parity Parity index (note: assumes that an offset from a non-zero dx is baked in)
+     @param[in] path Gauge link path
+     @param[in] length Length of gauge path
+     @param[in] dx Temporary shared memory storage for relative coordinate shift
+  */
+  template <typename Arg>
+  __device__ __host__ inline typename Arg::Link
+  computeGaugePath(const Arg &arg, int x[4], int parity, const int* path, int length, thread_array<int, 4>& dx)
```
maddyscientist: Suggest we make this interface slightly more generic by making `thread_array<int, 4>` a template type.  Then the function could be called with a regular C-style array or even the `array` object.

include/kernels/gauge_force.cuh
```diff
@@ -78,116 +27,72 @@ namespace quda {
...
-      case 3: GaugeForceKernel<Arg,3>(arg, x_cb, parity); break;
+      // update mom(x)
+      Link mom = arg.mom(dir, x_cb, parity);
+      if (arg.compute_force) {
+        mom = mom - arg.epsilon * link_prod;
+        makeAntiHerm(mom);
+      } else {
+        mom = mom + arg.epsilon * link_prod;
       }
+      arg.mom(dir, x_cb, parity) = mom;
```
maddyscientist, line 95: Nice cleanup in this file.  Good to confirm that there's not performance difference.

include/quda.h
```diff
@@ -1521,10 +1537,38 @@ extern "C" {
 
   /**
    * Computes the total, spatial and temporal plaquette averages of the loaded gauge configuration.
-   * @param Array for storing the averages (total, spatial, temporal)
+   * @param [out] Array for storing the averages (total, spatial, temporal)
```
maddyscientist: The doxygen here for the parameters, and in the lines below, has an erroneous space between `param` and `[in/out]`.  I haven't seen anywhere in the doxygen docs that this style is supported, so these spaces should be deleted.

include/quda_milc_interface.h
```diff
@@ -728,6 +728,48 @@ extern "C" {
   void qudaGaugeForcePhased(int precision, int num_loop_types, double milc_loop_coeff[3], double eb3,
                             QudaMILCSiteArg_t *arg, int phase_in);
 
+  /**
+   * Compute the real traces of gauge loops, with direct application to computing the gauge
+   * action.  All fields here are CPU fields in MILC order, and their precisions should
+   * match.
+   *
+   * @param precision The precision of the field (2 - double, 1 - single)
```
maddyscientist: Missing `[in]` / `[out]` specifiers for the Doxygen parameters in this file.

include/tunable_reduction.h
```diff
@@ -323,7 +323,7 @@ namespace quda
     {
       bool rtn;
       do {
-        rtn = Tunable::advanceBlockDim(param);
+        rtn = TunableReduction2D::advanceBlockDim(param);
```
maddyscientist, line 326: I think we may have discussed this, but did this unexpected prior implementation of calling `Tunable::` methods instead of `TunableReduction2D::` cause issue somewhere?

lib/gauge_loop_trace.cu
```diff
@@ -0,0 +1,74 @@
...
+      p(p)
+    {
+      if (p.num_paths != loop_traces.size()) errorQuda("Loop traces size %lu != number of paths %d", loop_traces.size(), p.num_paths);
+
+      strcat(aux, "num_paths=");
+      char loop_str[4];
+      u32toa(loop_str, p.num_paths);
+      strcat(aux, loop_str);
+
+      // we should add some hash of lengths, paths, etc...
```
maddyscientist: is this comment relevant?

include/gauge_path_helper.cuh
```diff
@@ -0,0 +1,115 @@
...
+      qudaMemcpy(buffer, path_h, bytes, qudaMemcpyHostToDevice);
+      host_free(path_h);
+
+      // finally set the pointers to the correct offsets in the buffer
+      for (int d=0; d < dim; d++) this->input_path[d] = buffer + d*num_paths*max_length;
+      length = buffer + dim*num_paths*max_length;
+      path_coeff = reinterpret_cast<double*>(buffer + dim * num_paths * max_length + num_paths + pad);
+    }
+
+    void free() {
```
maddyscientist, line 68: Is there some reason why we can't put the `free` function in the destructor to avoid having to call this?

maddyscientist: (I know this is legacy behaviour, but I can't remember)

tests/gauge_force_test.cpp
```diff
@@ -1,6 +1,7 @@
 #include <stdio.h>
```
maddyscientist, line 1: Maybe it's time we renamed this test?  Gauge loop test?

tests/gauge_force_test.cpp
```diff
@@ -214,6 +236,127 @@ void gauge_force_test(bool compute_force = true)
...
+  if (getTuning() == QUDA_TUNE_YES) {
+    computeGaugeLoopTraceQuda(traces, sitelink, trace_path_p, trace_loop_length_p, trace_loop_coeff_p, num_paths, max_length, scale_factor, &gauge_param);
+  }
+
+  struct timeval t0, t1;
+  double total_time = 0.0;
+  // Multiple execution to exclude warmup time in the first run
+
+  for (int i = 0; i < niter; i++) {
+    gettimeofday(&t0, NULL);
```
maddyscientist: replace this with `host_timer_t`

tests/gauge_force_test.cpp
```diff
@@ -214,6 +236,127 @@ void gauge_force_test(bool compute_force = true)
...
+    sitelink = U_milc->Gauge_p();
+  } else if (gauge_order == QUDA_QDP_GAUGE_ORDER) {
+    sitelink = U_qdp->Gauge_p();
+  } else {
+    errorQuda("Unsupported gauge order %d", gauge_order);
+  }
+
+  // storage for traces
+  using double_complex = double _Complex;
+  double_complex* traces = new double_complex[num_paths];
```
maddyscientist: replace with a `std::vector<double_complex>` to avoid needing to free at the end

tests/gauge_force_test.cpp
```diff
@@ -214,6 +236,127 @@ void gauge_force_test(bool compute_force = true)
...
+  gauge_param.t_boundary = QUDA_PERIODIC_T;
+
+  setDims(gauge_param.X);
+
+  double *trace_loop_coeff_p = &trace_loop_coeff_d[0];
+  int num_paths = sizeof(trace_path) / sizeof(trace_path[0]);
+  int *trace_loop_length_p = &trace_loop_length[0];
+
+  // b/c we can't cast int*[6] -> int**
+  int** trace_path_p = new int*[num_paths];
```
maddyscientist: Could make this a vector of vectors to remove the `delete` junk at the end

lib/milc_interface.cpp
```diff
@@ -645,6 +645,76 @@ void qudaGaugeForce(int precision, int num_loop_types, double milc_loop_coeff[3]
   qudaGaugeForcePhased(precision, num_loop_types, milc_loop_coeff, eb3, arg, 0);
 }
 
+/**
+ * @brief Reusable routine that creates a qudaGaugeParam for gauge-related observable measurements
+ *
+ * @param [in] precision MILC precision
```
maddyscientist: spaces between `param` and `[in]`

lib/milc_interface.cpp
```diff
@@ -645,6 +645,76 @@ void qudaGaugeForce(int precision, int num_loop_types, double milc_loop_coeff[3]
...
+{
+  qudamilc_called<true>(__func__);
+
+  QudaGaugeParam qudaGaugeParam = createGaugeParamForObservables(precision, arg, phase_in);
+  void *gauge = arg->site ? arg->site : arg->link;
+
+  computeGaugeLoopTraceQuda(reinterpret_cast<double _Complex*>(traces), gauge, input_path_buf, path_length, loop_coeff, num_paths, max_length, factor, &qudaGaugeParam);
+
+  qudamilc_called<false>(__func__);
+  return;
```
maddyscientist, line 721: style comment: I'm not a big fan of explicit `return` statements at the end of `void` functions, but to each, their own 😄

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+
+  public:
+    GaugeInsertTimeslice(GaugeField &u, const GaugeField &s, int timeslice) :
+      TunableKernel2D(u, 2),
+      u(u),
+      s(s),
+      timeslice(timeslice)
+    {
+      strcat(aux, ",4d_vol=");
+      strcat(aux, u.VolString());
```
maddyscientist: Why is this needed?  The 4-d volume will be included automatically already in the `TuneKey`

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    void apply(const qudaStream_t &stream)
+    {
+      TuneParam tp = tuneLaunch(*this, getTuning(), getVerbosity());
+      GaugeInsertTimesliceArg<Float, nColor, recon> arg(u, s, timeslice);
+      launch<InsertTimeslice>(tp, stream, arg);
+    }
+
+    long long flops() const
```
maddyscientist: Can delete this function, since the virtual parent already returns 0

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+    unsigned int minThreads() const { return product_field.LocalVolumeCB(); }
+
+  public:
+    GaugePolyakovLoopProduct(const GaugeField &u, GaugeField &product_field) :
+      TunableKernel2D(u, 2),
+      product_field(product_field),
+      u(u)
+    {
+      strcat(aux, ",4d_vol=");
+      strcat(aux, u.VolString());
```
maddyscientist: can delete the 4-d volume part of this since that will already be in the `TuneKey`

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+
+  public:
+    GaugePolyakovLoopTrace(const GaugeField &u, array<double, 2> &ploop) :
+      TunableReduction2D(u),
+      u(u),
+      ploop(ploop)
+    {
+      if (u.Geometry() != QUDA_SCALAR_GEOMETRY && u.Geometry() != QUDA_VECTOR_GEOMETRY)
+        errorQuda("Invalid geometry %d in Polyakov loop calculation", u.Geometry());
+      strcat(aux, ",4d_vol=");
```
maddyscientist: Same comment again - delete the 4-d volume part

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+      strcat(aux, u.VolString());
+      strcat(aux, ",geometry=");
+      char aux2[3];
+      u32toa(aux2, u.Geometry());
```
maddyscientist: And again, the geometry should already be in the `aux` key.

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+
+    }
+
+    const GaugeField& G = commDimPartitioned(dir) ? const_cast<const GaugeField&>(*condensed_field.get()) : u;
+
+    // Trace over remaining bits
+    profile.TPSTART(QUDA_PROFILE_COMPUTE);
+    instantiate<GaugePolyakovLoopTrace, ReconstructNo12>(G, loop);
+    // We normalize by the 3-d volume, times the 4-d communications dim to cancel out redundant counting
+    long vol3d = u.Volume() * comm_dim(0) * comm_dim(1) * comm_dim(2) * comm_dim(3) / u.X()[3];
```
maddyscientist: make this `auto`

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+          product_field_ref.copy_from_buffer(v_double_buffer[RECV_BUFFER]);
+          profile.TPSTOP(QUDA_PROFILE_H2D);
+
+          // insert
+          profile.TPSTART(QUDA_PROFILE_COMPUTE);
+          gaugeInsertTimeslice(condensed_field_ref, product_field_ref, (comm_coord(3) + t) % comm_dim(3));
+          profile.TPSTOP(QUDA_PROFILE_COMPUTE);
+
+          // swap buffers
+          SEND_BUFFER ^= 1;
```
maddyscientist: I think this is unnecessary optimization that only serves to obfuscate the meaning here.

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+      GaugeField& condensed_field_ref = reinterpret_cast<GaugeField&>(*condensed_field.get());
+      profile.TPSTOP(QUDA_PROFILE_INIT);
+
+      profile.TPSTART(QUDA_PROFILE_COMPUTE);
+
+      // Compute the rank-local product of loops in the temporal direction...
+      instantiate<GaugePolyakovLoopProduct, ReconstructNo12>(u, product_field_ref);
+      // ...and inject it into my own accumulation field
+      gaugeInsertTimeslice(condensed_field_ref, product_field_ref, comm_coord(3));
+      profile.TPSTOP(QUDA_PROFILE_COMPUTE);
```
maddyscientist, line 184: The perhaps unusual aspect of the below is that it's in a .cu file.  Possible to removes this comms handling stuff into a new cpp file?

lib/gauge_polyakov_loop.cu
```diff
@@ -0,0 +1,276 @@
...
+      gParam.x = x;
+      gParam.create = QUDA_NULL_FIELD_CREATE;
+      condensed_field = std::make_unique<cudaGaugeField>(gParam);
+      GaugeField& condensed_field_ref = reinterpret_cast<GaugeField&>(*condensed_field.get());
+      profile.TPSTOP(QUDA_PROFILE_INIT);
+
+      profile.TPSTART(QUDA_PROFILE_COMPUTE);
+
+      // Compute the rank-local product of loops in the temporal direction...
+      instantiate<GaugePolyakovLoopProduct, ReconstructNo12>(u, product_field_ref);
```
maddyscientist: I think the default for instantiate is `ReconstructNo12` arelady?

include/kernels/gauge_polyakov_loop.cuh
```diff
@@ -0,0 +1,221 @@
...
+#include <quda_matrix.h>
+#include <index_helper.cuh>
+#include <array.h>
+#include <reduction_kernel.h>
+
+namespace quda {
+
+  template <typename Float_, int nColor_, QudaReconstructType recon_>
+  struct GaugeInsertTimesliceArg : public kernel_param<> {
+    using Float = Float_;
```
maddyscientist: Can you update this and the other kernels in this PR to use `store_t` for the storage type template argument to the struct, `As above: change this to using real = typename mapper<store_t>::type;` for the compute type, and only expose the name `real` in the kernel?

include/kernels/gauge_polyakov_loop.cuh
```diff
@@ -0,0 +1,221 @@
...
+      // save loop
+      arg.P(0, x_cb, parity) = polyloop;
+
+    }
+
+  };
+
+  template <typename Float_, int nColor_, QudaReconstructType recon_>
+  struct GaugePolyakovLoopTraceArg : public ReduceArg<array<double, 2>> {
+    using Float = Float_;
```
maddyscientist: As above: change this to `using real = typename mapper<store_t>::type;`

include/kernels/gauge_polyakov_loop.cuh
```diff
@@ -0,0 +1,221 @@
...
+      hi_link = link; // promote
+      polyloop = polyloop * hi_link;
+      nbr_oddbit = nbr_oddbit ^ 1;
+    } // dt
+    return polyloop;
+  }
+
+  template <typename Float_, int nColor_, QudaReconstructType recon_>
+  struct GaugePolyakovLoopProductArg : public kernel_param<> {
+    using Float = Float_;
```
maddyscientist: As above: change this to using `real = typename mapper<store_t>::type;`

include/kernels/gauge_loop_trace.cuh
```diff
@@ -0,0 +1,95 @@
...
+namespace quda {
+
+  /**
+    @brief Return the batch block size used for multi reductions.
+  */
+  constexpr unsigned int max_n_batch_block_loop_trace() { return 8; }
+
+  template <typename Float_, int nColor_, QudaReconstructType recon_>
+  struct GaugeLoopTraceArg : public ReduceArg<array<double, 2>>  {
+    using Float = Float_;
```
maddyscientist: As mentioned elsewhere: change this to using `real = typename mapper<store_t>::type;` and change the template type to `store_t`

include/kernels/gauge_polyakov_loop.cuh
```diff
@@ -0,0 +1,221 @@
...
+      getCoords(x, x_cb, arg.X, parity);
+
+      // hack
+      if (x[3] > 0) return;
+
+      auto polyloop = computePolyakovLoop<3>(arg, x, parity);
+
+      // save loop
+      arg.P(0, x_cb, parity) = polyloop;
+
```
maddyscientist: (very minor) style comment: I'm not a fan of the extra space here before the closing brace.  It's asymmetric with the opening brace to the function.

lib/interface_quda.cpp
```diff
@@ -4146,6 +4154,74 @@ int computeGaugePathQuda(void *out, void *siteLink, int ***input_path_buf, int *
...
+  } else {
+    delete cudaSiteLink;
+    delete cudaGauge;
+  }
+
+  if (cpuSiteLink) delete cpuSiteLink;
+  profileGaugeLoopTrace.TPSTOP(QUDA_PROFILE_FREE);
+
+  profileGaugeLoopTrace.TPSTOP(QUDA_PROFILE_TOTAL);
+  return 0;
```
maddyscientist: What's the point of the `return 0`?

lib/interface_quda.cpp
```diff
@@ -5203,6 +5278,150 @@ void plaqQuda(double plaq[3])
   profilePlaq.TPSTOP(QUDA_PROFILE_TOTAL);
 }
 
+/*
+ * Computes the total, spatial and temporal plaquette averages after loading a gauge field.
+ */
+void plaqLoadGaugeQuda(double plaq[3], void *gauge, QudaGaugeParam *param)
```
maddyscientist: Can you explain what the point of this function is?  Since we already have it's constituents in the form of `loadGaugeQuda` and `plaqQuda`.  Am I missing something?

lib/interface_quda.cpp
```diff
@@ -5203,6 +5278,150 @@ void plaqQuda(double plaq[3])
...
+
+  profilePolyakovLoop.TPSTOP(QUDA_PROFILE_TOTAL);
+}
+
+/*
+ * Computes the trace of the Polyakov loop in a direction dir from a loaded field
+ */
+void polyakovLoopLoadGaugeQuda(double ploop[2], int dir, void *gauge, QudaGaugeParam *param)
+{
+
```
maddyscientist, line 5255: delete this space please

lib/interface_quda.cpp
```diff
@@ -5203,6 +5278,150 @@ void plaqQuda(double plaq[3])
   profilePlaq.TPSTOP(QUDA_PROFILE_TOTAL);
 }
 
+/*
+ * Computes the total, spatial and temporal plaquette averages after loading a gauge field.
+ */
+void plaqLoadGaugeQuda(double plaq[3], void *gauge, QudaGaugeParam *param)
+{
+
```
maddyscientist: delete space

maddyscientist:
Completed an initial visual review of this.  This looks like great work Evan and I'm looking forward to testing it.  I've left a variety of comments in the meantime 😉

include/tunable_reduction.h
weinbe2: Yes, the issue was in the gauge loop trace routine, where I tuned over a 3-d space (`x_cb`, parity, and loop id) and wanted to reduce only over the first two dimensions. This is the first time such a tuning space + reduction was introduced in the code, which is why it didn't get hit in the past. (The only other place a close pattern gets hit is the multi-reduction, which is a 2-d space of NXZ and NYW.)

include/gauge_path_helper.cuh
weinbe2: The theoretical issue with putting it in a destructor is if this got moved around as a copy by value as opposed to by reference. Since this object gets passed directly as a kernel argument, we can't just wrap the pointers in smart pointers to avoid this issue. I guess we could avoid it by creating a smart pointer (with a custom deallocator) outside the function, with some `static` function that'll do the math for you, and then passing the raw pointer in, but it's just extra plumbing. Happy to do it if you think that's best.

tests/gauge_force_test.cpp
weinbe2: Maybe gauge path test? Since the forces technically aren't loops. But, yes, I agree we should rename it.

maddyscientist: Good suggestion

include/gauge_path_helper.cuh
maddyscientist: Thanks for the reminder on this, the theoretical issue is exactly the issue at hand.  Just leave it as is for now (resolving).

lib/gauge_polyakov_loop.cu
maddyscientist: To be more coherent here: I am wondering if this communications pattern should be abstracted into a separate function in its own cpp file as opposed to being part of the Polyakov loop driver.

lib/gauge_loop_trace.cu
```diff
@@ -0,0 +1,74 @@
...
+  public:
+    // max block size of 8 is arbitrary for now
+    GaugeLoopTrace(const GaugeField &u, std::vector<reduce_t> &loop_traces, double factor, const paths<1>& p) :
+      TunableMultiReduction(u, 2u, p.num_paths, 8),
+      u(u),
+      loop_traces(loop_traces),
+      factor(factor),
+      p(p)
+    {
+      if (p.num_paths != loop_traces.size()) errorQuda("Loop traces size %lu != number of paths %d", loop_traces.size(), p.num_paths);
```
maddyscientist: Noting it's this line that is causing clang builds to fail (signed vs unsigned comparison)

include/gauge_tools.h
weinbe2: Done locally, will push soon

lib/gauge_loop_trace.cu
weinbe2: It's sort of a fix-me type deal that may be unnecessary. We include the number of loops in the tunekey, but nothing's stopping you from sending in multiple different sets of paths where it happens to be the same number of loops, in which case it isn't necessarily the ideal tuning... but that's a bit contrived/unlikely/this isn't super perf-critical anyway, so it can probably be ignored. What do you think?

tests/gauge_force_test.cpp
weinbe2: Done locally, will push soon

tests/gauge_force_test.cpp
weinbe2: Is it legal to pass that as a pointer in as `&traces[0]`? I think I ran into some nightmares with passing different flavors of complex numbers (C vs C++ vs ...) into `computeGaugeLoopTraceQuda`, but that could be revisited. I could also just follow the convention I have elsewhere of passing in an array/vector of doubles of length 2 * loops.

tests/gauge_force_test.cpp
weinbe2: Same issue as before, but I guess there's no harm in making everything use a `std::vector<...>`. Will do in a subsequent commit.

tests/gauge_force_test.cpp
maddyscientist: C and C++ complex numbers are defined by standard to be compatible, and even directly "castable" to each other.  This is one specific place where `reinterpret_cast` as opposed to `memcpy` is preferable.

To pass it as a pointer, one should use `traces.data()` as opposed to `&traces[0]`, with the latter indeed, likely being UB.

lib/milc_interface.cpp
weinbe2: Done locally, will push soon

lib/milc_interface.cpp
weinbe2: I don't disagree, I was just a sheeple following the MILC interface herd!

lib/gauge_polyakov_loop.cu
weinbe2: Done locally, will push soon

lib/gauge_polyakov_loop.cu
weinbe2: iirc it's because the `TuneKey` included the number of sites as opposed to the `##x##x##x##` volume?

lib/gauge_polyakov_loop.cu
weinbe2: Same as above, but this one's less critical since the relevant info can be extracted from the 3-d volume. Done locally.

lib/gauge_polyakov_loop.cu
weinbe2: Done locally

lib/gauge_polyakov_loop.cu
weinbe2: Me? Never! (Jokes aside, replaced by a `std::swap`)

lib/gauge_polyakov_loop.cu
weinbe2: Discussed offline, fixed locally.

lib/gauge_polyakov_loop.cu
weinbe2: Fixed locally, will push soon

lib/gauge_polyakov_loop.cu
weinbe2: Fixed locally, will push soon

lib/gauge_polyakov_loop.cu
weinbe2: Discussed online, we can revisit this if the need for a ring buffer pops up elsewhere, but there's no real need now.

include/kernels/gauge_polyakov_loop.cuh
weinbe2: Done locally, agreed, just never thought about it.

lib/interface_quda.cpp
weinbe2: It was just a copy + paste of what's done in the gauge path routine, but it isn't necessary there either, so I can poke that out in a separate commit.

lib/interface_quda.cpp
weinbe2: Two reasons: first, ultimately MILC, where I wanted to avoid using `loadGaugeQuda` so I didn't have to think about gauge residency. The second was `loadGaugeQuda` goes through a whole machinery of creating the field, plus sloppy fields, etc, which felt like an unnecessary overhead if some reconstructs, etc, aren't set properly.

Nothing that can't be book-kept around, but I ended up getting the feeling it'd take as many lines of book-keeping to make sure I don't introduce extra work via `loadGaugeQuda` as it did to just create a separate `plaqLoadGaugeQuda` flavor.

lib/interface_quda.cpp
weinbe2: Done locally

lib/interface_quda.cpp
weinbe2: Done locally

include/gauge_tools.h
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

include/kernels/gauge_force.cuh
weinbe2: I believe I checked and it was sub-percentile, if any.

include/quda.h
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/gauge_loop_trace.cu
weinbe2: Stripped out for now, in any case: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/milc_interface.cpp
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/gauge_polyakov_loop.cu
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/gauge_polyakov_loop.cu
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/gauge_polyakov_loop.cu
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/gauge_polyakov_loop.cu
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/gauge_polyakov_loop.cu
weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

lib/gauge_loop_trace.cu
weinbe2: Done locally, will push momentarily

weinbe2: https://github.com/lattice/quda/commit/4a893d83c4c5b4c15c9e0d07f8cc3cd1a0e1955a

include/quda_milc_interface.h
weinbe2: https://github.com/lattice/quda/pull/1308/commits/59cf5f0abf8e0c5cb25ef657c81d520720b48439

include/quda.h
weinbe2: https://github.com/lattice/quda/pull/1308/commits/59cf5f0abf8e0c5cb25ef657c81d520720b48439

include/gauge_path_helper.cuh
weinbe2: I think we could do something in the spirit of `std::pair<int,int>`, which would be a lot more readable. I do think it would be a lot of re-plumbing, though---there are a lot of arrays that look like `{ 0, 1, 7, 6 }` (but longer) that would need to become `{ {0, 1}, {1, 1}, {0, -1}, {1, -1} }` or something of the sort... the question is if the clarity added is worth the time spent debugging and propagating the changes out to the interface (and adding conversions to the MILC routines, etc)...

lib/gauge_polyakov_loop.cu
weinbe2: I don't think so? I don't see a place in `instantiate.h` where a default is specified (also, specifying a default like that could get awkward if recon 12 isn't enabled...) Happy to be corrected, though.

maddyscientist: https://github.com/lattice/quda/blob/develop/include/instantiate.h#L173 looks like the default is `ReconstructNo12` I think.

What do you mean regarding awkwardness?

weinbe2: Ah, yep, my eyes just glazed over that. Ignore the awkwardness thought. Rambles from drinking too much coffee too quickly.

weinbe2: https://github.com/lattice/quda/commit/58845f78894fa8a26a831a6d0a61b8a50b56a269

include/kernels/gauge_polyakov_loop.cuh
weinbe2: https://github.com/lattice/quda/commit/58845f78894fa8a26a831a6d0a61b8a50b56a269

include/kernels/gauge_polyakov_loop.cuh
weinbe2: https://github.com/lattice/quda/commit/58845f78894fa8a26a831a6d0a61b8a50b56a269

include/kernels/gauge_loop_trace.cuh
weinbe2: https://github.com/lattice/quda/commit/58845f78894fa8a26a831a6d0a61b8a50b56a269

include/kernels/gauge_polyakov_loop.cuh
weinbe2: https://github.com/lattice/quda/commit/58845f78894fa8a26a831a6d0a61b8a50b56a269

tests/gauge_force_test.cpp
weinbe2: https://github.com/lattice/quda/pull/1308/commits/1e3db7ab2f46d71cf3b88aa7e904b8b4f9c8c834

tests/gauge_force_test.cpp
weinbe2: https://github.com/lattice/quda/pull/1308/commits/1e3db7ab2f46d71cf3b88aa7e904b8b4f9c8c834

tests/gauge_force_test.cpp
weinbe2: Discussed offline that, because the QUDA interface needs to be C-friendly, there's no point in wrapping this in a `std::vector` just to unwrap it before passing it to a function.

include/gauge_path_helper.cuh
weinbe2: Done

lib/interface_quda.cpp
weinbe2: https://github.com/lattice/quda/pull/1308/commits/04f84f575b3a13a6d8e1bf0c520553f4c038c033

lib/interface_quda.cpp
weinbe2: Cleaned up in https://github.com/lattice/quda/pull/1308/commits/e24bc686fab8a6cf6eaa4a70bb12ce4fdf479cfa , thanks for the offline iterations!

tests/host_reference/gauge_force_reference.cpp
```diff
@@ -277,57 +301,102 @@ int gf_neighborIndexFullLattice(size_t i, int dx[], const lattice_t &lat)
...
+  su3_matrix tmat;
+  dcomplex accum;
+  memset(&accum, 0, sizeof(accum));
+  int dx[4];
+
+  for (size_t i = 0; i < lat.volume; i++) {
+    memset(dx, 0, sizeof(dx));
+    tmat = compute_gauge_path(sitelink, i, path, len, dx, lat);
+    auto tr = trace_su3(&tmat);
+    accum.real += tr.real;
```
maddyscientist: could not use `CADD` and `CMUL` a few lines below?

lib/milc_interface.cpp
```diff
@@ -645,6 +645,169 @@ void qudaGaugeForce(int precision, int num_loop_types, double milc_loop_coeff[3]
...
+    qudaGaugeParam.use_resident_gauge = false;
+  } else {
+    qudaGaugeParam.use_resident_gauge = true;
+    qudaGaugeParam.make_resident_gauge = true;
+  }
+
+  return qudaGaugeParam;
+}
+
+void qudaGaugeLoopTracePhased(int precision, double *traces, int** input_path_buf, int* path_length, double *loop_coeff, int num_paths,
```
maddyscientist: Wondering if we might want `qudaGaugeLoopTrace`, `qudaPlaquette`, `qudaPolyakovLoop` functions with the phased versions thereof being thin wrappers around those that just deal with the phasing aspect.  Not that important, just thinking out loud.

lib/check_params.h
```diff
@@ -1026,14 +1026,32 @@ void printQudaGaugeObservableParam(QudaGaugeObservableParam *param)
 #ifdef INIT_PARAM
   P(su_project, QUDA_BOOLEAN_FALSE);
   P(compute_plaquette, QUDA_BOOLEAN_FALSE);
+  P(compute_polyakov_loop, QUDA_BOOLEAN_FALSE);
+  P(compute_gauge_loop_trace, QUDA_BOOLEAN_FALSE);
+  P(traces, nullptr);
+  P(input_path_buff, nullptr);
+  P(path_length, nullptr);
+  P(loop_coeff, nullptr);
+  P(num_paths, 0);
```
maddyscientist: `num_paths`, `max_length` and `factor` should be invalid initialized here.  I think what should be done here is that lines 1047-1051 should just be moved below the `endif` on line 1055, and lines 1035-1037 should be deleted.

maddyscientist (commented):
I've left a few small things left to be addressed, but this PR is 99.9% there 😄   Once the last couple of things are addressed, I think we're good to merge.

lib/milc_interface.cpp
weinbe2: I was torn about this. We only added the `Phased` versions with @mathiaswagner 's residency work, and left the un-`Phased` versions for backwards compatibility/`C` not having default arguments. There's no backwards compatibility here, so having a non-phased version would just be confusing... (though maybe no more confusing than having `Phased` flavors of anything in the first place)

tests/host_reference/gauge_force_reference.cpp
weinbe2: `CADD` is a good call, there's no flavor for multiplying by a `double` but I popped something in locally, will push momentarily.

lib/check_params.h
weinbe2: Makes sense, changed locally, will push after some cursory testing

weinbe2: So in practice it couldn't be moved out of the `if`/`endif`/etc outright, due to different definitions of `P` and the local struct being `*param` or `ret`, but I think I found a happy medium.

weinbe2: https://github.com/lattice/quda/commit/0eacdf5a4fa5ffd1631db7509707665b0b59f8b4

tests/host_reference/gauge_force_reference.cpp
weinbe2: https://github.com/lattice/quda/commit/0eacdf5a4fa5ffd1631db7509707665b0b59f8b4

lib/milc_interface.cpp
maddyscientist: Sounds fine.  We can leave it as is, and address at a later date if ever we need to.

maddyscientist (approved):
Thanks for the last few fixes.  This looks good to go.

weinbe2:
> Thanks for the last few fixes. This looks good to go.

Excellent --- I'm doing a few last checks (just making sure `clang-format` didn't accidentally do something stupid), and when Jenkins completes we can get this merged, after which I'll do some last tests to the MILC PR.

maddyscientist:
@mathiaswagner anything to offer on this PR, or are we good to merge?

mathiaswagner:
I'll have another look today and guess we can then merge later.

tests/sanity_check.sh
```diff
@@ -8,7 +8,7 @@ fail_msg="*****************************Sanity check for quda failed!************
 function basic_sanity_check {
```
mathiaswagner, line 8: Not sure this file should be add here. If we want something like this we should add it to ctest.

include/quda.h
```diff
@@ -802,14 +802,25 @@ extern "C" {
...
+    double *loop_coeff;                   /**< Multiplicative factor for each loop */
+    int num_paths;                        /**< Total number of paths */
+    int max_length;                       /**< Maximum length of any path */
+    double factor;                        /**< Global multiplicative factor to apply to each loop trace */
+    QudaBoolean compute_qcharge;          /**< Whether to compute the topological charge and field energy */
+    double qcharge;                       /**< Computed topological charge */
+    double energy[3];                     /**< Total, spatial and temporal field energies, respectively */
+    QudaBoolean compute_qcharge_density;  /**< Whether to compute the topological charge density */
     void *qcharge_density; /**< Pointer to host array of length volume where the q-charge density will be copied */
+    QudaBoolean remove_staggered_phase; /**< Whether or not the resident gauge field has staggered phases applied and if they should be removed */
```
mathiaswagner: only reading the doxygen leacves me confused. The description mentions phase applied true/false and remove_phase true/false. But there is only a single bool here. Please clarify.

tests/gauge_path_test.cpp
```diff
@@ -214,6 +235,140 @@ void gauge_force_test(bool compute_force = true)
   delete Mom_ref_milc;
 }
 
+void gauge_loop_test()
+{
+  int max_length = 6;
+
+  setVerbosityQuda(QUDA_VERBOSE, "", stdout);
```
mathiaswagner: should that respect the commandline option?

tests/gauge_path_test.cpp
```diff
@@ -214,6 +235,140 @@ void gauge_force_test(bool compute_force = true)
...
+  int **trace_path_p = new int *[num_paths];
+  for (int i = 0; i < num_paths; i++) {
+    trace_path_p[i] = new int[max_length];
+    for (int j = 0; j < trace_loop_length[i]; j++) trace_path_p[i][j] = trace_path[i][j];
+  }
+
+  quda::GaugeFieldParam param(gauge_param);
+  param.create = QUDA_NULL_FIELD_CREATE;
+  param.order = QUDA_QDP_GAUGE_ORDER;
+  auto U_qdp = new quda::cpuGaugeField(param);
```
mathiaswagner: does this need to be a ptr?

include/kernels/gauge_force.cuh
```diff
@@ -5,70 +5,19 @@
...
   struct GaugeForceArg : kernel_param<> {
-    using Float = Float_;
+    using real = typename mapper<store_t>::type;
     static constexpr int nColor = nColor_;
     static constexpr bool compute_force = force_;
+    using Link = Matrix<complex<real>, nColor>;
     static_assert(nColor == 3, "Only nColor=3 enabled at this time");
-    typedef typename gauge_mapper<Float,recon_u>::type Gauge;
-    typedef typename gauge_mapper<Float,recon_m>::type Mom;
+    typedef typename gauge_mapper<real,recon_u>::type Gauge;
```
mathiaswagner: how about using using here for consistency?

include/kernels/gauge_loop_trace.cuh
```diff
@@ -0,0 +1,94 @@
...
+  template <typename store_t, int nColor_, QudaReconstructType recon_>
+  struct GaugeLoopTraceArg : public ReduceArg<array<double, 2>>  {
+    using real = typename mapper<store_t>::type;
+    using reduce_t = array<double, 2>;
+    static constexpr unsigned int max_n_batch_block = max_n_batch_block_loop_trace();
+    static constexpr int nColor = nColor_;
+    static constexpr QudaReconstructType recon = recon_;
+    using Link = Matrix<complex<real>, nColor>;
+    static_assert(nColor == 3, "Only nColor=3 enabled at this time");
+    typedef typename gauge_mapper<real,recon>::type Gauge;
```
mathiaswagner: using?

include/kernels/gauge_loop_trace.cuh
```diff
@@ -0,0 +1,94 @@
...
+
+    const Gauge u;
+
+    const double factor;
```
mathiaswagner: mayeb a comment what factor is used for?

include/gauge_path_helper.cuh
mathiaswagner: as I just had the same thought: This calls for being able to use some smart ptr ... Mayeb for now add a comment?

lib/interface_quda.cpp
```diff
@@ -4621,6 +4647,7 @@ void computeHISQForceQuda(void* const milc_momentum,
 
   profileHISQForce.TPSTART(QUDA_PROFILE_COMPUTE);
   hisqStaplesForce(*cudaOutForce, *cudaInForce, *cudaGauge, act_path_coeff);
+  qudaDeviceSynchronize();
```
mathiaswagner, line 4650: I guess all those qudaDeviceSynchronize() are maing sure we get correct timing but do they add overhead?

mathiaswagner:
Looks good to me but I still left a few minor comments. Thanks for adding that @weinbe2 !

mathiaswagner (approved):

tests/sanity_check.sh
weinbe2: To be fair, this was a bit of a `sed` adventure since I renamed `gauge_force_test` to `gauge_path_test`. I'm down with removing this file since it seems a bit legacy (and I've never used it), but I think it should be left for a separate PR. Any objections?

lib/interface_quda.cpp
weinbe2: That was a @maddyscientist addition so I defer to her, I didn't explicitly check that contribution, but any cost certainly gets amortized by the e2e speed-up.

include/quda.h
weinbe2: tl;dr this is a bit of a MILC-specific kludge to deal with this issue: https://github.com/lattice/quda/issues/1315 . I can add a direct reference to that issue in a comment.

tests/gauge_path_test.cpp
weinbe2: Not that this is a good excuse, that what was there before, but I can patch that up.

include/quda.h
weinbe2: Added locally, will push soon

tests/gauge_path_test.cpp
weinbe2: Addressed locally, will push soon.

tests/gauge_path_test.cpp
weinbe2: Fixed throughout the file, good call, will push momentarily

include/kernels/gauge_loop_trace.cuh
weinbe2: Fixed locally

include/kernels/gauge_loop_trace.cuh
weinbe2: Fixed locally

include/kernels/gauge_force.cuh
weinbe2: Fixed locally

lib/interface_quda.cpp
weinbe2: Discussed offline, cost is negligible (especially relative to 100s of force kernels inside the function)

include/kernels/gauge_loop_trace.cuh
weinbe2: https://github.com/lattice/quda/pull/1308/commits/ccb7dd9dfe1dbb485f0d2f05456c9276a79fbafb

include/kernels/gauge_loop_trace.cuh
weinbe2: https://github.com/lattice/quda/pull/1308/commits/ccb7dd9dfe1dbb485f0d2f05456c9276a79fbafb

include/kernels/gauge_force.cuh
weinbe2: https://github.com/lattice/quda/pull/1308/commits/ccb7dd9dfe1dbb485f0d2f05456c9276a79fbafb

tests/gauge_path_test.cpp
weinbe2: https://github.com/lattice/quda/pull/1308/commits/ccb7dd9dfe1dbb485f0d2f05456c9276a79fbafb

tests/gauge_path_test.cpp
weinbe2: https://github.com/lattice/quda/pull/1308/commits/ccb7dd9dfe1dbb485f0d2f05456c9276a79fbafb

include/quda.h
weinbe2: https://github.com/lattice/quda/pull/1308/commits/ccb7dd9dfe1dbb485f0d2f05456c9276a79fbafb

weinbe2:
All comments addressed @mathiaswagner , let me know if you have any further thoughts, otherwise I can get this merged once Jenkins finishes.

mathiaswagner:
I think with all comments addressed just feel free to use my pre-approval to merge once Jenkins is happy.

PR #1309, opened 2022-08-04
Hotfix/blas interface lu test

cpviolator:
This PR adds complex double and complex single batched, strided inversion via LU decomposition on the device. The test file 'blas_interface_test.cpp` invokes the test via:
```
./blas_interface_test --blas-test-type lu-inv --blas-lu-inv-mat-size <N>  --blas-data-type <C/Z>
``` 
where `<N>` is rank of the square matrix.

GEMM command line parameters and QUDA param struct values now contain the respective strings `-gemm-` and `_gemm_` to delineate.

For the LU inversion we allow a tolerance of 5000x of the `std::numeric_limits<float/double>::epsilon();` to account for the near degeneracy/singularities introduced by pure random data, as well as minor variations between host and device algorithms and order of arithmetic operations.

The ctest suite has been updated to include the new BLAS functions, and can be invoked (for the blas test only) via
```
ctest -R blas_interface_test -V
```

### DIFF
--- a/lib/check_params.h
+++ b/lib/check_params.h
@@ -1120,6 +1120,7 @@ void printQudaBLASParam(QudaBLASParam *param)
   P(batch_count, 1);
   P(data_type, QUDA_BLAS_DATATYPE_S);
   P(data_order, QUDA_BLAS_DATAORDER_ROW);
+  P(blas_type, QUDA_BLAS_INVALID);
   P(inv_mat_size, INVALID_INT);
 #else
   P(trans_a, QUDA_BLAS_OP_INVALID);
@@ -1139,6 +1140,7 @@ void printQudaBLASParam(QudaBLASParam *param)
   P(batch_count, INVALID_INT);
   P(data_type, QUDA_BLAS_DATATYPE_INVALID);
   P(data_order, QUDA_BLAS_DATAORDER_INVALID);
+  P(blas_type, QUDA_BLAS_INVALID);
   P(inv_mat_size, INVALID_INT);
 #endif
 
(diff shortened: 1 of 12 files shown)


weinbe2:
Thanks for the contribution, @cpviolator ! I've done a cursory review and this looks good, I'll do a more in-depth one plus a compile+run confirmation momentarily. I especially appreciate that you've extended the batched invert code to support complex double matrices, which has some applications for more stable Kahler-Dirac inverses.

lib/check_params.h
```diff
@@ -1120,6 +1120,7 @@ void printQudaBLASParam(QudaBLASParam *param)
   P(batch_count, 1);
   P(data_type, QUDA_BLAS_DATATYPE_S);
   P(data_order, QUDA_BLAS_DATAORDER_ROW);
+  P(inv_mat_size, INVALID_INT);
```
weinbe2, line 1124: This also needs `blas_type`

tests/blas_interface_test_gtest.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+  auto param = GetParam();
+  auto test_type = ::testing::get<0>(param);
+  auto data_type = ::testing::get<1>(param);
+  switch (test_type) {
+  case QUDA_BLAS_GEMM: {
+    auto deviation_gemm = gemm_test(param);
+    decltype(deviation_gemm) tol_gemm;
+    switch (data_type) {
+    case 0:
```
weinbe2: Can `QUDA_BLAS_DATATYPE_S`, etc, be used here to avoid the "magic number" 0, 1, 2, 3?

weinbe2:
This looks good, besides running the `ctest` I ran a quick MG workflow to sanity check the inversions required there. Pending some thoughts on my comments, this is good to go.

lib/check_params.h
cpviolator: a43b61988

tests/blas_interface_test_gtest.hpp
cpviolator: Swapped both instances, good catch. a43b61988

weinbe2 (approved):
All looks great, thanks @cpviolator !

PR #1310, opened 2022-08-04
Fused DWF + NVSHMEM

hummingtree:
~~This PR makes the DWF fused kernels run with NVSHMEM. Running on one node on Selene with 1x2x2x2 and Ls = 12, getting (performance numbers are in GFLOPS)~~ No there is no speed up.


maddyscientist (changes requested):
Nice to see there is a regime that gives speedup from NVSHMEM on Selene.
* Although it's never used, the `constantInv` function should be updated to handle an `active` argument, in case this code is ever used again.
* Also, perhaps the need for the `active` flag argument requirement should documented in the code comments

hummingtree:
Thanks @maddyscientist. I have added update the doxygen to cover `active`, and also applied the changes to `constantInv` in https://github.com/lattice/quda/pull/1310/commits/4040ff814f9a4e8aa4b3088313e8cea103557aef. I have also tested `constantInv` code path.

maddyscientist (approved):
Thanks for the additions @hummingtree.  This PR is good to go from my perspective.

weinbe2 (approved):
This is great --- I appreciate how "simple" the PR is from a lines-of-code standpoint, which is a statement of clean design.

PR #1312, opened 2022-08-05
Feature/quark smearing.pr

alexstrel:
PR introduces an implementation of the Gaussian smearing for staggered quarks. It also includes computation of two link field, simple test instruction: 
 mpirun -np 1 ./tests/staggered_gsmear_test --dslash-type asqtad --prec double --gridsize 1 1 1 1 --dim 16 16 16 48 --load-gauge path-togauge-config


lib/CMakeLists.txt
```diff
@@ -352,6 +352,10 @@ if(QUDA_LAPLACE)
   target_compile_definitions(quda PUBLIC GPU_LAPLACE)
 endif(QUDA_LAPLACE)
 
+if(QUDA_TWOLINK_GSMEAR)
```
mathiaswagner: is that used anywhere or can we just remove the `if` ?

tests/utils/command_line_params.cpp
```diff
@@ -1023,3 +1029,13 @@ void add_testing_option_group(std::shared_ptr<QUDAApp> quda_app)
   auto opgroup = quda_app->add_option_group("Testing", "Options controlling automated testing");
   opgroup->add_option("--enable-testing", enable_testing, "Enable automated testing (default false)");
 }
+
+void add_quark_smear_option_group(std::shared_ptr<QUDAApp> quda_app)
+{
+  auto opgroup = quda_app->add_option_group("Quark smearing", "Options controlling quark smearing testing");
+  opgroup->add_option("--compute_two_link", compute_two_link, "Compute two link field (default true)");
+  opgroup->add_option("--delete_two_link", delete_two_link, "Compute two link field (default true)");
```
mathiaswagner: the help text needs an update

lib/CMakeLists.txt
hwancheolJeong: Sorry for the slow response. We intend for users to turn on this flag when building QUDA only if they need it. We will create a wiki page for it.

mathiaswagner: I was more worried that there is no corresponding cmake option.

hwancheolJeong: Thanks for letting me know that! I have just added the cmake option.

include/dirac_quda.h
```diff
@@ -230,6 +230,16 @@ namespace quda {
...
 			    const double &k) const = 0;
 
+    /**
+       @brief Similar to the Xpay version of Dslash, but used only by the Laplace op for 
+       smearing.
+    */    
+    virtual void SmearOp(ColorSpinorField &, const ColorSpinorField &, 
+                         const double&, const double &, const int &,  const QudaParity) const 
+    {
+      errorQuda("Not implemented!\n");
```
maddyscientist: We don't need the trailing `\n` for `errorQuda` calls since it will apply this automatically.

include/dirac_quda.h
```diff
@@ -1326,6 +1336,8 @@ namespace quda {
      */
     void createCoarseOp(GaugeField &Y, GaugeField &X, const Transfer &T, double kappa, double mass, double mu = 0.,
                         double mu_factor = 0., bool allow_truncation = false) const;
+
+    void SmearOp(ColorSpinorField &out, const ColorSpinorField &in, const double &a, const double &b, const int &t0,  const QudaParity parity) const;
```
maddyscientist: doxygen for this function

include/dirac_quda.h
```diff
@@ -1556,6 +1568,8 @@ namespace quda {
       @param[in] stream Which stream to run the prefetch in (default 0)
     */
     virtual void prefetch(QudaFieldLocation mem_space, qudaStream_t stream = device::get_default_stream()) const;
+    
+    void SmearOp(ColorSpinorField &out, const ColorSpinorField &in, const double &a, const double &b, const int &t0,  const QudaParity parity) const;
```
maddyscientist: doxygen for this function

include/dirac_quda.h
```diff
@@ -230,6 +230,16 @@ namespace quda {
 			    const QudaParity parity, const ColorSpinorField &x,
 			    const double &k) const = 0;
 
+    /**
+       @brief Similar to the Xpay version of Dslash, but used only by the Laplace op for 
+       smearing.
+    */    
+    virtual void SmearOp(ColorSpinorField &, const ColorSpinorField &,
```
maddyscientist, line 256: What makes this different from the `DslashXpay` that it necessitates its own function?

include/dslash_quda.h
```diff
@@ -725,6 +725,17 @@ namespace quda
...
                               const int *comm_override, TimeProfile &profile);
+                              
+  /**
+     @brief Apply the 2-link staggered operator to a color-spinor field.
+     @param[out] out Result color-spinor field
+     @param[in] in Input color-spinor field
+     @param[in] dir Direction of the derivative 0,1,2,3 to omit (-1 is full 4D)
+     @param[in] a Scale factor applied to derivative
+     @param[in] b Scale factor applied to aux field
+  */
```
maddyscientist, line 741: The doxygen doesn't match the list of parameters passed to the function.

include/enum_quda.h
```diff
@@ -571,6 +571,12 @@ typedef enum QudaGaugeSmearType_s {
   QUDA_GAUGE_SMEAR_INVALID = QUDA_INVALID_ENUM
 } QudaGaugeSmearType;
 
+typedef enum QudaFermionSmearType_s {
```
maddyscientist, line 575: This addition also needs to be propagated to `enum_fortran_quda.h`

include/index_helper.cuh
```diff
@@ -121,6 +121,19 @@ namespace quda {
   {
     return linkIndexDn<I, -1>(x, X, mu);
   }
+  
+  /**
+     Compute the checkerboard 1-d index from the 4-d coordinate x[] -2 in the mu direction
+
+     @return 1-d checkerboard index
+     @param x 4-d lattice index
```
maddyscientist: Missing `[in]` / `[out]` labels from the parameters, and the `@return` should be after the parameter list.

include/index_helper.cuh
```diff
@@ -165,6 +178,20 @@ namespace quda {
   __device__ __host__ inline int linkIndexP1(const Coord &x, const I X[4], const int mu) {
     return linkIndexDn<I, 1>(x, X, mu);
   }
+  
+ /**
+     Compute the checkerboard 1-d index from the 4-d coordinate x[] +2 in the mu direction
+
+     @return 1-d checkerboard index
```
maddyscientist, line 210: Missing [in] / [out] labels from the parameters, and the @return should be after the parameter list.

include/kernels/staggered_quark_smearing.cuh
```diff
@@ -0,0 +1,229 @@
...
+#include <color_spinor.h>
+#include <dslash_helper.cuh>
+#include <index_helper.cuh>
+#include <kernels/dslash_pack.cuh> // for the packing kernel
+
+namespace quda
+{
+
+  /**
+     @brief Parameter structure for driving the covariatnt derivative operator
```
maddyscientist: typo (covariatnt -> covariant)

include/kernels/staggered_quark_smearing.cuh
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+     Applies the off-diagonal part of the covariant derivative operator
+
+     @param[out] out The out result field
+     @param[in,out] arg Parameter struct
+     @param[in] U The gauge field
+     @param[in] coord Site coordinate struct
+     @param[in] parity The site parity
+     @param[in] idx Thread index (equal to face index for exterior kernels)
```
maddyscientist: Doxygen list doesn't quite match the function parameters being passed

include/kernels/staggered_quark_smearing.cuh
```diff
@@ -0,0 +1,229 @@
...
+          // Forward gather - compute fwd offset for vector fetch
+          const bool ghost = (coord[d] + 2 >= arg.dim[d]) && isActive<kernel_type>(active, thread_dim, d, coord, arg);//1=>2
+	  
+          if (doHalo<kernel_type>(d) && ghost) {//?
+
+            const int ghost_idx = ghostFaceIndexStaggered<1>(coord, arg.dim, d, 2);//check nFace=2, requires improved staggered fields
+            const Link U = arg.U(d, coord.x_cb, parity);
+            const Vector in = arg.in.Ghost(d, 1, ghost_idx, their_spinor_parity);//?
+
+            out += U * in;
```
maddyscientist: I think this line, and equivalently on lines 127, 141 and 150 should be replaced with a call to `mv_add`.  This will ensure the operations are performed using the optimal number of `fma` instructions.  See for example [here](https://github.com/lattice/quda/blob/develop/include/kernels/dslash_staggered.cuh#L105).

include/kernels/staggered_quark_smearing.cuh
```diff
@@ -0,0 +1,229 @@
...
+  /**
+     @brief Parameter structure for driving the covariatnt derivative operator
+  */
+  template <typename Float, int nSpin_, int nColor_, int nDim, QudaReconstructType reconstruct_>
+  struct StaggeredQSmearArg : DslashArg<Float, nDim> {
+    static constexpr int nColor = 3;
+    static constexpr int nSpin  = 1;
+    static constexpr bool spin_project = false;
+    static constexpr bool spinor_direct_load = false; // false means texture load
+    typedef typename colorspinor_mapper<Float, nSpin, nColor, spin_project, spinor_direct_load>::type F;
```
maddyscientist: style note: prefer the use of `using` over `typedef`.

include/kernels/staggered_two_link.cuh
```diff
@@ -0,0 +1,102 @@
...
+        // compute the forward two links
+        int point_c = e_cb;
+        int point_d = linkIndexShift(x,dx,arg.E);
+
+        Link Ucd = arg.link(mu, point_c, parity);
+        Link Ude = arg.link(mu, point_d, 1-parity);
+
+        Link temp = Ucd*Ude;
+
+        arg.outA(mu, x_cb, parity) = temp;
```
maddyscientist: you could jut write `arg.outA(mu, x_cb, parity) = Ucd * Ude;`

include/ks_qsmear.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <quda_internal.h>
+#include <quda.h>
```
maddyscientist: Both the `quda.h` and `quda_internal.h` includes should be needed in this include.  We should be striving in general to only include what we use.

include/quda.h
```diff
@@ -1342,6 +1347,8 @@ extern "C" {
 
   void computeKSLinkQuda(void* fatlink, void* longlink, void* ulink, void* inlink,
                          double *path_coeff, QudaGaugeParam *param);
+                         
+  void computeTwoLinkQuda(void *twolink, void *inlink, QudaGaugeParam *param);
```
maddyscientist, line 1359: Add doxygen for this function.

include/quda.h
```diff
@@ -1701,6 +1708,21 @@ extern "C" {
...
+   *                    and operator which will be applied to the spinor
+   * @param n_steps     Number of steps to apply
+   * @param width       The width of the Gaussian
+   * @param compute_2link nonzero to compute two-link,
+   *                      zero to reuse gaugeSmeared (default = 1)
+   * @param delete_2link nonzero to delete two-link after smearing,
+   *                     zero to keep two-link for future use (default = 0)
+   * @param t0          Set if the input spinor is on a time slice (default = -1)
+   */
+  void performTwoLinkGaussianSmearNStep(void *h_in, QudaInvertParam *inv_param, const int n_steps, const double width, const int compute_2link /*= 1*/, const int delete_2link /*= 0*/, const int t0 /*= -1*/);
```
maddyscientist: Add missing `[in]` / `[out]` labels to this function.

CMakeLists.txt
```diff
@@ -157,6 +157,8 @@ option(QUDA_COVDEV "build code for covariant derivative" OFF)
 option(QUDA_LAPLACE "build laplace operator" OFF)
 option(QUDA_QIO "build QIO code for binary I/O" OFF)
 
+option(QUDA_TWOLINK_GSMEAR "build code for two-link Gaussian smearing" OFF)
```
maddyscientist: This option should have a name that will allow us to aggregate similar options together in the `ccmake` list.  E.g., `QUDA_SMEAR_GAUSS_TWOLINK` which would mean that if we have other `QUDA_SMEAR_....` options they will appear next to each other in the ccmake list.

lib/dirac_staggered.cpp
```diff
@@ -138,6 +138,33 @@ namespace quda {
...
+    if( is_time_slice && ( t0_local < 0 || t0_local >= in.X(3) ) ) t0_local = -1; // when source is not in this local lattice
+
+    int comm_dim[4] = {};
+    // only switch on comms needed for directions with a derivative
+    for (int i = 0; i < 4; i++) {
+      comm_dim[i] = comm_dim_partitioned(i);
+      if (laplace3D == i) comm_dim[i] = 0;
+    }
+
+    ApplyStaggeredQSmear(out, in, *gauge, t0_local, is_time_slice, QUDA_INVALID_PARITY, laplace3D, dagger, comm_dim, profile); // parity is not used
```
maddyscientist: `DiracStaggered::SmearOp` is missing the check for single parity that `DiracImprovedStaggered::SmearOp` has.

include/quda_milc_interface.h
```diff
@@ -594,6 +594,17 @@ extern "C" {
...
+  /**
+     Free the smeared gauge field allocated in QUDA.
+   */
+  void qudaFreeGaugeSmeared();
+  
+  /**
+     Alias of qudaFreeGaugeSmeared().     
+     Free the two-link field allocated in QUDA.
+   */
+  void qudaFreeTwoLink();
```
maddyscientist, line 601: Why introduce this alias?

tests/CMakeLists.txt
```diff
@@ -110,6 +110,11 @@ if(QUDA_DIRAC_STAGGERED)
   quda_checkbuildtest(staggered_eigensolve_test QUDA_BUILD_ALL_TESTS)
   install(TARGETS staggered_eigensolve_test ${QUDA_EXCLUDE_FROM_INSTALL} DESTINATION ${CMAKE_INSTALL_BINDIR})
 
+  add_executable(staggered_gsmear_test staggered_gsmear_test.cpp)
+  target_link_libraries(staggered_gsmear_test ${TEST_LIBS})
+  quda_checkbuildtest(staggered_gsmear_test QUDA_BUILD_ALL_TESTS)
+  install(TARGETS staggered_gsmear_test ${QUDA_EXCLUDE_FROM_INSTALL} DESTINATION ${CMAKE_INSTALL_BINDIR})
+
 endif()
```
maddyscientist: There needs to be a `ctest` instance added in this file here as well so that the `staggered_gsmear_test` will run as part of regression tests.

tests/utils/staggered_host_utils.cpp
```diff
@@ -324,6 +326,150 @@ void computeLongLinkCPU(void **longlink, void **sitelink, QudaPrecision prec, vo
...
+        }
+      }
+    } // 4-d volume
+  }   // right-hand-side
+  return;
+}
+
+void staggeredTwoLinkGaussianSmear(quda::ColorSpinorField &out, void *qdp_twolnk[], void** ghost_twolnk,  quda::ColorSpinorField &in, QudaGaugeParam* /*qudaGaugeParam*/, QudaInvertParam* /*inv_param*/, const int oddBit, const double /*width*/, const int /*t0*/, QudaPrecision prec)
+{
+  //const int nSrc = in.X(4);
```
maddyscientist: Delete this line

tests/utils/host_utils.cpp
```diff
@@ -906,6 +906,28 @@ int compare_floats(void *a, void *b, int len, double epsilon, QudaPrecision prec
...
+    if (diff > epsilon || std::isnan(diff)) {
+      //printfQuda("ERROR: i=%d, a[%d]=%f, b[%d]=%f\n", i, i, a[i], i, b[i]);
+      return diff;
+    }
+    global_diff = std::max(global_diff, diff);
+  }
+  return global_diff;
+}
+
+double compare_floats_v2(void *a, void *b, int len, double epsilon, QudaPrecision precision)
```
maddyscientist, line 924: Why is a `v2` needed?  Add some comments to explain what `v2` does different from the original.

tests/staggered_gsmear_test_utils.h
```diff
@@ -0,0 +1,405 @@
...
+  
+  void *milc_inlink       = nullptr;
+  void *milc_twolnk       = nullptr;
+  //
+  GaugeField *cpuTwoLink  = nullptr;
+
+  QudaGaugeParam  gauge_param;//
+  QudaInvertParam inv_param;
+  //
+  std::unique_ptr<ColorSpinorField> spinor;
```
maddyscientist: We never need to use smart pointers now for `ColorSpinorField` objects now that we have both default and move constructors.  These should be replaced with simply `ColorSpinorField` objects.

tests/staggered_gsmear_test_utils.h
```diff
@@ -0,0 +1,405 @@
...
+    printfQuda("Tuning...\n");
+    gsmearCUDA(1);
+
+
+    GSmearTime gsmear_time = gsmearCUDA(niter);
+    if(gtest_type == gsmear_test_type::GaussianSmear) *spinorRef = *spinor;
+
+    if (print_metrics) {
+      printfQuda("%fus per kernel call\n", 1e6 * gsmear_time.event_time / niter);
+#if 0
```
maddyscientist: Is the code below going to enabled at some point?

lib/interface_quda.cpp
```diff
@@ -5356,6 +5443,160 @@ void performWuppertalnStep(void *h_out, void *h_in, QudaInvertParam *inv_param,
...
+  cpuParam.nSpin = 1;
+  // QUDA style pointer for host data.
+  ColorSpinorField *in_h = ColorSpinorField::Create(cpuParam);
+
+  // Device side data.
+  ColorSpinorParam cudaParam(cpuParam);
+  cudaParam.location = QUDA_CUDA_FIELD_LOCATION;
+  cudaParam.create   = QUDA_ZERO_FIELD_CREATE;
+  cudaParam.setPrecision(inv_param->cuda_prec, inv_param->cuda_prec, true);
+  ColorSpinorField *in    = ColorSpinorField::Create(cudaParam);
```
maddyscientist: Can replace these with `ColorSpinorField in(cudaParam)`, etc.

lib/interface_quda.cpp
```diff
@@ -5356,6 +5443,160 @@ void performWuppertalnStep(void *h_out, void *h_in, QudaInvertParam *inv_param,
...
+    qsmear_op.Expose()->SmearOp(*out, *in, a, 0.0, t0, parity);
+    if (getVerbosity() >= QUDA_DEBUG_VERBOSE) {
+      double norm = blas::norm2(*out);
+      printfQuda("Step %d, vector norm %e\n", i, norm);
+    }
+    blas::xpay(*temp1, -1.0, *out);
+    blas::zero(*temp1);
+  }
+
+#if 0
```
maddyscientist: delete if not going to be used

lib/interface_quda.cpp
```diff
@@ -5356,6 +5443,160 @@ void performWuppertalnStep(void *h_out, void *h_in, QudaInvertParam *inv_param,
 
   profileWuppertal.TPSTOP(QUDA_PROFILE_TOTAL);
 }
+ 
+
+void performTwoLinkGaussianSmearNStep(void *h_in, QudaInvertParam *inv_param, const int n_steps, const double width, const int compute_2link = 1, const int delete_2link = 0, const int t0 = -1)
```
maddyscientist: It would be better to collect the parameters for this function into either a new struct , or to augment the `QudaInvertParam` with these new parameters.

include/quda.h
maddyscientist: The doxygen mentions that the parameters have default values, but the signature shows these are not present.  Indeed, default parameters are not allowed with regular C.

lib/interface_quda.cpp
maddyscientist: Moreover, placing the parameters into the `QudaInvertParam` struct would allow for defaults to be set for these parameteres.

tests/staggered_gsmear_test_utils.h
```diff
@@ -0,0 +1,405 @@
...
+      spinor.reset();
+      spinorRef.reset();
+    }
+
+    freeGaugeQuda();
+
+    commDimPartitionedReset();
+  }
+
+  GSmearTime gsmearCUDA(int niter) //niter
```
maddyscientist: Probably this function should not be called `gsmearCUDA` since QUDA runs on more than CUDA nowadays 😉

lib/staggered_quark_smearing.cu
```diff
@@ -0,0 +1,221 @@
...
+          if( !(arg.commDim[i]) ) continue;
+          arg.threadDimMapLower[i] = ( prev>=0 ? arg.threadDimMapUpper[prev] : 0 );
+          arg.threadDimMapUpper[i] = arg.threadDimMapLower[i] + this->Nface() * (in.GhostFaceCB())[i];
+          prev = i;
+        }
+      }
+
+      // operator is Hermitian so do not instantiate dagger
+      if (arg.nParity == 1) {
+        if(arg.is_t0_kernel) //FIXME we reuse xpay non-type template parameter for t0 option!
```
maddyscientist: Is this really a `FIXME` or just helpful information?  I think abusing the `xpay` template like this seems like a reasonable thing to do.

lib/staggered_two_link_quda.cu
```diff
@@ -0,0 +1,74 @@
...
+      }
+    };
+#if defined(GPU_STAGGERED_DIRAC) && defined(GPU_TWOLINK_GSMEAR)
+  void computeTwoLink(GaugeField &newTwoLink, const GaugeField &link)
+    {
+      checkNative(newTwoLink, link);
+      checkLocation(newTwoLink, link);
+      checkPrecision(newTwoLink, link);
+      instantiate<ComputeTwoLink, WilsonReconstruct>(link, newTwoLink);
+
```
maddyscientist: delete lines 65-66 (general style it to leave trivial `return` statements as implicit)

maddyscientist (changes requested):
I've left initial review comments.  This is really great work @alexstrel, @hwancheolJeong and @stevengottlieb, and is a great example of the Dslash framework in action 😄   Thank you for this.

The main things to be addressed are documentation, cleaning up the doxygen comments, some adaptions for modern QUDA (e.g., just use `ColorSpinorField` objects instead of (smart) pointers), and finally getting some tests added to the `ctest` list.

include/dirac_quda.h
alexstrel: It's an additional method, since requires a two-link field in addition to the regular staggered field. It's also optional, while dslash Xpay is mandatory.  (It is similar in terms of argument list , not the implementation, sorry for the confusion)

maddyscientist: Ah thanks for the clarification @alexstrel.

lib/staggered_quark_smearing.cu
maddyscientist: Actually thinking about this a bit more, I think we might want to fix this, so that it doesn't use the `xpay` template.  

(The `xpay` template is something that I think I might want to explore removing in the future.  For example, making it a run-time option when fast-compilation for the dslash is enabled).

For the present case here, maybe just make the `is_t0_kernel` parameter a run-time variable.  Looking at the code, it doesn't look like this really needs to be a compile-time parameter and run-time should be sufficient.  This will of course also reduce compilation time.

include/dirac_quda.h
alexstrel: Done.

include/dirac_quda.h
alexstrel: done

include/dirac_quda.h
alexstrel: done.

include/enum_quda.h
alexstrel: done

include/index_helper.cuh
alexstrel: fixed (also other places in the file)

include/kernels/staggered_quark_smearing.cuh
alexstrel: fixed

include/kernels/staggered_quark_smearing.cuh
alexstrel: fixed

include/kernels/staggered_quark_smearing.cuh
alexstrel: done (no difference in performance, though)

include/kernels/staggered_quark_smearing.cuh
alexstrel: fixed

include/kernels/staggered_two_link.cuh
alexstrel: sure, done.

include/ks_qsmear.h
alexstrel: removed `quda.h`

include/quda.h
alexstrel: done

include/quda.h
alexstrel: as per our previous discussions, all params are now grouped in a new structure

CMakeLists.txt
alexstrel: done

lib/dirac_staggered.cpp
alexstrel: done

include/quda_milc_interface.h
alexstrel: instead, deleted qudaFreeGaugeSmeared()

tests/utils/staggered_host_utils.cpp
alexstrel: done

tests/utils/host_utils.cpp
alexstrel: it returns computed deviation rather then an int flag.

tests/staggered_gsmear_test_utils.h
alexstrel: not sure since it is in fact a conditional instantiation (e.g., the object is not needed for pure two-link tests)

tests/staggered_gsmear_test_utils.h
alexstrel: done

lib/interface_quda.cpp
alexstrel: agreed, fixed

lib/interface_quda.cpp
alexstrel: fixed

lib/interface_quda.cpp
alexstrel: done

tests/staggered_gsmear_test_utils.h
alexstrel: sure, done.

lib/staggered_quark_smearing.cu
alexstrel: fixed

lib/staggered_two_link_quda.cu
alexstrel: done

tests/utils/command_line_params.cpp
```diff
@@ -272,6 +272,12 @@ double eofa_mq3 = 1.0;
 
 QudaContractType contract_type = QUDA_CONTRACT_TYPE_OPEN;
 
+int n_steps = 50;
```
maddyscientist: Can you put all these variables in a struct, so that it's clear what option group they belong to?

tests/utils/host_utils.cpp
maddyscientist: Ok, can you add a comment on this difference in the code then?

tests/staggered_gsmear_test_utils.h
maddyscientist: That doesn't matter, since `ColorSpinorField` has a default constructor, so it's easy enough to conditionally create the object.

include/kernels/staggered_quark_smearing.cuh
maddyscientist: This still isn't quite fixed (`active` is missing)

include/quda.h
maddyscientist: Missing `in/out` on `param` descriptor here

include/quda.h
maddyscientist: This looks much cleaner now 😄

include/dirac_quda.h
maddyscientist: Thanks, but there's still some missing `in`/`out` attributes and the ones for `in` and `out` should be `@param[out] out` as opposed to `@param out[out]`

maddyscientist:
Thanks @alexstrel for the updates to the PR.  I left some more minor comments.  Beyond these, the main outstanding issue is the lack of including the test in the `ctest` list.

include/dirac_quda.h
alexstrel: fixed

tests/utils/command_line_params.cpp
alexstrel: Thanks @maddyscientist  for the review. Not sure I understand the last point. Would you like me to create a common structure that will include all related variables? Currently I just left a comment above the group while all variables in the group remain plain build-in types. (Similar to other test vars, e.g. deflation-related ones)

maddyscientist: How about just prefixing these variables, e.g., `smear_n_steps`, `smear_t0`, etc. so it's clear which group they belong to?

alexstrel: sure, done.

weinbe2:
@alexstrel do you have a MILC workflow leveraging this that you can share? It can be a dummy one, no need to share anything special. As part of testing on my end it'd just be nice to see it "in action".

tests/host_reference/gsmear_test_helpers.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
```
mathiaswagner: Do we need that as a separate header?

lib/dirac_improved_staggered.cpp
```diff
@@ -127,6 +127,37 @@ namespace quda {
   {
     // do nothing
   }
+  
+  void DiracImprovedStaggered::SmearOp(ColorSpinorField &out, const ColorSpinorField &in, const double &, const double &,
```
mathiaswagner: is this an exact code duplication from `dirac_staggered.cpp` ? Just wondering if we can avoid that.

lib/CMakeLists.txt
```diff
@@ -287,6 +287,22 @@ if(QUDA_INTERFACE_TIFR OR QUDA_INTERFACE_ALL)
   target_compile_definitions(quda PUBLIC BUILD_TIFR_INTERFACE)
 endif(QUDA_INTERFACE_TIFR OR QUDA_INTERFACE_ALL)
 
+if(QUDA_CONTRACT)
```
mathiaswagner: looks like the target_compile_definitions should not be added back. IIRC @maddyscientist removed those in favor of a configured header.

include/dirac_quda.h
mathiaswagner: Why do we have a `b (ignored)` parameter?

mathiaswagner: Why do we pass doubles as const ref instead of by value?

include/dirac_quda.h
```diff
@@ -1361,6 +1371,18 @@ namespace quda {
...
+     * @brief Create two-link staggered quark smearing operator
+     *
+     * @param[out] out output smeared field
+     * @param[in] in input field 
+     * @param[in] a smearing parameter
+     * @param     b (ignored)
+     * @param[in] t0 time-slice index
+     * @param[in] parity Parity flag
+     */
+    void SmearOp(ColorSpinorField &out, const ColorSpinorField &in, const double &a, const double &b, const int &t0,  const QudaParity parity) const;
```
mathiaswagner: Why do we pass doubles as const ref instead of by value?

include/dirac_quda.h
mathiaswagner: Please also clarify this in the doxygen comment.

lib/interface_quda.cpp
```diff
@@ -3932,6 +3945,80 @@ void computeKSLinkQuda(void *fatlink, void *longlink, void *ulink, void *inlink,
...
+  computeTwoLink(*gaugeSmeared, *cudaInLinkEx);
+  gaugeSmeared->exchangeGhost();
+
+  profileGaussianSmear.TPSTOP(QUDA_PROFILE_COMPUTE);
+  //
+  gaugeSmeared->saveCPUField(cpuTwoLink, profileGaussianSmear);
+
+  profileGaussianSmear.TPSTART(QUDA_PROFILE_FREE);
+  
+  delete gaugeSmeared;gaugeSmeared = nullptr;
```
mathiaswagner: that should not be a single line but I think we need a clang-format on the diff anyway

lib/CMakeLists.txt
maddyscientist: Ah, good catch @mathiaswagner

hwancheolJeong:
> @alexstrel do you have a MILC workflow leveraging this that you can share? It can be a dummy one, no need to share anything special. As part of testing on my end it'd just be nice to see it "in action".

@weinbe2 Here is my test app: https://github.com/hwancheolJeong/milc_qcd/tree/feature/gaussianSmearing/ks_gauss_smear
It is my fork of MILC code. It uses an interface defined in generic_ks/gauss_smear_ks_QUDA.c that will be merged into develop branch later (once this pr is done.), but this test app is solely for personal use. 
You can build the test app by "make ks_gauss_smear", and run it with an input file (refer to test/ks_gauss_smear.2.sample-in). It will produce a smeared source in Scidac format as well as binary format (for easy comparison).

lib/dirac_improved_staggered.cpp
hwancheolJeong: They are identical. To avoid duplication, I think we should introduce an intermediate class between the Dirac class and them (DiracStaggered and DiracImprovedStaggered). Or, if we remove (or replace) checkSpinorAlias() defined in the Dirac class, it would also be possible to create a shared function outside the class (or in another mother class).

include/dirac_quda.h
hwancheolJeong: In fact, both `a` and `b` parameters do nothing in `Dirac(Improved)Staggered::SmearOP()`'s, since we pull them out as an overall factor. (For clarity, we omit their names in the definitions of these functions.)
However, we keep them in the input structure to accommodate the future addition of `GaugeLaplace::SmearOp()`.
 (See https://github.com/lattice/quda/blob/feature/quark-smearing/include/dirac_quda.h#L1959. I suppose Dean worked on it before us.)
If you want to remove them for now, that is okay with us.

hwancheolJeong: Regarding calling by reference, again, we follow the structure of `GaugeLaplace::SmearOp()` that started before this work. I don't think we have any preference for it.

include/dirac_quda.h
hwancheolJeong: Replied at https://github.com/lattice/quda/pull/1312#discussion_r1025695761

lib/CMakeLists.txt
hwancheolJeong: If I understand correctly, it's not just a problem for this pr to fix but a general issue. Right?

lib/interface_quda.cpp
hwancheolJeong: Thanks for finding it out. Split into two lines. https://github.com/lattice/quda/commit/8e7ad58492d3a863a6af6f868dd28d5354de57e8

include/dirac_quda.h
hwancheolJeong: Okay. Alexei and I discussed it, and we decided to change these `a`,`b`,`t0` parameters to be called by value and fixed doxygen description for `a`. Thank you for bringing it up.
https://github.com/lattice/quda/commit/8bdf739483e267317bc4e7af7117b202907ecd35
https://github.com/lattice/quda/commit/ff3af36ad06315d7f9dea719009a3078fcbe85ce

lib/CMakeLists.txt
maddyscientist: Resolved with [daa6244](https://github.com/lattice/quda/pull/1312/commits/daa6244e041e55034a1adbe8cc0a47bd86b790b6)

lib/dirac_improved_staggered.cpp
maddyscientist: I think we can leave this as it is for now and consider cleanup / consolidation after merging.

tests/utils/command_line_params.cpp
maddyscientist: [35d2002](https://github.com/lattice/quda/pull/1312/commits/35d2002fcc6ce30ed86e8c59e8cd517365815db1)

maddyscientist:
Thanks for the latest updates @alexstrel and @hwancheolJeong.  I've done some tests with the latest version, and things are mostly good.  A few outstanding issues before we can merge:
* the automatic ctest only runs in single precision, can you add a double precision version as well?  (see what is done with `gauge_path_test` for example for how to deal with this, where it is run in a loop over enabled precision)
* reconstruct options seem to be ignored by the test, and the kernel is always run with no reconstruct
* can you merge in #1345 into your PR?  This will be merged shortly into develop, but in the meantime it will fix the presently broken Jenkins build with CUDA 12 / GCC-12 / Clang 14

There's some other cleanup that I will apply, but this will be for after the branch is merged.

maddyscientist:
....and lastly, this PR needs a clang-format.

tests/staggered_gsmear_test_utils.h
alexstrel: done

tests/host_reference/gsmear_test_helpers.h
alexstrel: fixed, enumerator is moved into a destination header

include/quda.h
alexstrel: fixed

maddyscientist (approved):
Thank you @alexstrel and @hwancheolJeong for continuing to get this PR in good shape.  I'm finally approving 😉 

@weinbe2 can you take this PR for a spin before we merge?

weinbe2:
@alexstrel and @hwancheolJeong , this is all really nice! I'm doing some testing and a little bit of cleanup here and there, but this should be good to go soon.

I had one question --- I noticed that you created a routine in `milc_interface.cpp`, exposed in the appropriate header, to free the two link gauge field, but no routines to create the field or apply the smearing. Is this intentional?

include/kernels/staggered_quark_smearing.cuh
```diff
@@ -0,0 +1,222 @@
...
+      using Vector = ColorSpinor<real, Arg::nColor, 1>;
+
+      // is thread active (non-trival for fused kernel only)
+      bool active = mykernel_type == EXTERIOR_KERNEL_ALL ? false : true;
+
+      // which dimension is thread working on (fused kernel only)
+      int thread_dim;
+
+      if (arg.is_t0_kernel) {
+        if (arg.t0 < 0) return;
```
weinbe2, line 176: I'm worried that this `return;` could imply that `arg.out(...);` is never assigned, which may happen if this is a case where line 218:
```
if (kernel_type != EXTERIOR_KERNEL_ALL || active) arg.out(coord.x_cb, my_spinor_parity) = out;
```
never gets hit.

I think we should refactor the logic here in a way to at least guarantee that `arg.out` always has something sane assigned to it.

lib/staggered_quark_smearing.cu
```diff
@@ -0,0 +1,223 @@
...
+    using Dslash = Dslash<staggered_qsmear, Arg>;
+    using Dslash::arg;
+    using Dslash::in;
+
+  public:
+    StaggeredQSmear(Arg &arg, const ColorSpinorField &out, const ColorSpinorField &in) : Dslash(arg, out, in) { }
+
+    void apply(const qudaStream_t &stream)
+    {
+      TuneParam tp = tuneLaunch(*this, getTuning(), getVerbosity());
```
weinbe2, line 33: While this isn't going to break correctness, when `t0 == -1` on rank 0 the tuned values for the dslash part of this are going to end up very meaningless, since the kernel isn't doing any work...

tests/staggered_gsmear_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  initQuda(device_ordinal);
+
+  // Ensure gtest prints only from rank 0
+  ::testing::TestEventListeners &listeners = ::testing::UnitTest::GetInstance()->listeners();
+  if (comm_rank() != 0) { delete listeners.Release(listeners.default_result_printer()); }
+
+  display_test_info();
+
+  if (link_recon != QUDA_RECONSTRUCT_NO) {
```
weinbe2, line 74: Out of curiosity, what's the reason for enforcing this, for at least `computeTwoLink`? As far as I can tell, `computeTwoLink` supports reconstruct 12 and 8 (and you instantiate it).

I can see how it wouldn't work for `staggeredTwoLinkGaussianSmear`, though.

hwancheolJeong:
> Thank you @alexstrel and @hwancheolJeong for continuing to get this PR in good shape. I'm finally approving 😉
> 
> @weinbe2 can you take this PR for a spin before we merge?

@maddyscientist I appreciate you! for reviewing it and for your advice and help.

hwancheolJeong:
> @alexstrel and @hwancheolJeong , this is all really nice! I'm doing some testing and a little bit of cleanup here and there, but this should be good to go soon.
> 
> I had one question --- I noticed that you created a routine in `milc_interface.cpp`, exposed in the appropriate header, to free the two link gauge field, but no routines to create the field or apply the smearing. Is this intentional?

Thanks, @weinbe2 .
I made an interface inside the MILC code: https://github.com/hwancheolJeong/milc_qcd/blob/feature/gaussianSmearing/generic_ks/gauss_smear_ks_QUDA.c,
which will be merged into the develop branch of the MILC code once this PR is done (I may need to update a little to accommodate recent updates in QUDA).
We may add a similar interface routine in milc_interface.cpp later.

include/kernels/staggered_quark_smearing.cuh
hwancheolJeong: Regarding your first concern, I am afraid I don't know much about other backends. What can happen on them due to `return;`?

Regarding your second concern, if I understand correctly and my memory is correct, I recall other kernels are defined in a similar way. (e.g., https://github.com/lattice/quda/blob/develop/include/kernels/dslash_staggered.cuh#L206)
And arg.out is assigned here: https://github.com/lattice/quda/blob/feature/quark-smearing.pr/include/kernels/staggered_quark_smearing.cuh#L50.

hwancheolJeong: For the latter one, are you concerned when ColorSpinorField instance `out` is created with NULL_FIELD_CREATE option? It can happen in general.. though we are creating `out` with ZERO_FIELD_CREATE in the `performTwoLinkGaussianSmearNStep()`..

lib/staggered_quark_smearing.cu
hwancheolJeong: Thanks. I (or we) didn't think about it... Now, I am concerned.. What would happen when the second source's time slice is on a different rank from the first source's?

include/kernels/staggered_quark_smearing.cuh
hwancheolJeong: And I believe I checked all available dslash policies... (nothing in memory though..)

maddyscientist: @hwancheolJeong I think @weinbe2's concern with other targets is misplaced, and there's no issue there.  And I agree with you on the second point as well, there's no issue with `arg.out` not being hit here, and this is consistent with what happens in the other dslash kernels.  Marking as resolved.

lib/staggered_quark_smearing.cu
maddyscientist: I just had a chance to look into this, and I share @weinbe2's concern here.  I'm not sure what the solution is to this.  Perhaps we could have an interface where we can override which rank does the kernel tuning and broadcasts the tune cache from?  Then here we'd set this to being the first rank with the time slice in it.

lib/staggered_two_link_quda.cu
```diff
@@ -0,0 +1,67 @@
...
+      TwoLink_<decltype(arg)> twolnk(arg, link, newTwoLink);
+    }
+  };
+#if defined(GPU_STAGGERED_DIRAC) && defined(GPU_TWOLINK_GSMEAR)
+  void computeTwoLink(GaugeField &newTwoLink, const GaugeField &link)
+  {
+    checkNative(newTwoLink, link);
+    checkLocation(newTwoLink, link);
+    checkPrecision(newTwoLink, link);
+    instantiate<ComputeTwoLink, WilsonReconstruct>(link, newTwoLink);
```
maddyscientist: Should this be `WilsonReconstruct` here or is `StaggeredReconstruct` more appropriate?

weinbe2: I dug into this and realized `WilsonReconstruct` does make sense at face value, since the smearing does indeed seem to be done with the non-phased links. @hwancheolJeong I did see you have a comment in your test executable and your MILC code that `recon-12` isn't working, though. If that's the case, can you replace the `WilsonReconstruct` with `ReconstructNo` (so it'll be recon-18 only) and file an issue for it in this repo so it's revisited in the future?

weinbe2:
> > @alexstrel and @hwancheolJeong , this is all really nice! I'm doing some testing and a little bit of cleanup here and there, but this should be good to go soon.
> > I had one question --- I noticed that you created a routine in `milc_interface.cpp`, exposed in the appropriate header, to free the two link gauge field, but no routines to create the field or apply the smearing. Is this intentional?
> 
> Thanks, @weinbe2 . I made an interface inside the MILC code: https://github.com/hwancheolJeong/milc_qcd/blob/feature/gaussianSmearing/generic_ks/gauss_smear_ks_QUDA.c, which will be merged into the develop branch of the MILC code once this PR is done (I may need to update a little to accommodate recent updates in QUDA). We may add a similar interface routine in milc_interface.cpp later.

Understood, thank you for clarifying. That's not the typical approach but I'm certainly not going to hold up the PR on this! Similar to my other ask, can you please file an issue noting that this needs to be ported into the MILC Interface in the future?

lib/staggered_two_link_quda.cu
alexstrel: thanks @weinbe2 I agree it makes sense to restrict the reconstruction here as well , (with a fixme note)

tests/staggered_gsmear_test.cpp
alexstrel: Reconstruction options are left for the future optimization work.

weinbe2 (approved):
After some offline discussion, I'd say this is ready to go, and the outstanding questions (which are ultimately harmless for this PR's correctness/functionality) can be addressed in a future PR. Awesome work @alexstrel and @hwancheolJeong !

weinbe2:
I'll merge this after Jenkins completes

PR #1316, opened 2022-08-27
Hotfix/arpack

maddyscientist:
This PR is fix some outstanding issues with ARPACK integration with QUDA
* Fix handling of MPI communicators (@hummingtree)
* Fix building when `QUDA_DOWNLOAD_ARPACK=ON`

@mathiaswagner as we discussed, can you look into adding CPM support to better handle building ARPACK when `QUDA_DOWNLOAD_ARPACK=ON` is enabled?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -562,6 +562,7 @@ if(QUDA_ARPACK)
   if(QUDA_DOWNLOAD_ARPACK)
   #TODO: switch to CPM
     include(GNUInstallDirs)
+    include(ExternalProject)
     ExternalProject_Add(
       ARPACK-NG
       GIT_REPOSITORY https://github.com/opencollab/arpack-ng.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -619,6 +619,7 @@ if(QUDA_OPENBLAS)
   if(QUDA_DOWNLOAD_OPENBLAS)
   #TODO: switch to CPM
     include(GNUInstallDirs)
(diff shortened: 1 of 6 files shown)


mathiaswagner:
Yes, this is in progress.

mathiaswagner (approved):
Not sure whether we want to merge that now to have it fixed and make CPM a follow up.But changes look good to me.

CMakeLists.txt
```diff
@@ -562,6 +562,7 @@ if(QUDA_ARPACK)
   if(QUDA_DOWNLOAD_ARPACK)
   #TODO: switch to CPM
     include(GNUInstallDirs)
+    include(ExternalProject)
```
mathiaswagner: Do we also need to add that when downloading OpenBLAS?

mathiaswagner:
I added initial CPM support for ARPACK-NG. @maddyscientist  @cpviolator  please give it a spin. I did some build tests and things looked good. 

One thought: I left the non QUDA_DOWNLOAD_ARPACK path untouched so this could probably also benefit from `find_package(arpack-ng)` or something like that but IIRC the logging functionality might be why we want to keep arpack (without ng).

OpenBLAS would be next but their cmake support is not quite there yet.

maddyscientist:
Thanks @mathiaswagner.  I just tested this and it's working great on my workstation.  I'm happy leaving the manual download version as is to allow ARPACK to be classic to be used if desirable.  We can revisit that in the future.

PR #1266, opened 2022-04-20
feature/omptarget

jxy:
The OpenMP target backend here is still a work in progress.  We welcome any suggestions.

As of now this port uses a few Intel extensions, contains hacks specifically for Intel architectures, and it only works on Intel GPUs.

For a quick test, try
```
cmake\
        -DCMAKE_BUILD_TYPE=RELEASE\
        -DQUDA_TARGET_TYPE=OMPTARGET\
        -DQUDA_DOWNLOAD_USQCD=on\
        -DQUDA_QMP=on\
        -DQUDA_QIO=on\
        -DQUDA_DIRAC_DEFAULT_OFF=on\
        -DQUDA_DIRAC_STAGGERED=on\
        -DQUDA_PRECISION=8\
        -DQUDA_RECONSTRUCT=4\
        -DQUDA_FAST_COMPILE_REDUCE=on\
        -DQUDA_FAST_COMPILE_DSLASH=on\
        -DQUDA_BUILD_NATIVE_LAPACK=off\
        -DCMAKE_CXX_COMPILER=mpic++\
        -DCMAKE_C_COMPILER=mpicc\
        ../quda
```

PR #1317, opened 2022-09-01
add some SYCL build infrastructure

jcosborn:
include a few other minor fixes

### DIFF
--- a/include/kernels/unitarize_links.cuh
+++ b/include/kernels/unitarize_links.cuh
@@ -101,7 +101,7 @@ namespace quda {
     real s = c[1] * static_cast<real>(1.0 / 3.0) - c[0] * c[0] * static_cast<real>(1.0 / 18.0);
 
     real cosTheta;
-    if (fabs(s) >= arg.unitarize_eps) { // faster when this conditional is removed?
+    if (s >= arg.unitarize_eps) { // faster when this conditional is removed?
       const real rsqrt_s = quda::rsqrt(s);
       r = c[2]*0.5 - (c[0] * static_cast<real>(1.0 / 3.0)) * (c[1] - c[0] * c[0] * static_cast<real>(1.0 / 9.0));
       cosTheta = r*rsqrt_s*rsqrt_s*rsqrt_s;
--- a/include/targets/generic/mrg32k3a.h
+++ b/include/targets/generic/mrg32k3a.h
@@ -1,0 +1,191 @@
+/*
+   An implementation of MRG32k3a based on constexpr.
+   Original algorithm from
+      Pierre L'Ecuyer
+      Good Parameter Sets for Combined Multiple Recursive Random Number Generators
+      Operations Research, 47, 1 (1999), 159-164.
+ */
+
+#pragma once
+
+#include <cmath>
+#include <cstdint>
+#include <iostream>
+#include <limits>
+
+namespace quda
+{
+  namespace target
+  {
+    namespace rng
+    {
+      template <typename E, int N> struct Arr {
+        constexpr E &operator[](int i) { return d[i]; }
+        constexpr const E &operator[](int i) const { return d[i]; }
+        E d[N];
+      };
+
+      template <typename E> struct Arr3 {
+        constexpr Arr3() {};
+        constexpr Arr3(E x, E y, E z) : d {x, y, z} {};
+        constexpr E &operator[](int i) { return d[i]; }
+        constexpr const E &operator[](int i) const { return d[i]; }
+        constexpr bool operator==(const Arr3 &x) const { return d[0] == x.d[0] && d[1] == x.d[1] && d[2] == x.d[2]; }
+        E d[3];
+      };
+
+      template <typename E> inline std::ostream &operator<<(std::ostream &o, const Arr3<E> &a)
+      {
+        return o << '[' << a[0] << ' ' << a[1] << ' ' << a[2] << ']';
+      }
+
+      using State = Arr3<uint32_t>;
+      using Trans = Arr3<State>;
+
+      struct MRG32k3a {
+        State s1, s2;
+      };
+
+      inline std::ostream &operator<<(std::ostream &o, const MRG32k3a &prn)
+      {
+        return o << "MRG32k3a(" << prn.s1 << ' ' << prn.s2 << ')';
+      }
+
+      inline constexpr Trans squaremod(Trans a, uint64_t m)
+      {
+        Trans x;
+        for (int i = 0; i < 3; ++i) {
+          Arr3<uint64_t> t {0u, 0u, 0u};
+          for (int k = 0; k < 3; ++k) {
+            uint64_t aik = (uint64_t)a[i][k];
+            t[0] += aik * (uint64_t)a[k][0] % m;
+            t[1] += aik * (uint64_t)a[k][1] % m;
+            t[2] += aik * (uint64_t)a[k][2] % m;
+          }
+          x[i][0] = (uint32_t)(t[0] % m);
+          x[i][1] = (uint32_t)(t[1] % m);
+          x[i][2] = (uint32_t)(t[2] % m);
+        }
+        return x;
+      }
+
+      template <int p> inline constexpr Arr<Trans, p> squaremodarray(Trans a, uint64_t m)
+      {
+        Arr<Trans, p> x;
+        x[0] = a;
+        for (int i = 1; i < p; ++i) x[i] = squaremod(x[i - 1], m);
+        return x;
+      }
+
+      inline void matvecmod(Trans a, State &v, uint64_t m)
+      {
+        uint64_t v0 = (uint64_t)v[0];
+        uint64_t v1 = (uint64_t)v[1];
+        uint64_t v2 = (uint64_t)v[2];
+        v[0]
+          = (uint32_t)((((uint64_t)a[0][0] * v0) % m + ((uint64_t)a[0][1] * v1) % m + ((uint64_t)a[0][2] * v2) % m) % m);
+        v[1]
+          = (uint32_t)((((uint64_t)a[1][0] * v0) % m + ((uint64_t)a[1][1] * v1) % m + ((uint64_t)a[1][2] * v2) % m) % m);
+        v[2]
+          = (uint32_t)((((uint64_t)a[2][0] * v0) % m + ((uint64_t)a[2][1] * v1) % m + ((uint64_t)a[2][2] * v2) % m) % m);
+      }
+
+      constexpr double norm = 2.328306549295728e-10;
+      constexpr double m1 = 4294967087.0;
+      constexpr double m2 = 4294944443.0;
+      constexpr double a12 = 1403580.0;
+      constexpr double a13n = 810728.0;
+      constexpr double a21 = 527612.0;
+      constexpr double a23n = 1370589.0;
+      constexpr uint32_t defaultSEED = 12345U;
+      constexpr int subsequenceBase = 76;
+
+      constexpr Trans a1 {State {0u, 1u, 0u}, State {0u, 0u, 1u}, State {(uint32_t)(m1 - a13n), (uint32_t)a12, 0u}};
+      constexpr Trans a2 {State {0u, 1u, 0u}, State {0u, 0u, 1u}, State {(uint32_t)(m2 - a23n), 0u, (uint32_t)a21}};
+
+      constexpr int maxpower2 = 190;
+      constexpr Arr<Trans, maxpower2> a1sq = squaremodarray<maxpower2>(a1, (uint64_t)m1);
+      constexpr Arr<Trans, maxpower2> a2sq = squaremodarray<maxpower2>(a2, (uint64_t)m2);
+
+      static_assert(a1sq[76]
+                      == Trans {State {82758667u, 1871391091u, 4127413238u}, State {3672831523u, 69195019u, 1871391091u},
+                                State {3672091415u, 3528743235u, 69195019u}},
+                    "a1sq[76] wrong!");
+      static_assert(a2sq[76]
+                      == Trans {State {1511326704u, 3759209742u, 1610795712u},
+                                State {4292754251u, 1511326704u, 3889917532u},
+                                State {3859662829u, 4292754251u, 3708466080u}},
+                    "a2sq[76] wrong!");
+
+      inline void skip(MRG32k3a &prn, uint64_t offset, int base = 0)
+      {
+        int i = 0;
+        uint64_t s = offset;
+        while (s > 0) {
+          if (s & 1u) {
+            matvecmod(a1sq[base + i], prn.s1, (uint64_t)m1);
+            matvecmod(a2sq[base + i], prn.s2, (uint64_t)m2);
+          }
+          s >>= 1u;
+          ++i;
+        }
+      }
+
+      inline void seed(MRG32k3a &prn, uint64_t seed, uint64_t subsequence)
+      {
+        if (seed) {
+          const uint64_t d1 = (uint64_t)defaultSEED * (uint64_t)((uint32_t)seed ^ 0x55555555u);
+          const uint64_t d2 = (uint64_t)defaultSEED * (uint64_t)((uint32_t)(seed >> 32u) ^ 0xAAAAAAAAu);
+          prn.s1[0] = (uint32_t)(d1 % (uint64_t)m1);
+          prn.s1[1] = (uint32_t)(d2 % (uint64_t)m1);
+          prn.s1[2] = (uint32_t)(d1 % (uint64_t)m1);
+          prn.s2[0] = (uint32_t)(d2 % (uint64_t)m2);
+          prn.s2[1] = (uint32_t)(d1 % (uint64_t)m2);
+          prn.s2[2] = (uint32_t)(d2 % (uint64_t)m2);
+        } else {
+          prn.s1[0] = defaultSEED;
+          prn.s1[1] = defaultSEED;
+          prn.s1[2] = defaultSEED;
+          prn.s2[0] = defaultSEED;
+          prn.s2[1] = defaultSEED;
+          prn.s2[2] = defaultSEED;
+        }
+        skip(prn, subsequence, subsequenceBase);
+      }
+
+      inline double uniform(MRG32k3a &prn)
+      {
+        double p1, p2;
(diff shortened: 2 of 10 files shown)


include/targets/generic/mrg32k3a.h
```diff
@@ -0,0 +1,191 @@
+/*
```
maddyscientist, line 1: @jcosborn, I'm curious: have you tried using this with the CUDA target?  I imagine you'd just need to delete the `include/cuda/random_helper.h` file.

include/kernels/unitarize_links.cuh
```diff
@@ -101,7 +101,7 @@ namespace quda {
     real s = c[1] * static_cast<real>(1.0 / 3.0) - c[0] * c[0] * static_cast<real>(1.0 / 18.0);
 
     real cosTheta;
-    if (fabs(s) >= arg.unitarize_eps) { // faster when this conditional is removed?
+    if (s >= arg.unitarize_eps) { // faster when this conditional is removed?
```
maddyscientist, line 104: why this change?

maddyscientist (commented):
Apart from a couple of questions I have, this looks good to me.

@mathiaswagner, FWIW, the point of this PR is to get the QUDA cmake build system ready for the SYCL PR.

maddyscientist (approved):

include/kernels/unitarize_links.cuh
jcosborn: fabs is unnecessary.  If s<0 then that code (rsqrt) won't make sense either.  I don't know if that would ever happen, but there is no reason to keep fabs here.

include/targets/generic/mrg32k3a.h
jcosborn: I checked that it compiles and passes tests.  I didn't check if it produces the same numbers, but it should.

tests/CMakeLists.txt
```diff
@@ -19,11 +19,15 @@ endif()
 
 set(TEST_LIBS quda_test)
 
+# custom target for building all tests (useful if QUDA_BUILD_ALL_TESTS is OFF)
+add_custom_target(buildtests)
```
mathiaswagner: I like the funcitonality but wondering if `build tests` is a good name. How about something like `all_tests` or `quda_tests`. 
Buildtests to me sound like this targets builds tests that test whether the build works.

lib/CMakeLists.txt
```diff
@@ -203,9 +206,9 @@ include(CheckLinkerFlag)
 check_linker_flag(CXX "-Wl,--compress-debug-sections=zlib" QUDA_LINKER_COMPRESS)
 target_link_options(quda PRIVATE $<$<CONFIG:DEVEL>:$<${QUDA_LINKER_COMPRESS}:-Wl,--compress-debug-sections=zlib>>)
 
-# set up QUDA compile options
+# set up QUDA compile options, put them before target specific options
 target_compile_options(
-  quda PRIVATE $<$<COMPILE_LANGUAGE:CXX,C,Fortran>: 
+  quda BEFORE PRIVATE $<$<COMPILE_LANGUAGE:CXX,C,Fortran>:
```
mathiaswagner, line 211: just curious: Is this style or are there technical reasons with sycl?

maddyscientist: It was a technical reason.  Some warnings need to be disabled on SYCL at present, and so any target-specific warning disablement has to be done after the general warning are enabled.

maddyscientist:
@jcosborn a couple more things to address:
* does [include/targets/generic/mrg32k3a.h](https://github.com/lattice/quda/pull/1317/files/49bdef08c332cb17c80bb8c5e308b666d88f4474#diff-7f2eae9d8bc377076b1efbcb6d708c5d8304fa7dd68784aafc25c187ef2a8c93) come with any license considerations?
* can you replace the C-style casts with `static_cast`?

tests/CMakeLists.txt
jcosborn: quda_test (without 's') is already an internal target.  all_tests sounds good though.

jcosborn: 086686eaa45f53ca785f69cfc8f6a441516af9b0

mathiaswagner: Looks good. Thanks!

jcosborn:
I believe [include/targets/generic/mrg32k3a.h](https://github.com/lattice/quda/pull/1317/files/49bdef08c332cb17c80bb8c5e308b666d88f4474#diff-7f2eae9d8bc377076b1efbcb6d708c5d8304fa7dd68784aafc25c187ef2a8c93) is @jxy's own code (from the OMP branch [include/targets/omptarget/mrg32k3a.h](https://github.com/lattice/quda/pull/1266/files#diff-2e5c53a1e4888e2c0eee06ce710be311f7e61d85b17e6ee68786608ea6a3495d)).
Is that correct @jxy?

include/targets/generic/mrg32k3a.h
jcosborn: Actually, I didn't test that correctly (I just moved the source file, without reconfiguring).  With a clean build it fails with nvcc (note #2703-D: cannot call non-constexpr function "quda::target::rng::Arr<E, N>::Arr() [with E=quda::target::rng::Trans, N=190]).  Explicitly adding a constexpr constructor (constexpr Arr() {};) doesn't help.  The original code works fine with dpcpp (which is clang-based), so I'm not sure what the problem is.

jxy: I have tested the code with gcc and clang.  It should work with `-std=c++17`, though I can't recall what versions of gcc or clang I tried.  Maybe there's some issue with nvcc's support of constexpr functions?

maddyscientist: I have got this working with `nvcc`.  Partly because `nvcc` is more exacting about `constexpr` conformance, and partly because of an undefined use case.  I'll push some fixes.

maddyscientist:
Ok, I've pushed the changes to enable the generic RNG to work with nvcc:
* For default constructors that are `constexpr` all members must be initialized
* One cannot (reliably) declare `constexpr` arrays and access said array from the device.
  * All arrays must work when using a base address plus offset for the cases where the compiler cannot perform compile-time evaluation.
  * When the compiler fails to do the consteval, it will naively dereference the array address which is only valid on the host which obviously leads to badness.
  * The solution is to maintain a separate device copy of the array, with the explicit decoration `__device__`
* Any functions that are to be compiled on the device must be `constexpr` else they require explicit `__device__` decoration

@jcosborn @jxy can you verify these changes are ok from your side?

jcosborn:
The RNG updates work fine for me.

maddyscientist:
@mathiaswagner I think all items have been addressed now.  Ok to merge?

mathiaswagner:
I have not seen a confirmation from @jxy on the license considerations. Apart from that good to merge.

jxy:
The RNG code is my clean room implementation.  I first implemented it in Nim, https://github.com/jcosborn/qex/blob/devel/src/rng/mrg32k3a.nim, and later translated it to C++.  QUDA's MIT license applies.

I don't fully understand the reason of the changes though.

The implicit default constructor should take care of the array type.  Is it actually a bug in nvcc to require explicit initialization of the member array?

`squaremod` and `squaremodarray` are used called both at compile time and at runtime.  If we remove `inline`, wouldn't the binary contains multiple copies of the function body binary and cause duplicated symbols during linking?  Maybe device functions are different, but host functions would be?

The functions that changed from `inline` to `constexpr` are only used with runtime variables.  Maybe we really want them to be `__host__ __device__ inline`?  It's odd that `constexpr` effectively works like `__device__` though.  And the same question as above, wouldn't the final executable contain multiple copies of the function body binary if they are used on the host?

jxy:
I found half of my answers here, https://en.cppreference.com/w/cpp/language/constexpr

> A constexpr specifier used in an object declaration or non-static member function (until C++14) implies const. A constexpr specifier used in a function or [static](https://en.cppreference.com/w/cpp/language/static) data member (since C++17) declaration implies inline. If any declaration of a function or function template has a constexpr specifier, then every declaration must contain that specifier.

so we don't need `inline` if it's `constexpr`.  I guess it somehow implies `__device__` for nvcc?

I think the changes are fine, and don't really change the compiletime and runtime behaviors.

mathiaswagner:
Thanks for the confirmation on the licensing questions @jxy . As MRG32k3a is something quite common just wanted to be on the safe side here. We often add 3rd party code to QUDA but make sure we only do this if the license allows it and add the required information to QUDA's LICENSE file.
@maddyscientist  should comment on the code questions.

Good to merge for me.

jxy:
I am not sure if it matters, but the `seed` function follows what @jcosborn did in the sycl branch, https://github.com/lattice/quda/blob/a82a09351d23a54d55678d17c797ccca61df7c4f/include/targets/sycl/random_helper_mkl.h#L23-L38, which I believe is specifically made to be compatible with cublas.

maddyscientist:
> The implicit default constructor should take care of the array type. Is it actually a bug in nvcc to require explicit initialization of the member array?

I should have said above, all *explicit* default constructors require the member initialization.  It's not a bug in `nvc++`, the issue also arises with g++ and clang++, and is part of the standard. 

The issue wasn't `Arr`, which has an implicit default constructor, rather `Arr3`, which has a user-defined default constructor and requires explicit initialization of the member array.
```c++
struct foo {
  constexpr foo() { }
  int x[4];
};

int main()
{
  constexpr foo bar = foo();
}
```
Compiling with `g++ -std=c++17` gives
```bash
$ g++-11 test.cpp 
test.cpp: In constructor ‘constexpr foo::foo()’:
test.cpp:2:21: error: member ‘foo::x’ must be initialized by mem-initializer in ‘constexpr’ constructor
    2 |   constexpr foo() { }
      |                     ^
test.cpp:3:7: note: declared here
    3 |   int x[4];
      |       ^
test.cpp: In function ‘int main()’:
test.cpp:8:27: error: ‘constexpr foo::foo()’ called in a constant expression
    8 |   constexpr foo bar = foo();
      |                           ^
test.cpp:2:13: note: ‘constexpr foo::foo()’ declared here
    2 |   constexpr foo() { }
```
and clang++ gives
```bash
$ clang++-15 test.cpp  -std=c++17
test.cpp:2:13: warning: constexpr constructor that does not initialize all members is a C++20 extension [-Wc++20-extensions]
  constexpr foo() { }
            ^
test.cpp:3:7: note: member not initialized by constructor
  int x[4];
      ^
test.cpp:8:17: error: constexpr variable 'bar' must be initialized by a constant expression
  constexpr foo bar = foo();
                ^     ~~~~~
test.cpp:8:17: note: subobject of type 'int' is not initialized
test.cpp:3:7: note: subobject declared here
  int x[4];
      ^
1 warning and 1 error generated.
```
With the issue being resolved with adding the explicit initialization to the member array.

I just added the initialization to `Arr` to keep it visually consistent with `Arr3`, it isn't needed for compilation to proceed.

maddyscientist:
> so we don't need inline if it's constexpr. I guess it somehow implies __device__ for nvcc?

Yes, `inline` is implied by `constexpr`.  Moreover, if a function is a valid `constexpr` function, then nvcc can compile it for the device without needing the `__device__` decoration.

I changed some function to be `constexpr` that don't need to be made, e.g., the ones in `random_helper.cuh`.  This is just following the "_constexpr all things_" mantra, but also means that the code remains pure C++ without the CUDA decorations .

jxy:
Thanks for the detailed explanation.  I tried to reduce the mrg32k3a code, and it seems the code without explicit initialization works fine with clang-14 and fails with gcc-12.2.  Here is the code I'm testing.
```
#include<cmath>
#include<cstdint>

template <typename E,int N>
struct Arr {
  constexpr E& operator[](int i) {return d[i];}
  constexpr const E& operator[](int i) const {return d[i];}
  E d[N];
};

template <typename E>
struct Arr3 {
  constexpr Arr3() {};
  constexpr Arr3(E x,E y,E z):d{x,y,z} {};
  constexpr E& operator[](int i) {return d[i];}
  constexpr const E& operator[](int i) const {return d[i];}
  constexpr bool operator==(const Arr3& x) const {return d[0]==x.d[0] && d[1]==x.d[1] && d[2]==x.d[2];}
  E d[3];
};

using State = Arr3<uint32_t>;
using Trans = Arr3<State>;

struct MRG32k3a {
  State s1,s2;
};

constexpr Trans squaremod(Trans a, uint64_t m)
{
  Trans x;
  for(int i=0;i<3;++i){
    Arr3<uint64_t> t{0u,0u,0u};
    for(int k=0;k<3;++k){
      uint64_t aik = (uint64_t)a[i][k];
      t[0] += aik*(uint64_t)a[k][0] % m;
      t[1] += aik*(uint64_t)a[k][1] % m;
      t[2] += aik*(uint64_t)a[k][2] % m;
    }
    x[i][0] = (uint32_t)(t[0] % m);
    x[i][1] = (uint32_t)(t[1] % m);
    x[i][2] = (uint32_t)(t[2] % m);
  }
  return x;
}

template <int p>
constexpr Arr<Trans,p> squaremodarray(Trans a, uint64_t m)
{
  Arr<Trans,p> x;
  x[0] = a;
  for(int i=1;i<p;++i)
    x[i] = squaremod(x[i-1], m);
  return x;
}

inline void matvecmod(Trans a, State& v, uint64_t m)
{
  uint64_t v0 = (uint64_t)v[0];
  uint64_t v1 = (uint64_t)v[1];
  uint64_t v2 = (uint64_t)v[2];
  v[0] = (uint32_t)((((uint64_t)a[0][0]*v0) % m + ((uint64_t)a[0][1]*v1) % m + ((uint64_t)a[0][2]*v2) % m) % m);
  v[1] = (uint32_t)((((uint64_t)a[1][0]*v0) % m + ((uint64_t)a[1][1]*v1) % m + ((uint64_t)a[1][2]*v2) % m) % m);
  v[2] = (uint32_t)((((uint64_t)a[2][0]*v0) % m + ((uint64_t)a[2][1]*v1) % m + ((uint64_t)a[2][2]*v2) % m) % m);
}

constexpr double m1 = 4294967087.0;
constexpr double m2 = 4294944443.0;
constexpr double a12 = 1403580.0;
constexpr double a13n = 810728.0;
constexpr double a21 = 527612.0;
constexpr double a23n = 1370589.0;

constexpr Trans a1 {
  State {0u,                  1u,            0u},
  State {0u,                  0u,            1u},
  State {(uint32_t)(m1-a13n), (uint32_t)a12, 0u}};
constexpr Trans a2 {
  State {0u,                  1u, 0u},
  State {0u,                  0u, 1u},
  State {(uint32_t)(m2-a23n), 0u, (uint32_t)a21}};

int main()
{
  constexpr int maxpower2 = 190;
  constexpr Arr<Trans,maxpower2> a1sq = squaremodarray<maxpower2>(a1, (uint64_t)m1);
  constexpr Arr<Trans,maxpower2> a2sq = squaremodarray<maxpower2>(a2, (uint64_t)m2);

  static_assert(
    a1sq[76] ==
      Trans {
        State {82758667u,   1871391091u, 4127413238u},
        State {3672831523u, 69195019u,   1871391091u},
        State {3672091415u, 3528743235u, 69195019u}},
    "a1sq[76] wrong!");
  static_assert(
    a2sq[76] ==
      Trans {
        State {1511326704u, 3759209742u, 1610795712u},
        State {4292754251u, 1511326704u, 3889917532u},
        State {3859662829u, 4292754251u, 3708466080u}},
    "a2sq[76] wrong!");
}
```

```
% clang++-mp-14 -std=c++17 cexpr_test.cc        
% g++-mp-12 -std=c++17 cexpr_test.cc        
cexpr_test.cc: In function 'int main()':
cexpr_test.cc:85:66:   in 'constexpr' expansion of 'squaremodarray<190>(a1, (uint64_t)(double)m1)'
cexpr_test.cc:49:16:   in 'constexpr' expansion of 'x.Arr<Arr3<Arr3<unsigned int> >, 190>::Arr()'
cexpr_test.cc:85:83:   in 'constexpr' expansion of 'Arr3<Arr3<unsigned int> >()'
cexpr_test.cc:85:83: error: 'constexpr Arr3<E>::Arr3() [with E = unsigned int]' called in a constant expression
   85 |   constexpr Arr<Trans,maxpower2> a1sq = squaremodarray<maxpower2>(a1, (uint64_t)m1);
      |                                                                                   ^
cexpr_test.cc:13:13: note: 'constexpr Arr3<E>::Arr3() [with E = unsigned int]' is not usable as a 'constexpr' function because:
   13 |   constexpr Arr3() {};
      |             ^~~~
cexpr_test.cc:13:13: error: member 'Arr3<unsigned int>::d' must be initialized by mem-initializer in 'constexpr' constructor
cexpr_test.cc:18:5: note: declared here
   18 |   E d[3];
      |     ^
cexpr_test.cc:86:66:   in 'constexpr' expansion of 'squaremodarray<190>(a2, (uint64_t)(double)m2)'
cexpr_test.cc:49:16:   in 'constexpr' expansion of 'x.Arr<Arr3<Arr3<unsigned int> >, 190>::Arr()'
cexpr_test.cc:86:83:   in 'constexpr' expansion of 'Arr3<Arr3<unsigned int> >()'
cexpr_test.cc:86:83: error: 'constexpr Arr3<E>::Arr3() [with E = unsigned int]' called in a constant expression
   86 |   constexpr Arr<Trans,maxpower2> a2sq = squaremodarray<maxpower2>(a2, (uint64_t)m2);
      |                                                                                   ^
cexpr_test.cc:89:14: error: non-constant condition for static assertion
   89 |     a1sq[76] ==
cexpr_test.cc:96:14: error: non-constant condition for static assertion
   96 |     a2sq[76] ==
```
I think it's a clang bug at least in version 14.

maddyscientist:
Anyway, it seems like we're all on the same now, so I'll go ahead and merge this PR now.

jxy:
I solved my own mystery.  I have been testing with `c++20` only.  With `-std=c++20` gcc happily compiles the code without a hiccup.  It only gives error with `-std=c++17`.

maddyscientist:
Ah, that makes sense.  Glad the mystery is solved 😄

jxy:
I guess it's something to do with this proposal, https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1331r2.pdf

> so long as uninitialized values are not read from, such states should be permitted in constexpr in both heap and stack allocated scenarios.

Issue #1319, opened 2022-09-14
The normal distribution is not standard in gaugeGauss.


SaltyChiang:
In https://github.com/lattice/quda/blob/develop/include/kernels/gauge_random.cuh#L45, QUDA attempts to generate some normally distributed random variables by applying the Box-Muller transform.

Comparing the implementation with the one on Wikipedia, it seems that these variables are distributed with $\mathcal{N}(0,\frac{1}{\sqrt{2}})$ instead of the standard normal distribution $\mathcal{N}(0,1)$, because the line `radius[i] = sqrt(-log(rand2[i]))` lacks a `2` before `log`.

Is it by design or a bug? The meaning of `sigma` parameter for `gaugeGaussQuda` (which is called `sigma` in the docstring but `epsilon` as the actual arg name) should be modified if it is by design.

jxy:
NO.  I believe that is by design such that the su(3) matrix has an expectation value of norm squared equal to 8.

maddyscientist:
Thanks @jxy.  What are your thoughts @SaltyChiang?  Do you need the factor of two to match some other code?  If so, I wonder if we should make this a parameter to support different conventions.

SaltyChiang:
@jxy Thank you for your information. The problem is that there exists a parameter to adjust the expectation value, instead of just returning the result.

In spinor_noise.cuh#L29, we have no chance to change the sigma of the distribution, but for su3 field, we do have the parameter. If this is by design, I think some comments should be there to point out that the sigma is not the one in the normal distribution.

jxy:
This is su(3).  We are not talking about Normal distribution in R.  We are talking about Normal distribution in su(3), which has dimension 8.  So 8 is the correct normalization.  The comment about sigma being the width is not wrong.

SaltyChiang:
I see. That sounds reasonable and I will revert the commit. Thank you for the comment.

Also, this issue should be closed as there is no problem in the current code.

PR #1321, opened 2022-09-15
Add gaussMomQuda to quda.h.

SaltyChiang:
Recently I'm trying to implement a full GPU HMC routine with QUDA. I can only access functions in `quda.h` as I'm not using C++ to write the code. I noticed `gaussGauge` can be applied to the momentum field but there is no interface in `quda.h`.

This PR added a function to call `gaussGauge` on the resident momentum field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,7 @@ Advanced Scientific Computing (PASC21) [arXiv:2104.05615[hep-lat]].
 *  Steven Gottlieb (Indiana University) 
 *  Kyriakos Hadjiyiannakou (Cyprus)
 *  Dean Howarth (Lawrence Livermore Lab, Lawrence Berkeley Lab)
+*  Xiangyu Jiang (Chinese Academy of Sciences)
 *  Balint Joo (OLCF, Oak Ridge National Laboratory, formerly Jefferson Lab)
 *  Hyung-Jin Kim (Samsung Advanced Institute of Technology)
 *  Bartosz Kostrzewa (HPC/A-Lab, University of Bonn)
--- a/include/kernels/gauge_random.cuh
+++ b/include/kernels/gauge_random.cuh
@@ -46,7 +46,7 @@ namespace quda {
       rand1[i] = uniform<real>::rand(localState);
       rand2[i] = uniform<real>::rand(localState);
       phi[i] = 2.0 * rand1[i];
-      radius[i] = sqrt(-2.0 * log(rand2[i]));
+      radius[i] = sqrt(-log(rand2[i]));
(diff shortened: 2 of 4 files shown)


mathiaswagner:

maddyscientist:
Thanks for this contribution @SaltyChiang.  This looks good to me.

Just wondering if you might want to add a fix to the issue #1319 to this PR, since I imagine the fix will be trivial?

One other thing: please feel free to add your name to the contributor list in the README.md so you are counted as an official QUDA author 😄

SaltyChiang:
@maddyscientist Yes fixing #1319 is trivial, but I'm still not sure if it's truly a bug, and maybe someone has already used this as a feature. For example, the branch experimental/hmc uses `gaussGauge(sigma=1.0)` to generate the momentum field, while the coefficient is often $\frac{1}{\sqrt{2}}$ in other libraries like chroma.

SaltyChiang:
Should close #1319. Keeping the same behavior is important.

maddyscientist:
Thanks for the fix in [00e4b30](https://github.com/lattice/quda/pull/1321/commits/00e4b309093d691013ad02059e9429c439c0bf22).  I think this is fine to change, because I don't think anyone is relying on repro  The experimental hmc branch you note is experimental, so it's fine to make fixes at the moment.

jxy:
NO. It is a typical convention that the su(3) matrix has an expectation of its norm squared being 8.  You can see the same thing the spinor noise, https://github.com/lattice/quda/blob/605e92c83a32b68cd8f889b69aee8c66172b5010/include/kernels/spinor_noise.cuh#L29,  where the convention is the norm squared of the complex number being 1.

SaltyChiang:
The modification on gauss_su3 is reverted. 

@jxy convinced me the current code is correct, and see #1319 for details.

maddyscientist:
@SaltyChiang do you want to add you name to the author list before we merge?

maddyscientist (approved):

cpviolator:
@SaltyChiang Once these commits are merged I was thinking of adding the plumbing to QUDA to enable user choices for normalisation. They won't affect your commits as we can default to the desired behaviour and then allow for other standard choices, or even custom `sigma` values.

SaltyChiang:
> @SaltyChiang do you want to add you name to the author list before we merge?

So glad to receive the invitation!

I also removed some unnecessary variables in gaussGaugeQuda().

maddyscientist (approved):

maddyscientist:
Thanks @SaltyChiang 😄

PR #1322, opened 2022-09-16
Fix crash while using computeGaugeLoopTraceQuda.

SaltyChiang:
QUDA will crash when calling `computeGaugeLoopTraceQuda` if `extendedGaugeResident` is not set. For example, this will happen if you call `computeGaugeLoopTraceQuda` just after `loadGaugeQuda`.

This PR fixed the issue.


mathiaswagner:

maddyscientist (approved):

Issue #1323, opened 2022-09-19
MG Setup (or refresh) fails with "too many heavy quark residual restarts"


kostrzewa:
This is a very strange issue that has come up quite recently and which might require another bisection, unless there's a smoking gun that I'm missing or it's simply an instance of me doing something stupid.

Somewhat randomly, QUDA, as used in tmLQCD's HMC, is having trouble either generating the MG setup or refreshing it. I say randomly because it sometimes works fine, suggesting that likely unitialised memory is used or memory holding a param struct is overwritten. It's also strange that the same exact version of both tmLQCD and QUDA sometimes works / has worked for prolonged periods of time. Perhaps there is some bad interaction with the driver.

I have been able to reproduce the issue on two machines: Juwels Booster and a small GPU Cluster at Bonn University sporting A40s.

* **Juwels Booster** seems to run: Driver Version: 515.65.01    CUDA Version: 11.7 (from nvidia-smi)
* QUDA and tmLQCD are compiled with:
  -  GCC/11.2.0
  - OpenMPI/4.1.2
  - CUDA/11.5

* **The A40 cluster** runs: Driver Version: 515.48.07    CUDA Version: 11.7 (from nvidia-smi)
* QUDA and tmLQCD are compiled with:
  -  GCC/11.2.0
  - OpenMPI/4.1.2-GCC-11.2.0
  - UCX-CUDA/1.11.2-GCCcore-11.2.0-CUDA-11.4.1
  - CUDA/11.4.1

### Juwels Booster

I've observed the following behaviour with 227ff8c8b (commit from the end of March) on Juwels Booster. I have also reproduced the same behaviour with 64998f1a (mid-September) so if it's a QUDA problem, it's one which has been present for quite a while.

1) The MG setup at the beginning of the trajectory is generated correctly

```
# TM_QUDA: Performing MG Preconditioner Setup for gauge_id: 0.000000
MG level 0 (GPU): Using randStateMRG32k3a
MG level 0 (GPU): Creating a CG solver
MG level 0 (GPU): Running vectors setup on level 0 iter 1 of 1
MG level 0 (GPU): Tuned block=(800,1,1), grid=(172,1,1), shared_bytes=0, aux=(-1,-1,-1,-1) giving 514.12 Gflop/s, 1028.23 GB/s for N4quda4b
las5Norm2IdfEE with GPU-offline,nParity=2,vol=442368,precision=4,order=4,Ns=4,Nc=3,TwistFlavour=1
[...]

PR #1324, opened 2022-09-19
Hotfix for weird heavy quark residual behavior during MG setup

weinbe2:
This narrow PR is a hotfix for the issue reported in #1323. I had hit this issue previously and fixed it in #1283 (which has not yet been merged), commit https://github.com/lattice/quda/commit/f954cec3b97ba65add2736301eae3b5d02b575ae . I assumed, at the time, that the need for that fix was related to my other changes in MG setup and not something already lurking in `develop` (or maybe something introduced since then), which is why I didn't make a hotfix branch in the first place.

That commit is much bigger than just the narrow heavy quark residual fix so it can't just be "cherry-pick"ed out.

While this issue appearing is indeed weird, I think the logic of the fix is reasonable, so I don't see a need to investigate it more deeply at this time.

### DIFF
--- a/lib/inv_cg_quda.cpp
+++ b/lib/inv_cg_quda.cpp
@@ -567,7 +567,7 @@ namespace quda {
         }
 
         // if L2 broke down already we turn off reliable updates and restart the CG
-        if (ru.reliable_heavy_quark_break(L2breakdown, heavy_quark_res, heavy_quark_res_old, heavy_quark_restart)) {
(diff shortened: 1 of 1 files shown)


maddyscientist (approved):
Do you know when this issue was introduced?  For posterity, it would be good to document when.

Regardless: approved

mathiaswagner:
Looks likely to be https://github.com/lattice/quda/commit/3ab9b4e46306f267d808eb74120da8fc0de5ed51

mathiaswagner (approved):

PR #1327, opened 2022-09-28
correction of small typo in comment

simone-romiti:


mathiaswagner:

maddyscientist (approved):
@simone-romiti in general, best to put trivial fixes like this as part of larger PRs.  If nothing else it helps saves cycles on our CI testing.  Regardless, thanks for the fix.

kostrzewa:
@maddyscientist Thanks for explaining that. We weren't sure if you preferred to have stuff like this in isolated PRs or, as you now said, as part of a larger one. Maybe a mechanism for a submitter to mark a PR to bypass the CI pipeline until manually triggered by someone in quda_core would be a way to have both isolated PRs for small stuff and avoid needlessly using CI cycles?

Anyway, we'll package further things of this sort as part of larger PRs.

PR #1328, opened 2022-10-06
The NVHPC CUDA toolkit layout moves math library headers and librarie…

james-simone:
…s to a sibling directory.

FindCUDAToolkit provides the variable CUDAToolkit_MATH_INCLUDE_DIR to selectively add the math lib includes.


maddyscientist (approved):
Thanks for the contribution @james-simone.  Just noting that this change will only have an effect of CMake 3.23 and above, since `CUDAToolkit_MATH_INCLUDE_DIR` is only defined from that version upwards.

PR #1330, opened 2022-10-12
computeCloverForceQuda

Marcogarofalo:
I am trying to use the function 
computeCloverForceQuda. 
The first error I encountered was 
```
ERROR: Unexpected pc_type -2147483648 (rank 0, host lnode14.cluster.hiskp, color_spinor_field.cpp:141 in void quda::ColorSpinorField::create(const quda::ColorSpinorParam&)())
       last kernel called was (name=cudaMemset,volume=bytes=147456,aux=cudaGaugeField,cuda_gauge_field.cpp,54)
```
that I try to solve with
https://github.com/lattice/quda/blob/434cacafda68cbe3ef3643698b460bb8f160fcb2/lib/interface_quda.cpp#L4765-L4766

but then I get an error at 
https://github.com/lattice/quda/blob/434cacafda68cbe3ef3643698b460bb8f160fcb2/lib/interface_quda.cpp#L4832-L4836

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -4762,6 +4762,9 @@ void computeCloverForceQuda(void *h_mom, double dt, void **h_x, void **, double 
   qParam.siteSubset = QUDA_FULL_SITE_SUBSET;
   qParam.siteOrder = QUDA_EVEN_ODD_SITE_ORDER;
   qParam.nDim = 4;
+  qParam.pc_type = QUDA_4D_PC;
+  if (inv_param->dslash_type == QUDA_DOMAIN_WALL_DSLASH) { qParam.pc_type = QUDA_5D_PC; }
+
   qParam.setPrecision(fParam.Precision());
   qParam.pad = 0;
   for(int dir=0; dir<4; ++dir) qParam.x[dir] = fParam.x[dir];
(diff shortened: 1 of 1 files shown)


mathiaswagner:
@jenkins test this please

maddyscientist:
Hi @Marcogarofalo.  I suspect this function is suffering from code rot given the lack of a regression test for it.  I'll take a look at this when I get a chance.  Have you checked if the `x` field has been allocated already?

Marcogarofalo:
I think that x is allocated, the problem is that `cpuQuarkX` has different volume, that's the error I get:
```
So far so good: volume x : 256  , volume cpuQuarkX: 128
ERROR: volumes do not match: 256 128 (rank 0, host lnode14.cluster.hiskp, color_spinor_field.cpp:728 in static void quda::ColorSpinorField::checkField(const quda::ColorSpinorField&, const quda::ColorSpinorField&)())
```
@simone-romiti noticed that https://github.com/lattice/quda/blob/d6d7da1f860ea38602864899157ab10349a14523/lib/interface_quda.cpp#L4821
could be the reason.

SaltyChiang:
> @simone-romiti noticed that 
> 
> https://github.com/lattice/quda/blob/d6d7da1f860ea38602864899157ab10349a14523/lib/interface_quda.cpp#L4821
> 
> 
> 
> 
> could be the reason.

I think this line is definitely an issue. You can expect that `x[0]` will be divided by 2 again and again.
Just deleting this line might fix the problem, as `x[0]` has already been divided in line #L4781. Not tested yet.

maddyscientist:
@Marcogarofalo I assume we can close this PR in favor of #1338?

Marcogarofalo:
Yes please.

PR #1331, opened 2022-10-24
Removed '#include <complex.h>'

bjoo:
To avoid a #define conflict between <complex.h> and gtest.h when using hipcc compiler. This fixes a local issue Venkitesh Ayyar raised on Crysger at OLCF which prevented `gauge_path_test.cpp` from compiling with the following #define clash:

```
shared/ayyar/builds_crusher/install_oct17_2022/QUDA/src/quda/tests/googletest/include/gtest/internal/gtest-internal.h:1154:21: error: expected ')'
template <size_t... I, size_t sizeofT>
                    ^
/usr/include/complex.h:53:11: note: expanded from macro 'I'
#define I _Complex_I
          ^
/usr/include/complex.h:48:21: note: expanded from macro '_Complex_I'
#define _Complex_I      (__extension__ 1.0iF)
                         ^
/gpfs/alpine/lgt104/proj-shared/ayyar/builds_crusher/install_oct17_2022/QUDA/src/quda/tests/googletest/include/gtest/internal/gtest-internal.h:1154:21: note: to match this '('
/usr/include/complex.h:53:11: note: expanded from macro 'I'
#define I _Complex_I
          ^
/usr/include/complex.h:48:20: note: expanded from macro '_Complex_I'
#define _Complex_I      (__extension__ 1.0iF)
```

Also I changed from `#include <complex.h>` to the more C++
`#include <complex>`

### DIFF
--- a/tests/gauge_path_test.cpp
+++ b/tests/gauge_path_test.cpp
@@ -11,7 +11,6 @@
 #include <gauge_path_quda.h>
 #include <timer.h>
 #include <gtest/gtest.h>
-#include <complex>
 
 static QudaGaugeFieldOrder gauge_order = QUDA_QDP_GAUGE_ORDER;
 
--- a/tests/gauge_path_test.cpp
+++ b/tests/gauge_path_test.cpp
@@ -1,7 +1,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <complex.h>
 
 #include <quda.h>
 #include <host_utils.h>
@@ -12,6 +11,7 @@
 #include <gauge_path_quda.h>
 #include <timer.h>
 #include <gtest/gtest.h>
+#include <complex>
 
 static QudaGaugeFieldOrder gauge_order = QUDA_QDP_GAUGE_ORDER;
 


maddyscientist:
@bjoo thanks for the fix.  

Note it was intentional to include the C complex header not the C++ one, since the QUDA interface is C based.  This test might not even need this header at all though?

mathiaswagner:
Reminds me that we should clean up headers using IWYU at some point.

bjoo:
Turns out it wasn't needed. I commented it and the code compiled fine. So I just ended up removing it. Makes for a cleaner PR

maddyscientist (approved):
Note this will also fix gcc 7 compilation with QUDA that I recently realized is broken (not on our testing matrix)

maddyscientist:
Just noting that this issue was a gtest issue which has since been [fixed](https://github.com/google/googletest/pull/2293]).  We should update the version of gtest we package with QUDA, or perhaps use CPM if it's available?

kostrzewa:
@bjoo Thanks for this. Also hit it recently but didn't get around to figuring out where it was coming from.

PR #1333, opened 2022-11-08
Feature/multi rhs kernel

maddyscientist:
This PR is centered around the addition of multi-RHS capability to the coarse dslash kernel:
* Cleanup of the fine-grain accessors to reduce their memory footprint, and separate the ghost access into a parent class.
* Add multi-RHS capability to the coarse dslash.  Up to 64 RHS are supported in a single kernel (limited by kernel arg size at 4 KiB).
* Expose as cmake options, the field ordering used for fine grid fixed point fields (`QUDA_ORDER_FP`), coarse grid fields in single precision (`QUDA_ORDER_SP_MG`) and coarse grid fields in fixed point (`QUDA_ORDER_FP_MG`).  This will be explored in future optimizations.
* Allow for batched ghost packing with generic ghost packer using a higher dimension color-spinor field to act as a halo container.
* Extend the `FieldTmp` interface to allow for custom keys and utilize this to provide cached halo container fields
* Extend `getFieldTmp` to allow for the creation of vectors of temporaries.  For example, deployed in multi-RHS `Dirac::MdagM` for the temporary. 
* `multigrid_benchmark_test` now extended to do self-contingency tests of the multi-RHS coarse dslash kernel for a correctness test.  We now have to specify the number of colours used, e.g. with `--mg-nvec 0 24` for 24 colours, and the number of RHS is given by `--nsrc N` for `N` RHS
* Improvements to eigensolver parameter checking.
* Improve robustness of `blas::copy` when field pointers alias.
* Hotfix for Kahler-Dirac preconditioning (thanks to @weinbe2).

While multi-RHS deployment is still highly limited, I have confirmed that the multi-RHS coarse dslash kernel works very well to accelerate the block TRLM when using it to find the eigenvalues / eigenvectors of the coarse grid operator to perform coarse grid deflation.

Review suggestions:
* @bjoo can you report on performance scaling on red-team architectures?
* @weinbe2 can you test the block TRLM solver with HISQ MG?
* @kostrzewa can you test the block TRLM solver with TMC MG?


CMakeLists.txt
```diff
@@ -194,7 +194,9 @@ option(QUDA_BUILD_ALL_TESTS "build tests by default" ON)
 option(QUDA_INSTALL_ALL_TESTS "install tests by default" ON)
 option(QUDA_BUILD_SHAREDLIB "build quda as a shared lib" ON)
 
-option(QUDA_FLOAT8 "enable float-8 ordered fixed-point fields?" ON)
+set(QUDA_ORDER_FP "8" CACHE STRING "which data order to use for Wilson, gauge (recon-8/9) and clover fixed-point fields (4, 8)")
+set(QUDA_ORDER_SP_MG "2" CACHE STRING "which data order to use for fp32 MG fields (2, 4)")
+set(QUDA_ORDER_FP_MG "2" CACHE STRING "which data order to use for fixed-point MG fields (2, 4, 8)")
```
mathiaswagner, line 199: should we use set_propery(CACHE ...) here to make the possible values show up in ccmake ?

tests/multigrid_benchmark_test.cpp
```diff
@@ -44,43 +43,31 @@ void initFields(QudaPrecision prec)
...
-  param.location = QUDA_CPU_FIELD_LOCATION;
-
-  xH = std::make_unique<ColorSpinorField>(param);
-  yH = std::make_unique<ColorSpinorField>(param);
-
-  // Now set the parameters for the cuda fields
-  // param.pad = xdim*ydim*zdim/2;
-
-  if (param.nSpin == 4) param.gammaBasis = QUDA_UKQCD_GAMMA_BASIS;
+  param.gammaBasis = param.nSpin == 4 ? QUDA_UKQCD_GAMMA_BASIS : QUDA_DEGRAND_ROSSI_GAMMA_BASIS;
```
mathiaswagner: looks like param.Nspin is hardcoded above now so why do we need that?

tests/multigrid_benchmark_test.cpp
```diff
@@ -109,6 +96,19 @@ void initFields(QudaPrecision prec)
   X_h = new cpuGaugeField(gParam);
   Xinv_h = new cpuGaugeField(gParam);
 
+  // insert random noise into the gauge fields
+  for (auto d = 0; d < 8; d++) {
+    for (auto i = 0u; i < Y_h->Length()/8; i++) {
+      ((float**)Y_h->Gauge_p())[d][i] = double(rand()) / RAND_MAX;
```
mathiaswagner: float** and double(rand()) gives me some headaches here.

weinbe2:
If possible, it would be nice if the new auto-generated files could be inserted earlier in the order of compilation, as opposed to at the end, so the compile kicks off earlier and can get hidden behind the other files.

mathiaswagner:
@Jenkins test this please

weinbe2 (approved):
After dragging this PR through the mud, I can say everything is looking squeaky clean! Pending `clang-format` and sanity checking merging in `develop` this is good to go. Awesome work @maddyscientist !

include/convert.h
```diff
@@ -182,4 +182,16 @@ namespace quda
...
+  {
+#if 1
+    if constexpr (sizeof(fixed_t) < 4) {
+      return static_cast<fixed_t>(target::dispatch<f2i>(f));
+    } else {
+      return static_cast<fixed_t>(rint(f));
+    }
+#else
+    return static_cast<fixed_t>(rint(f));
+#endif
```
hummingtree: Should we actually just call `rint`/`rintf` regardless of `sizeof(fixed_t)`, since in `f2i` we use `rintf` anyways?

include/targets/cuda/mdw_dslash5_tensor_core.cuh
```diff
@@ -367,7 +366,8 @@ namespace quda
     a = __hmul2(sm_b[(threadIdx.y * 4 + 3) * N_sm_d2 + 3 * threadIdx.x + 1], max_i_div_max2_);
     b = __hmul2(sm_b[(threadIdx.y * 4 + 3) * N_sm_d2 + 3 * threadIdx.x + 2], max_i_div_max2_);
     out[sid + 5 * output.volumeCB] = __2half22integer4_rn<storage_vec>(a, b);
-
+#else
+    static_assert(QUDA_ORDER_FP == 8 || QUDA_ORDER_FP == 4, "Invalid QUDA_ORDER_FP");
```
hummingtree: I guess we do not need this check here, since it has already been checked generally in the other file?

include/convert.h
maddyscientist: I intentionally left it as it is, just in case we want to use the alternative path for some reason.

include/targets/cuda/mdw_dslash5_tensor_core.cuh
maddyscientist: Agreed.  Changed here: 5357496edb2acea31f5d50614e6b3036d04299e7

include/convert.h
hummingtree: Looks like Jenkins is not happy about this when both `fixed_t` and `float_t` are `double`. Not sure how that happens though.

hummingtree (approved):
I have given this PR a quick walk through, and have done some tests by extending the changes here to use MMA to perform the coarse dslash with multi-rhs, and things look all good. The only remaining issue from me is that the max deviation test in `tests/multigrid_benchmark_test.cpp` gives `nan` when the max deviation is exactly zero.

bjoo (approved):
I checked it built fine on HiP wirh ROCm/5.1.0. if correctness issues arise I will submit an issue to deal with it

PR #1334, opened 2022-11-09
Update to trove to fix UB caused by type punning. 

maddyscientist:
This fixes trove when using device debugging, allowing us to remove the prior WAR of disabling trove when debugging

### DIFF
--- a/include/targets/cuda/aos.h
+++ b/include/targets/cuda/aos.h
@@ -19,20 +19,18 @@ namespace quda
   };
 
   /**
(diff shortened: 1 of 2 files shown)


mathiaswagner (approved):

maddyscientist:
Thanks for the quick review @mathiaswagner.

For sake of historicity: this change corresponds to this [PR](https://github.com/bryancatanzaro/trove/pull/15)  in trove.

PR #1337, opened 2022-11-18
Hotfix to fix compilation on Summit

weinbe2:
This narrow hotfix PR addresses an issue with compilation that only seemed to creep up on Summit, where the cmake internal variable `CXX_OPT` caused compiler errors (seemingly unrelated to the specialization of `CXX_OPT` for Power9). The issue appeared with https://github.com/lattice/quda/commit/738e4be0f0ce56a7ee33c6cf8af2d70f8a53fa3b (which needs the subsequent commit, https://github.com/lattice/quda/commit/63182d6bfe58ed1a96399f86ebdce13bbd176792 , to compile).

I confirmed that this fixes the compile-time issue on Summit, and "does no harm" on `x86` as verified by looking at the outputs from a `--verbose` compile of QUDA through `nvcc`.

I also fixed a few harmless typos in `CMakeLists.txt`.

### DIFF
--- a/lib/targets/cuda/target_cuda.cmake
+++ b/lib/targets/cuda/target_cuda.cmake
@@ -47,17 +47,17 @@ set(CMAKE_CUDA_FLAGS_STRICT
     "-O3"
     CACHE STRING "Flags used by the CUDA compiler during strict jenkins builds.")
 set(CMAKE_CUDA_FLAGS_RELEASE
-    "-O3 ${CXX_OPT}"
+    "-O3 -Xcompiler \"${CXX_OPT}\""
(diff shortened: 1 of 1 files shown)


weinbe2:
~Do not merge yet...~ This is good to go, the issue I hit was because I merged this into a different branch.

weinbe2:
I confirmed that all other build types (`DEVEL`, `DEBUG`, `HOSTDEBUG`, `STRICT`, `SANITIZE`) compile without an issue. As a note for future reference, all of these tests were with CUDA 11.0, which is the current default on Summit.

maddyscientist (approved):

PR #1338, opened 2022-11-18
minimal trial implementation for a twisted clover determinant derivative (pre-draft)

kostrzewa:
This is a temporary copy of `computeCloverForceQuda` with (almost) the minimum amount of functionality to implement the derivative of a twisted clover determinant monomial as a starting point for debugging


kostrzewa:
@Marcogarofalo @simone-romiti fyi

kostrzewa:
This is similar to @Marcogarofalo's #1330 but I want to focus on getting this to work with tmLQCD first and then generalize the original function from there. As a result, the present PR will be temporary and only to share progress.

kostrzewa:
The problem https://github.com/lattice/quda/pull/1330#issuecomment-1285797127 is clearly fixed by removing the extra `qParam.x[0] /= 2;`.

However, the next problem crops up when the Dslash is supposed to be applied:

```
# QUDA: Dslash(x.Odd(), x.Even(), QUDA_ODD_PARITY)
# QUDA: ERROR: Parity spinor volume 16384 doesn't match clover checkboard volume 16384 (rank 0, host node-03.bender, dirac_twisted_clover.cpp:43 in virtual void quda::DiracTwistedClover::checkParitySpinor(const quda::ColorSpinorField&, const quda::ColorSpinorField&) const())
# QUDA:        last kernel called was (name=N4quda15CopyColorSpinorILi4ELi3ENS_11colorspinor11FloatNOrderIdLi4ELi3ELi2ELb0ELb0EEENS1_21SpaceSpinorColorOrderIdLi4ELi3EEESt5tupleIJRNS_16ColorSpinorFieldERKS7_19QudaFieldLocation_sPdPKdEEEE,volume=8x16x16x8,aux=GPU-offline,vol=16384,parity=1,precision=8,order=9,Ns=4,Nc=3vol=16384,parity=1,precision=8,order=2,Ns=4,Nc=3,NonRelBasis)
```

and this I find quite strange since `16384 == 16384` :smile:

kostrzewa:
The issue above occurs because of 

```
  void DiracTwistedClover::checkParitySpinor(const ColorSpinorField &out, const ColorSpinorField &in) const
  {
    Dirac::checkParitySpinor(out, in);

    if (out.TwistFlavor() == QUDA_TWIST_SINGLET) {
      if (out.Volume() != clover->VolumeCB())
        errorQuda("Parity spinor volume %lu doesn't match clover checkboard volume %lu", out.Volume(),
                  clover->VolumeCB());
    } else {
      //
      if (out.Volume() / 2 != clover->VolumeCB())
        errorQuda("Parity spinor volume %lu doesn't match clover checkboard volume %lu", out.Volume(),                                     
                  clover->VolumeCB());
    }
  }
```

and `qParam.twist_flavor` not being set when `quarkX` and `quarkP` are created.

I can update the error message above to indicate that this might be a possibility, but to be honest, maybe we generally should check more frequently if `twistFlavor` is set and compatible?

kostrzewa:
The next small step. Here I do not really understand where the problem comes from.

```
# QUDA: computeCloverForce(cudaForce, gaugePrecise, quarkX, quarkP, force_coeff);
# QUDA: ERROR: qudaEventSynchronize_ returned CUDA_ERROR_ILLEGAL_ADDRESS
 (timer.h:107 in peek())
 (rank 0, host node-02.bender, quda_api.cpp:72 in void quda::target::cuda::set_driver_error(CUresult, const char*, const char*, const char*, const char*, bool)())
# QUDA:        last kernel called was (name=N4quda11CloverForceIdLi3EL21QudaReconstructType_s18EEE,volume=16x16x16x8,aux=GPU-offline,vol=32768stride=16384precision=8geometry=4Nc=3,exterior,dir=1)
--------------------------------------------------------------------------
```

SaltyChiang:
> The next small step. Here I do not really understand where the problem comes from.
> 
> ```
> # QUDA: computeCloverForce(cudaForce, gaugePrecise, quarkX, quarkP, force_coeff);
> # QUDA: ERROR: qudaEventSynchronize_ returned CUDA_ERROR_ILLEGAL_ADDRESS
>  (timer.h:107 in peek())
>  (rank 0, host node-02.bender, quda_api.cpp:72 in void quda::target::cuda::set_driver_error(CUresult, const char*, const char*, const char*, const char*, bool)())
> # QUDA:        last kernel called was (name=N4quda11CloverForceIdLi3EL21QudaReconstructType_s18EEE,volume=16x16x16x8,aux=GPU-offline,vol=32768stride=16384precision=8geometry=4Nc=3,exterior,dir=1)
> --------------------------------------------------------------------------
> ```

I got a similar error while trying to use `computeCloverForceQuda`. I checked the [doc of CUDA](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT_1g949aa42b30ae9e622f6ba0787129ff22) and find that `cudaEventSynchronize` will not return `cudaErrorIllegalAddress`, which is pretty strange.

When I try to disable autotuning via setting `QUDA_ENABLE_TUNING=0`, the same error code (`cudaErrorIllegalAddress`) happens at another place.

kostrzewa:
I suspect that the problem is in the spinor comms of `computeCloverForce` and the error message is erroneous. I just realised I've been running with a RELEASE build of QUDA and will investigate with a DEBUG build first of all :)

kostrzewa:
```
Thread 1 "hmc_tm" received signal CUDA_EXCEPTION_14, Warp Illegal Address.
[Switching focus to CUDA kernel 0, grid 4320, block (0,0,0), thread (0,0,0), device 0, sm 0, warp 0, lane 0]
quda::Kernel1D<quda::Exterior, quda::CloverForceArg<double, 3, (QudaReconstructType_s)18, 3>, false><<<(64,1,1),(32,1,1)>>> ()
    at /home/bkostrze/code/quda-tm_force/lib/../include/color_spinor.h:1017 in _ZN4quda18outerProdSpinTraceIdLi3ELi4EEENS_6MatrixINS_7complexIT_EEXT0_EEERKNS_11ColorSpinorIS3_XT0_EXT1_EEES9_ inlined from color_spinor.h:115
```

it seems that the error does not quite occur where it is reported above, as far as I can tell from a run with `cuda-gdb`.

kostrzewa:
Thanks for the tip about disabling tuning @SaltyChiang, this makes it quite obvious where the issue is:

```
# QUDA: ERROR: qudaDeviceSynchronize_ returned CUDA_ERROR_ILLEGAL_ADDRESS
 (clover_outer_product.cu:100 in exchangeGhost())
 (rank 2, host node-02.bender, quda_api.cpp:72 in void quda::target::cuda::set_driver_error(CUresult, const char*, const char*, const char*, const char*, bool)())
# QUDA:        last kernel called was (name=N4quda11CloverForceIdLi3EL21QudaReconstructType_s18EEE,volume=16x16x16x8,aux=GPU-offline,vol=32768stride=16384precision=8geometry=4Nc=3,exterior,dir=1)
```

kostrzewa:
I think I can confirm that the error in https://github.com/lattice/quda/pull/1338#issuecomment-1321203371 stems from `exchangeGhost` of `computeCloverForce` since it runs without erroring out with a single GPU (this is output from tmLQCD's https://github.com/etmc/tmLQCD/tree/quda_work_tm_force branch running a simple HMC with a twisted clover clover determinant):

```
# QUDA: Dslash(x.Odd(), x.Even(), QUDA_ODD_PARITY)
# QUDA: gamma5(tmp, x.Even())
# QUDA: M(p.Even(), tmp)
# QUDA: Dslash(p.Odd(), p.Even(), QUDA_ODD_PARITY)
# QUDA: computeCloverForce(cudaForce, gaugePrecise, quarkX, quarkP, force_coeff);
# QUDA: computeCloverSigmaTrace(oprod, *cloverPrecise, k_csw_ov_8);
# QUDA: computeCloverSigmaOprod(oprod, quarkX, quarkP, ferm_epsilon)
# QUDA: cloverDerivative(cudaForce, gaugeEx, *oprodEx, 1.0, QUDA_ODD_PARITY)
# QUDA: cloverDerivative(cudaForce, gaugeEx, *oprodEx, 1.0, QUDA_EVEN_PARITY)
# QUDA: updateMomentum(gpuMom, -1.0, cudaForce, "tmclover")
# TM_QUDA: Time for computeTMCloverForceQuda 3.765968e-02 s level: 3 proc_id: 0 /HMC/cloverdetlight:cloverdet_derivative/compute_cloverdet_derivative_quda/computeTMCloverForceQuda
```

maddyscientist:
I will get more involved with this debug in the next day or so.  One thought though: when debugging it’s best to run with `CUDA_LAUNCH_BLOCKING=1`, this will ensure that all kernels are synchronized with the host and if an illegal memory access, etc., occurs you will be guaranteed to told exactly which kernel is misbehaving by QUDA’s error checking.

SaltyChiang:
> I think I can confirm that the error in [#1338 (comment)](https://github.com/lattice/quda/pull/1338#issuecomment-1321203371) stems from `exchangeGhost` of `computeCloverForce` since it runs without erroring out with a single GPU (this is output from tmLQCD's https://github.com/etmc/tmLQCD/tree/quda_work_tm_force branch running a simple HMC with a twisted clover clover determinant):

Pretty strange. I can also see some `cudaErrorIllegalAddress` even though I'm using a single-GPU build.

kostrzewa:
> One thought though: when debugging it’s best to run with CUDA_LAUNCH_BLOCKING=1, this will ensure that all kernels are synchronized with the host and if an illegal memory access, etc., occurs you will be guaranteed to told exactly which kernel is misbehaving by QUDA’s error checking.

Thanks a lot, I keep forgetting about this...

On 2 GPUs:

```
# QUDA: ERROR: qudaLaunchKernel returned an illegal memory access was encountered
 (/home/bkostrze/code/quda-tm_force/lib/targets/cuda/quda_api.cpp:152 in qudaLaunchKernel())
 (rank 1, host node-04.bender, quda_api.cpp:58 in void quda::target::cuda::set_runtime_error(cudaError_t, const char*, const char*, const char*, const char*, bool)())
# QUDA:        last kernel called was (name=N4quda11CloverForceIdLi3EL21QudaReconstructType_s18EEE,volume=16x16x16x16,aux=GPU-offline,vol=65536stride=32768precision=8geometry=4Nc=3,exterior,dir=1)
```

maddyscientist:
> The issue above occurs because of
> 
> ```
>   void DiracTwistedClover::checkParitySpinor(const ColorSpinorField &out, const ColorSpinorField &in) const
>   {
>     Dirac::checkParitySpinor(out, in);
> 
>     if (out.TwistFlavor() == QUDA_TWIST_SINGLET) {
>       if (out.Volume() != clover->VolumeCB())
>         errorQuda("Parity spinor volume %lu doesn't match clover checkboard volume %lu", out.Volume(),
>                   clover->VolumeCB());
>     } else {
>       //
>       if (out.Volume() / 2 != clover->VolumeCB())
>         errorQuda("Parity spinor volume %lu doesn't match clover checkboard volume %lu", out.Volume(),                                     
>                   clover->VolumeCB());
>     }
>   }
> ```
> 
> and `qParam.twist_flavor` not being set when `quarkX` and `quarkP` are created.
> 
> I can update the error message above to indicate that this might be a possibility, but to be honest, maybe we generally should check more frequently if `twistFlavor` is set and compatible?

To be honest I never really liked this design where a singlet fermion field is distinct from a non-twisted field.  The dimensions are identical, and the field can be used as intended, so I don't see the need for the separate field type.  I'm not sure what the original design motivation was for this (perhaps @alexstrel knows, as I think he was the originator, but this may be lost in the mists of time?)

A better approach here may be to dump the distinction between `QUDA_TWIST_NO` and `QUDA_TWIST_SINGLET` and allow these fields to be used interchangeably.   This would remove some of this pointless checking, and then we only need to worry if we have a doublet field or not.  Thoughts?

maddyscientist:
> > One thought though: when debugging it’s best to run with CUDA_LAUNCH_BLOCKING=1, this will ensure that all kernels are synchronized with the host and if an illegal memory access, etc., occurs you will be guaranteed to told exactly which kernel is misbehaving by QUDA’s error checking.
> 
> Thanks a lot, I keep forgetting about this...
> 
> On 2 GPUs:
> 
> ```
> # QUDA: ERROR: qudaLaunchKernel returned an illegal memory access was encountered
>  (/home/bkostrze/code/quda-tm_force/lib/targets/cuda/quda_api.cpp:152 in qudaLaunchKernel())
>  (rank 1, host node-04.bender, quda_api.cpp:58 in void quda::target::cuda::set_runtime_error(cudaError_t, const char*, const char*, const char*, const char*, bool)())
> # QUDA:        last kernel called was (name=N4quda11CloverForceIdLi3EL21QudaReconstructType_s18EEE,volume=16x16x16x16,aux=GPU-offline,vol=65536stride=32768precision=8geometry=4Nc=3,exterior,dir=1)
> ```

Ok, so this suggests the issue isn't the `exchangeGhost` and is the `CloverForce` kernel.  Moreover, we can see from the signature that it's the halo kernel (that's what the `exterior` part means).  So that would explain why it's a multi-GPU issue. 
 Maybe I missed it, but what happens what happens when you run on a single GPU?

kostrzewa:
>  Maybe I missed it, but what happens what happens when you run on a single GPU?

It works fine (I haven't checked yet whether it's correct, but at least it doesn't crash). I will be able to work some more on this next week hopefully and make some more concrete progress.

SaltyChiang:
I think it is the problem in `exchangeGhost`.

Please see [this commit](https://github.com/lattice/quda/pull/1339/commits/8e1504286db9e9aa48bebcb858176bf0dcf66ed6) in #1339.

https://github.com/lattice/quda/blob/8e1504286db9e9aa48bebcb858176bf0dcf66ed6/lib/clover_outer_product.cu#L145-L158

After using `exchangeGhost` in `ColorSpinorField` instead of one in `clover_outer_product.cu`, this function doesn't raise `cudaErrorIllegalAddress` for 2 GPUs.

`exchangeGhost` in `clover_outer_product.cu` seems to make the `inB.Ghost()` invalid for device function, still don't know why. At least we can go to the next step.

kostrzewa:
> A better approach here may be to dump the distinction between QUDA_TWIST_NO and QUDA_TWIST_SINGLET and allow these fields to be used interchangeably. This would remove some of this pointless checking, and then we only need to worry if we have a doublet field or not. Thoughts?

I agree. The name is also a bit fundamentally confusing in some sense, since I would consider the "flavour" to correspond to either +mu or -mu or equivalently +r or -r in the "physical basis", rather than the distinction between single-flavour and doublet. That is a whole other story, however, and the present solution works well in practice.

On the other hand, having a flag indicating that one is dealing with a twisted mass operator (rather than having to check whether mu is non-zero against best practice, even though also this is done here and there) is not a bad thing. This fact is useful in a couple of places:

1) clover field and its inverse
2) implicit use in `transfer->Vectors().TwistFlavor()` (I think)

Before getting rid of it, perhaps it's worth checking if this would not introduce lots of local and equivalent variables (which will likely be `bool twisted = param.mu != 0.0` (or a safer variant thereof). One might conclude, of course, that having these defined locally is better for one reason or another.

maddyscientist:
All good points @kostrzewa regarding `TwistFlavour()`.  I will have a think about this and some hacking and see what solution makes sense.

I've taken a look at the clover_outer_product.cu file a bit more to see if I could visually isolate what's going on here.  One thing I do see is that the halo exchange code is rather nasty, and is badly need of an update to clean this up.  It wouldn't surprise me if this code has bit rotted, and that in turn could lead to the crashes you were seeing, e.g., if the ghost buffers weren't being sized correctly, or some such.@SaltyChiang tried replacing with `ColorSpinorField::exchangeGhost` in #1339 but that isn't quite the correct thing to do, at least in isolation, since kernel changes would be needed as well.

May I suggest the following (which I think is what you were trending towards anyway):
* we focus on getting single GPU to work correctly
* once we have that, it will make debugging the multi-GPU variant much easier and more self contained
* moreover, in order to remove the dynamic clover aspect as a source of error, we focus on non-dynamic clover

What's the best way for me to run this function myself?  I assume you're calling this from tmLQCD?

kostrzewa:
> May I suggest the following (which I think is what you were trending towards anyway):
>
> 
> *  we focus on getting single GPU to work correctly
> *  once we have that, it will make debugging the multi-GPU variant much easier and more self contained
> *  moreover, in order to remove the dynamic clover aspect as a source of error, we focus on non-dynamic clover

I agree, this is exactly what I was trying to do to get properly familiar with this part of QUDA and to iron out crashes first so that actual step by step testing can be done.

> What's the best way for me to run this function myself? I assume you're calling this from tmLQCD?

If you'd like to test right away, I can send instructions on how to compile tmLQCD+QUDA and to provide an input file which will result in the function being called in the derivative of a twisted clover determinant monomial.

In the meantime perhaps we can get started on a test harness in QUDA in the hope that we can get this under test coverage eventually.

kostrzewa:
Disregard what was written in https://github.com/lattice/quda/pull/1338#issuecomment-1324701045 before about `inv_param` not being correct. We just checked with @Marcogarofalo and in fact all should be fine.

maddyscientist:
Getting started on a test harness in QUDA does sound like a good immediate step.  That would then allow, for example, me to focus on crashes with multi-GPU.

kostrzewa:
@Marcogarofalo as discussed I merged in @SaltyChiang's additions from #1339 such that we can proceed step by step here.

maddyscientist:
> computeTMCloverForceQuda now gives the same result of tmLQCD

that sounds like a good thing 😄 🍾

Marcogarofalo:
Thanks, but it is still working only with one mpi process.

include/kernels/clover_trace.cuh
```diff
@@ -157,7 +186,7 @@ namespace quda {
     __device__ __host__ inline void operator()(int x_cb)
     {
       // odd parity
-      cloverSigmaTraceCompute<Arg>(arg, x_cb, 1);
+      cloverSigmaTraceCompute<Arg>(arg, x_cb, 0);
```
Marcogarofalo: should I try to make this a parameter that one can select from interface_quda.cpp ?

maddyscientist: > should I try to make this a parameter that one can select from interface_quda.cpp ?

Sounds good to me

maddyscientist:
> Thanks, but it is still working only with one mpi process.

In that case, wondering if now is a good time to work on an internal QUDA test for this function?  With that done, I could then focus on making the multi-GPU results consistent with the single-GPU.  @kostrzewa you noted before you could add a numerical derivative for this.

kostrzewa:
> In that case, wondering if now is a good time to work on an internal QUDA test for this function? With that done, I could then focus on making the multi-GPU results consistent with the single-GPU. @kostrzewa you noted before you could add a numerical derivative for this.

@maddyscientist I would very much like to proceed on this but until the beginning of March I have a number of important deadlines that require full dedication unfortunately.

lib/clover_outer_product.cu
```diff
@@ -142,18 +142,21 @@ namespace quda {
...
+      // p[i]->Odd().allocateGhostBuffer(1);
 
       for (int parity=0; parity<2; parity++) {
 	ColorSpinorField& inA = (parity&1) ? p[i]->Odd() : p[i]->Even();
 	ColorSpinorField& inB = (parity&1) ? x[i]->Even(): x[i]->Odd();
 	ColorSpinorField& inC = (parity&1) ? x[i]->Odd() : x[i]->Even();
 	ColorSpinorField& inD = (parity&1) ? p[i]->Even(): p[i]->Odd();
 
+        static constexpr int nFace = 1;
+        inB.exchangeGhost((QudaParity)(1-parity), nFace, dag);
```
Marcogarofalo: By chance, I found that `exchangeGhost` method avoids the `CUDA_ERROR_ILLEGAL_ADDRESS` I was encountering before.
However, I suppose there is now a double exchange of the ghost.

maddyscientist: With this use of `exchangeGhost` does it now work and give the correct answer when running multi-GPU?

Marcogarofalo: Yes.

tests/host_reference/clover_reference.cpp
```diff
@@ -312,18 +312,19 @@ void tmc_dslash(void *out, void **gauge, void *in, void *clover, void *cInv, dou
...
-                      precision);
-    if (matpc_type == QUDA_MATPC_EVEN_EVEN_ASYMMETRIC || matpc_type == QUDA_MATPC_ODD_ODD_ASYMMETRIC) {
-      wil_dslash(tmp2, gauge, tmp1, parity, dagger, precision, param);
-      twistCloverGamma5(out, tmp2, clover, cInv, dagger, kappa, mu, flavor, parity, QUDA_TWIST_GAMMA5_INVERSE, precision);
-    } else {
-      wil_dslash(out, gauge, tmp1, parity, dagger, precision, param);
-    }
-  } else {
+
+  if (matpc_type == QUDA_MATPC_EVEN_EVEN_ASYMMETRIC || matpc_type == QUDA_MATPC_ODD_ODD_ASYMMETRIC) {
```
Marcogarofalo, line 316: I noticed that these two tests were failing
```
mpirun -n 1 ./tests/dslash_test  \
        --dslash-type twisted-clover --compute-clover 1 \
		--matpc even-even-asym \
	   	--dim  4 4 4 4 --prec ${prec} \
		--gridsize 1 1 1 1\
		--niter 2 \
		--compute-clover-trlog 1 \
		--dagger \
 		--gtest_output=xml:gauge_path_test_${prec}.xml | tee test-even-even-asym-dag.txt
```
```
mpirun -n 1 ./tests/dslash_test  \
        --dslash-type twisted-clover --compute-clover 1 \
		--matpc odd-odd-asym \
	   	--dim  4 4 4 4 --prec ${prec} \
		--gridsize 1 1 1 1\
		--niter 2 \
		--compute-clover-trlog 1 \
		--dagger \
 		--gtest_output=xml:gauge_path_test_${prec}.xml | tee test-odd-odd-asym-dag.txt
```
with
```
1.000000e-01 Failures: 0 / 3072  = 0.000000e+00
1.000000e-02 Failures: 2010 / 3072  = 6.542969e-01
1.000000e-03 Failures: 2944 / 3072  = 9.583333e-01
1.000000e-04 Failures: 3060 / 3072  = 9.960938e-01
``` 
this change should fix those two tests. For completeness, I try all combinations of 
```
--matpc even-even odd-odd even-even-asym odd-odd-asym
```
with the `--dagger` option on and off

Marcogarofalo: The Reference implementation of  `computeTMCloverForceQuda` works when tested with
```
mpirun -n 2  ./tests/TMCloverForce_test  --dslash-type twisted-clover --compute-clover 1 \
        --matpc even-even-asym   --dim  4 4 4 4 --prec double   --gridsize 2 1 1 1\
        --niter 1  --compute-clover-trlog 1  --verbosity verbose  --kappa 1  --mu 1 \
        --clover-csw 1   --gtest_output=xml:gauge_path_test_${prec}.xml
```

maddyscientist:
@Marcogarofalo since it seems you have a working test now, and the code is working against tmLQCD, does this mean this is ready for review now?

Marcogarofalo:
Hi, thanks for the reply. I am unsure what requirements to meet before starting a review. Maybe I can give a first clean-up before starting.

maddyscientist:
@Marcogarofalo ideally we have:
* all build tests passing
* latest develop branch merged in
* doxygen applied for all new functions (in the headers)
* any unit tests added to the ctest list

I'm happy to help you get this over the line.  Most of the hard work is done I think.

lib/interface_quda.cpp
```diff
@@ -5073,6 +5092,179 @@ void computeCloverForceQuda(void *h_mom, double dt, void **h_x, void **, double
   profileCloverForce.TPSTOP(QUDA_PROFILE_TOTAL);
 }
 
+void computeTMCloverForceQuda(void *h_mom, void **h_x, double *coeff, int nvector, 
+    void *h_gauge, QudaGaugeParam *gauge_param, QudaInvertParam *inv_param)
```
maddyscientist: @Marcogarofalo  This `h_gauge` variable here is causing an unused variable warning which results in the build tests failing (CI builds use `STRICT` building which promotes all warnings to errors).  To fix this, either:
* delete the variable from the argument list
* just write it as `void *` instead of `void *h_gauge`, e.g., leave it unnamed
I suspect we can rely on having a resident gauge field, so you could just delete the variable from the argument list.

maddyscientist: (To see the compiler warnings, click on a failing build (the red cross) and you should be able to see the console build output after a few other clicks.  From there you can see the build failures, to allow you to fix any failures.)

Marcogarofalo: Thank you for the explanation. I didn't fix it immediately because I got distracted by the fact that the tuner was not working after merging with develop
```
ERROR: Failed to find key entry (N4quda11CloverForceIdLi3EL21QudaReconstructType_s18EEE:4x4x4x4:GPU-offline,vol=256stride=128precision=8geometry=4Nc=3,exterior) (rank 1, host lnode17.cluster.hiskp, tune.cpp:1065 in quda::TuneParam quda::tuneLaunch(quda::Tunable&, QudaTune, QudaVerbosity)())
       last kernel called was (name=N4quda11CloverForceIdLi3EL21QudaReconstructType_s18EEE,volume=4x4x4x4,aux=GPU-offline,vol=256stride=128precision=8geometry=4Nc=3,exterior)
```

Marcogarofalo: The tuner issue went away by making a new build directory. Still `computeTMCloverForceQuda` is not working in single precision and the test  I think is not compiling if `MULTI_GPU` is not enabled

Marcogarofalo: @maddyscientist I tried to fix the errors in the integration tests but I think the last one is not under my control.

maddyscientist: Ok, mind if I take a pass at cleaning up these issues then?

Marcogarofalo: Please go ahead, thanks.

maddyscientist:
Ok, I've started to work on it, to address some of the issues.  I've found a few limitations, that would be good to fix:
* Add support for Wilson-clover, e.g., no twisted mass term
* Support for even-even / odd-odd preconditioning, and not just asymmetric versions
* Test code hard codes sloppy, refinement and precondition preconditions (I can fix this)

maddyscientist:
Also, I am finding that if I run the test command line above:
```
mpirun -n 2  ./tests/TMCloverForce_test --dslash-type clover --compute-clover 1 --matpc even-even-asym --dim  4 4 4 4 --prec double --gridsize 2 1 1 1 --niter 1 --compute-clover-trlog 1 --verbosity verbose --kappa 1 --mu 1 --clover-csw 1
```
I get an error:
```
clover.TrLog()[0]=-nan, clover.TrLog()[1]=-nan
ERROR: Clover trlog has returned -nan, likey due to the clover matrix being singular. (rank 0, host nvsocal2, clover_invert.cu:26 in quda::CloverInvert<store_t>::CloverInvert(quda::CloverField&, bool) [with store_t = double]())
       last kernel called was (name=N4quda12CloverInvertIdEE,volume=4x4x4x4,aux=GPU-offline,vol=256precision=8Nc=3,trlog=true,twist=false)
```
But I guess you don't get this problem?

SaltyChiang:
> I get an error:
> 
> ```
> clover.TrLog()[0]=-nan, clover.TrLog()[1]=-nan
> ERROR: Clover trlog has returned -nan, likey due to the clover matrix being singular. (rank 0, host nvsocal2, clover_invert.cu:26 in quda::CloverInvert<store_t>::CloverInvert(quda::CloverField&, bool) [with store_t = double]())
>        last kernel called was (name=N4quda12CloverInvertIdEE,volume=4x4x4x4,aux=GPU-offline,vol=256precision=8Nc=3,trlog=true,twist=false)
> ```

@maddyscientist This often happens when I'm trying to build the clover term for a random gauge generated by gaussGaugeQuda. No idea why this happens, but decreasing sigma works.

And I noticed @Marcogarofalo calls `computeCloverForce` with `... quarkP, quarkX, ...` instead of `... quarkX, quarkP` (the latter one should be the correct order described by docstring). I tried to do a similar thing (you can see the modification [here](https://github.com/CLQCD/quda/commit/b47950dd3026e2356298e41199e235d68f593746), which is almost a swap between `A,B` and `C,D`), and get a reasonable deltaH in a clover HMC process. If this is really an issue, maybe we should change the implementation of `computeCloverForce` to match the docstring.

Also, the dynamic inverse of the clover term in clover_trace.cuh now differs from the one in clover_invert.cuh. And I'm wondering what will happen if we disable dynamic clover term?

Marcogarofalo:
@maddyscientist  I do not get this problem, maybe setting `--compute-clover-trlog 0` will avoid the problem for the moment.

@SaltyChiang I saw https://github.com/CLQCD/quda/commit/b47950dd3026e2356298e41199e235d68f593746 but then also the exterior kernel needs to be modified with the same swap.

SaltyChiang:
@Marcogarofalo The modification I made in that commit is not a good one, but just use that to make HMC on a single GPU work. I think it's better to swap `x` and `p` [here](https://github.com/lattice/quda/blob/acadf0de23391bf00e2c8c4dc32ab16cd366a5a0/lib/clover_outer_product.cu#L151).

Marcogarofalo:
> Ok, I've started to work on it, to address some of the issues. I've found a few limitations, that would be good to fix:
> 
> * Add support for Wilson-clover, e.g., no twisted mass term

Add support for wilson-clover should be possible, in principle just setting mu=0 should work. A problem could be whether to use the Hermitian operator or not.

> * Support for even-even / odd-odd preconditioning, and not just asymmetric versions

I can type the symmetric version but I do not have any existing reference to test against it

> * Test code hard codes sloppy, refinement and precondition preconditions (I can fix this)

Regarding the  error



> Also, I am finding that if I run the test command line above:
> 
> ```
> mpirun -n 2  ./tests/TMCloverForce_test --dslash-type clover --compute-clover 1 --matpc even-even-asym --dim  4 4 4 4 --prec double --gridsize 2 1 1 1 --niter 1 --compute-clover-trlog 1 --verbosity verbose --kappa 1 --mu 1 --clover-csw 1
> ```
> 
> I get an error:
> 
> ```
> clover.TrLog()[0]=-nan, clover.TrLog()[1]=-nan
> ERROR: Clover trlog has returned -nan, likey due to the clover matrix being singular. (rank 0, host nvsocal2, clover_invert.cu:26 in quda::CloverInvert<store_t>::CloverInvert(quda::CloverField&, bool) [with store_t = double]())
>        last kernel called was (name=N4quda12CloverInvertIdEE,volume=4x4x4x4,aux=GPU-offline,vol=256precision=8Nc=3,trlog=true,twist=false)
> ```
> 
> But I guess you don't get this problem?

I did not get the program because I was using `--dslash-type twisted-clover`, with `--dslash-type clover`  the value of kappa used is too big.  With the default kappa value `--kappa 0.121951` the test run but it fails.

Marcogarofalo:
> Ok, I've started to work on it, to address some of the issues. I've found a few limitations, that would be good to fix:
> 
> * Add support for Wilson-clover, e.g., no twisted mass term
> * Support for even-even / odd-odd preconditioning, and not just asymmetric versions

The problem with even-even / odd-odd symmetric preconditioning is that the force is not obtained from the asymmetric version with a change of the operator, but more algebra is needed, at least from what I understand fromAppendix A of https://arxiv.org/pdf/hep-lat/0112051.pdf.
I do not have a reference implementation in tmLQCD to test against it. I would prefer to return an error if `computeTMCloverForceQuda` is called with even-even / odd-odd preconditioning and in the future someone interested can implement it. Does this strategy seem reasonable to you?

maddyscientist:
> The problem with even-even / odd-odd symmetric preconditioning is that the force is not obtained from the asymmetric version with a change of the operator, but more algebra is needed, at least from what I understand fromAppendix A of https://arxiv.org/pdf/hep-lat/0112051.pdf. I do not have a reference implementation in tmLQCD to test against it. I would prefer to return an error if `computeTMCloverForceQuda` is called with even-even / odd-odd preconditioning and in the future someone interested can implement it. Does this strategy seem reasonable to you?

Understood @Marcogarofalo.  Returning an error for the symmetric case is fine I think.

Marcogarofalo:
> Understood @Marcogarofalo. Returning an error for the symmetric case is fine I think.

I implemented it. Maybe it can help if I list the problems that I did not manage to solve:

-  The ghost exchange in computeCloverForce is not optimal https://github.com/lattice/quda/blob/30e4f52b0f45b259c19ddb7367f04185be51d252/lib/clover_outer_product.cu#L157-L160   It works but there are unnecessary communications done. Also, notice that the order of this exchange is important, for example, the following does not work  
   ```
          inB.exchangeGhost((QudaParity)(1-parity), nFace, dag);
          inD.exchangeGhost((QudaParity)(1-parity), nFace, 1-dag);
          exchangeGhost(inB, parity, dag);
          exchangeGhost(inD, parity, 1-dag);
    ```
   I think it should be something like the two spinors B and D use the same ghost array.

- In the host reference implementation I did not manage to create an extended `oprod`  (a `cudaGaugeField` with `QUDA_TENSOR_GEOMETRY`) so I used the device routines to exchange the ghost and then I downloaded on the  host https://github.com/lattice/quda/blob/30e4f52b0f45b259c19ddb7367f04185be51d252/tests/host_reference/TMCloverForce_reference.cpp#L197-L203
- This is not really a problem but, I found that the `cloverInvert` function in https://github.com/lattice/quda/blob/feature/tm_force/include/kernels/clover_invert.cuh#L62-L64 does not really compute the clover inverse for twisted mass. It is missing this part that I add in `computeCloverSigmaTrace`
  https://github.com/lattice/quda/blob/30e4f52b0f45b259c19ddb7367f04185be51d252/include/kernels/clover_trace.cuh#L59-L61
but I guess that changing this is tedious, one needs to change the dirac operator accordingly

kostrzewa:
@maddyscientist Do you think that we could discuss the first point that @Marcogarofalo raises above in https://github.com/lattice/quda/pull/1338#issuecomment-1790508776 about the ghost exchange? We just discussed and he suggested that perhaps splitting the computation to do one checkerboard first and then the other would provide an easy fix for the fact that currently communication is done twice. Our next "physics" target is to get the force working for our rational approximation for the "1+1" sector (which uses the two-flavour ND operator) as this is the single largest remaining bottleneck and it would be great if we could benefit there from doing the least amount of comms.

maddyscientist:
@kostrzewa yes this would be good to discuss.  I'm planning to push a few updates shortly to this branch, namely to do some cleanup and merge in the latest develop branch.  Let's discuss this after I've done this (which I'll hopefully get to in the next day or two).

maddyscientist:
@kostrzewa @Marcogarofalo I've done further testing and cleanup:
* Merged in latest develop branch
* Fixed the force when the clover inverse is explicitly pre-computed and stored
* Cleanup code to use objects instead of pointers

Regarding the redundant communications in exchangeGhost: can you explain what unnecessary comms are being done here?

I've found a few outstanding issues that it need to be fixed:
* The test should test multiple quark fields, e.g., as in RHMC
* The test should test with a determinant ratio (ctest should automatically test both paths)
* There's a segmentation fault when enabling partitioning: segmentation is in the test code itself (e.g., run with `--partition 12` on a single GPU)

Marcogarofalo:
@maddyscientist  thank you for the help.

> Regarding the redundant communications in exchangeGhost: can you explain what unnecessary comms are being done here?

I think that it is here
https://github.com/lattice/quda/blob/901ddedfb572385c4f6c8b1a14df392268f70cf7/lib/clover_outer_product.cu#L153-L156

it looks like there is an exchange with `inB.exchangeGhost` and then with `exchangeGhost`, I do not fully understand this part so maybe I am wrong. With only line 154 the kernel gives `CUDA_ERROR_ILLEGAL_ADDRESS`. I add the line 153 in 
https://github.com/lattice/quda/pull/1338/commits/1e1af7f561c266efbda76e1a604218b13b30dc22
I also noticed that swapping any of the lines above gives a wrong result.

> 
> I've found a few outstanding issues that it need to be fixed:
> 
> * The test should test multiple quark fields, e.g., as in RHMC

Sorry I was ignoring this part because in tmLQCD the RHMC it is different, needs the non-degenerate operator.
I can work on it

> * The test should test with a determinant ratio (ctest should automatically test both paths)
ok
> * There's a segmentation fault when enabling partitioning: segmentation is in the test code itself (e.g., run with `--partition 12` on a single GPU)

I will have a look

Marcogarofalo:
@maddyscientist  I hope I fix the issue you listed, I see a failure in the CI but looking at the logs I see
`slurmstepd: error: *** STEP 50331581.1 ON nid04394 CANCELLED AT 2023-12-05T20:00:47 DUE TO TIME LIMIT ***`

maddyscientist:
@Marcogarofalo Thanks for the latest fixes.  All looks good, and I wouldn't worry about the CI failure at CSCS - this target is a bit temperamental and sometimes takes longer than the maximum time, hence the failure.

I've identified the issue with the exchange, and working a on a fix now.

maddyscientist:
Ok, I have pushed the fix for the quark field exchange.  Issue was that the ghost pointers in the accessor were not being set correctly.  This was caused by two ghost pointer getter methods: `ColorSpinorField::Ghost()` and `ColorSpinorField::Ghost2()`, which do slightly different things (some long-term cruft that I need to address, and unify these two).

I've also pushed some rudimentary OMP parallelization of the host test code which allows for easier testing of larger problem sizes.

maddyscientist:
@Marcogarofalo @kostrzewa what from your end needs to be done on this PR?  I assume the non degenerate force is for a later PR?

Marcogarofalo:
@maddyscientist I also think that it is better to keep the non degenerate force for another PR.
My only doubt remaining here is in the reference implementation. I used the device routine to create the extended version of oprod
https://github.com/lattice/quda/blob/c1ee1f7d0df32ad4329d6ef4531d5bc37f7b69a4/tests/host_reference/TMCloverForce_reference.cpp#L236
I didn't manage to make the host version with `QUDA_TENSOR_GEOMETRY`, I am not sure whether the current state is acceptable.

maddyscientist:
Thanks @Marcogarofalo for your thoughts.  I'll take a look at the reference code, though I don't view this as critical.

I'm curious: compared to the reference tmLQCD implementation, how is QUDA performance looking?  I'll push some more performance updates to this branch shortly.

Marcogarofalo:
From a quick test on a `64^3 x 128` lattice using 4 nodes of juwels-booster
https://apps.fz-juelich.de/jsc/hps/juwels/booster-overview.html
assuming that I am using the machine in a reasonable way I get
- tmLQCD native implementation 3.60 s
- computeTMCloverForceQuda  0.35 s 

It is not a physical ensemble, however, it should not matter for this test.

urbach:
> From a quick test on a `64^3 x 128` lattice using 4 nodes of juwels-booster
> https://apps.fz-juelich.de/jsc/hps/juwels/booster-overview.html
> assuming that I am using the machine in a reasonable way I get
> - tmLQCD native implementation 3.60 s
> - computeTMCloverForceQuda  0.35 s 

wow! excellent!

maddyscientist:
I take it from @urbach’s comment then that 10x is good enough.

A word on performance: the most compute intensive function is the clover term derivative kernel.  On my Volta this is achieving about 50% of peak so is likely about as optimal as one would hope.

There’s more performance to be squeezed out yet, specifically with respect to the outer product over multiple quark fields.  Since this PR isn’t really concerned with RHMC performance for TMC I guess this optimization isn’t too important for this PR.

maddyscientist:
And the other big optimization would be to avoid the host device transfers but that’s also likely not super critical until we’re thinking about RHMC.

kostrzewa:
> I take it from @urbach’s comment then that 10x is good enough.

The speedup is not quite that high but it still leads to a major reduction in total time spent in the light sector of our simulations.

The force computation in tmLQCD for a determinant or determinant ratio takes around 5 seconds (not including the solve itself) in this particular test case (on 4 nodes, on 4 x 48 Zen2 cores). Offloading to to QUDA reduces this to 1.8 seconds.

The host-device transfer overhead for our RHMC part (two flavours) is around 0.5 to 1.2 seconds (depending on the number of shifts) in this test case on 4 nodes. This should be compared to 3-15 seconds spent in the solves for the different monomials. The gain of reducing host-device transfers would not be very large.

There is one major improvement that we're still missing: the calculation of the clover term and its inverse, which are always done on the CPU *and* on the GPU. The CPU computation is a total waste of time and should be offloaded and transferred back, it's "just" a matter of setting everything up which we haven't got around to yet...

kostrzewa:
A nice side-effect: @Marcogarofalo caught an OMP stupidity which we had in our host-device transfer of our momentum field (thanks to yours truly), see https://github.com/etmc/tmLQCD/commit/020dfa3632ec214a11c72905ff15da0b23ceadde. This will lead to a major reduction in cost in the gauge sector saving hundreds of seconds per trajectory in some cases...

maddyscientist:
When running multi-node, the limiter might be the comms for the quark as part of the outer product.  This is presently written in a very dumb way and doesn’t use NVLink or any overlapping at all, with excessive synchronization.  In my next push I’ll make sure that is properly included in the profile summary to stdout so we can verify this.

include/kernels/clover_sigma_outer_product.cuh
```diff
@@ -21,21 +21,20 @@ namespace quda
...
-      inA{*inA[0]},
-      inB{*inB[0]}
+      oprod(oprod)
     {
       for (int i = 0; i < nvector; i++) {
-        coeff[i][0] = coeff_[i][0];
-        coeff[i][1] = coeff_[i][1];
+        this->inA[i] = inA[i];
+        this->inB[i] = inB[i];
+        coeff[i] = {coeff_[i][0], coeff_[i][1]};
```
kostrzewa: this leads to a narrowing conversion error, needs a cast to `real`

maddyscientist: Yep, I'll catch all the CI errors on my next push.

maddyscientist:
I think this PR is ready for wider review from the QUDA core.  @weinbe2 can you take a look?

A few outstanding comments:
* Eventually I'd like to remove the redundancy between the `computeCloverForceQuda` and `computeTMCloverForceQuda` functions and have a single interface for doing this.  The former function is catered towards the conventions of MILC and the latter is catered to tmLQCD.  I think that can wait until subsequent cleanup but happy to hear opinions about this.
* @Marcogarofalo noted some of the reference code had some issues, so there's a workaround in place at present with respect to the outer product field extension.  This is something I still intend to take a look at.

There's quite a bit of optimization left on the table but these are all things that can be done at a later date:
* The outer product computation is presently done as a serial loop, with separate comms and kernel for each quark field.  Ideally this would be deployed as a multi-RHS kernel.  This will be easy to deploy once we have multi-RHS Dslash kernels in place.
* There's some redundant communication being done at the moment (the `gauge` and `oprod` fields are being over extended)
* I suspect the end result will be a 2-3x boost in performance, depending on degree of scaling.

@Marcogarofalo @kostrzewa one thought from that link to tmLQCD you posted above regarding the OMP snafu.  It looks like you are doing reordering in tmLQCD on the host.  Ideally you'd be doing this in QUDA on the device, as GPUs are much faster at doing this reordering than CPUs.  Any reason you can't just give the raw fields to QUDA, and let it handle this?

kostrzewa:
> @Marcogarofalo @kostrzewa one thought from that link to tmLQCD you posted above regarding the OMP snafu. It looks like you are doing reordering in tmLQCD on the host. Ideally you'd be doing this in QUDA on the device, as GPUs are much faster at doing this reordering than CPUs. Any reason you can't just give the raw fields to QUDA, and let it handle this?

It's a historical thing: this is just what we started with at the time. Because the overhead of doing this was / is minor compared to everything else on most machines, we never put it high on our list of priorities (less than1% of total runtime). Now that the light force and especially once the 1+1 sector RHMC force is done on the GPU, however, we should move this to the GPU eventually, same as the calculation of the clover term and its inverse. The latter are even computed redundantly right now...

kostrzewa:
> Eventually I'd like to remove the redundancy between the computeCloverForceQuda and computeTMCloverForceQuda functions and have a single interface for doing this. The former function is catered towards the conventions of MILC and the latter is catered to tmLQCD. I think that can wait until subsequent cleanup but happy to hear opinions about this.

I completely agree. This was the original plan but time constraints on my side have prevented this, unfortunately.

include/clover_field.h
```diff
@@ -10,6 +10,19 @@ namespace quda {
   namespace clover
   {
 
+    /**
+       @brief Helper function that returns whether we have enabled
+       clover fermions.
+     */
+    constexpr bool is_enabled()
```
weinbe2: This isn't a hill I'll die on, but I think I'd personally prefer if this was called `is_clover_enabled()`, even though I see it's in the `clover` namespace, to absolutely minimize potential ambiguity.

This would be consistent with precedent set in `instatiate.h`, where the enabled functions either include a specifier in the function name or require an explicit template parameter that serves the same self-clarifying purpose.

include/kernels/clover_trace.cuh
```diff
@@ -4,146 +4,155 @@
...
     // load the clover term into memory
-    for (int mu=0; mu<4; mu++) {
-      for (int nu=0; nu<mu; nu++) {
+#pragma unroll
+    for (int ch = 0; ch < 2; ch++) {
+      A[ch] = arg.clover_inv(x, arg.parity, ch);
+      A[ch] *= static_cast<real>(2.0); // factor of two is inherent to QUDA clover storage
+
+      if constexpr (Arg::dynamic_clover) {
+        if (Arg::twist) { // Compute (T^2 + mu2 - epsilon2) first, then invert
```
weinbe2: `if constexpr`, because we can, and to match the `if constexpr(Arg::twist)` below.

include/quda.h
```diff
@@ -1531,6 +1531,22 @@ extern "C" {
   void computeCloverForceQuda(void *mom, double dt, void **x, void **p, double *coeff, double kappa2, double ck,
 			      int nvector, double multiplicity, void *gauge,
 			      QudaGaugeParam *gauge_param, QudaInvertParam *inv_param);
+  
+  /** this is a minimal test example of a function which computes the force
```
weinbe2: I'm guessing this doxygen needs an update, since this isn't a minimal test example (anymore?)

lib/dslash_policy.cuh
```diff
@@ -646,7 +646,7 @@ namespace quda
 
             // Query if comms has finished
             if (!pattern.commsCompleted[2 * i + dir]) {
-              if (commsComplete(*in, dslash, i, dir, true, true, false)) {
+              if (commsComplete(in, dslash, i, dir, true, true, false)) {
                 ;
```
weinbe2: For the sheer virtue that there's a lot of other cleanup in this PR, can we delete this `;` as part of some other commit?

lib/interface_quda.cpp
```diff
@@ -4515,135 +4522,226 @@ void computeCloverForceQuda(void *h_mom, double dt, void **h_x, void **, double
...
+      ColorSpinorParam cpuParam(h_x[i], *inv_param, fParam.x, true, inv_param->input_location);
+      ColorSpinorField cpuQuarkX(cpuParam);
+      x[i].Even() = cpuQuarkX;
+      gamma5(x[i].Even(), x[i].Even());
+    } else {
+      x[i].Even() = solutionResident[i];
+    }
+  }
 
-  bool pc_solve = (inv_param->solve_type == QUDA_DIRECT_PC_SOLVE) ||
```
weinbe2, line 4549: Should there still be some type of check that everything passed in is full parity (based on other behavior in the function)?

weinbe2: (FYI, I edited this a few times, my original question was meaningless.)

weinbe2:
In `interface_quda.cpp`: is there a way we could rewrite `computeCloverForceQuda` and `computeTMCloverForceQuda` to be thin wrappers around some more general function, say, `computeGeneralForceClover` that uses info in `gauge_param` and/or `inv_param` to deduce if the regular or twisted details need to be called?

The reason I suggest this is because there seems to be a huge amount of preamble/epilogue boilerplate in common between these two functions, and this could be a good place to start cutting that boilerplate reuse down.

weinbe2:
The file `tests/host_reference/CloverForce_reference.h` and its twisted friends should be renamed `clover_force_reference.h`, etc, to match the conventions of other files.

This isn't a hill I'll die on, but I think there's reasonable scope to combine `clover_force_reference.cpp` and its twisted friend in a single file.

tests/host_reference/CloverForce_reference.h
```diff
@@ -0,0 +1,985 @@
...
+#include <command_line_params.h>
+#include <dslash_reference.h>
+#include <gauge_field.h>
+#include <host_utils.h>
+#include <quda.h>
+// #include <quda_matrix.h>
+// #include <complex_quda.h>
+#include <Eigen/Dense>
+
+// FIXME: this was copied from  wilson_dslash_reference.cpp maybe it is better to create a separate file with the projection
```
weinbe2: I agree :)

tests/host_reference/CloverForce_reference.h
```diff
@@ -0,0 +1,985 @@
...
+
+#include "gauge_force_reference.h"
+#include "misc.h"
+#include <color_spinor_field.h> // convenient quark field container
+#include <command_line_params.h>
+#include <dslash_reference.h>
+#include <gauge_field.h>
+#include <host_utils.h>
+#include <quda.h>
+// #include <quda_matrix.h>
```
weinbe2: May as well remove these comments

tests/host_reference/CloverForce_reference.h
```diff
@@ -0,0 +1,985 @@
...
+  {
+    {{1,0}, {0,0}, {1,0}, {0,0}},
+    {{0,0}, {1,0}, {0,0}, {1,0}},
+    {{1,0}, {0,0}, {1,0}, {0,0}},
+    {{0,0}, {1,0}, {0,0}, {1,0}}
+  }
+};
+// clang-format on
+
+static const double local_gamma[4][4][4][2] = {{// x
```
weinbe2: Missing `// clang-format off` and/or the `// clang-format on` above can be removed

tests/utils/command_line_params.cpp
```diff
@@ -1116,3 +1118,9 @@ void add_quark_smear_option_group(std::shared_ptr<QUDAApp> quda_app)
   opgroup->add_option("--smear-nsteps", smear_n_steps, "Number of smearing steps (default 50)");
   opgroup->add_option("--smear-t0", smear_t0, "Index of the time slice (default -1)");
 }
+
+void add_clover_force_option_group(std::shared_ptr<QUDAApp> quda_app)
+{
+  auto opgroup = quda_app->add_option_group("Clover force", "Options controlling clover force testing");
+  opgroup->add_option("--detratio", detratio, "Test a ratio of determinants. Default is false");
```
weinbe2: I think `--determinant-ratio` would be clearer, I won't push on renaming `detratio` everywhere though.

weinbe2:
I just completed a visual review of this PR, this is a fantastic contribution. I left a handful of comments but nothing crazy (in my opinion, anyway). I'll note that a `clang-format` is still outstanding, but that's just to have doing it written down somewhere, it'd be a last step.

I'll get to a due-diligence compile+run soon enough, I know @maddyscientist still needs to push a few small things.

tests/host_reference/CloverForce_reference.h
maddyscientist: 4954aff24fdf497bf043b86ccdbd3bafeb3d319c

tests/host_reference/CloverForce_reference.h
maddyscientist: 4954aff24fdf497bf043b86ccdbd3bafeb3d319c

tests/utils/command_line_params.cpp
maddyscientist: 45eaa3d85351e8bfd22e51ab22338ab7afac555b

lib/dslash_policy.cuh
maddyscientist: 7c4a177f21901ac2003b750df85f9f19e90ce20d

include/quda.h
maddyscientist: 014913e504d47481b3c15b71084b2986d3339831

include/kernels/clover_trace.cuh
maddyscientist: 014913e504d47481b3c15b71084b2986d3339831

tests/host_reference/CloverForce_reference.h
maddyscientist: https://github.com/lattice/quda/commit/4954aff24fdf497bf043b86ccdbd3bafeb3d319c

include/clover_field.h
maddyscientist: I had a different design in mind, the `clover::is_enabled()` was one of expediency.  I might as well just push the real one....
``` c++
constexpr bool is_enabled(QudaDiracType);
```
Will add this in my next round of pushes.

maddyscientist: After hacking around for a while, and realizing that a global edit for deploying `is_enabled()` throughout for Dirac type enablement is not appropriate for this PR, I've gone with a narrow fix.

6cefd1c4816e9c3750ad89126450e59a887620ee

lib/interface_quda.cpp
maddyscientist: Not sure what this question is here?  The quark fields that are passed in are all single parity.

maddyscientist:
> In `interface_quda.cpp`: is there a way we could rewrite `computeCloverForceQuda` and `computeTMCloverForceQuda` to be thin wrappers around some more general function, say, `computeGeneralForceClover` that uses info in `gauge_param` and/or `inv_param` to deduce if the regular or twisted details need to be called?
> 
> The reason I suggest this is because there seems to be a huge amount of preamble/epilogue boilerplate in common between these two functions, and this could be a good place to start cutting that boilerplate reuse down.

I've pushed some level of this cleanup in 1aa953ea6d411b5e99dafa89896b632f0a5edef8.  The main code that's left out of here is the construction of the quark solution vectors.  At some point we can clean this up and get it moved into the new clover_force.cpp file, but not for this PR I think.

lib/interface_quda.cpp
weinbe2: We discussed this offline, @maddyscientist is going to add a check to make sure the _solution_ type is `matpc` or `matpc_matpcdag` or whatnot

maddyscientist: [461f1b6](https://github.com/lattice/quda/pull/1338/commits/461f1b6615ac5c3de998f907a2342294b600de12) addresses this.  @Marcogarofalo @kostrzewa note that you will need to now set `QudaInvertParam::solution_type = QUDA_MATPCDAG_MATPC_SOLUTION` when calling the clover force term.

maddyscientist:
> The file `tests/host_reference/CloverForce_reference.h` and its twisted friends should be renamed `clover_force_reference.h`, etc, to match the conventions of other files.
> 
> This isn't a hill I'll die on, but I think there's reasonable scope to combine `clover_force_reference.cpp` and its twisted friend in a single file.

@weinbe2 c0fdc7bdcb1ed48121263129fb11913a20821805 address this suggestion: merging CloverForce_reference.h and TMCloverForce_reference.cpp into a single file clover_force_reference.cpp.

lib/interface_quda.cpp
Marcogarofalo: @maddyscientist thank you for the warning, I enforce `QUDA_MATPCDAG_MATPC_SOLUTION` in tmLQCD before calling the clover force in [9f7a6d5](https://github.com/etmc/tmLQCD/pull/549/commits/9f7a6d5906fe66faee0d5b5b361f538f45c76386). Before [9f7a6d5](https://github.com/etmc/tmLQCD/pull/549/commits/9f7a6d5906fe66faee0d5b5b361f538f45c76386) we were inheriting the `solution_type` from the inversion, which was `QUDA_MATPC_SOLUTION` when using the MG.

maddyscientist: Great, thanks for the fast update @Marcogarofalo .  I will also add a check that enforces `QudaInvertParam::matpc_type = EVEN_EVEN_ASYM` since the present design implicitly requires this.  Generalizing this to support to both even-even-asym and odd-odd-asym will be part of the update to fuse the two separate clover force interfaces.

One other thing I noticed from looking at your code: I see that you're doing the momentum accumulation on the host and not in QUDA.  With this is mind, you could avoid the zeroing on the host and the download of the field to the gpu by setting `QudaInvertParam::overwrite_mom = 1` (I already applied this optimization to the test code).  Should give a wee speedup.  Alternatively of course, you could just delegate the update to QUDA and delete the accumulation on the host.

Marcogarofalo: Thank you for the wee tip, for the moment I made the easiest solution of the two [b6e5a6d](https://github.com/etmc/tmLQCD/pull/549/commits/b6e5a6d8632ae00eb6009b9a3970d8925c69ca4e).
To do the accumulation on QUDA I'll need to rearrange the momentum before and after the QUDA call.

Marcogarofalo: > Great, thanks for the fast update @Marcogarofalo . I will also add a check that enforces `QudaInvertParam::matpc_type = EVEN_EVEN_ASYM` since the present design implicitly requires this. Generalizing this to support to both even-even-asym and odd-odd-asym will be part of the update to fuse the two separate clover force interfaces.
> 
@maddyscientist sorry for the confusion in tmLQCD we need `QUDA_MATPC_ODD_ODD_ASYMMETRIC`

maddyscientist: That makes sense, when I looked at the code deeper I realized this must be the case 🙂.  Thanks for confirming.

Marcogarofalo:
@maddyscientist I just noticed that here 
https://github.com/lattice/quda/blob/7bd79ed0fff902871dd063e099994900009cfedf/lib/interface_quda.cpp#L4661-L4663
we could remove completely the `tmp` spinor. Can I push this change or this part is under refurbishment?

maddyscientist:
> @maddyscientist I just noticed that here
> 
> https://github.com/lattice/quda/blob/7bd79ed0fff902871dd063e099994900009cfedf/lib/interface_quda.cpp#L4661-L4663
> 
> 
> we could remove completely the `tmp` spinor. Can I push this change or this part is under refurbishment?

Sounds like a good idea, thanks.

maddyscientist:
@Marcogarofalo @kostrzewa just making sure I understand the conventions here: does tmLQCD solve for MdagM or MMdag when doing HMC forces?  Was having another look at parametrizing the differences between tmLQCD and MILC interfaces to simplify the code: apart from the parity differences, the only other main difference is that one seems to be the Hermitian conjugate of the other.

maddyscientist:
I was also confused by comments like this in the tmLQCD interface code:
```
    // QUDA applies the MMdag operator, we need QpQm^{-1) in the end
    // so we want QUDA to use the MdagM operator
    inv_param.dagger = QUDA_DAG_YES;
 ```
 In general the default with QUDA CG for example is MdagM not MMdag, so wanting to make sure we're on the same page.

Marcogarofalo:
My understanding is that in tmLQCD we need $(Q_+Q_-)^{-1}$ where $Q_{\pm}=\gamma_5M_{\pm}$. The sign $\pm$ is the sign of the twisted mass $\mu$. When we use the QUDA CG we are  doing $\gamma_5(M_+M_+^\dagger)^{-1}\gamma_5=(Q_+Q_-)^{-1}$ since $M_+^\dagger=\gamma_5M_-\gamma_5$. We call QUDA  with `QUDA_DAG_YES` to do  $(M_+M_+^\dagger)^{-1}$, the $\gamma_5$ are applied in tmLQCD before and after the inversion.

Then inside the force at https://github.com/lattice/quda/blob/19ce4dd47ff8f2711d85463f60f12210b995837a/lib/interface_quda.cpp#L4657-L4660 
We want $(M_-^{ee})^{-1}M^{eo}$ so in QUDA we are doing $\gamma_5 (M_+^{ee})^{-1}M^{oe} \gamma_5$.
With $M_\pm=M_\pm^{oo}-M^{oe}(M_\pm^{ee})^{-1}M^{eo}$.

Maybe some comments are misleading

maddyscientist:
@Marcogarofalo I've now parametrized the differences between MILC and tmLQCD.  For tmLQCD, I'm now expecting that `QudaInvertParam::matpc_type = QUDA_ODD_ODD_ASYMMETRIC` and `QudaInvertParam::dagger = QUDA_DAG_YES`.  Can you have a look at [0b99a1d](https://github.com/lattice/quda/pull/1338/commits/0b99a1d93dcd2f9d34958251518ee1d05396cdf3) you agree?

Marcogarofalo:
@maddyscientist Yes they are the ones that are passed when in tmLQCD we call QUDA CG, for the MG they are different but I set them after the MG and before calling the force function.

maddyscientist:
@weinbe2 this is ready for final sign off.  I'll push clang-format shortly.

weinbe2 (approved):
Hugely impressive work, thank you @kostrzewa , @Marcogarofalo , + @maddyscientist for additional clean-up and contributions. Merging momentarily!

Marcogarofalo:
Thank you All for making this possible!

PR #1345, opened 2022-12-21
Hotfix/cmake housekeeping

mathiaswagner:
update CPM
use more of FindCUDAToolkit and remove legacy workarounds
use NVTX3 (and drop support for legacy NVTX)
removed unused NVML stuff


maddyscientist:
Looks good to me.  Only thought is if we want to remove the driver version reporting with NVML.  That might be something we should be doing by default?

mathiaswagner:
It looked to me like enabling that from cmake was broken. Happy to make it the default.

maddyscientist:
It may well have been, it's been a while since I tried it.  😄   NVML is something we should always link to anyway I think, since it has a multitude of uses.

maddyscientist (approved):
Approving.

I pushed various minor fixes to get the CUDA 12 build passing.  One weird aspect of this is that `-isystem` was seemingly being ignored for CLI11.hpp, triggering a warning.  Not sure what was going on there, but I ended up just fixing the source code, which I think is reasonable since IIRC we have a modified version already anyway.

mathiaswagner:
Thanks. The missing `system` is a bit weird. At some point we should try to get to using an updated version of CLI11 but that is not high priority for now.

PR #1346, opened 2022-12-22
Fixing expsu3 for special case of zero input

sbacchio:
This is a patch for the expsu3 function for the edge case of an input zero matrix.
In that situation the algorithm would return `NaN`s instead of unity matrix.
This is because of illigal operations such the `log(0)` below.

The fix proposed adds only one extra if-statement for non-zero cases where the value of `tmp` is checked.
If that is zero then I make sure that the whole matrix is zero (just to catch other edge cases where `tmp` may get a zero value). Finally unity matrix is returned.


include/quda_matrix.h
```diff
@@ -1068,7 +1069,24 @@ namespace quda {
...
+
+      // If the matrix q is zero, this algorithm produces nan, instead of unity.
+      // The first invalid operation is the log hereafter. Therefore, we check
+      // if its argument (tmp) is zero and then return unity.
+
+      Complex tmp = a3 + sg2h3;
+      if (tmp.x == 0 and tmp.y==0) {
+	// Making sure q is a zero matrix
+	for(int i=0; i<9; i++) {
+	  assert(q(i).x==0 and q(i).y==0);
```
weinbe2: I looked at the PTX/SASS for a smaller code tested in the compiler explorer, and this does generate a non-inlined function call. Whether or not it's harmless to end-to-end performance, I think it might be cleaner to not use `assert`.

sbacchio: @weinbe2 thank you for having a look, then I would do the following (see updated code):
- check if the matrix is actually zero and return identity
- otherwise let the algorithm go through and one will get `nan` in the output. This then should be a signal to the user that something wrong happened. I believe it is better `nan` than a wrong result.

include/quda_matrix.h
```diff
@@ -1068,7 +1069,30 @@ namespace quda {
...
 
       Complex sg2h3 = sqrt(a3 * a3 - (Float)4. * a2 * a2 * a2);
-      Complex cp = exp(log((Float)0.5 * (a3 + sg2h3)) / (Float)3.0);
+
+      // If the matrix q is zero, this algorithm produces nan, instead of unity.
+      // The first invalid operation is the log hereafter. Therefore, we check
+      // if its argument (tmp) is zero and then return unity.
+
+      Complex tmp = a3 + sg2h3;
+      if (tmp.x == 0 and tmp.y==0) {
```
maddyscientist: Can you switch these `.x` to `.real()` and `.y` to `.imag()` here.  Eventually I'd like to have `std::complex` be a compatible drop in replacement for `quda::complex` (I know there's lots more instances of `.x` and `.y` spread all the place).

include/quda_matrix.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <assert.h>
```
maddyscientist: Can remove this file inclusion now.

maddyscientist (approved):
Aside from addressing the quick changes I've suggested (`x` -> `real()`, `y` -> `imag()`, and deleting the unnecessary include), and a clang format, this PR is good to go.

PR #1348, opened 2023-01-13
Adding gauge shift

sbacchio:
This is a (partial) implementation of shifting for gauge field. The CPU kernel is missing, but in light of the unification of cpuGaugeField and cudaGaugeField, I would avoid implementing it.


include/gauge_field.h
```diff
@@ -393,6 +393,13 @@ namespace quda {
      */
     virtual void copy(const GaugeField &src) = 0;
 
+    /**
+     * @brief Generic gauge field shift
+     * @param[in] src Source from which we are shifting (extended field in case of MPI)
+     * @param[in] dx Host array of shifts to apply to the field
+     */
+    virtual void shift(const GaugeField &src, const int *dx) = 0;
```
maddyscientist: I think we should use the  QUDA `array` class here for `dx`, rather than passing `const int *`.  This is a change I've started to propagate in other parts already.

include/gauge_field.h
```diff
@@ -543,6 +550,13 @@ namespace quda {
      */
     void copy(const GaugeField &src);
 
+    /**
+     * @brief Generic gauge field shift
+     * @param[in] src Source from which we are shifting (extended field in case of MPI)
+     * @param[in] dx Host array of shifts to apply to the field
+     */
+    void shift(const GaugeField &src, const int *dx);
```
maddyscientist: `const int *` -> `const array<int, 4> &`

include/gauge_field.h
```diff
@@ -680,6 +694,13 @@ namespace quda {
...
```
maddyscientist: const int * -> const array<int, 4> &

include/kernels/gauge_shift.cuh
```diff
@@ -0,0 +1,78 @@
...
+	X[i] = out.X()[i];
+	E[i] = in.X()[i];
+	border[i] = (E[i] - X[i])/2;
+	P += dx[i];
+      }
+      P = std::abs(P)%2;
+    }
+  };
+
+  template <typename Arg, int dir>
```
maddyscientist: There's no need here for `dir` to be a template.  It's not being used to index into a thread-local array, so making it a template only really adds bloat.

maddyscientist (changes requested):
Looks mostly good.  
* How about a unit test for this?
* Will need a clang format
* Also see my inline comments

include/gauge_field.h
```diff
@@ -872,4 +893,12 @@ namespace quda {
 
 #define checkReconstruct(...) Reconstruct_(__func__, __FILE__, __LINE__, __VA_ARGS__)
 
+  /**
+   * @brief Generic gauge field shift
+   * @param[out] dst Gauge field to store output
+   * @param[in] srd Source from which we are shifting (extended field in case of MPI)
+   * @param[in] dx Host array of shifts to apply to the field
+   */
+  void gaugeShift(GaugeField &dst, const GaugeField &src, const array<int, 4> &dx);
```
sbacchio, line 894: Should this be moved to e.g. `gauge_tools.h`?

sbacchio:
@maddyscientist thank you for the review. I applied the suggested changes and run clang format via `git-clang-format` (hopefully the right way).

Only the unit testing would be missing. Could you please point me to one of the current tests to extend / base it on?

include/gauge_field.h
maddyscientist: Yeah, that seems like a consistent place for it.

maddyscientist:
> @maddyscientist thank you for the review. I applied the suggested changes and run clang format via `git-clang-format` (hopefully the right way).

Looks good I think.

> Only the unit testing would be missing. Could you please point me to one of the current tests to extend / base it on?

I guess the easiest one to hack from as a base is `plaq_test`.  It's not quite what you want since this proposed test would be done using native QUDA types rather through the interface.

The bigger question is what the test would consist of.  Some ideas, OTOH:
* check that the plaquette is invariant after shifting
* check that performing a closed loop of shifts leads to a bitwise identical guage field to the initial field

Anything else come to mind?

include/gauge_field.h
maddyscientist: [a304a3a](https://github.com/lattice/quda/pull/1348/commits/a304a3ab0225e254f0afb7eb1be7475f39a9bd6b)

include/gauge_field.h
maddyscientist: [a304a3a](https://github.com/lattice/quda/pull/1348/commits/a304a3ab0225e254f0afb7eb1be7475f39a9bd6b)

include/gauge_field.h
maddyscientist: [a304a3a](https://github.com/lattice/quda/pull/1348/commits/a304a3ab0225e254f0afb7eb1be7475f39a9bd6b)

include/kernels/gauge_shift.cuh
maddyscientist: https://github.com/lattice/quda/pull/1348/commits/666a38e49176dbd6e365d42b1e45872fb76b25c9

PR #1351, opened 2023-01-17
Remove MILC gauge order from HISQ force verification code

weinbe2:
This narrow PR removes the plumbing for the MILC gauge order in the HISQ force verification code, since it was unused (and didn't work in the multi-GPU case) anyways.

FWIW, this was originally going to be a `hotfix` branch due to linker errors with an `extern` variable, but around the same time that I finished the work for this PR, it turns out the linker error was just due to a stale build directory. In any case, it doesn't hurt to do this cleanup.


mathiaswagner (approved):
Looks good and seems to enable a shared libquda_test

PR #1352, opened 2023-01-17
Hotfix/libquda test shared

mathiaswagner:
Some improvements for development workflow:
shared libquda_test
compress debug information for develop build (unless using QUDA_BACKWARDS as the stracktrace does break with it).

Should help with link times and disk space requirements.

Thanks to Evan for #1351 that enabled this.

Should go through some testing to make sure the shared libquda_test does not break anything -- at least it can't break apps using QUDA.


mathiaswagner:
A simple build with not a lot of things enabled had its build directory going from 3GB to 2.1Gb with this.

maddyscientist (approved):
This looks great, and the reduced test footprint is appreciated.

weinbe2 (approved):
I did a cursory check and everything looks good on my end as well. This includes testing the executables as built in the `[...]/build/test` directory, as well as making sure the shared test library is copied to the right place when running `make install` (or `ninja install` as appropriate).

PR #1353, opened 2023-01-26
some small fixes from sycl branch

jcosborn:
fix some warnings from clang++
fix some headers needed in sycl branch

### DIFF
--- a/include/eigen_helper.h
+++ b/include/eigen_helper.h
@@ -9,6 +9,7 @@
 #define EIGEN_DONT_VECTORIZE
 #endif
 
+#include <math.h>
 #include <Eigen/Eigenvalues>
 #include <Eigen/Dense>
 #include <Eigen/LU>
--- a/include/eigen_helper.h
+++ b/include/eigen_helper.h
@@ -9,7 +9,7 @@
 #define EIGEN_DONT_VECTORIZE
 #endif
 
-#include <math.h>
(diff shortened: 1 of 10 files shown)


include/eigen_helper.h
```diff
@@ -9,6 +9,7 @@
 #define EIGEN_DONT_VECTORIZE
 #endif
 
+#include <math.h>
```
maddyscientist: Probably should be `cmath` here since we're a C++ library 😄

include/targets/generic/load_store.h
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <target_device.h>
```
maddyscientist, line 3: Super pedant comment here (sorry): prefer a line space between these two lines.

maddyscientist (approved):
Looks good.  Some a couple of trivial fixes, but we can merge once those are done.

include/eigen_helper.h
jcosborn: 062dc04e98a324dc503822ac992771308e9cb8ab

include/targets/generic/load_store.h
jcosborn: 062dc04e98a324dc503822ac992771308e9cb8ab

maddyscientist (approved):

PR #1344, opened 2022-12-16
Feature/multi rhs kernel block ortho (DO NOT MERGE YET)

maddyscientist:
This is a reminder to myself to test this addition by @cpviolator.  This will be for after #1333 is merged.

PR #1357, opened 2023-02-06
Block TRLM tweaks

cpviolator:
This PR adds a new parameter `eig_max_ortho_attempts` to the eigensolver orthonormalisation routines. The solver will now error out if it exceeds this number of attempts (default 10).

It also changes the criteria for convergence on the block solver by summing all elements of the block beta array at the restart step, preventing the solver from falsely reporting full convergence while the largest eigenvectors are still not fully converged.


maddyscientist:
I have merged in the changes from #1344 into this PR since it concerns related aspects of the block TRLM solver.  Namely, the addition of an optimized Gram Schmidt block orthogonalization routine through the addition of a new interface parameter `ortho_block_size`.

I will aim to test this PR's features with block TRLM for coarse grid deflation acceleration.

maddyscientist (approved):
These changes are good.  I've verified things are working as expected for both of these new parameters introduced.

Block Lanczos on the coarse grid has a nice boost in performance due to the increased kernel fusion in the block orthogonalization with no real loss in stability.

I note that there is still scope for further improvement, since we could do the block orthogonalization using a fully fused classical algorithm, using Cholseky or equivalent.  This can be left for future work however.

PR #1360, opened 2023-02-20
Commended out hipDeviceSetCacheConfig call 

bjoo:
In a recent ROCm, this call stopped returning hipSuccess. Since we check the error code, it basically stops the code from running right at the outset. I am commenting it out here. I am told it will be fixed in a future ROCm, but actually I am not sure if it does anything in the first place.

### DIFF
--- a/lib/targets/hip/device.cpp
+++ b/lib/targets/hip/device.cpp
@@ -41,8 +41,9 @@ namespace quda
 
       CHECK_HIP_ERROR(hipSetDevice(dev));
 
-      CHECK_HIP_ERROR(hipDeviceSetCacheConfig(hipFuncCachePreferL1));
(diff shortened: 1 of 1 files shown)


kostrzewa:
Thanks! I bumped into this on LUMI-G with ROCm > 5.2.x but didn't want to bother you without first checking with our contacts there, which I haven't gotten around to yet. Good to know that we can test with newer versions without worrying about this.

maddyscientist (approved):

bjoo:
@kostrzewa can you please approve this PR in your role in target_hip_maintainers? I cannot approve my own PR....

kostrzewa (approved):

kostrzewa:
@bjoo should I push the button?

PR #1362, opened 2023-02-23
QUDA interface `freeGaugeQuda` clean-up

weinbe2:
This PR does a good deal of clean-up/removal of code duplication related to how gauge fields are `free`d in `interface_quda.cpp`, largely by introducing a new function `freeUniqueGaugeQuda(...);` which narrowly frees the "regular" gauge fields, the fat fields, the long fields, and the smeared fields depending on the argument passed in.

This is also a hotfix PR because it fixes a bug introduced in the MILC gauge observable PR. In this PR, the new MILC interface observable routines loaded the gauge field via `loadGaugeQuda`, allocating the precise field and making aliases for the sloppy, refinement, etc field. This conflicted with the existing routine to update the gauge fields, which as part of maintaining/updating residency would narrowly `delete` just the precise field without any consistent clean-up of the sloppy fields, etc. In many cases this skated by on UB --- the host structure memory which was allocated incidentally remained "uncleaned" so the sloppy fields could still be accessed. In recent tests, though, this triggered segfaults due to accessing already free'd memory (as it should).

Since this PR touches the infrastructure to free internal fields in many different places, it needs to be tested across many applications before being merged, so I've tagged multiple people who work on a range of different applications.

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -666,20 +666,59 @@ void loadGaugeQuda(void *h_gauge, QudaGaugeParam *param)
   // creating sloppy fields isn't really compute, but it is work done on the gpu
   profileGauge.TPSTART(QUDA_PROFILE_COMPUTE);
 
-  // Create pointers to store fields in
+  // switch the parameters for creating the mirror sloppy cuda gauge field
+  gauge_param.reconstruct = param->reconstruct_sloppy;
+  gauge_param.setPrecision(param->cuda_prec_sloppy, true);
   cudaGaugeField *sloppy = nullptr;
+  if (param->cuda_prec == param->cuda_prec_sloppy && param->reconstruct == param->reconstruct_sloppy) {
+    sloppy = precise;
+  } else {
+    sloppy = new cudaGaugeField(gauge_param);
+    sloppy->copy(*precise);
+  }
+
+  // switch the parameters for creating the mirror preconditioner cuda gauge field
+  gauge_param.reconstruct = param->reconstruct_precondition;
+  gauge_param.setPrecision(param->cuda_prec_precondition, true);
   cudaGaugeField *precondition = nullptr;
+  if (param->cuda_prec == param->cuda_prec_precondition && param->reconstruct == param->reconstruct_precondition) {
+    precondition = precise;
+  } else if (param->cuda_prec_sloppy == param->cuda_prec_precondition
+             && param->reconstruct_sloppy == param->reconstruct_precondition) {
+    precondition = sloppy;
+  } else {
+    precondition = new cudaGaugeField(gauge_param);
+    precondition->copy(*precise);
+  }
+
+  // switch the parameters for creating the refinement cuda gauge field
+  gauge_param.reconstruct = param->reconstruct_refinement_sloppy;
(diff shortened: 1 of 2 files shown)


lib/interface_quda.cpp
```diff
@@ -523,6 +523,10 @@ void initQuda(int dev)
 // possible flag to indicate we need to recompute the clover field
 static bool invalidate_clover = true;
 
+// pre-declare some cleanup utilities
+void freeUniqueSloppyGaugeUtility(cudaGaugeField*&, cudaGaugeField*&, cudaGaugeField*&, cudaGaugeField*&, cudaGaugeField*&);
+void freeUniqueGaugeUtility(cudaGaugeField*&, cudaGaugeField*&, cudaGaugeField*&, cudaGaugeField*&, cudaGaugeField*&, cudaGaugeField*&, QudaBoolean);
+
```
hummingtree: Wonder where are these two functions implemented?

hummingtree: Also might be worthwhile to give them a doxygen.

weinbe2: Haha, yes, good catch. They're implemented further down (closer to the other `free` functions), but they also have utility for the `load` functions.

weinbe2: https://github.com/lattice/quda/pull/1362/commits/02950b82682235b89f67a80b762b1d62986535ce

lib/interface_quda.cpp
```diff
@@ -523,6 +523,38 @@ void initQuda(int dev)
...
+ * @param precise[in/out] Reference to the pointer of a given "precise" field.
+ * @param sloppy[in/out] Reference to the pointer of a given "sloppy" field.
+ * @param precondition[in/out] Reference to the pointer of a given "precondition" field.
+ * @param refinement[in/out] Reference to the pointer of a given "refinement" field.
+ * @param eigensolver[in/out] Reference to the pointer of a given "eigensolver" field.
+ * @param extended[in/out] Reference to the pointer of a given "extended" field.
+ * @param preserve_precise[in] Whether (true) or not (false) to preserve the precise field.
+ */
+void freeUniqueGaugeUtility(cudaGaugeField*& precise, cudaGaugeField*& sloppy, cudaGaugeField*& precondition,
+                            cudaGaugeField*& refinement, cudaGaugeField*& eigensolver, cudaGaugeField*& extended, QudaBoolean preserve_precise);
```
maddyscientist: Since this is in internal function, `preserve_precise` can just be a regular `bool`.

maddyscientist (approved):
Looks good (apart from the `QudaBoolean` to `bool` change).

lib/interface_quda.cpp
weinbe2: https://github.com/lattice/quda/pull/1362/commits/1b2b21bc9eb68b19fb12095af331758950f5761b

weinbe2:
@maddyscientist @mathiaswagner when y'all have a moment, can you do a follow-up visual review of the code? @hummingtree helped me find a typo (thanks!), so I feel a few more independent reviews would be nice...

maddyscientist (approved):

PR #1364, opened 2023-03-13
Feature/large kernel args

maddyscientist:
This PR adds support for large kernel argument support to QUDA
* On Volta and above, CUDA now supports kernel arguments sized up 32,764 KiB.
* This functionality was introduced with CUDA 12.1, and requires NVIDIA graphics driver >= 530
* This feature provides a significant performance boost for short-running kernels that require large kernel arguments, since now the pre-kernel h-d memcpy is no longer needed, and we avoid a CUDA API call
* The QUDA CMake advanced option `QUDA_LARGE_KERNEL_ARG` enables utilization of this feature (for now left as `OFF` by default)
* When enabled, the `constant_kernel_arg.h` header serves no purpose, and its contents are hidden from the compiler


CMakeLists.txt
```diff
@@ -133,6 +133,8 @@ message(STATUS "QUDA target: ${QUDA_TARGET_TYPE}")
 # ######################################################################################################################
 
 # build options
+option(QUDA_LARGE_KERNEL_ARG "enable large kernel arg support" OFF)
```
weinbe2: I think this should be marked as `advanced` --- with my main motivation more being that I'd like to avoid folks "accidentally" turning it on.

include/multi_blas_helper.cuh
```diff
@@ -208,13 +208,12 @@ namespace quda
...
 
       constexpr auto arg_known_size = (sizeof(kernel_param<>)                                        // kernel_param parent
-				       + sizeof(int)                                                 // NYW parameter
 				       + sizeof(SpinorX[NXZ])                                        // SpinorX array
 				       + (Functor::use_z ? sizeof(SpinorZ[NXZ]) : sizeof(SpinorZ *)) // SpinorZ array
 				       + sizeof(Functor)                                             // functor
-				       + sizeof(dim3)                                                // threads parameter
 				       + (!Functor::use_w ? sizeof(SpinorW *) : 0)                   // subtract pointer if not using W
 				       + (Functor::reducer ? sizeof(ReduceArg<device_reduce_t>) : 0) // reduction buffers
+                                       + 12                                                          // alignment guard
```
weinbe2: I think we should have a better approach to the alignment guard then updating this number every time we have to change something --- perhaps we can do something more like:

```
constexpr auto arg_known_size_naive = (sizeof(kernel_param<>)                                        // kernel_param parent
				       + [...]
				       + (Functor::reducer ? sizeof(ReduceArg<device_reduce_t>) : 0));
constexpr size_t align_factor = 16; // representative, not sure if it's actually 16 bytes
constexpr auto arg_known_size = ((arg_known_size_naive + align_factor - 1)/(align_factor) * align_factor;
```
With the "floor" function perhaps more sanely wrapped in a `constexpr` function, but details.

include/multi_blas_helper.cuh
```diff
@@ -263,14 +262,13 @@ namespace quda
...
-				   + sizeof(int)                                                 // NYW parameter
 				   + (NXZ * spinor_x_size)                                       // SpinorX array
 				   + (Functor::use_z ? NXZ * spinor_z_size : sizeof(void *))     // SpinorZ array (else dummy pointer)
 				   + sizeof(Functor)                                             // functor
-				   + sizeof(dim3)                                                // threads parameter
 				   + (!Functor::use_w ? sizeof(void *) : 0)                      // subtract dummy pointer if not using W
 				   + (Functor::reducer ? sizeof(ReduceArg<device_reduce_t>) : 0) // reduction buffers
-				   );
+                                   + 12                                                          // alignment guard
+                                   );
```
weinbe2, line 271: Same alignment wrapper comment above

include/multi_blas_helper.cuh
weinbe2: (I'm aware this is ultimately orthogonal to this PR, but I've seen magic padding factors change a few too many times at this point)

weinbe2:
Should large kernel args be added to the tunekey string, at least for kernels whose argument is larger than 4KB? Or do the kernel arg vs constant memory versions of the functions have appropriately different kernel names?

weinbe2:
I have a few outstanding questions + comments, but overall I'm excited about this PR. I'm taking it for a spin now, and pending issues there and thoughts on my various comments this should be good to go.

include/multi_blas_helper.cuh
maddyscientist: [9f897ac](https://github.com/lattice/quda/pull/1364/commits/9f897ac922ab8f846d3c52a1df6b09cd6ae314f1)

include/multi_blas_helper.cuh
maddyscientist: [9f897ac](https://github.com/lattice/quda/pull/1364/commits/9f897ac922ab8f846d3c52a1df6b09cd6ae314f1)

CMakeLists.txt
maddyscientist: [9f897ac](https://github.com/lattice/quda/pull/1364/commits/9f897ac922ab8f846d3c52a1df6b09cd6ae314f1)

maddyscientist:
> Should large kernel args be added to the tunekey string, at least for kernels whose argument is larger than 4KB? Or do the kernel arg vs constant memory versions of the functions have appropriately different kernel names?

Thanks Evan for this suggestion.  I've implemented this idea here: https://github.com/lattice/quda/pull/1364/commits/bad9df8302f1e739721dcb6ca3d5b79193355bdd

maddyscientist:
I've now done some testing of the effect of this optimization on Ampere, and the speedup is impressive.  Below is for the block caxpy kernel running in single precision with 32x32 (performance in GFLOPS):

| L | constant | large arg |
| - | -- | - |
| 2 |  64.4 |  100.0  |
| 4 |  977 |  1500  |
| 6 |  2720 |  3640  |
| 8 |  4760 |  6080  |
| 10 | 4680 |  6440 |
| 12 |  3170 | 5120  |
| 14 | 3130 | 4970  |
| 16 |  3190|  4960 |

Moreover, the minimum runtime decreases from ~12 us to <8 us.

CMakeLists.txt
```diff
@@ -133,6 +133,9 @@ message(STATUS "QUDA target: ${QUDA_TARGET_TYPE}")
...
```
mathiaswagner: just wondering if this should move to the target cuda specific part and if we should check for CUDA >= 12.1 there as well

maddyscientist: e312d10888a28f9c1ab9f63a2f73ff63e0f9e2c5

weinbe2 (approved):
Everything looks good, and from some offline conversations I'm satisfied with the testing done on this PR. I personally consider it good to go.

PR #1365, opened 2023-03-14
fix tests for SYCL target

jcosborn:


include/eigen_helper.h
```diff
@@ -9,7 +9,7 @@
 #define EIGEN_DONT_VECTORIZE
 #endif
 
-#include <cmath>
+#include <math.h>
```
maddyscientist, line 12: why this change?

tests/invert_test_gtest.hpp
```diff
@@ -205,6 +206,7 @@ INSTANTIATE_TEST_SUITE_P(MultiShiftEvenOdd, InvertTest,
                                  no_schwarz),
                          gettestname);
 
+#ifdef QUDA_MMA_AVAILABLE
```
maddyscientist: How about we skip this these tests rather than comment them out if MMA isn't available? e.g., `GTEST_SKIP`

lib/gauge_fix_ovr.cu
```diff
@@ -71,7 +71,10 @@ namespace quda {
     bool advanceAux(TuneParam &param) const
     {
       param.aux.x = (param.aux.x + 1) % 6;
-       // mu must be contained in the block, types 0, 1, 2 have mu = 8 and 3, 4, 5 have mu = 4
+      if (!device::shared_memory_atomic_supported()) { // 1, 4 use shared memory atomics
```
maddyscientist, line 74: Is lack of shared memory atomics a thing on some SYCL platforms?

maddyscientist (commented):
Looks good.  Just a couple of minor queries, but we should be able to merge this pronto.

include/eigen_helper.h
jcosborn: Actually, it needs the functions (isinf, isnan) in the root namespace, not std::.

maddyscientist: What needs the functions in the root namespace?

lib/gauge_fix_ovr.cu
jcosborn: This is mainly for testing.  I don't think there is actually a way to query it, so is seems to expect it to be supported.

include/eigen_helper.h
jcosborn: I think this is the cause: https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/MathFunctions.h#L984

tests/invert_test_gtest.hpp
jcosborn: a3e633450853ed8c31bcc392632b69adb61cdf1e

maddyscientist (approved):

PR #1367, opened 2023-03-17
Fat/long force optimizations

weinbe2:
This PR constitutes a large refactor and optimization of the implementation of the fat/long force computation in QUDA. It introduces multiple different groups of optimizations:

**Five- and seven-link terms**
- Fully fuses the forward and backward seven-link term in the "rho" direction into the subsequent middle five-link term in the "nu" direction
- Fuses the middle five-link "nu" direction contribution into the side-link contribution in the next "nu" direction

**Three-link and Lepage terms**
- Fully fuses the middle three-link term and the middle and side Lepage terms (which enter in the force from the first part of the HISQ force chain rule)
- Fuses the middle three-link "mu" direction into the side-link contribution in the next "mu" direction

**Ancillary detriments and benefits from fusion**
- The above-noted fusions of middle-link force terms with the next side-link force terms requires additional "double buffer" fields
- However, this PR also removes the temporary `Qmu` field, which was ultimately a shifted version of the original `U` gauge link field, so the memory bloat from this PR is relatively minimal.
- The above fusion incidentally eliminates a few routines which required looping over every site in the depth-two extended field
- ...though it's worth noting that depth-two extended fields are still needed as intermediates in many cases, so while this reduces the total number of launched threads it does not lead to a commensurate reduction in the depth of the temporary fields

**Gauge field compression**
- Enables recon-13 for the input gauge field, which is applicable for both the "U" field in the first level of ASQTAD smearing and the "W" field in the second level of ASQTAD smearing (as part of the overall HISQ force)
- Extends the usage of recon-13 to the MILC interface, based on the existing environment variables for enabling recon-13 for outer solves

**Testing and timing**
- This PR does non-trivial cleanup of unit testing and timing for the HISQ force
- In particular, it splits the test into two separate components which emulate the first step (Lepage and long link) and second step (traditional ASQTAD smearing only) of the HISQ force

**General clean-up of FLOPS/bytes counts**
- This PR both streamlines and fixes some bugs in the FLOPS/bytes counts in the sub-components of the HISQ force

There is still some remaining clean-up to be done in this PR, none of which block opening this PR sooner as opposed to later:

- [x] Updating the FLOPS/bytes counts in the CPU unit tests (straightforward, just needs to be copied from the GPU counts)
- [x] Cleaning up (or eliminating?) the "low memory" code path, which removes some of the intermediate fields at the expense of removing some kernel fusion
- [x] `clang-format`

With regards to gauge compression, there is still outstanding work to enable recon-12 for the second step of the HISQ force chain rule because the `U` field is an SU(3) field (as opposed to the first step featuring the U(3) `W` field). Since this corresponds to a marginal gain in memory savings (relative to enabling recon-13) at the expense of a decent amount of coding headaches, we're going to punt this to a subsequent PR.

As a function of the problem size _and_ the geometry of the decomposition, this can lead to a ~30+% performance boost in the computation of the HISQ force. This is due to the large amount of kernel fusion (and the corresponding cache reuse from it) as well as the introduction of gauge reconstruction for the `U`/`W` field and a reduction of the depth of the halo that needs to be explicitly looped over as part of the computation.


include/instantiate.h
```diff
@@ -410,6 +410,78 @@ namespace quda
...
   }
 
+  /**
+     @brief This instantiate function is used to instantiate combinations of reconstruct
+     and phase for pure-gauge routines using staggered phases
+     @param[in] U Gauge field
+     @param[in,out] args Additional arguments for kernels
+  */
+  template <template <typename, int, QudaReconstructType, QudaStaggeredPhase> class Apply, typename store_t, int nColor, typename G, typename... Args>
+  constexpr void instantiateGaugeHisq(G &U, Args &&...args)
```
maddyscientist: Given the description, perhaps this should be named `instantiateGaugeStaggered`.  There's nothing specifically HISQ here is there?

tests/utils/host_utils.h
```diff
@@ -246,6 +271,24 @@ inline double getTolerance(QudaPrecision prec)
   return 1.0;
 }
 
+/**
+  @brief Return the negative exponent of the tolerances defined in `getTolerance` above
+  @param[in] prec Precision
+  @return Exponent of the base-10 reasonably expected tolerance
+*/
+inline int getNegLog10Tolerance(QudaPrecision prec)
```
maddyscientist, line 262: I think you should rewrite `getTolerance` to use `getNegLog10Tolerance` (or vice versa), to ensure that these are always consistent.

lib/hisq_paths_force_quda.cu
```diff
@@ -11,304 +11,545 @@ namespace quda {
...
+    public:
+      OneLinkForce(Arg &arg, const GaugeField &link, const GaugeField &force) :
+        TunableKernel3D(link, 2, 4),
+        arg(arg),
+        force(force),
+        link(link)
+      {
+        char aux2[16];
+        strcat(aux, comm_dim_partitioned_string());
+        strcat(aux, ",threads=");
```
maddyscientist: The `threads` number isn't needed in the `tuneKey` anymore, it can be nuked.  This was previously used to lift the degeneracy between say an 8^4 single-GPU local volume and a (4 + 2 * R)^4, with R = 2, multi-GPU local volume.  Since we now have `comm_dim_partitioned_string` in the `tuneKey` the `threads` label can be safely nuked.

lib/hisq_paths_force_quda.cu
```diff
@@ -11,304 +11,545 @@ namespace quda {
...
-        case FORCE_SIDE_LINK_SHORT:    strcat(aux, ",SIDE_LINK_SHORT");    break;
-        default: errorQuda("Undefined force type %d", type);
+        arg.sig = sig.dim;
+        arg.mu = mu.dim;
+        arg.mu_next = mu_next.dim;
+        arg.compute_lepage = has_lepage ? 1 : 0;
+
+        char aux2[16];
+        strcat(aux, comm_dim_partitioned_string());
+        strcat(aux, ",threads=");
```
maddyscientist: nuke `threads`

lib/hisq_paths_force_quda.cu
```diff
@@ -11,304 +11,545 @@ namespace quda {
...
+              break;
+          }
+        } else {
+          arg.rho = -1;
         }
+
+        char aux2[16];
+        strcat(aux, comm_dim_partitioned_string());
+        strcat(aux, ",threads=");
+        u32toa(aux2, arg.threads.x);
```
maddyscientist: Nuke `threads`

lib/hisq_paths_force_quda.cu
```diff
@@ -350,90 +602,63 @@ namespace quda {
...
         force(force),
-        meta(meta),
-        type(type)
+        meta(meta)
       {
-        arg.sig = sig;
-        arg.mu = mu;
+        char aux2[16];
+        strcat(aux, comm_dim_partitioned_string());
+        strcat(aux, ",threads=");
```
maddyscientist: nuke `threads`

lib/hisq_paths_force_quda.cu
```diff
@@ -452,12 +677,66 @@ namespace quda {
...
+    public:
+      HisqCompleteLinkForce(Arg &arg, GaugeField &force, const GaugeField &meta) :
+        TunableKernel2D(meta, 2),
+        arg(arg),
+        force(force),
+        meta(meta)
+      {
+        char aux2[16];
+        strcat(aux, comm_dim_partitioned_string());
+        strcat(aux, ",threads=");
```
maddyscientist: Nuke `threads`

maddyscientist (commented):
Initial visual review, just a few minor quibbles.  Will have more when I've actually tested it.

maddyscientist:
~When building, I'm getting the following warning (CUDA 12.0 / GCC 11)~
```
/home/kate/github/quda-develop-old/tests/host_reference/hisq_force_reference.cpp: In instantiation of ‘su3_matrix* get_su3_matrix(int, su3_matrix*, int, int) [with su3_matrix = fsu3_matrix]’:
/home/kate/github/quda-develop-old/tests/host_reference/hisq_force_reference.cpp:109:37:   required from ‘void computeLinkOrderedOuterProduct(su3_vector*, su3_matrix*, size_t, int) [with su3_vector = fsu3_vector; su3_matrix = fsu3_matrix; size_t = long unsigned int]’
/home/kate/github/quda-develop-old/tests/host_reference/hisq_force_reference.cpp:118:35:   required from here
/home/kate/github/quda-develop-old/tests/host_reference/hisq_force_reference.cpp:87:63: warning: unused parameter ‘gauge_order’ [-Wunused-parameter]
   87 | template <typename su3_matrix> su3_matrix *get_su3_matrix(int gauge_order, su3_matrix *p, int idx, int dir)
   ```
   
Edit: this was caused by me using a stale local copy of the branch.  Error is not present in HEAD.

maddyscientist:
Something I just noticed from testing, is that the `hisq_paths_force_test` test will wrongly complain of failing if `--verify false` is pass.  The correctness check should not be applied in this case.

include/instantiate.h
weinbe2: Ultimately this will be for recon-18/13/12 (once I get that working), which is why I made it separate. I can add a comment to this sort, or a separate error path with a `recon-12`-specific error.

weinbe2:
> Something I just noticed from testing, is that the `hisq_paths_force_test` test will wrongly complain of failing if `--verify false` is pass. The correctness check should not be applied in this case.

https://github.com/lattice/quda/pull/1367/commits/9e9845e5827a35ac61349c89a718f0ad192014e7

include/instantiate.h
weinbe2: https://github.com/lattice/quda/pull/1367/commits/49435f4ff06d1e6ddd5d04699d2922e08a21e016 --- let me know if this is good

tests/utils/host_utils.h
weinbe2: Rewrote `getTolerance` as a function of `getNegLog10Tolerance` because `int` -> `double` seemed safer than the other way around. I checked the C++11+ standard and confirmed `int` will get promoted to `double` in this context. (I also confirmed it via a `ctest`.)

https://github.com/lattice/quda/pull/1367/commits/4722ed6db404269e99e7b9b6dc9d1b313bb4b4be

lib/hisq_paths_force_quda.cu
weinbe2: https://github.com/lattice/quda/pull/1367/commits/0f06ef4407130c0df053db25f74c7c7c3fe70e39

lib/hisq_paths_force_quda.cu
weinbe2: https://github.com/lattice/quda/pull/1367/commits/0f06ef4407130c0df053db25f74c7c7c3fe70e39

lib/hisq_paths_force_quda.cu
weinbe2: https://github.com/lattice/quda/pull/1367/commits/0f06ef4407130c0df053db25f74c7c7c3fe70e39

lib/hisq_paths_force_quda.cu
weinbe2: https://github.com/lattice/quda/pull/1367/commits/0f06ef4407130c0df053db25f74c7c7c3fe70e39

lib/hisq_paths_force_quda.cu
weinbe2: https://github.com/lattice/quda/pull/1367/commits/0f06ef4407130c0df053db25f74c7c7c3fe70e39

include/instantiate.h
maddyscientist: My point was here that the function name `instantiateGaugeHisq` isn't the best name, and something like `instantiateGaugeStaggered` would be more descriptive, since it instantiates templates for reconstruct and staggered phase (as well as color and precision).

maddyscientist (approved):
Thanks the fixes @weinbe2.  Aside from one minor [quibble](https://github.com/lattice/quda/pull/1367#discussion_r1149992456), this looks good to go.  We can merge this once that's taken care of.

maddyscientist:
@mathiaswagner are you wanting to review this before we merge?

include/instantiate.h
weinbe2: Ah, okay, understood

weinbe2: https://github.com/lattice/quda/pull/1367/commits/26d81fcf5af38cab8913e5ca750be024c98eed84

mathiaswagner:
I guess this has seen enough testing and I am not sure I'll have cycles this week so feel free to go ahead with the merge.

PR #1368, opened 2023-03-17
Feature/hisq fgi

cpviolator:
This WIP add functionality to allow for offload in the FGI RHMC routines in https://github.com/callat-qcd/milc_qcd/tree/feature/nested-force-gradient-integrator. It simply creates copy gauge and momentum fields consistent with MILC's QUDA offload sequences.

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -87,6 +87,8 @@ extern "C" {
     int return_result_gauge; /**< Return the result gauge field */
     int return_result_mom;   /**< Return the result momentum field */
 
+    int use_fgi;             /**< add FGI terms */
+
     size_t gauge_offset; /**< Offset into MILC site struct to the gauge field (only if gauge_order=MILC_SITE_GAUGE_ORDER) */
     size_t mom_offset; /**< Offset into MILC site struct to the momentum field (only if gauge_order=MILC_SITE_GAUGE_ORDER) */
     size_t site_size; /**< Size of MILC site struct (only if gauge_order=MILC_SITE_GAUGE_ORDER) */
@@ -1366,6 +1368,18 @@ extern "C" {
    * @param[in] param The parameters of the external field
    */
   void momResidentQuda(void *mom, QudaGaugeParam *param);
+
+  /**
+   * For the FGI HMC step
+   * make a copy of the gauge field
+   * restore the gauge field
+   * Zero out the momResident field (after making a copy)
+   * and restore the momResident field
+   */
+  void copyGaugeQuda();
+  void restoreGaugeQuda();
+  void momResidentZeroQuda();
+  void momResidentRestoreQuda();
 
   /**
    * Compute the gauge force and update the momentum field
--- a/lib/check_params.h
+++ b/lib/check_params.h
@@ -124,6 +124,7 @@ void printQudaGaugeParam(QudaGaugeParam *param) {
   P(use_resident_mom, 0);
   P(make_resident_gauge, 0);
   P(make_resident_mom, 0);
+  P(use_fgi, 1);// FGI: This needs to be 0, and set to 1 with MILC/QUDA interface
   P(return_result_gauge, 1);
   P(return_result_mom, 1);
   P(gauge_offset, 0);
@@ -135,6 +136,7 @@ void printQudaGaugeParam(QudaGaugeParam *param) {
   P(use_resident_mom, INVALID_INT);
   P(make_resident_gauge, INVALID_INT);
   P(make_resident_mom, INVALID_INT);
+  P(use_fgi, INVALID_INT);// FGI: This needs to be 0, and set to 1 with MILC/QUDA interface
   P(return_result_gauge, INVALID_INT);
   P(return_result_mom, INVALID_INT);
   P(gauge_offset, (size_t)INVALID_INT);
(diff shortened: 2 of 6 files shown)


lib/check_params.h
```diff
@@ -124,6 +124,7 @@ void printQudaGaugeParam(QudaGaugeParam *param) {
   P(use_resident_mom, 0);
   P(make_resident_gauge, 0);
   P(make_resident_mom, 0);
+  P(use_fgi, 1);// FGI: This needs to be 0, and set to 1 with MILC/QUDA interface
```
weinbe2, line 127: Should be set back to `0`, I'm guessing?

lib/interface_quda.cpp
```diff
@@ -104,6 +103,32 @@ CloverField *cloverEigensolver = nullptr;
 cudaGaugeField *momResident = nullptr;
 cudaGaugeField *extendedGaugeResident = nullptr;
 
+// These Copy fields are to support the Force Gradient Integrator
```
weinbe2, line 106: I'm not sure if I like the term "Copy"... maybe "Backup" would be a better term? "Previous"? I guess that would be best informed by the language of the FGI algorithm --- and if "copy" is the common language, that's good enough to me. (This also isn't a hill I'll die on!)

lib/interface_quda.cpp
```diff
@@ -2086,8 +2142,11 @@ quda::cudaGaugeField *checkGauge(QudaInvertParam *param)
                                     param->cuda_prec_refinement_sloppy, param->cuda_prec_eigensolver};
       QudaReconstructType recon[4] = {gaugeSloppy->Reconstruct(), gaugePrecondition->Reconstruct(),
                                       gaugeRefinement->Reconstruct(), gaugeEigensolver->Reconstruct()};
+
+      printfQuda("QudaInvertParam has different precisions than loadGaugeQuda, so recomputing sloppy fields\n");
```
weinbe2, line 2146: I like the addition of this for logging reasons---I'd put this in a `logQuda(QUDA_VERBOSE, [...]);`, I think.

lib/interface_quda.cpp
```diff
@@ -2120,8 +2179,10 @@ quda::cudaGaugeField *checkGauge(QudaInvertParam *param)
       // recon is always no for fat links, so just use long reconstructs here
       QudaReconstructType recon[4] = {gaugeLongSloppy->Reconstruct(), gaugeLongPrecondition->Reconstruct(),
                                       gaugeLongRefinement->Reconstruct(), gaugeLongEigensolver->Reconstruct()};
+      printfQuda("QudaInvertParam has different precisions than loadGaugeQuda, so recomputing sloppy fields\n");
```
weinbe2, line 2182: Same as above

lib/interface_quda.cpp
```diff
@@ -4286,6 +4347,262 @@ void momResidentQuda(void *mom, QudaGaugeParam *param)
...
+  profileGaugeForce.TPSTOP(QUDA_PROFILE_TOTAL);
+}
+
+void copyGaugeQuda()
+{ // FGI make a copy of the precise gauge field on device
+  profileGaugeForce.TPSTART(QUDA_PROFILE_TOTAL);
+
+  // Wilson Links
+  //----------------------------------------------------------
+  if(gaugePrecise) {
```
weinbe2, line 4368: Just confirming --- it's intentional that this doesn't throw an `errorQuda` if `gaugePrecise` is unset, right? I see that it's fully safe code on its own.

lib/interface_quda.cpp
```diff
@@ -4286,6 +4347,262 @@ void momResidentQuda(void *mom, QudaGaugeParam *param)
...
+  profileGaugeForce.TPSTOP(QUDA_PROFILE_TOTAL);
+}
+
+void restoreGaugeQuda()
+{ // FGI restore Gauge field
+  profileGaugeForce.TPSTART(QUDA_PROFILE_TOTAL);
+
+  // Wilson Links
+  //----------------------------------------------------------
+  if(gaugePrecise) gaugePrecise->copy(*gaugeCopyPrecise);
```
weinbe2, line 4520: `if (gaugePrecise && gaugeCopyPrecise)`

lib/interface_quda.cpp
```diff
@@ -1381,6 +1406,37 @@ void endQuda(void)
 
   if(momResident) delete momResident;
 
+  // Delete MILC FGI fields
```
weinbe2, line 1409: I know we discussed moving these into `freeUniqueGauge`, etc, offline --- just adding this comment so it can be marked as "resolved" at some point.

include/quda.h
```diff
@@ -87,6 +87,8 @@ extern "C" {
     int return_result_gauge; /**< Return the result gauge field */
     int return_result_mom;   /**< Return the result momentum field */
 
+    int use_fgi;             /**< add FGI terms */
```
maddyscientist, line 90: Perhaps this should be a `bool` instead?  (Note that I'm going to kill `QudaBoolean` at some point, and replace it with `bool` - this is fine since `bool` is supported on C99 which we require anyway for the QUDA C interface.)

include/quda.h
```diff
@@ -1373,6 +1375,18 @@ extern "C" {
    */
   void momResidentQuda(void *mom, QudaGaugeParam *param);
 
+  /**
```
maddyscientist, line 1378: Some punctuation in this comment might be nice :P

maddyscientist:
Nothing controversial in this commit.  One thought on the design used here: would it be a better fit to `push` and `pop` gauge / momentum fields.  E.g., instead of backing up the resident fields and zeroing them, and the later restoring them, we just `push` a new field (which pushes the resident field onto the proverbial stack) and when we're done with this temporary, we simply `pop` it, freeing the temporary and restoring the prior resident field off of the stack.

Just a thought.

Issue #1341, opened 2022-11-23
Add appropriate `const` modifiers to member functions of the shared memory cache


weinbe2:
Many of the member functions of the shared memory cache helper ( https://github.com/lattice/quda/blob/develop/include/targets/cuda/shared_memory_cache_helper.h ) lack `const` modifiers, which leads to compiler errors when a shared memory cache is passed to a function via a constant reference.

This isn't a blocking issue as it's easy to work around, but it would be a good quality-of-life/best practices update nonetheless.

PR #1369, opened 2023-04-01
Hotfix/stackframe

maddyscientist:
This PR primarily serves to remove the induced stack frame in a number of kernels
* Use flavor/chirality swizzle trick in non-degenerate Twisted-mass preconditioned clover application kernel
* Laplace and Staggered quark smearing kernel with NVSHMEM (lack of `__force_inline__`)
* Use `SharedMemoryCache` to act as virtual registers (Symanzik improved Wilson-flow and STOUT kernels)
  * In doing so we add support for multiple concurrent dynamic cache objects through use of an offset to the base shared-memory pointer
* Reordering of clover derivative kernel to reduce register pressure

Some other minor changes:
* Fix for HIP compilation which was broken by #1364 
* Remove some unused legacy `Matrix` functions

### DIFF
--- a/include/kernels/dslash_clover_helper.cuh
+++ b/include/kernels/dslash_clover_helper.cuh
@@ -4,6 +4,7 @@
 #include <clover_field_order.h>
 #include "color_spinor.h"
 #include <linalg.cuh>
+#include "shared_memory_cache_helper.h"
 #include "kernel.h"
 
 namespace quda {
@@ -44,7 +45,8 @@ namespace quda {
     CloverArg(ColorSpinorField &out, const ColorSpinorField &in, const CloverField &clover,
 	      int parity, real kappa=0.0, real mu=0.0, real epsilon = 0.0,
 	      bool dagger = false, QudaTwistGamma5Type twist = QUDA_TWIST_GAMMA5_INVALID) :
-      kernel_param(dim3(in.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ? in.VolumeCB()/2 : in.VolumeCB(), in.SiteSubset())),
+      kernel_param(dim3(in.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ? in.VolumeCB()/2 : in.VolumeCB(),
+                        in.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ? 2 : 1, in.SiteSubset())),
       out(out), in(in),
       clover(clover, inverse && !dynamic_clover && twist == QUDA_TWIST_GAMMA5_INVALID), // only inverse if non-twisted clover and !dynamic
       cloverInv(clover, !dynamic_clover), // only inverse if !dynamic
@@ -90,7 +92,7 @@ namespace quda {
     constexpr CloverApply(const Arg &arg) : arg(arg) {}
     static constexpr const char* filename() { return KERNEL_FILE; }
 
-    __device__ __host__ inline void operator()(int x_cb, int parity)
+    __device__ __host__ inline void operator()(int x_cb, int, int parity)
     {
       using namespace linalg; // for Cholesky
       int clover_parity = arg.nParity == 2 ? parity : arg.parity;
@@ -133,7 +135,7 @@ namespace quda {
     constexpr TwistCloverApply(const Arg &arg) : arg(arg) {}
     static constexpr const char* filename() { return KERNEL_FILE; }
 
-    __device__ __host__ inline void operator()(int x_cb, int parity)
+    __device__ __host__ inline void operator()(int x_cb, int, int parity)
     {
       using namespace linalg; // for Cholesky
       int clover_parity = arg.nParity == 2 ? parity : arg.parity;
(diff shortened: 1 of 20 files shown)


kostrzewa:
Doesn't seem to break anything on 8xP100 in the tmLQCD HMC.

include/kernels/dslash_clover_helper.cuh
```diff
@@ -184,63 +186,68 @@ namespace quda {
...
+      SharedMemoryCache<half_fermion> cache(target::block_dim());
 
+      half_fermion in_chi[n_flavor]; // flavor array of chirally projected fermion
 #pragma unroll
-        for (int flavor = 0; flavor < n_flavor; flavor++) {
-          out_chi[flavor] = A * in_chi[flavor];
-          out_chi[flavor] += (flavor == 0 ? a : -a) * in_chi[flavor];
-          out_chi[flavor] += arg.b * in_chi[1 - flavor];
-        }
+      for (int i = 0; i < n_flavor; i++) in_chi[i] = in.chiral_project(i);
```
weinbe2, line 210: Should we put `#pragma unroll` here just for full diligence? I know this loop should be more than simple enough for a compiler to unroll, but we may as well give it more information if we can.

include/kernels/dslash_clover_helper.cuh
```diff
@@ -184,63 +186,68 @@ namespace quda {
...
-        }
+      auto swizzle = [&](half_fermion x[2], int chirality, int reverse) {
+        if (chirality == 0) cache.save_y(x[1], reverse);
+        else                cache.save_y(x[0], 1 - reverse);
+        cache.sync();
+        if (chirality == 0) x[1] = cache.load_y(1 - reverse);
+        else                x[0] = cache.load_y(reverse);
+      };
 
+      swizzle(in_chi, chirality, 0); // apply the flavor-chirality swizzle between threads
```
weinbe2: Thoughts on creating a file-scope `enum` that's `FORWARD`/`BACKWARD` or something of the sort that's more explicit than `0` and `1`?

weinbe2: For the future

include/kernels/gauge_stout.cuh
```diff
@@ -146,7 +148,7 @@ namespace quda
       // Compute Omega_{mu}=[Sum_{mu neq nu}rho_{mu,nu}C_{mu,nu}]*U_{mu}^dag
       //-------------------------------------------------------------------
       // Compute \rho * staple_coeff * S - \rho * rectangle_coeff * R
-      Q = ((arg.staple_coeff * Stap) - (arg.rectangle_coeff * Rect)) * conj(U);
+      Q = ((arg.staple_coeff * static_cast<const Link &>(Stap)) - (arg.rectangle_coeff * static_cast<const Link &>(Rect))) * conj(U);
```
weinbe2, line 151: Pedantic, but should the `static_cast` be a `reinterpret_cast`? (Also, as a point of future work, is there a way to get rid of the need for the cast outside of a full expression template engine? I can see how multiplying by a scalar, on face value, requires the explicit cast.)

include/kernels/gauge_wilson_flow.cuh
```diff
@@ -60,20 +60,21 @@ namespace quda
...
       break;
     case QUDA_GAUGE_SMEAR_SYMANZIK_FLOW :
       // This function gets stap = S_{mu,nu} i.e., the staple of length 3,
       // and the 1x2 and 2x1 rectangles of length 5. From the following paper:
       // https://arxiv.org/abs/0801.1165
+      SharedMemoryCache<Link> Stap(target::block_dim());
+      SharedMemoryCache<Link> Rect(target::block_dim(), sizeof(Link)); // offset to ensure non-overlapping allocations
       computeStapleRectangle(arg, x, arg.E, parity, dir, Stap, Rect, Arg::wflow_dim);
-      Z = (arg.coeff1x1 * Stap + arg.coeff2x1 * Rect);
+      Z = arg.coeff1x1 * static_cast<const Link &>(Stap) + arg.coeff2x1 * static_cast<const Link &>(Rect);
```
weinbe2, line 77: `reinterpret_cast`?

include/quda_matrix.h
```diff
@@ -671,91 +671,29 @@ namespace quda {
...
-    __device__ __host__ inline
-    void setIdentity(Matrix<double2,N>* m){
-
+   __device__ __host__ inline void setIdentity(Matrix<T,N>* m)
+  {
 #pragma unroll
-      for (int i=0; i<N; ++i){
-        (*m)(i,i) = make_double2(1,0);
+    for (int i=0; i<N; ++i){
+      (*m)(i,i) = 1;
```
weinbe2: I'm surprised this works in all cases; in particular, I don't think you can initialize a `std::complex<float/double>` with an integer... but I guess we'll cross that bridge if/when we get there.

include/targets/cuda/shared_memory_cache_helper.h
```diff
@@ -84,12 +100,22 @@ namespace quda
       }
     };
 
-    template <bool dynamic_shared> __device__ __host__ inline std::enable_if_t<dynamic_shared, atom_t *> cache()
+    template <bool dynamic_shared> __device__ __host__ inline std::enable_if_t<dynamic_shared, atom_t*> cache()
```
weinbe2: Can this have `const` applied to it (i.e., `[...] cache() const`)? Considering the object returned can be modified but it doesn't actually modify anything within this class. Or, alternatively, is the function below all that's needed?

The reason I ask is so we can potentially close this issue: https://github.com/lattice/quda/issues/1341

include/targets/cuda/shared_memory_cache_helper.h
```diff
@@ -256,6 +285,67 @@ namespace quda
...
+       @brief Cast operator to allow cache objects to be used where T
+       is expected
+     */
+    __device__ __host__ operator T() const { return load(); }
+
+    /**
+       @brief Assignment operator to allow cache objects to be used on
+       the lhs where T is otherwise expected.
+     */
+    __device__ __host__ void operator=(const T& src) { save(src); }
```
weinbe2: Can this also have `const` applied to it?

lib/gauge_wilson_flow.cu
```diff
@@ -17,11 +17,14 @@ namespace quda {
     const QudaGaugeSmearType wflow_type;
     const WFlowStepType step_type;
 
-    bool tuneSharedBytes() const { return false; }
     unsigned int minThreads() const { return in.LocalVolumeCB(); }
-    unsigned int maxBlockSize(const TuneParam &) const { return 32; }
-    int blockStep() const { return 8; }
-    int blockMin() const { return 8; }
+    unsigned int maxSharedBytesPerBlock() const { return maxDynamicSharedBytesPerBlock(); }
```
weinbe2: Should this be dependent on `wflow_type == QUDA_GAUGE_SMEAR_SYMANZIK_FLOW` as well?

include/quda_matrix.h
maddyscientist: This isn't new code, I just deleted some unused routines.  But note that we use `quda::complex` not `std::complex`.

include/kernels/gauge_wilson_flow.cuh
maddyscientist: discussed offline: `static_cast` is correct here.

include/kernels/gauge_stout.cuh
maddyscientist: discussed offline: `static_cast` is correct here.

include/kernels/dslash_clover_helper.cuh
maddyscientist: acc2681e59785b58ce3e7a50acd2b6f697854540

include/targets/cuda/shared_memory_cache_helper.h
maddyscientist: 4334841b9ced537246a6d2a4565ae283ac291055

include/targets/cuda/shared_memory_cache_helper.h
maddyscientist: 4334841b9ced537246a6d2a4565ae283ac291055

lib/gauge_wilson_flow.cu
maddyscientist: b4317464f0a7be097777a50d6afd723b964b9e4e

weinbe2 (approved):
Looks great, thanks @maddyscientist ! Pending Jenkins completing this is good to merge.

PR #1370, opened 2023-04-05
Add "-lstdc++" to c_interface_test

hummingtree:
On Summit with CUDA 11.0.3 and GCC 9.1.0, when `c_interface_test` is linked against `libquda_test.so` it gives the following error:
```
libquda_test.so: undefined reference to `std::__cxx11::basic_stringstream<char, std::char_traits<char>, std::allocator<char> >::basic_stringstream()@GLIBCXX_3.4.26'
libquda_test.so: undefined reference to `std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::basic_ostringstream()@GLIBCXX_3.4.26'
```

Adding the "-lstdc++" flag fixes it.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -47,7 +47,7 @@ endif()
 
 # define tests
 add_executable(c_interface_test c_interface_test.c)
-target_link_libraries(c_interface_test ${TEST_LIBS})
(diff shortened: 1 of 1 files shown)


maddyscientist (approved):

maddyscientist:
Noting we may want to revisit in this future if we are considering different STL implementations.

PR #1374, opened 2023-04-19
Update device.cpp

### LINKED ISSUE #1373: Typo in macro in develop branch, file  lib/targets/hip/device.cpp prevents HIP compilation
Here is the fix

diff --git a/lib/targets/hip/device.cpp b/lib/targets/hip/device.cpp
index 38afb0d09..21366628c 100644
--- a/lib/targets/hip/device.cpp
+++ b/lib/targets/hip/device.cpp
@@ -54,7 +54,7 @@ namespace quda
     void init_thread()
     {
       if (device_id == -1) errorQuda("No HIP device has been initialized for this process");
-      CHECK_CUDA_ERROR(hipSetDevice(device_id));
+      CHECK_HIP_ERROR(hipSetDevice(device_id));
     }
 
     int get_device_count()

maddyscientist:
Thanks to @detar for the fix.  Closes #1373

### DIFF
--- a/lib/targets/hip/device.cpp
+++ b/lib/targets/hip/device.cpp
@@ -54,7 +54,7 @@ namespace quda
     void init_thread()
     {
       if (device_id == -1) errorQuda("No HIP device has been initialized for this process");
-      CHECK_CUDA_ERROR(hipSetDevice(device_id));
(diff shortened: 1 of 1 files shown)


bjoo (approved):
Trivial change, but I checked the code built anyway. Looks good.

PR #1382, opened 2023-05-24
CMakeLists.txt upgraded for Spack

bjoo:
Setting `MPI_XXX_COMPILERS` from `MPICXX`, `MPICC` etc.environment variables breaks Spack based builds.All I have done here is if one adds -DQUDA_SPACK_BUILD to CMake it won't look at those env variables


maddyscientist (approved):

mathiaswagner:
Thanks. I guess the whole logic there is more a legacy thing from the last decade. Would hope we can remove all of this but never bothered so far.

PR #1383, opened 2023-05-24
Sniff-testing ROCm build gh action workflow

dmcdougall:
Don't merge this yet, I'm still ironing out the mechanics.

### DIFF
--- a/.github/workflows/amd-build-ci.yml
+++ b/.github/workflows/amd-build-ci.yml
@@ -1,0 +1,57 @@
+name: amd-build-ci
+run-name: ${{ github.actor }} is kicking off a ROCm build
+on: pull_request
+jobs:
+  rocm-build:
+    runs-on: [self-hosted, amd]
+    steps:
+      - uses: actions/checkout@v3
+      - run: |
+          export ROCM_PATH=/opt/rocm-5.5.0
+          export PATH=${ROCM_PATH}/bin:${ROCM_PATH}/llvm/bin:${PATH}
+          SRCROOT=`pwd`
+          BUILDROOT=`mktemp -d build-XXXXXXXX`
+          INSTALLROOT=`mktemp -d install-XXXXXXXX`
+          QUDA_GPU_ARCH=gfx90a
+          cmake ${SRCROOT} \
+            -B ${BUILDROOT} \
+            -DQUDA_TARGET_TYPE="HIP" \
+            -DQUDA_GPU_ARCH=${QUDA_GPU_ARCH} \
+            -DROCM_PATH=${ROCM_PATH} \
+            -DQUDA_DIRAC_CLOVER=ON \
+            -DQUDA_DIRAC_CLOVER_HASENBUSCH=OFF \
+            -DQUDA_DIRAC_DOMAIN_WALL=OFF \
+            -DQUDA_DIRAC_NDEG_TWISTED_MASS=OFF \
+            -DQUDA_DIRAC_STAGGERED=ON \
+            -DQUDA_DIRAC_TWISTED_MASS=OFF \
+            -DQUDA_DIRAC_TWISTED_CLOVER=OFF \
+            -DQUDA_DIRAC_WILSON=ON \
+            -DQUDA_CLOVER_DYNAMIC=ON \
+            -DQUDA_FORCE_HISQ=ON \
+            -DQUDA_QDPJIT=OFF \
+            -DQUDA_INTERFACE_QDPJIT=OFF \
+            -DQUDA_INTERFACE_MILC=ON \
+            -DQUDA_INTERFACE_CPS=OFF \
+            -DQUDA_INTERFACE_QDP=ON \
+            -DQUDA_INTERFACE_TIFR=OFF \
+            -DQUDA_QMP=ON \
+            -DQUDA_DOWNLOAD_USQCD=ON \
+            -DQUDA_OPENMP=OFF \
+            -DQUDA_MULTIGRID=ON \
+            -DQUDA_MAX_MULTI_BLAS_N=9 \
+            -DQUDA_DOWNLOAD_EIGEN=ON \
+            -DQUDA_PRECISION=8 \
+            -DCMAKE_INSTALL_PREFIX=${INSTALLROOT} \
+            -DCMAKE_BUILD_TYPE="DEBUG" \
+            -DCMAKE_CXX_COMPILER="${ROCM_PATH}/llvm/bin/clang++" \
+            -DCMAKE_C_COMPILER="${ROCM_PATH}/llvm/bin/clang" \
+            -DCMAKE_HIP_COMPILER="${ROCM_PATH}/llvm/bin/clang++" \
+            -DBUILD_SHARED_LIBS=ON \
+            -DQUDA_BUILD_SHAREDLIB=ON \
+            -DQUDA_BUILD_ALL_TESTS=ON \
+            -DQUDA_CTEST_DISABLE_BENCHMARKS=ON \
+            -DCMAKE_C_STANDARD=99
+          cmake --build ${BUILDROOT} -j 16
+          cmake --install ${BUILDROOT}
+          rm -rf ${BUILDROOT}
+          rm -rf ${INSTALLROOT}
--- a/.github/workflows/rocm-build-ci.yml
+++ b/.github/workflows/rocm-build-ci.yml
@@ -1,4 +1,4 @@
-name: amd-build-ci
+name: rocm-build-ci
 run-name: ${{ github.actor }} is kicking off a ROCm build
 on: pull_request
 jobs:
--- a/.github/workflows/rocm-build-ci.yml
+++ b/.github/workflows/rocm-build-ci.yml
@@ -40,7 +40,7 @@ jobs:
             -DQUDA_DOWNLOAD_EIGEN=ON \
             -DQUDA_PRECISION=14 \
             -DCMAKE_INSTALL_PREFIX=${INSTALLROOT} \
-            -DCMAKE_BUILD_TYPE="RELEASE" \
+            -DCMAKE_BUILD_TYPE="DEVEL" \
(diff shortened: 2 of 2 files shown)


mathiaswagner:

mathiaswagner:

dmcdougall:
Can someone approve this so the workflow runs?  I think I have a recipe sorted out now, and the action triggers on my fork.  Just want to make sure it runs correctly here too.

dmcdougall:
Nice, the build passed!  Feedback welcome, but I'm happy for this to be merged now.

mathiaswagner:
Question as I have not looked it up yet: Could we do target specific subdirectories in the `workflows` directory ? Like '.github/workflows/hip/...yaml' ?

dmcdougall:
It seems Technically Doable but [quite awkward](https://github.com/orgs/community/discussions/18055).

A sensible compromise might be to simply come up with a naming convention for the filenames:

`<target>-<id1>-<id2>-...-<idN>.yml`

I'm open to other ideas.

.github/workflows/amd-build-ci.yml
```diff
@@ -0,0 +1,57 @@
...
+            -DQUDA_INTERFACE_CPS=OFF \
+            -DQUDA_INTERFACE_QDP=ON \
+            -DQUDA_INTERFACE_TIFR=OFF \
+            -DQUDA_QMP=ON \
+            -DQUDA_DOWNLOAD_USQCD=ON \
+            -DQUDA_OPENMP=OFF \
+            -DQUDA_MULTIGRID=ON \
+            -DQUDA_MAX_MULTI_BLAS_N=9 \
+            -DQUDA_DOWNLOAD_EIGEN=ON \
+            -DQUDA_PRECISION=8 \
```
maddyscientist: Depending on how much coverage you want, this would ideally be `14` (double, single and half precision), but if you only want one precision to be enabled `4` is best (this is pure single, for which nearly all kernels are compiled for)

.github/workflows/amd-build-ci.yml
```diff
@@ -0,0 +1,57 @@
...
+            -DQUDA_DIRAC_CLOVER=ON \
+            -DQUDA_DIRAC_CLOVER_HASENBUSCH=OFF \
+            -DQUDA_DIRAC_DOMAIN_WALL=OFF \
+            -DQUDA_DIRAC_NDEG_TWISTED_MASS=OFF \
+            -DQUDA_DIRAC_STAGGERED=ON \
+            -DQUDA_DIRAC_TWISTED_MASS=OFF \
+            -DQUDA_DIRAC_TWISTED_CLOVER=OFF \
+            -DQUDA_DIRAC_WILSON=ON \
+            -DQUDA_CLOVER_DYNAMIC=ON \
+            -DQUDA_FORCE_HISQ=ON \
```
maddyscientist: This flag is deprecated and doesn't do anything now.

.github/workflows/amd-build-ci.yml
```diff
@@ -0,0 +1,57 @@
...
+            -DQUDA_INTERFACE_QDPJIT=OFF \
+            -DQUDA_INTERFACE_MILC=ON \
+            -DQUDA_INTERFACE_CPS=OFF \
+            -DQUDA_INTERFACE_QDP=ON \
+            -DQUDA_INTERFACE_TIFR=OFF \
+            -DQUDA_QMP=ON \
+            -DQUDA_DOWNLOAD_USQCD=ON \
+            -DQUDA_OPENMP=OFF \
+            -DQUDA_MULTIGRID=ON \
+            -DQUDA_MAX_MULTI_BLAS_N=9 \
```
maddyscientist: This flag is mostly pointless now, and the default setting is sufficient.  This will just add unnecessary build time.

.github/workflows/amd-build-ci.yml
```diff
@@ -0,0 +1,57 @@
...
+            -DQUDA_DOWNLOAD_EIGEN=ON \
+            -DQUDA_PRECISION=8 \
+            -DCMAKE_INSTALL_PREFIX=${INSTALLROOT} \
+            -DCMAKE_BUILD_TYPE="DEBUG" \
```
maddyscientist: For our CUDA-targeted builds we use `STRICT` here, since that treats warnings as errors.  Whether that is ok to do on ROCm, I don't know.

maddyscientist (changes requested):
Looks good @dmcdougall.  Beyond the comments I've left, the only other change is to maybe change the name of your yml file, maybe to `rocm-build-ci.yml`.  I guess in the future we may have different targets or versions; we can decide that point whether to have different files for each (with a more descriptive filename) or to have all `rocm` builds done in this one file, etc.

Thank you for putting this together.

.github/workflows/amd-build-ci.yml
dmcdougall: Thanks, that's a good idea.  We're emitting warnings right now; unused compiler flags.  I'm not sure why and I need to investigate it.  I'd like to build `STRICT` in principle as we shouldn't have any compiler warnings at all.  For now, we can leave this as `DEBUG` and I'll definitely revisit once the warnings are addressed.

.github/workflows/amd-build-ci.yml
dmcdougall: Updated.

.github/workflows/amd-build-ci.yml
dmcdougall: Removed.

.github/workflows/amd-build-ci.yml
dmcdougall: Removed.

dmcdougall:
> maybe change the name of your yml file, maybe to `rocm-build-ci.yml`. 

Done.

> I guess in the future we may have different targets or versions; we can decide that point whether to have different files for each (with a more descriptive filename) or to have all `rocm` builds done in this one file, etc.

Right now there's a limitation of [one github workflow per yaml file](https://github.com/orgs/community/discussions/25342), but changing the filename doesn't change the name of the workflow, so I realised I had to update both the name of the file and the name of the workflow _in_ the file.

I agree with what you're saying.  This is something we'll just have to iterate on as github actions features evolve.

maddyscientist:
Maybe a DEVELOP build would be best.  All warnings will be enabled so they can be seen in the CI logs m.  A RELEASE build will hide all warnings which is not what we want for CI.

What are the warnings cause badness with STRICT build?

dmcdougall:
So I'm little confused on the build types.  I'm experimenting locally, and I'm expecting I'll probably need to open another issue to track some of these:

- DEBUG builds fail for single precision
  - Some kernels allocate too much LDS
  - Not sure why, it's possible that the compiler does no inlining which causes LDS bloat and hence the compiler failure
  - I can see in the error message that the failed compile was for gfx803, but I never asked for a gfx803 build.  I'd need to look at the build system to understand when and how CMake consumes HIP architectures
- RELEASE builds pass for all precisions
  - This is good, since it probably tells me that the LDS failure above is linked to the optimisation level passed to the compiler.  That's something we can report internally and get addressed
- STRICT builds fail because ROCm is currently warning about an upcoming header file reorganisation
  - If you want this fixed at the QUDA level today, we just need to point to `/opt/rocm/include` instead of `/opt/rocm/<component>/include`
  - It's not clear to me whether we can control all of these; some components are picked up by CMake and not controlled by the user directly
  - Those CMake elements will be addressed in the appropriate ROCm release

I'm trying a DEVELOP build now.

maddyscientist:
Debug: no idea, I guess that’s for you to work out 🤷‍♀️ 

Develop, release and strict should be mostly equivalent in terms of code generation I think: strict has -Werror and it also explicitly disabled one warning on clang about lack of loop unrolling which is unavoidable it seems; release hides all warnings.  With a strict build is there a compiler flag that can be passed to disable this header related warning?

dmcdougall:
> Maybe a DEVELOP build would be best

I tried this and realised it is a typo.  The build type is `DEVEL`.

> Debug: no idea, I guess that’s for you to work out 🤷‍♀️

Yes.

> With a strict build is there a compiler flag that can be passed to disable this header related warning?

No.  And it's not the only warning, it's just the first one I hit which caused the build to fail which means I didn't see any ones that may have come after the first one.  A DEVEL build shows there's a litany of other compiler warnings related to failed loop unrolls, and various other things, but it does compile at least.  This will give us something to base further changes off of.  I'd like to be able to address all of those warnings.

Are you expecting a warning-free build for all of the red, blue, and green team architectures?

maddyscientist:
For the clang loop unrolling warning, you can grab the incantation to disable it from the cuda cmake target file.

> Are you expecting a warning-free build for all of the red, blue, and green team architectures?

Yes, that’s the expectation and what we have for green team.  I accept that compiler warning disablement maybe needed for a given compiler if the issue can’t be dealt with in the code (eg clang loop unrolling).

maddyscientist:
And sorry, I always mix up `DEVEL` build with the `develop` branch name 😖

dmcdougall:
> I accept that compiler warning disablement maybe needed for a given compiler if the issue can’t be dealt with in the code (eg clang loop unrolling).

Thanks.  You beat me to the punch :)

Ok, I think I've addressed all your feedback here.  Let me know if you have more comments.

maddyscientist (approved):

maddyscientist:
@Jenkins ok to test

jxy:
Are you getting spams of my omp branch failing to bulid? How do I turn it off for the omp branch only?

dmcdougall:
@maddyscientist Thanks for your help!

@jxy I haven't looked at your branch.  Your build log is here: https://github.com/lattice/quda/actions/runs/5248198720/jobs/9479356861

jxy:
@dmcdougall It's not the issue with HIP/ROCM. For now, I can't guarantee my branch won't break other backends. I would like to turn the github workflow off for the omp branch, so that it doesn't unnecessary do the build test. How do I turn it off for the specific branch only?

PR #1384, opened 2023-05-26
Feature/gaugefield unity

maddyscientist:
This PR's primary purpose is to unify the `cpuGaugeField` and `cudaGaugeField` classes, along with an assortment of other cleanups.
* Removed `cpuGaugeField` and `cudaGaugeField` classes, we have now only `GaugeField`
  * Similar to `ColorSpinorField` the communiciation routines for host and device are still separate within the `GaugeField` class, unification of these is deferred until later
  * Added move and copy constructors and assignment operators for `GaugeField` to facilitate significant cleanup
  * Raw field pointer is now obtained using the `data()` method.  This is to align with STL design.  Though here we also have an optional template parameter that can be used to perform a cast to the desired type.  (Also applied to `ColorSpinorField` and `CloverField`)
* Introduced a smart pointer class `quda_ptr`, which provides heterogenous memory allocation
  * Deployed in `GaugeField`, `ColorSpinorField` and `CloverField`
* Introduced `qudaMemcpy` and `qudaMemset` functions that accept a `quda_ptr`
  * This allows for unified calls to memcpy and memset without needing to worry about the location of the field
* Introduced the concept of a profile stack, with `pushProfile` and `popProfile` functions, together with `getProfile` for getting a reference to the current top of stack profile.
   * This facilitates a significant cleanup of the profiling, we can now profile in arbitrary functions without needing to pass the profile to it explicitly
* Profiles now no longer complain about being double started: we now have a ref counter to allow things to just work (as long as the number of starts and stops match)
* Significant cleanup of the interface for the non-solver functions to utilize the technology introduced as part of this PR
* Jitify
  * Updated the version of Jitify included with QUDA
  * Some fixes for issues that creeped in (e.g., moved some function definitions out of headers to cpp files to avoid the run-time compiler from parsing them)
  

**Outstanding items**
- [x]  Add `quda_ptr` for ROCm
- [x]  Apply clang-format


lib/coarsecoarse_op_mma.in.cu
```diff
@@ -40,10 +40,10 @@ namespace quda {
       } else {
         GaugeFieldParam param(X);
         param.order = order;
-        output = cudaGaugeField::Create(param);
+        output = new GaugeField(param);
         if (copy_content) output->copy(X);
       }
-      return static_cast<cudaGaugeField *>(output);
+      return static_cast<GaugeField *>(output);
```
hummingtree: Since now there is only `GaugeField`, we no longer need the `static_cast` anymore?

lib/coarse_op_preconditioned.in.cu
```diff
@@ -160,7 +160,7 @@ namespace quda
           GaugeFieldParam param(X);
           param.order = gOrder_milc;
           param.setPrecision(X.Precision() < QUDA_SINGLE_PRECISION ? QUDA_SINGLE_PRECISION : X.Precision());
-          output = cudaGaugeField::Create(param);
+          output = new GaugeField(param);
```
hummingtree, line 169: I guess handling these raw pointers with smart pointers is out of the scope of this PR?

lib/dirac_coarse.cpp
```diff
@@ -33,10 +33,10 @@ namespace quda {
...
   }
 
-  DiracCoarse::DiracCoarse(const DiracParam &param, cpuGaugeField *Y_h, cpuGaugeField *X_h, cpuGaugeField *Xinv_h,
-                           cpuGaugeField *Yhat_h, // cpu link fields
-                           cudaGaugeField *Y_d, cudaGaugeField *X_d, cudaGaugeField *Xinv_d,
-                           cudaGaugeField *Yhat_d) // gpu link field
+  DiracCoarse::DiracCoarse(const DiracParam &param, GaugeField *Y_h, GaugeField *X_h, GaugeField *Xinv_h,
+                           GaugeField *Yhat_h, // cpu link fields
+                           GaugeField *Y_d, GaugeField *X_d, GaugeField *Xinv_d,
+                           GaugeField *Yhat_d) // gpu link field
```
hummingtree: With the MMA PR merged, these pointers are now `shared_ptr` so there are some merge needed here ...

include/clover_field.h
```diff
@@ -213,12 +214,18 @@ namespace quda {
...
     CloverField(const CloverFieldParam &param);
-    virtual ~CloverField();
 
     static CloverField *Create(const CloverFieldParam &param);
 
-    void* V(bool inverse=false) { return inverse ? cloverInv : clover; }
-    const void* V(bool inverse=false) const { return inverse ? cloverInv : clover; }
+    template <typename T = void *> auto data(bool inverse = false) const
+    {
+      return inverse ? reinterpret_cast<T>(cloverInv.data()) : reinterpret_cast<T>(clover.data());
```
weinbe2, line 222: Pedantic/in part for my own comprehension of how you implemented `.data()`---could this also be implemented as:

```
return inverse ? cloverInv.data<T*>() : clover.data<T*>();
```

I know it'd all be a means to the same end, I figured I'd ask for my own understanding.

include/enum_quda_fortran.h
```diff
@@ -17,11 +17,6 @@
 #define QUDA_ERROR 1
 #define QUDA_ERROR_UNINITIALIZED 2
 
-#define QUDA_MEMORY_DEVICE 0
```
weinbe2, line 20: Just confirming this full delete was intentional, and there's no need to expose these `enum` values via the fortran interface?

lib/coarse_op_preconditioned.in.cu
maddyscientist: Indeed, out of scope.  And smart pointer will go the way of the 🦤 anyway since now we have move and default constructors for the `GaugeField`.

lib/dirac_coarse.cpp
maddyscientist: 71a5335284050cc416a4937ca57e72fd990c680b

include/enum_quda_fortran.h
maddyscientist: Exactly.

include/clover_field.h
maddyscientist: No, because the underlying type here is a `quda_ptr` which is not templated.

lib/coarsecoarse_op_mma.in.cu
maddyscientist: Indeed (dd66595e1a5e2f61a04d5ec4dbc494b3da5840b4).

lib/gauge_field.cpp
```diff
@@ -274,12 +888,234 @@ namespace quda {
     }
   }
 
-  std::ostream& operator<<(std::ostream& output, const GaugeFieldParam& param) {
+  void *create_gauge_buffer(size_t bytes, QudaGaugeFieldOrder order, QudaFieldGeometry geometry)
```
weinbe2, line 908: Should there be a CUDA/CPU toggle here to change between allocating device vs host buffers? (Same question for the next few functions as well.)

tests/host_reference/gauge_force_reference.cpp
```diff
@@ -489,19 +492,24 @@ void gauge_force_reference(void *refMom, double eb3, void **sitelink, QudaPrecis
   param.t_boundary = QUDA_PERIODIC_T;
 
   auto qdp_ex = quda::createExtendedGauge((void **)sitelink, param, R);
+  //quda::TimeProfile dummy("blah");
```
weinbe2: Could be deleted :)

weinbe2:
I've reviewed every file not called `interface_quda.cpp` and I'm a big fan of all of the cleanup. I'll need to go through `interface_quda.cpp` with a fine-tipped comb before I feel good about it!

lib/gauge_field.cpp
maddyscientist: Nope, at least not in the present form.  These buffers are only used for device allocation.  I eventually plan to clean this stuff up, but in the spirit of the internals of `GaugeField` be similar to what was in `cuda`/`cpu` fields prior, this is deferred.

tests/host_reference/gauge_force_reference.cpp
maddyscientist: 50987b1b96a55143adae9810e1e45b38b7f93b13

lib/interface_quda.cpp
```diff
@@ -552,27 +552,26 @@ void freeUniqueSloppyGaugeUtility(cudaGaugeField *&precise, cudaGaugeField *&slo
...
-
   profileGauge.TPSTART(QUDA_PROFILE_INIT);
   // Set the specific input parameters and create the cpu gauge field
   GaugeFieldParam gauge_param(*param, h_gauge);
 
   if (gauge_param.order <= 4) gauge_param.ghostExchange = QUDA_GHOST_EXCHANGE_NO;
   GaugeField *in = (param->location == QUDA_CPU_FIELD_LOCATION) ?
-    static_cast<GaugeField*>(new cpuGaugeField(gauge_param)) :
-    static_cast<GaugeField*>(new cudaGaugeField(gauge_param));
+    static_cast<GaugeField*>(new GaugeField(gauge_param)) :
```
weinbe2: As far as I can tell, this conditional is now irrelevant and you can just do `GaugeField *in = new GaugeField(gauge_param);`

lib/interface_quda.cpp
```diff
@@ -3826,43 +3790,37 @@ void invertMultiShiftQuda(void **hp_x, void *hp_b, QudaInvertParam *param)
...
-  profileMulti.TPSTOP(QUDA_PROFILE_TOTAL);
-
   profilerStop(__func__);
+  popProfile();
 }
 
 void computeKSLinkQuda(void *fatlink, void *longlink, void *ulink, void *inlink, double *path_coeff, QudaGaugeParam *param)
 {
-  profileFatLink.TPSTART(QUDA_PROFILE_TOTAL);
-  profileFatLink.TPSTART(QUDA_PROFILE_INIT);
```
weinbe2, line 3837: Why was profiling `QUDA_PROFILE_INIT` deleted here?

weinbe2:
Why was timing `QUDA_PROFILE_INIT` removed in a handful of interface functions? (I know I mentioned this in one review comment, but it seems to appear in multiple places.)

lib/interface_quda.cpp
```diff
@@ -3981,101 +3901,55 @@ void computeTwoLinkQuda(void *twolink, void *inlink, QudaGaugeParam *param)
...
+  gParam.location = QUDA_CUDA_FIELD_LOCATION;
+  gParam.create = QUDA_COPY_FIELD_CREATE;
+  gParam.field = &cpuSiteLink;
+  gParam.reconstruct = qudaGaugeParam->reconstruct;
+  gParam.setPrecision(qudaGaugeParam->cuda_prec, true);
+  GaugeField cudaSiteLink = qudaGaugeParam->use_resident_gauge ? gaugePrecise->create_alias() : GaugeField(gParam);
 
   GaugeFieldParam gParamMom(*qudaGaugeParam, mom, QUDA_ASQTAD_MOM_LINKS);
   gParamMom.location = QUDA_CPU_FIELD_LOCATION;
-  if (gParamMom.order == QUDA_TIFR_GAUGE_ORDER || gParamMom.order == QUDA_TIFR_PADDED_GAUGE_ORDER)
```
weinbe2, line 4060: I assume this is now enforced internally?

lib/interface_quda.cpp
```diff
@@ -3981,101 +3901,55 @@ void computeTwoLinkQuda(void *twolink, void *inlink, QudaGaugeParam *param)
...
-    gParamMom.create = QUDA_ZERO_FIELD_CREATE;
-    cudaMom = new cudaGaugeField(gParamMom);
-    profileGaugeForce.TPSTOP(QUDA_PROFILE_INIT);
-    if (!qudaGaugeParam->overwrite_mom) {
-      profileGaugeForce.TPSTART(QUDA_PROFILE_H2D);
-      cudaMom->loadCPUField(*cpuMom);
-      profileGaugeForce.TPSTOP(QUDA_PROFILE_H2D);
-    }
-  }
+  if (qudaGaugeParam->use_resident_mom && !momResident.Volume()) errorQuda("No resident momentum field to use");
```
weinbe2: I'm not sure if I like the use of `.Volume()` as a proxy for figuring out if there's a resident field or not---it's textually reasonable, but not very clear. Maybe we could add a `.Initialized()` or `.Allocated()` or `isDummy()` or something of the sort that maps to `Volume() == [or !=] 0` as appropriate?

lib/interface_quda.cpp
```diff
@@ -4092,109 +3966,69 @@ int computeGaugeForceQuda(void* mom, void* siteLink,  int*** input_path_buf, int
...
-  }
+  if (qudaGaugeParam->return_result_mom) cpuMom.copy(cudaMom);
 
-  profileGaugeForce.TPSTART(QUDA_PROFILE_FREE);
   if (qudaGaugeParam->make_resident_gauge) {
-    if (gaugePrecise && gaugePrecise != cudaSiteLink) freeUniqueGaugeQuda(QUDA_WILSON_LINKS);
-    gaugePrecise = cudaSiteLink;
-  } else {
-    delete cudaSiteLink;
+    if (gaugePrecise && !qudaGaugeParam->use_resident_gauge) freeUniqueGaugeQuda(QUDA_WILSON_LINKS);
```
weinbe2: ~Was replacing `make_resident_gauge` with `!use_resident_gauge` intentional? We can't generally assume they're redundant. (Whether or not we should be able to make that assumption is another story.)~

Nevermind, time to stop reviewing for the day.

lib/interface_quda.cpp
```diff
@@ -5068,347 +4773,214 @@ void computeCloverForceQuda(void *h_mom, double dt, void **h_x, void **, double
...
   GaugeFieldParam gParam(*param, gauge, QUDA_SU3_LINKS);
   gParam.location = QUDA_CPU_FIELD_LOCATION;
-  gParam.site_offset = param->gauge_offset;
-  gParam.site_size = param->site_size;
   bool need_cpu = !param->use_resident_gauge || param->return_result_gauge;
-  cpuGaugeField *cpuGauge = need_cpu ? new cpuGaugeField(gParam) : nullptr;
+  GaugeField cpuGauge = need_cpu ? GaugeField(gParam) : GaugeField();
 
-  GaugeFieldParam gParamMom(*param, momentum);
-  gParamMom.reconstruct = (gParamMom.order == QUDA_TIFR_GAUGE_ORDER || gParamMom.order == QUDA_TIFR_PADDED_GAUGE_ORDER) ?
```
weinbe2, line 5111: Similar comment as above, I assume this logic has been shifted elsewhere/internally?

lib/interface_quda.cpp
```diff
@@ -5068,347 +4773,214 @@ void computeCloverForceQuda(void *h_mom, double dt, void **h_x, void **, double
...
   gParam.link_type = QUDA_ASQTAD_MOM_LINKS;
   gParam.reconstruct = QUDA_RECONSTRUCT_10;
   gParam.ghostExchange = QUDA_GHOST_EXCHANGE_NO;
   gParam.pad = 0;
-  cudaGaugeField *cudaMom = !param->use_resident_mom ? new cudaGaugeField(gParam) : nullptr;
+  GaugeField cudaMom = param->use_resident_mom ? momResident.create_alias() : GaugeField(gParam);
 
+  if (param->use_resident_gauge && !gaugePrecise) errorQuda("No resident gauge field allocated");
   gParam.link_type = QUDA_SU3_LINKS;
   gParam.reconstruct = param->reconstruct;
```
weinbe2, line 4675: Technically orthogonal to this PR, but if we're updating the reconstruct, don't we (and haven't we always) also need(ed) a `setPrecision` to force native fields?

lib/interface_quda.cpp
```diff
@@ -5539,12 +5111,11 @@ void performWuppertalnStep(void *h_out, void *h_in, QudaInvertParam *inv_param,
 
 void performTwoLinkGaussianSmearNStep(void *h_in, QudaQuarkSmearParam *smear_param)
 {
-  if(smear_param->n_steps == 0) return;
+  if (smear_param->n_steps == 0) return;
+  pushProfile(profileGaussianSmear);
+  profileGaussianSmear.TPSTART(QUDA_PROFILE_INIT);
```
weinbe2: If `QUDA_PROFILE_INIT` has been removed elsewhere, why is it still included here?

weinbe2:
I've completed my visual review of `interface_quda.cpp` and left a handful of comments/questions, most of which probably only beg clarification. I'll start doing some explicit tests soon.

lib/interface_quda.cpp
maddyscientist: indeed 😄

lib/interface_quda.cpp
maddyscientist: This was intentional.  Profiling of memory allocations is now done automatically inside `quda_ptr`.

maddyscientist: Though I'm glad you asked now, as this made me find an unrelated bug 😉

lib/interface_quda.cpp
maddyscientist: While I had intentionally engineered ref counting, so that we can double start the same timer, to avoid any breakages, given you comment here, I'll just go ahead and remove all unnecessary `QUDA_PROFILE_INIT` start and stops in `interface_quda.cpp`.

maddyscientist: [3963f63](https://github.com/lattice/quda/pull/1384/commits/3963f6329ef8af6d3cc56d1ce6b44859bdf77dbb)

lib/interface_quda.cpp
maddyscientist: https://github.com/lattice/quda/pull/1384/commits/06d2dcbeb259fb35a4b0e95561d594167d278daa

lib/interface_quda.cpp
maddyscientist: Yep, it's set now in the constructor for `GaugeFieldParam`.

lib/interface_quda.cpp
maddyscientist: Yep.

lib/interface_quda.cpp
maddyscientist: This is a good point.  I found a few more cases like this in interface_quda.cpp, so I've gone and fixed this (426b59a579dd093e76940ecc539f17a3f927b9aa).

lib/interface_quda.cpp
maddyscientist: I've added `empty()` which adds this functionality (`empty()` being chosen since that's consistent with STL's containers for the same purpose).

maddyscientist:
@weinbe2 Ok, all your comments should be addressed now.

include/gauge_field.h
```diff
@@ -37,109 +37,70 @@ namespace quda {
...
 
-    double anisotropy;
-    double tadpole;
-    void *gauge; // used when we use a reference to an external field
+    QudaFieldCreate create = QUDA_REFERENCE_FIELD_CREATE; // used to determine the type of field created
 
-    QudaFieldCreate create; // used to determine the type of field created
-
-    QudaFieldGeometry geometry; // whether the field is a scale, vector or tensor
+    QudaFieldGeometry geometry = QUDA_VECTOR_GEOMETRY; // whether the field is a scale, vector or tensor
```
weinbe2: typo `scale` -> `scalar`

maddyscientist: [b19fe54](https://github.com/lattice/quda/pull/1384/commits/b19fe5443e6b9b8a2787cc684d3e65500cc19268)

maddyscientist:
This has now been tested with MILC RHMC, and things are looking good.

hummingtree (approved):
Have verified (again) that CPS works good with this branch.

lib/color_spinor_field.cpp
```diff
@@ -186,7 +169,7 @@ namespace quda
       components.reserve(composite_descr.dim);
       for (int cid = 0; cid < composite_descr.dim; cid++) {
         param.component_id = cid;
-        param.v = static_cast<void *>(static_cast<char *>(v) + cid * bytes / composite_descr.dim);
+        param.v = static_cast<void *>(static_cast<char *>(v.data()) + cid * bytes / composite_descr.dim);
```
weinbe2, line 172: `v.data<char*>()`?

lib/color_spinor_field.cpp
```diff
@@ -203,7 +186,7 @@ namespace quda
       param.is_component = composite_descr.is_component;
       param.component_id = composite_descr.id;
       even = new ColorSpinorField(param);
-      param.v = static_cast<char *>(v) + bytes / 2;
+      param.v = !ghost_only ? static_cast<char *>(v.data()) + bytes / 2 : nullptr;
```
weinbe2, line 189: `v.data<char*>()`

weinbe2:
Tentative approve, I noticed a few cosmetic things, and this needs a fresh merge of `develop` + `clang-format`. After that it's good to go!

lib/color_spinor_field.cpp
maddyscientist: No applicable here.  There is no template option for `quda_ptr::data()` (unlike for the higher-level object constructs `data` member function).

lib/color_spinor_field.cpp
maddyscientist: No applicable here.  There is no template option for `quda_ptr::data()` (unlike for the higher-level object constructs `data` member function).

maddyscientist:
@SaltyChiang I have merged  in your changes into this branch, that will shortly be merged into develop.  

One line I have not included is this one:
```
gauge[d] = (param.gauge) ? ((void **)param.gauge)[d] : nullptr;
```
as this line doesn't make sense to me.  If it's a reference to a pre-existing field, why should the value be a `nullptr`?  Can you explain why this change is needed?

Thanks

SaltyChiang:
Hi @maddyscientist, there are some `if` statements about the input gauge link, like  https://github.com/lattice/quda/blob/6e90b49a2d12891f2f68e35b7cfa7defbd506526/lib/interface_quda.cpp#L3873 and https://github.com/lattice/quda/blob/6e90b49a2d12891f2f68e35b7cfa7defbd506526/lib/interface_quda.cpp#L3903. I believe the logic here is originally used for QUDA_MILC_GAUGE_ORDER, but this will fail if QUDA_QDP_GAUGE_ORDER is used here. For example, if the `ulink` is `nullptr` here, the unitarized fat7 link will not be calculated, which is right; but `cpuUnitarizedLink` defined in https://github.com/lattice/quda/blob/6e90b49a2d12891f2f68e35b7cfa7defbd506526/lib/interface_quda.cpp#L3847 fails now as you cannot get `((void **)nullptr)[d]`. A null `QUDA_QDP_GAUGE_ORDER` field should be something like `void *ulink[4] = {nullptr, nullptr, nullptr, nullptr}` but now the `if (ulink)` will return true which we do not expect.

So I made all four reference fields `nullptr` when the input value itself is `nullptr` and I think this will not affect the existing logic.

maddyscientist:
@SaltyChiang thanks for the explanation.  I have fixed my branch accordingly (https://github.com/lattice/quda/pull/1384/commits/4c308f6f72a2239e97f3423137459e049dc56b03), so it should work for your usecase.  If you could test this branch, that would be appreciated, thanks.

weinbe2 (approved):
Looks great, thanks @maddyscientist !

SaltyChiang:
Hi @maddyscientist, very sorry for the late feedback. I think it should be `!unitarizedLink.StaggeredPhaseApplied() && param->staggered_phase_applied` here. https://github.com/lattice/quda/blob/0c016af709ad6a3c249540fdad809c8d18abc226/lib/interface_quda.cpp#L3748-L3749 

Edit: I believe `gauge_param.location` should be changed to `QUDA_CUDA_FIELD_LOCATION` here.
https://github.com/lattice/quda/blob/0c016af709ad6a3c249540fdad809c8d18abc226/lib/interface_quda.cpp#L765-L774

maddyscientist:
@SaltyChiang thanks for pointing put my bug.  I've pushed a fix to my next PR that will be merged (#1393).  This is a fairly simple PR, so will likely be merged soon.

SaltyChiang:
@maddyscientist thanks for the fix. You might miss another issue that will happen while saving the smeared gauge through saveGaugeQuda(). I updated the comment above and I repeat it here. 


> Edit: I believe `gauge_param.location` should be changed to `QUDA_CUDA_FIELD_LOCATION` here.
> 
> https://github.com/lattice/quda/blob/0c016af709ad6a3c249540fdad809c8d18abc226/lib/interface_quda.cpp#L765-L774

maddyscientist:
Thanks again @SaltyChiang.  Fix is here 44c41006158e493517452210a7a1d1e55412d63f (in #1393).

PR #1390, opened 2023-07-07
Preconditioner cleanup

weinbe2:
This narrow PR cleans up various bits and bops related to how preconditioners are created and handled in QUDA, plus other various helpful utilities:

- Creates an abstract `getInverterType()` function within the base `Solver` class, forcing each derived class to specify its `QudaInverterType` enum.
- Created a single `Solver` function `fillInnertParam`/`extractInnerParam` to handle copying preconditioner properties in/out, which removed redundant code in `GCR` and `PCG`
- Unified creating preconditioners, plus wrapping them in `std::shared_ptr` to simplify cleanup
- ...which requires some trickery, a shared pointer with a custom no-op deallocator, to handle `MG` preconditioners
- Threaded in support for an MG preconditioner for PCG as a proof-of-concept (even though there's no formal operator with a Hermitian coarse op in QUDA)

Outstanding bits:

- [x] `clang-format`
- [x] `ctest`


weinbe2:
This now passes `ctest`. @maddyscientist , assuming this now passes Jenkins, can you give this a look (and I can subsequently `clang-format` it)?

include/invert_quda.h
```diff
@@ -569,7 +569,7 @@ namespace quda {
     /**
        @return Whether the solver is only for Hermitian systems
      */
-    virtual bool hermitian() = 0;
+    virtual bool hermitian() const noexcept = 0;
```
maddyscientist: Curious: why do we need this to be `noexcept`?

tests/eigensolve_test_gtest.hpp
```diff
@@ -41,7 +41,7 @@ TEST_P(EigensolveTest, verify)
   // The solution to avoid this is to use a Krylov space (eig-n-kr) about 3-4 times the
   // size of the search space (eig-n-ev), or use a well chosen Chebyshev polynomial,
   // or use a tighter than necessary tolerance.
-  if (eig_param.eig_type == QUDA_EIG_IR_ARNOLDI || eig_param.eig_type == QUDA_EIG_BLK_IR_ARNOLDI) factor *= 5;
+  if (eig_param.eig_type == QUDA_EIG_IR_ARNOLDI || eig_param.eig_type == QUDA_EIG_BLK_IR_ARNOLDI) factor *= 10;
```
maddyscientist, line 44: Why was this needed?

include/invert_quda.h
weinbe2: Seemed like a best-practice type thing, there's no reason for it to not be (I guess), but I can certainly revert it.

maddyscientist:
This looks fairly inoffensive.  Regarding `ctest` passing: does this pass on multiple GPUs?  Checking, since this PR touches things with Schwarz solvers, just checking that the Schwarz-preconditioned tests are working ok (Mobius PCG and Wilson GCR).....

tests/eigensolve_test_gtest.hpp
weinbe2: Four operator tests (during a build including all dslash) each had one or two eigenvalues (of 16 or 32, whichever it was) that jumped just over the buffer factor of 5, but were covered by a factor of 10. Since this happened for only a few random operators and a few eigenvalues each, with no systematic-looking patterns, I just bumped the factor a bit and didn't feel a need to dig in deeper.

weinbe2:
> This looks fairly inoffensive. Regarding `ctest` passing: does this pass on multiple GPUs? Checking, since this PR touches things with Schwarz solvers, just checking that the Schwarz-preconditioned tests are working ok (Mobius PCG and Wilson GCR).....

Yes, my `ctest` experiments were all with `-DQUDA_MPI=ON` builds running on 8xGPUs, 1x2x2x2 grid partitioning.

tests/eigensolve_test_gtest.hpp
maddyscientist: Ok.  I hope that improved reduction precision will improve the precision of Eigen solvers in this regard.  Anyway, not a blocker here.

include/invert_quda.h
maddyscientist: I see no reason to use `noexcept` here.  The only reason I've really seen where it matters is with respect to `move` semantics (e.g., ensuring move constructor is used over copy constructor) which obviously doesn't apply here.  I would prefer this can be reverted unless there's reason not to.

weinbe2: Sounds good, I can get rid of it.

weinbe2: https://github.com/lattice/quda/pull/1390/commits/9d1e63f9abde0ee62284db1fe1b156e2d902acf9

maddyscientist (approved):
Thanks @weinbe2 for this PR.  Approving.

PR #1393, opened 2023-07-18
Block dilution

maddyscientist:
This simple PR add supports for local disjoint hyper-cubic block dilution.  I developed this as part of some experiments with smoothers.  This effort is on ice, but since this is a nice self-contained feature addition, I figured it's best to get this upstreamed.  This work makes some small changes to the `colorspinor::FloatNOrder` accessor, which will be helpful for multi-RHS.

Note this PR is dependent on #1384 and so should not be considered for merging until after that is merged.


weinbe2:
This is a nice PR @maddyscientist , thanks!

One concern I have is that, when running `dilution_test` by hand, there's an OOM error (on an 80GB card). This is reasonable on the merits because the default values are a volume of `24x24x24x24` and the default dilution size is `1x1x1x1`. I know that "under the hood" the maximum number of dilution blocks is 128, but the OOM error hits during the allocations before that.

This is not an issue during `ctest` since the local volume, and the dilution block size, is sufficiently small.

In the interest of the test executables "just working", I recommend changing the default value for `--dilution-block-size` to `8 8 8 8`, since it "just works", though I'm not married to any particular value so long as it doesn't OOM on a reasonable GPU and the number of dilution vectors is <= 128.

weinbe2 (approved):
Pending Jenkins, approved!

PR #1395, opened 2023-07-19
add github actions build for CUDA

mathiaswagner:
This adds some minimal CUDA build using GitHub Actions using GitHub runners to QUDA. 

I had to workaround some warnings to make it pass the build (esp clang) so happy to see if we can come up with better solutions.


maddyscientist (approved):

maddyscientist:
Noting @weinbe2 has also approved this PR in offline chat.  Merging...

PR #1397, opened 2023-08-01
Add some sanity checking on the CPM.cmake download

dmcdougall:
It's possible for this download to fail.  An example failure scenario is when CMake is not built with OpenSSL support, and so the https protocol is not supported.

After this change, when it fails the user will be informed:
1.  that it actually failed, as opposed to silently continuing with an empty CPM.cmake file; and
2.  why CMake thinks the failure happened.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -15,7 +15,13 @@ if(NOT (EXISTS ${CPM_DOWNLOAD_LOCATION}))
   file(DOWNLOAD
        https://github.com/cpm-cmake/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake
        ${CPM_DOWNLOAD_LOCATION}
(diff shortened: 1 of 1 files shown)


mathiaswagner:

mathiaswagner:

mathiaswagner:
This file is included following the instructions from https://github.com/cpm-cmake/CPM.cmake#adding-cpm and potentially overwritten when updating CPM.
I am hesitant to modify the version that comes with QUDA for that reason. Maybe this can be addressed directly in CPM?

mathiaswagner:
Maybe [this](https://github.com/cpm-cmake/CPM.cmake/pull/474/files) is addressing it already ?

dmcdougall:
Ah, thanks.  I didn't realise it came from upstream CPM.

mathiaswagner:
Might still be worth having an issue to track this.

PR #1400, opened 2023-08-16
Extend invert_test_gtest.hpp with testing for HQ CG solves.

maddyscientist:
This PR adds support for testing heavy-quark CG solves in the gtest solver testing.  In doing so, I have seemingly discovered a bug in the HQ solver, where it fails to converge in mixed-precision.

The following command
```
tests/invert_test --dslash-type wilson --ngcrkrylov 8 --dim 2 4 6 8 --prec double --prec-sloppy single --tol 1e-7 --tolhq 1e-7
```
fails to converge.
```
CG: Convergence at 161 iterations, L2 relative residual: iterated = 2.738192e+05, true = 2.738192e+05 (requested = 1.000000e-07), heavy-quark residual = 1.085493e+00 (requested = 1.000000e-07)
```
This is seemingly a mixed-precision bug in the HQ CG solver, as convergence is fine for the regular CG solver, and also proceeds as expected in the HQ solver in uni-precision.

@weinbe2 handing this to you to fix this bug, so we can get this test merged in.

### DIFF
--- a/lib/inv_cg_quda.cpp
+++ b/lib/inv_cg_quda.cpp
@@ -941,13 +941,13 @@ namespace quda {
         // we "reset" the solve in a different way.
         if (heavy_quark_restart) {
           // If we're in the HQ residual part of the solve, we just do a hard CG restart.
-          logQuda(QUDA_SUMMARIZE, "HQ restart == hard CG restart\n");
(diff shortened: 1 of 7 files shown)


weinbe2 (approved):
PR looks good! Thanks for adding it, especially because it caught one of my own bugs =D

PR #1401, opened 2023-08-17
Specifying granular masses and tolerances for multishift tests

weinbe2:
This very narrow PR adds command line arguments for specifying custom masses, tolerances, and HQ tolerances for each shift in multishift solves for both staggered and Wilson-type solves. The default behavior is to base masses off the mass (with extended shifts adding `0.01 * i * i`, as before) and to use `tol` and `tol_hq` for all of the shifts, but the flags `--multishift-masses`, `--multishift-tol`, and `multishift-tols-hq` can be used to override this behavior.

Examples below are for staggered; remove the `--dslash-type staggered --compute-fat-long true --test 5` to test Wilson-type operators.

4-shift solve with default masses, tols:
```
./staggered_invert_test --dim 8 8 8 8 --dslash-type staggered --compute-fat-long true --prec double --prec-sloppy single --test 5 --multishift 4
```

4-shift solve with custom masses, default tols:
```
./staggered_invert_test --dim 8 8 8 8 --dslash-type staggered --compute-fat-long true --prec double --prec-sloppy single --test 5 --multishift 4 --multishift-masses 0.01 0.02 0.03 0.04
```

4-shift solve with custom masses and tolerances:
```
./staggered_invert_test --dim 8 8 8 8 --dslash-type staggered --compute-fat-long true --prec double --prec-sloppy single --test 5 --multishift 4 --multishift-masses 0.01 0.02 0.03 0.04 --multishift-tols 1e-4 1e-5 1e-6 1e-7
```

If the number of masses, etc, does not agree with the argument to `--multishift` the tests will intentionally error out.

Outstanding work is minimal:
- [ ] `clang-format`

### DIFF
--- a/tests/invert_test.cpp
+++ b/tests/invert_test.cpp
@@ -239,14 +239,22 @@ std::vector<std::array<double, 2>> solve(test_t param)
   if (multishift > 1) {
     if (use_split_grid) { errorQuda("Split grid does not work with multishift yet."); }
     inv_param.num_offset = multishift;
+
+    // Consistency check for masses, tols, tols_hq size if we're setting custom values
+    if (multishift_masses.size() != 0 && multishift_masses.size() != static_cast<unsigned long>(multishift))
+      errorQuda("Multishift mass count %d does not agree with number of masses passed in %lu\n", multishift, multishift_masses.size());
+    if (multishift_tols.size() != 0 && multishift_tols.size() != static_cast<unsigned long>(multishift))
+      errorQuda("Multishift tolerance count %d does not agree with number of masses passed in %lu\n", multishift, multishift_tols.size());
+    if (multishift_tols_hq.size() != 0 && multishift_tols_hq.size() != static_cast<unsigned long>(multishift))
+      errorQuda("Multishift hq tolerance count %d does not agree with number of masses passed in %lu\n", multishift, multishift_tols_hq.size());
+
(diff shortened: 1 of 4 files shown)


maddyscientist:
The term `masses` isn’t appropriate for Wilson and it would be more appropriate to call these offsets or shifts.  Similarly, the scaling that’s been applied in `invert_test` doesn’t make sense, since we would want ideally to just specify raw numbers from the poles from RHMC.

How about a flag `—multishift-offset` to specify the raw (unscaled) shifts for Wilson like solvers and for staggered it could work with either the offset or masses option, with the rescaling only being applied when the masses option is specified?

weinbe2:
> The term `masses` isn’t appropriate for Wilson and it would be more appropriate to call these offsets or shifts. Similarly, the scaling that’s been applied in `invert_test` doesn’t make sense, since we would want ideally to just specify raw numbers from the poles from RHMC.
> 
> How about a flag `—multishift-offset` to specify the raw (unscaled) shifts for Wilson like solvers and for staggered it could work with either the offset or masses option, with the rescaling only being applied when the masses option is specified?

I applied your suggestion for Wilson-type fermions. There are too many conventions under the hood for staggered fermions for me to want to "poke the bear" on handling shifts vs masses at this time...

maddyscientist (approved):
Nice wee cleanup

PR #1414, opened 2023-11-06
Feature/openqxd

chaoos:
* Adds interface openQxD <--> QUDA
* Adds capability for C* boundaries in 1,2 or 3 spatial directions
* Fixed CPU reordering
* Adds openQCD gamma matrix basis (and basis transformations)
* Adds order class for spinors
* Adds order class for clover field (U(1) clover field is transfered from openQxD)
* Adds order class for gauge field
* Changed all the comments in quda.h to C89 style comments
* Rebased on your current `develop` branch


lib/color_spinor_field.cpp
```diff
@@ -454,7 +454,7 @@ namespace quda
 
       if (reorder_location() == QUDA_CPU_FIELD_LOCATION) { // reorder on the host
         void *buffer = pool_pinned_malloc(bytes);
-        qudaMemcpy(buffer, v.data(), bytes, qudaMemcpyDefault);
+        qudaMemcpy(buffer, src.v.data(), bytes, qudaMemcpyDefault);
```
chaoos: I observed that reordering on the CPU always results in vectors that are zero everywhere. I think there is a typo, since this fixed reordering for me.

include/clover_field_order.h
```diff
@@ -11,6 +11,7 @@
 #include <convert.h>
 #include <clover_field.h>
 #include <complex_quda.h>
+#include <index_helper.cuh>
```
chaoos, line 14: I need getCoords as well as my openqcd index functions in the order class

include/communicator_quda.h
```diff
@@ -41,6 +41,7 @@ namespace quda
     int (*coords)[QUDA_MAX_DIM];
     int my_rank;
     int my_coords[QUDA_MAX_DIM];
+    int cstar; // number of C* direction as per openQxD convention
```
chaoos, line 44: We extended the Topology struct with a cstar member to indicate how many spatial directions have C* boundaries. Maybe you want to have that as a option in the `QudaInvertParam` struct? However, we need this setting here, since it influences in the process grid topology.

include/communicator_quda.h
```diff
@@ -126,9 +127,26 @@ namespace quda
...
+          (i==1 && topo->cstar >= 2) ||
+          (i==2 && topo->cstar >= 3)
+      )) {
+        // if we go over the boundary and have a shifted boundary condition,
+        // we shift Nx/2 ranks in x-direction:
+        // shift_integer in {-1, 0, 1}
+        shift_integer = (comm_coords(topo)[i] + displacement[i] + comm_dims(topo)[i]) / comm_dims(topo)[i];
+        Nx_displacement += (shift_integer - 1) * (comm_dims(topo)[0]/2);
+      }
+      coords[i] = (i < topo->ndim) ? mod(comm_coords(topo)[i] + displacement[i] + (i==0 ? Nx_displacement :0), comm_dims(topo)[i]) : 0;
```
chaoos: Here we change the rank-topology based on how many C* directions we have. If there is a C* direction, we calculate the rank displacement in X-direction (`NX_displacement`) accordingly. `topo->cstar=0` recovers legacy behavior.

include/dslash_quda.h
```diff
@@ -808,4 +808,10 @@ namespace quda
   */
   void gamma5(ColorSpinorField &out, const ColorSpinorField &in);
 
+  /* RG: I have added these */
+  void gamma0(ColorSpinorField &out, const ColorSpinorField &in);
+  void gamma1(ColorSpinorField &out, const ColorSpinorField &in);
+  void gamma2(ColorSpinorField &out, const ColorSpinorField &in);
+  void gamma3(ColorSpinorField &out, const ColorSpinorField &in);
+
```
chaoos: I needed gamma0 to gamma3 for tests in openQxD side. Let me know if there is an intended way to apply a gamma matrix other than gamm5 to a spinor field (I didn't find anything).

include/enum_quda.h
```diff
@@ -18,18 +18,18 @@ typedef enum QudaMemoryType_s {
   QUDA_MEMORY_INVALID = QUDA_INVALID_ENUM
 } QudaMemoryType;
 
-//
-// Types used in QudaGaugeParam
-//
+/**
```
chaoos: Also here I changed all the comments to C89 style ones to be able to compile openQxD in C89. People don't want to change that standard ...

include/gauge_field_order.h
```diff
@@ -23,6 +23,7 @@
 #include <load_store.h>
 #include <aos.h>
 #include <transform_reduce.h>
+#include <communicator_quda.h>
```
chaoos: I need the number of ranks in all 4 directions (see `nproc` in the order class) to calculate offsets for the gauge, clover and spinor fields. Let me know if there is a better way to do that. in openQCD these are compile time constants.

include/index_helper.cuh
```diff
@@ -1106,4 +1106,302 @@ namespace quda {
     return (((x[3]*X[2] + x[2])*X[1] + x[1])*X[0] + x[0]) >> 1;
   }
 
+
```
chaoos: Here in a namespace `openqcd`, I added all indexing helper function that are needed by the order classes to reduce code duplication.

include/quda.h
```diff
@@ -16,7 +16,7 @@
 
 #ifndef __CUDACC_RTC__
 #define double_complex double _Complex
-#else // keep NVRTC happy since it can't handle C types
+#else /* keep NVRTC happy since it can't handle C types */
```
chaoos: Again the C89 style comments

include/quda.h
```diff
@@ -980,7 +979,7 @@ extern "C" {
    * initQuda.  Calling initQudaMemory requires that the user has
    * previously called initQudaDevice.
    */
-  void initQudaMemory();
+  void initQudaMemory(void);
```
chaoos, line 1018: C89 complains that without the void, this is not a valid function signature.

lib/comm_common.cpp
```diff
@@ -102,6 +102,13 @@ namespace quda
 
     Topology *topo = new Topology;
 
+#ifdef BUILD_OPENQCD_INTERFACE
+    int *data = static_cast<int *>(map_data);
+    topo->cstar = data[0];
+#else
+    topo->cstar = 0;
+#endif
+
```
chaoos, line 110: Here we set `cstar` to the value that comes from openQxD. We set `cstar=0` else, to recover legacy behavior.

lib/dslash_gamma_helper.cu
```diff
@@ -103,4 +103,9 @@ namespace quda {
   // Applies a gamma5 matrix to a spinor (wrapper to ApplyGamma)
   void gamma5(ColorSpinorField &out, const ColorSpinorField &in) { ApplyGamma(out,in,4); }
 
+  /* RG: I have added these */
+  void gamma0(ColorSpinorField &out, const ColorSpinorField &in) { ApplyGamma(out,in,0); }
+  void gamma1(ColorSpinorField &out, const ColorSpinorField &in) { ApplyGamma(out,in,1); }
+  void gamma2(ColorSpinorField &out, const ColorSpinorField &in) { ApplyGamma(out,in,2); }
+  void gamma3(ColorSpinorField &out, const ColorSpinorField &in) { ApplyGamma(out,in,3); }
```
chaoos: Again, let me know it I should remove them.

lib/gauge_field.cpp
```diff
@@ -151,6 +151,25 @@ namespace quda {
...
+        /* analogue to BNDRY in openQCD:include/global.h */
+        long int bndry = 0;
+        bndry += (1-(comm_dim(0)%2))*x[1]*x[2]*x[3];
+        bndry += (1-(comm_dim(1)%2))*x[0]*x[2]*x[3];
+        bndry += (1-(comm_dim(2)%2))*x[0]*x[1]*x[3];
+        bndry += (1-(comm_dim(3)%2))*x[0]*x[1]*x[2];
+        bndry *= 2;
+
+        length += 18*7*bndry/4;
+        bytes = length * precision;
```
chaoos, line 173: For openQxD, we need to copy a little more data for the gauge field, because the gauge fields of even boundary points in positive direction are stored in `base_pointer + 4*VOLUME + some_offset`

include/clover_field.h
```diff
@@ -133,6 +133,8 @@ namespace quda {
       inverse(param.inverse),
       clover(param.clover),
       cloverInv(param.cloverInv),
+      csw(param.csw),
+      coeff(param.coeff),
```
donion74, line 170: What does this `coeff` do?

include/index_helper.cuh
```diff
@@ -1106,4 +1106,302 @@ namespace quda {
...
+     *
+     * @return     Volume
+     */
+    __device__ __host__ inline int vol(const int *X)
+    {
+      return X[0]*X[1]*X[2]*X[3];
+    }
+
+
+    /**
```
donion74, line 1260: Could you write out what `cbs`does?

include/index_helper.cuh
```diff
@@ -1106,4 +1106,302 @@ namespace quda {
...
+    {
+      return 2*(((1-(nproc[0]%2))*surface(L, 0))
+              + ((1-(nproc[1]%2))*surface(L, 1))
+              + ((1-(nproc[2]%2))*surface(L, 2))
+              + ((1-(nproc[3]%2))*surface(L, 3))
+      );
+    }
+
+
+    __device__ __host__ inline int ifc(const int *L, const int *nproc, const int mu)
```
donion74: Could you add a small comment what this does?

include/index_helper.cuh
```diff
@@ -1106,4 +1106,302 @@ namespace quda {
...
+          ret = surface(X, 0);
+        }
+      }
+
+      return ret;
+    }
+
+
+    /**
+     * @brief      Pure implementation of iup[ix][mu]
```
donion74: Maybe add: Returns neighbouring point in positive $\mu$ direction.

include/quda_openqcd_interface.h
```diff
@@ -0,0 +1,343 @@
...
+
+#ifdef double2_ORIGINAL
+#define double2 double2_ORIGINAL
+#undef double2_ORIGINAL
+#endif
+
+#ifdef __CUDACC_RTC_ORIGINAL__
+#define __CUDACC_RTC__ __CUDACC_RTC_ORIGINAL__
+#undef __CUDACC_RTC_ORIGINAL__
+#endif
```
chaoos, line 47: I know that looks ugly, but it redefines the datatype `double2` to the struct above, such that in `quda.h` the `double_complex` is redefined to `double2` (which in turn is `openqcd_complex_dble`), which is something that C89 is not complaining about. I agree that we should probably solve that differently :-), but by this we can still build on the openqxd side with `-std=c89 -pedantic -Werror`.

maddyscientist:
@Jenkins test this please

maddyscientist:
Thanks for this contribution @chaoos.  This looks like a great PR and it will be great to OpenQCD support plumbed into QUDA 😄 

@mathiaswagner and I are traveling at the moment, so reviewing this will likely take longer than normal.  In the meantime, can you fix the broken builds in the CI matrix? (click on the `Details` link next to the failed build, and you should be able to find the reason for the failure).  Also, please add your name to the `README` under the list of contributors.

havogt:
cscs-ci run

include/clover_field.h
chaoos: This is set to `kappa*csw`. in `loadCloverQuda()` and later used in `computeClover()`.

include/index_helper.cuh
chaoos: Comment added

include/index_helper.cuh
chaoos: Comment added

include/index_helper.cuh
chaoos: Comment added

chaoos:
@maddyscientist Could you trigger the pipeline again? Locally I fixed the problems.
Thank you,
Roman

include/clover_field.h
donion74: Do you want to add this comment in the code?

mathiaswagner:
@Jenkins test this please.

include/gauge_field_order.h
chaoos: This line is removed.

include/clover_field_order.h
```diff
@@ -1015,6 +1016,102 @@ namespace quda {
         size_t Bytes() const { return length*sizeof(Float); }
       };
 
+      /**
+       * OpenQCD ordering for clover fields
+       */
+      template <typename Float, int length = 72> struct OpenQCDOrder {
```
maddyscientist, line 994: Just curious, whether QUDA has sufficient capability to create the clover term for OpenQ*D's use case or whether you have specific requirements that we are lacking?  E.g., exponential clover or anisotropy?

include/color_spinor_field_order.h
```diff
@@ -1811,6 +1811,90 @@ namespace quda
...
+
+      OpenQCDDiracOrder(const ColorSpinorField &a, int = 1, Float *field_ = 0, float * = 0) :
+        field(field_ ? field_ : a.data<Float *>()),
+        offset(a.Bytes() / (2 * sizeof(Float))),  // TODO: What's this for??
+        volumeCB(a.VolumeCB()),
+        nParity(a.SiteSubset()),
+        dim {a.X(0), a.X(1), a.X(2), a.X(3)}, // *local* lattice dimensions, xyzt
+        L   {a.X(3), a.X(0), a.X(1), a.X(2)}  // *local* lattice dimensions, txyz
+      {
+        if constexpr (length != 24) {
```
maddyscientist: Make this a `static_assert`?

include/enum_quda.h
maddyscientist: That's unfortunate, and will be a difficult constraint going forward as we have been planning on making the interface fully embrace C99 (e.g., use of `bool`).

What's the reason for this desire to stick to a 35 year old standard?

include/gauge_field_order.h
```diff
@@ -2347,8 +2347,124 @@ namespace quda {
...
+
+      OpenQCDOrder(const GaugeField &u, Float *gauge_ = 0, Float **ghost_ = 0) :
+        LegacyOrder<Float, length>(u, ghost_),
+        gauge(gauge_ ? gauge_ : (Float *)u.data()), // pointer to the gauge field on CPU
+        volumeCB(u.VolumeCB()), // Volume and VolumeCB refer to the global lattice, if VolumeLocal, then local lattice
+        dim {u.X()[0], u.X()[1], u.X()[2], u.X()[3]}, // *local* lattice dimensions, xyzt
+        L   {u.X()[3], u.X()[0], u.X()[1], u.X()[2]}, // *local* lattice dimensions, txyz
+        nproc {comm_dim(3), comm_dim(0), comm_dim(1), comm_dim(2)} // txyz
+      {
+        if constexpr (length != 18) {
```
maddyscientist: `static_assert` here?

include/index_helper.cuh
```diff
@@ -1106,4 +1106,323 @@ namespace quda {
...
+     *
+     * @return     iup[ix][mu]
+     */
+    __device__ __host__ inline int iup(const int *x, const int mu, const int *L, const int *nproc)
+    {
+      int i, ret, xb[4], xn[4];
+
+      if ((x[mu]==(L[mu]-1))&&(nproc[mu]>1)) {
+
+        int cbs[4] = {setup_cbs(0, L), setup_cbs(1, L), setup_cbs(2, L), setup_cbs(3, L)};
```
maddyscientist, line 1378: Might not be practical, but in an ideal world the `cbs` numbers would be pre-computed on the host in `int_fastdiv` types.  This would allow for much faster division in the kernels.  Likely not performance critical, just mentioning it out of completeness.

include/quda_openqcd_interface.h
maddyscientist: This makes me wonder what hoops might be necessary when we switch to using `bool` in the interface...

lib/color_spinor_field.cpp
maddyscientist: Thanks for catching that.  The CPU ordering path isn't really used much, except for debugging.  I likely introduced this bug in #1384.

lib/comm_common.cpp
maddyscientist: To avoid possible issues in the future, best to make sure that any local variables in each path here are scoped or simply write the first path here as follows to avoid any local variable:
```c++
topo->cstar = static_cast<int *>(map_data)[0];
```

lib/copy_gauge_extended.cu
```diff
@@ -110,6 +110,15 @@ namespace quda {
       errorQuda("TIFR interface has not been built\n");
 #endif
 
+    } else if (out.Order() == QUDA_OPENQCD_GAUGE_ORDER) {
+
+#ifdef BUILD_OPENQCD_INTERFACE
+      using G = OpenQCDOrder<FloatOut, length>;
+      CopyGaugeEx<FloatOut, FloatIn, length, G, InOrder>(out, in, location, Out, In);
+#else
+      errorQuda("OPENQCD interface has not been built\n");
```
maddyscientist: General comment: `errorQuda` includes a line break (`\n`), so these should be removed.

lib/gauge_field.cpp
```diff
@@ -151,6 +151,25 @@ namespace quda {
         bytes = site_dim * (x[0] + 4) * (x[1] + 2) * (x[2] + 2) * (x[3] + 2) * nInternal * precision;
       } else if (order == QUDA_MILC_SITE_GAUGE_ORDER) {
         bytes = volume * site_size;
+      } else if (order == QUDA_OPENQCD_GAUGE_ORDER) {
+        /**
+         * With an openQCD gauge field, we need all the links of even lattice
+         * points in positive direction. These are links that lie in the buffer
+         * space that spans 7*BNDRY/4 gauge fields. These boundary fields are
```
maddyscientist, line 159: Curious where the factor of 7 comes from?

include/dslash_quda.h
maddyscientist: I don't think we've needed anything other in gamma5 in the past, so I never bothered to instantiate them.  No problem adding them and this is the correct place to do so.

I do question the difference in notation here though, for C-style counting from zero for gamma_1..gamma_4 and Fortran-style / physics notation for gamma_5.  Should we make these consistent?

include/communicator_quda.h
maddyscientist: Forgive my ignorance, but are the C* boundaries only relevant for fermions?

lib/openqcd_interface.cpp
```diff
@@ -0,0 +1,1360 @@
...
+  {"QUDA_MEMORY_DEVICE_PINNED",             std::to_string(QUDA_MEMORY_DEVICE_PINNED)},
+  {"QUDA_MEMORY_HOST",                      std::to_string(QUDA_MEMORY_HOST)},
+  {"QUDA_MEMORY_HOST_PINNED",               std::to_string(QUDA_MEMORY_HOST_PINNED)},
+  {"QUDA_MEMORY_MAPPED",                    std::to_string(QUDA_MEMORY_MAPPED)},
+  {"QUDA_MEMORY_MANAGED",                   std::to_string(QUDA_MEMORY_MANAGED)},
+  {"QUDA_MEMORY_INVALID",                   std::to_string(QUDA_MEMORY_INVALID)},
+  {"QUDA_CUSOLVE_EXTLIB",                   std::to_string(QUDA_CUSOLVE_EXTLIB)},
+  {"QUDA_EIGEN_EXTLIB",                     std::to_string(QUDA_EIGEN_EXTLIB)},
+  {"QUDA_EXTLIB_INVALID",                   std::to_string(QUDA_EXTLIB_INVALID)}
+};
```
maddyscientist: This is something we should probably standardize this for all enums and make available library wide.  @mathiaswagner did I imagine us talking about something like this a few years ago?

maddyscientist (commented):
Thank you for this PR, this is a great contribution to QUDA.  I have done an initial visual review at this point, and left a bunch of comments.  My only real concern of note is the demotion of the interface to being C89 instead of C99, but those concerns may be unwarranted.

Outside of those comments, somethings that will need to be done:

- [ ] Apply clang-format (see [here](https://github.com/lattice/quda/wiki/Coding-Conventions-and-Style#coding-style-and-clang-format))
- [ ] Create a wiki page of how to use OpenQ*D with QUDA, similar for example to what we have for [MILC}(https://github.com/lattice/quda/wiki/MILC-with-QUDA)

Is there any specific functionality that is added to QUDA that we should be adding to the internal tests?

lib/openqcd_interface.cpp
mathiaswagner: Yes, we had discussed that but seems we lost track.

chaoos:
> Thank you for this PR, this is a great contribution to QUDA. I have done an initial visual review at this point, and left a bunch of comments. My only real concern of note is the demotion of the interface to being C89 instead of C99, but those concerns may be unwarranted.
> 
> Outside of those comments, somethings that will need to be done:
> 
>     * [ ]  Apply clang-format (see [here](https://github.com/lattice/quda/wiki/Coding-Conventions-and-Style#coding-style-and-clang-format))
> 
>     * [ ]  Create a wiki page of how to use OpenQ*D with QUDA, similar for example to what we have for [MILC}(https://github.com/lattice/quda/wiki/MILC-with-QUDA)
> 
> 
> Is there any specific functionality that is added to QUDA that we should be adding to the internal tests?

Dear Kate,

I dealt with the two points: 

* see commit https://github.com/lattice/quda/pull/1414/commits/568dd3734e4aa20b99e0e18bd1f7d344e5e5bb9a for the clang-formatting
* The page for the wiki is written, but I cannot commit it, permission issue. I attach the patch here [openqxd.zip](https://github.com/lattice/quda/files/14418468/openqxd.zip)

Best,
Roman

maddyscientist:
Hi Roman,

I've added you to the lattice github group, so you should have permission to edit the wiki now (and push to branches in the lattice org).

Thx

maddyscientist:
@Jenkins ok to test

chaoos:
I was able to commit to the wiki, thanks for giving me access!
The CI/CD also seems to be fixed.

havogt:
cscs-ci run

include/dirac_quda.h
```diff
@@ -2606,6 +2606,8 @@ namespace quda {
           && (pc_type == QUDA_MATPC_EVEN_EVEN || pc_type == QUDA_MATPC_EVEN_EVEN_ASYMMETRIC))
         return true;
 
+      if (dirac_type == QUDA_WILSON_DIRAC || dirac_type == QUDA_CLOVER_DIRAC) return true;
```
maddyscientist: This addition here looks bogus: the Wilson operator is not Hermitian.  Is this addition intentional?

include/instantiate_dslash.h
```diff
@@ -38,6 +38,20 @@ namespace quda
 #else
       errorQuda("QUDA_RECONSTRUCT=%d does not enable reconstruct-8/9", QUDA_RECONSTRUCT);
 #endif
+#ifdef BUILD_OPENQCD_INTERFACE
```
maddyscientist: @kostrzewa noting that we could augment this for tmLQCD after this is merged to give recon 9/13 support, which I guess will help your perf?

maddyscientist (changes requested):
I've had a review of the latest code.  This is a great contribution, but a few things left outstanding to be resolved.  I have left some comments, but beyond those:
* The C89 changes are of concern, since we will be embracing a C99 interface, e.g., use of `bool`.
* The new openQCD interface is rather opaque to me.  Can this be better documented?
* Thanks for the wiki additions.  Is there a test input file or equivalent to allow us to run a test with this?

chaoos:
> The C89 changes are of concern, since we will be embracing a C99 interface, e.g., use of `bool`.

Dear Kate,
* The C99 thing is ok from our side. Should I just revert all changes I made in quda.h regarding this?
* We are currently in the phase of documenting, so give me some days for this. Do you want more comments in the code (I think every function/method has a javadoc-decoration), or conceptual things in the wiki?
* Regarding testing of openqxd functionality; we have some testing on the side of openqxd. If you want to test this in this repo, then you need to clone and compile the openqxd repo and run the tests from there. We are working on making this repo public, but there is still some politics going on: that's not in my power.

include/dirac_quda.h
chaoos: This is in the class `DiracG5M` (i.e. gamma5 times the Dirac operator), which is Hermitian. I had to change this, since the eigensolver didn't accept a non-Hermitian operator else. So this should be correct (only for `gamma5*D` and `D*gamma5` though).

maddyscientist:
@chaoos apologies for the slow reply at my end.  GTC and other travel got in the way.....

If you can revert the C99 -> C89 changes that would be great.  It would also narrow the scope of the PR which would also be helpful.

Regarding documentation, the basic rule is that all new functions should have doxygen markup next to their declarations, together with overall conceptual documentation in the wiki.  If you can write up the documentation to include cloning the openq*d and compiling the code to work with QUDA that would be great.  Understood that the repo isn't public yet, and that's fine for now.  Does that sound ok to you?

chaoos:
Hello

I'm sorry for the long wait! Finally I was able to finish this up.

I've reverted all the C89 changes, appled clang-format again, implemented CI/CD testing for all this on our side, and fixed the issues we still had.

So nothing really stops this anymore from being merged. Please review again. Let me know if there is something I can do to settle this.

Thanks and best wishes,
Roman

PS: I've created a spack package for QUDA (pull-request awaits approval, see https://github.com/spack/spack/pull/48939). This became very useful for our CSCS CI/CD workflow. Maybe it's also useful for others, and can be put in the wiki or so?

weinbe2:
@chaoos following up on last week's meeting---if you can resolve the new merge conflicts, we should be able to get this merged this week

include/clover_field_order.h
weinbe2: @maddyscientist filling this in for future reference, this is relevant for QCD+QED simulations where the QCD and QED clovers need to be separately constructed from the SU(3) and U(1) fields, respectively, then combined. We (QUDA) would need to be able to explicitly handle standalone U(1) fields.

The reason we can't do this directly from the unified U(3) = SU(3)xU(1) fields is because you can't (naively) disentangle the SU(3) and U(1) components. (...I guess you can divide the U(1) contribution out of the SU(3) field if you know the right relative scaling coefficient, but it would be pretty gross)

include/dslash_quda.h
weinbe2: I think a better solution could be creating `enum` for `QUDA_GAMMA_X`, `QUDA_GAMMA_Y`, `QUDA_GAMMA_Z`, `QUDA_GAMMA_T`, and `QUDA_GAMMA_5` (or something of that sort), then exposing a version of `ApplyGamma` that takes in these arguments. This can then be a thin wrapper around the `ApplyGamma` function that takes in 0 ... 4, and the thin wrapper can handle mapping the `enum` to the appropriate integers internally.

include/enum_quda_fortran.h
```diff
@@ -47,6 +47,7 @@
 #define QUDA_BQCD_GAUGE_ORDER 15 // expect *gauge mu even-odd spacetime+halos row-column order
 #define QUDA_TIFR_GAUGE_ORDER 16
 #define QUDA_TIFR_PADDED_GAUGE_ORDER 17
+#define QUDA_OPENQCD_GAUGE_ORDER 18
```
weinbe2: This (and other `enum`) will need to be updated once `feature/vectorize` is merged

include/quda_openqcd_interface.h
```diff
@@ -0,0 +1,525 @@
...
+ * interfacing between QUDA and the OpenQCD software.
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * OpenQCD inmitations of structs.
+ * Copied from flags.h, su3.h
```
weinbe2: We can't have code directly copied from `OpenQCD`. However, I see that the data layouts are publicly documented in the `docs` subdirectory of the OpenQCD code, as is noted in the `README` file, which I didn't need to look at the code to see.

Can you remove this explicit copy+paste and implement data structures, such as `openQCD_bc_params_t`, that match the data layout in the documentation, and cite the documentation appropriately?

weinbe2:
@chaoos are there ever cases where OpenQ*D would need both recon-13/9 _and_ recon-12/8 simultaneously? I'm wondering if we should make QCD+QED fields a compile-time option where, if it's enabled, Wilson-type ops are recon 18/13/9 and otherwise are 18/12/8. Otherwise we might need to go through some plumbing headaches to make sure, for ex, 18/13/12/9/8 are _all_ unit tested in a single build.

@maddyscientist what are your thoughts? I think it's good that we have support for QCD+QED workflows more broadly, but supporting five different reconstructs in a single build is unit-testing nightmare fuel.

include/dirac_quda.h
weinbe2: So this technically isn't proper, because `hermitian()` is for Hermitian positive definite, not just general Hermitian. This check is used for things like the Lanczos eigensolver, which is for HPD only. (If Lanczos is working for Hermitian-indefinite operators, that's actually... insane, and we should talk about it separately @chaoos )

To handle gamma5 ops correctly, you need to go through the Arnoldi solver, which is documented here: https://github.com/lattice/quda/wiki/QUDA%27s-eigensolvers#computing-the-gamma5-hermitian-spectrum

lib/openqcd_interface.cpp
```diff
@@ -0,0 +1,2228 @@
...
+    if (layout.N[dir] % 2 != 0) {
+      errorQuda("Error: Odd lattice dimensions are not supported\n");
+      exit(1);
+    }
+  }
+
+  setMPICommHandleQuda(&layout.quda_comm);
+
+#ifdef MULTI_GPU
+/* TODO: would we ever want to run with QMP COMMS? */
```
weinbe2: If `OpenQCD`/`OpenQxD` doesn't support QMP comms, then I think it's best to make the build error out if you try to compile with both the OpenQCD interface and QMP comms---if anything because there's no means to test it. More ideally, this check would be done at `cmake` config time.

If some need for OpenQCD + QMP comes up in the future, we can revisit it then.

lib/openqcd_interface.cpp
```diff
@@ -0,0 +1,2228 @@
...
+#include <color_spinor_field.h>
+#include <dslash_quda.h>
+#include <invert_quda.h>
+#include <index_helper.cuh>
+#include <multigrid.h>
+#include <mpi.h>
+
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+
+#define WITH_COMM(expr) do { if (in_comm()) { expr; } } while (0)
```
weinbe2, line 27: Two things:

1. You should put `#undef WITH_COMM` at the end of this file, just as a best-practices.
2. Out of curiosity, why does this need to be in `do {...} while (0)`?

lib/eig_trlm.cpp
```diff
@@ -26,7 +26,8 @@ namespace quda
     beta.resize(n_kr, 0.0);
 
     // Thick restart specific checks
-    if (n_kr < n_ev + 6) errorQuda("n_kr=%d must be greater than n_ev+6=%d\n", n_kr, n_ev + 6);
+    if (n_kr < n_ev + 6) errorQuda("n_kr=%d must be greater than or equal to n_ev+6=%d\n", n_kr, n_ev + 6);
+    if (n_kr < n_conv + 12) errorQuda("n_kr=%d must be greater than or equal to n_conv+12=%d\n", n_kr, n_conv + 12);
```
weinbe2, line 30: What was the reason for this? Was this a check that was missing all along, or was it empirical?

weinbe2:
@chaoos sorry this stalled for a while, I added some review comments. We can also schedule a call at some point to go over it.

include/quda_openqcd_interface.h
chaoos: I changed the names and implmementations of these structs + I remove an unneccessary one. They don't have to be alignment compatible anymore and a reference to the docs is added.

chaoos:
> are there ever cases where OpenQ*D would need both recon-13/9 _and_ recon-12/8 simultaneously?

Originally yes, but I saw that it makes stuff increasingly complicated. So I added a new Cmake flag `QUDA_QCD_PLUS_QED=ON/OFF` as you proposed to build Wilson ops with `QUDA_RECONSTRUCT_NO/8/12` if set to `OFF` (the default) or `QUDA_RECONSTRUCT_NO/9/13` if set to `ON`.

See https://github.com/lattice/quda/pull/1414/commits/452f6e67d8922262ce6c34013220d1efbe156832

lib/openqcd_interface.cpp
chaoos: True, I removed the QMP part and added a cmake condition to error if `QUDA_QMP=ON` and `QUDA_INTERFACE_OPENQCD=ON`.

lib/openqcd_interface.cpp
chaoos: 1. Added, thanks!
2. That one of these C shenanigans one has to be careful. If you have
```
if (x)
  WITH_COMM(foo());
else
  bar();
```
and the macro without the `do { } while (0)` the `else` binds to the inner `if`, not the outer one.

lib/eig_trlm.cpp
chaoos: That's long ago and I forgot, but I found my old commit message, see https://github.com/chaoos/quda/commit/ae170fe25f1bd80dec57ae1a39ea5d328f9ec20c
Trusting my past-me, it actually makes sense if yo look at https://github.com/chaoos/quda/blob/feature/openqxd/lib/eig_trlm.cpp#L127

include/dirac_quda.h
chaoos: I see, removed https://github.com/lattice/quda/pull/1414/commits/49211c40dcea8c1ef21c0910bc70748053520a8b

include/dslash_quda.h
chaoos: Good point, I removed the functions `gamma{0,1,2,3}` and wrote a `ApplyGamma` that takes an enum as direction, see https://github.com/lattice/quda/pull/1414/commits/d7fe0a237022b7f6cda97c8624a7d723ae39a5bb and https://github.com/lattice/quda/pull/1414/commits/bce45981bdbec211f522a2fc2f21a8c95ab88a2f

include/index_helper.cuh
```diff
@@ -1109,6 +1109,295 @@ namespace quda {
...
+
+    /**
+     * @brief      Returns the surface in direction mu
+     *
+     * @param      X     Extent in all 4 directions
+     * @param[in]  mu    Direction
+     *
+     * @return     Surface
+     */
+    __device__ __host__ inline int surface(const int *X, const int mu)
```
weinbe2: Can we replace `const int *X` with `const int X[4]` to be consistent with other kernels in `include/index_helper.cuh`, or is there a specific reason this needs to be a pointer? (Similar question for other routines below.)

include/quda.h
weinbe2: Just double-checking, is this still a concern, or has OpenQxD moved to C99?

include/quda_openqcd_interface.h
```diff
@@ -0,0 +1,522 @@
+#pragma once
+
+#include "mpi.h"
```
weinbe2: Can these be changed to `<mpi.h>` and `<pthread.h>`, or is this an OpenQxD requirement?

Also, if these are being baked in, these should also be guarded by a check that QUDA was specifically compiled with MPI, since QUDA does still support single GPU/single MPI rank compilation.

lib/copy_gauge_extended.cu
```diff
@@ -65,7 +65,7 @@ namespace quda {
 #else
         errorQuda("QUDA_RECONSTRUCT=%d does not enable reconstruct-8", QUDA_RECONSTRUCT);
 #endif
-#ifdef GPU_STAGGERED_DIRAC
+#if defined(GPU_STAGGERED_DIRAC) || defined(BUILD_OPENQCD_INTERFACE)
```
weinbe2: `defined(BUILD_OPENQCD_INTERFACE)` should be replaced by `defined(QUDA_QCD_PLUS_QED)` as appropriate

lib/copy_gauge_extended.cu
```diff
@@ -111,10 +111,18 @@ namespace quda {
       errorQuda("TIFR interface has not been built\n");
 #endif
 
+    } else if (out.Order() == QUDA_OPENQCD_GAUGE_ORDER) {
+
+#ifdef BUILD_OPENQCD_INTERFACE
```
weinbe2, line 116: And this is an exception for where we'd need it, of course...

lib/eig_trlm.cpp
weinbe2: Oh awesome, good catch!

lib/openqcd_interface.cpp
weinbe2: Wow, okay, thank you for explaining this.

include/quda.h
```diff
@@ -70,10 +70,8 @@ extern "C" {
     int staple_pad;   /**< Used by link fattening */
     int llfat_ga_pad; /**< Used by link fattening */
     int mom_ga_pad;   /**< Used by the gauge and fermion forces */
-    union {
-      bool use_split_gauge_bkup; /**< Used by gauge split buffers (default=true keep split gauge after usage)*/
-      int pad;                   /**< Forces 4-byte alignment */
-    };
+    bool use_split_gauge_bkup; /**< Used by gauge split buffers (default=true keep split gauge after usage)*/
```
weinbe2: I did some digging into the C/C++ standard and apparently there's no guarantee that `bool` is 1 byte. In practice, this seems to be the case, but since it's technically implementation-defined I'd like to make this more robust.

Can this be replaced by an `int` and then in various codes it's explicitly compared with 0 to check `true`/`false`? In that case it'd be safer...

weinbe2:
I'm happy with this PR from a visual standpoint. I still need to do build/run tests, in particular with QCD+QED, but that's more to make sure nothing unexpectedly broke, there are a few things I'd like to try beyond what's covered by our CI. This is awesome work, @chaoos .

Based on an offline convo, I'll take on updating the Wilson dslash tests to properly test recon-13 and -9 in a separate PR. This would be done by generating random U(3) instead of SU(3) fields. Since I've had to deal with this relatively recently with HISQ (as is relevant for long links) this should be relatively straightforward for me.

@maddyscientist , I believe you wanted to do some further review.

include/quda.h
chaoos: Unfortunately yes. Also in C99 with strict-prototypes this will error. I would prefer to leave it if that's no problem. But I added the void also in the function definition (I missed doing that). Now it's consistent at least. This way it is C-idiomatic (although probably not C++-idiomatic).

include/index_helper.cuh
chaoos: Done for all routines https://github.com/lattice/quda/pull/1414/commits/be3ba61cfbd69f71c7e7c384113aa48a5761a49b

include/quda_openqcd_interface.h
chaoos: Changed and added a cmake flag to build openqcd interface always with MPI enabled, see https://github.com/lattice/quda/pull/1414/commits/c7e3f74f60b5876c634887e0ad4b849f67ef58f2

lib/copy_gauge_extended.cu
chaoos: True, changed https://github.com/lattice/quda/pull/1414/commits/abbedf7dfa83805b282f86ecd2b53fc8cfde377f, https://github.com/lattice/quda/pull/1414/commits/9b7525d2639c4648843dfabfc7f42f5e01c612d0

include/quda.h
chaoos: I changed it to an int and replaced occurences approriately, see https://github.com/lattice/quda/pull/1414/commits/ddc9730a760ee688a30a22211f51af4759549951

include/quda.h
weinbe2: I'll need to make sure it doesn't break MILC for some strange reason (I think it'll be fine), but otherwise it is what it is and I accept it!

maddyscientist: Just wondering if we could augment the C99 compilation test (`tests/c_interface_test.c`) to enable strict prototypes to ensure this isn't accidentally regressed?

maddyscientist (approved):
Thank you @chaoos for all the work you have done and going through the review process to get this in shape to merge.

I have one small comment (regarding enabling strict prototyping for the C interface test), but beyond that, this is good to go.  Approving.

include/quda.h
weinbe2: MILC compiled just fine!

weinbe2:
> Thank you @chaoos for all the work you have done and going through the review process to get this in shape to merge.
> 
> I have one small comment (regarding enabling strict prototyping for the C interface test), but beyond that, this is good to go. Approving.

I'll take point on addressing the strict prototypic for the C interface as part of updating the unit tests to explicitly check recon-9/13 when QCD+QED is enabled, I'm happy to merge this. Thanks @chaoos for all of your work (and patience when I was slow to review it)!

weinbe2 (approved):

PR #1415, opened 2023-11-09
Make QUDA use new ROCm header locations

dmcdougall:
These header files are going to move in a future ROCm release, so we're just preparing for that change here.  These new locations have been available for several ROCm releases already, so this shouldn't be a disruptive change.

### DIFF
--- a/include/targets/hip/FFT_Plans.h
+++ b/include/targets/hip/FFT_Plans.h
@@ -2,7 +2,7 @@
 
 #include <quda_hip_api.h>
 #include <quda_internal.h>
-#include <hipfft.h>
+#include <hipfft/hipfft.h>
 
 #define FFT_FORWARD HIPFFT_FORWARD
 #define FFT_INVERSE HIPFFT_BACKWARD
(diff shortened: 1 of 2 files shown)


maddyscientist:
@dmcdougall will this fix the warnings regarding header location, and thus allow us to switch the ROCm CI target to use `STRICT` builds?

dmcdougall:
It'll likely fix the warnings regarding header location.  Switching to `STRICT` would need that there are no _other_ warnings.  I don't remember if we do or do not trigger compiler warnings in other parts of the codebase.

maddyscientist:
@dmcdougall there is a warning from a ROCm header, which is unexpected:
```
/opt/rocm-5.5.0/include/rocrand/rocrand_common.h:75:6: warning: "Disabled inline asm, because the build target does not support it." [-W#warnings]
    #warning "Disabled inline asm, because the build target does not support it."
 ```
 Are ROCm headers not included as system headers, which should suppress such warnings?

dmcdougall:
I don't know why we're warning there at all, to be honest.  It's not like we can expect the user to do anything meaningful if we don't generate inline assembly somewhere inside of rocrand for a specific hardware architecture.

Fortunately, it looks like the warning got taken away in 5.6: https://github.com/ROCmSoftwarePlatform/rocRAND/blob/release/rocm-rel-5.6/library/include/rocrand/rocrand_common.h#L69-L72

That preprocessor logic looks strange to me for other reasons, though.  I'll reach out to the rocrand folks to follow up on that.

maddyscientist (approved):

PR #1418, opened 2023-11-30
Prepare for hipMemoryType changes in ROCm 6

dmcdougall:
This is a breaking change happening in ROCm 6.  We guard on the new enum value.

The name change has been around since 5.5, so that's why we're not guarding it here.


maddyscientist (approved):

dmcdougall:
Thanks for the merge.  I'll work on cleaning up the compiler warnings.

PR #1419, opened 2023-11-30
Add more ROCm versions to CI

dmcdougall:
Let's see if this works... :)

### DIFF
--- a/.github/workflows/rocm-build-ci.yml
+++ b/.github/workflows/rocm-build-ci.yml
@@ -4,10 +4,13 @@ on: pull_request
 jobs:
   rocm-build:
     runs-on: [self-hosted, amd]
+    strategy:
+      matrix:
+        rocm: [ 5.5.0, 5.6.1, 5.7.2 ]
     steps:
       - uses: actions/checkout@v3
       - run: |
-          export ROCM_PATH=/opt/rocm-5.5.0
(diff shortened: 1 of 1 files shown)


maddyscientist:
@dmcdougall I see the inline assembly warnings are now gone with 5.6.  Is there any reason to keep around 5.5 or can we dump this and make the builds `STRICT` now?

dmcdougall:
> Is there any reason to keep around 5.5 or can we dump this and make the builds STRICT now?

You took the wind out of my sails :)

That is indeed my next step, I just wanted to sanity check the multiple version infrastructure did what I expected it to do.

Don't merge this yet; I'll update this PR with more commits.

dmcdougall:
@maddyscientist Can you kick off another build please?

dmcdougall:
Hmmm, these `STRICT` builds succeeded but there are still warnings.  I'll dig.

dmcdougall:
Oh [this](https://github.com/lattice/quda/blob/develop/lib/targets/hip/target_hip.cmake#L106) turns off the error (`STRICT` users `-Werror`).  It doesn't turn off the warning.

maddyscientist (approved):

maddyscientist:
@dmcdougall good to merge now from your point of view?

dmcdougall:
Yep.  Thanks.

PR #1427, opened 2024-01-09
start tagging kernels

jcosborn:
This is an initial set of changes to start tagging kernel operations that require special resources (e.g. shared memory) or other treatment (e.g. block synchronization).  This initial version only enables tagging for ThreadLocalCache and thread_array.


include/kernels/clover_deriv.cuh
```diff
@@ -33,9 +33,11 @@ namespace quda
     }
   };
 
-  template <typename Link, typename Force, typename Arg>
-  __device__ __host__ void computeForce(Force &force_total, const Arg &arg, int xIndex, int parity, int mu, int nu)
+  using computeForceOps = KernelOps<thread_array<int, 4>>;
```
weinbe2, line 36: I think we can get rid of this extra abstraction and just have `CloverDerivative` inherit directly from `KernelOps<thread_array<int, 4>>`

include/kernels/field_strength_tensor.cuh
```diff
@@ -169,14 +172,18 @@ namespace quda
...
       // 36 floating point operations here
     }
-    
+
     int munu_idx = (mu * (mu - 1)) / 2 + nu; // lower-triangular indexing
     arg.f(munu_idx, idx, parity) = F;
   }
 
-  template <typename Arg> struct ComputeFmunu {
+  template <typename Arg_> struct ComputeFmunu : computeFmunuCoreOps {
```
weinbe2, line 180: Similar comment as with clover derivatives, I think we should just have this inherit from `KernelOps<thread_array<int, 4>>`

include/kernels/gauge_utils.cuh
```diff
@@ -19,14 +20,16 @@ namespace quda
   // matrix+matrix = 18 floating-point ops
   // => Total number of floating point ops per function call
   // dims * (2*18 + 4*198) = dims*828
-  using computeStapleOps = thread_array<int, 4>;
-  template <typename Arg, typename Staple, typename Int>
-  __host__ __device__ inline void computeStaple(const Arg &arg, const int *x, const Int *X, const int parity, const int nu, Staple &staple, const int dir_ignore)
+  using computeStapleOps = KernelOps<thread_array<int, 4>>;
```
weinbe2, line 23: I appreciate the idea of declaring this once and using it across multiple files (`gauge_ape.cuh`, etc), but for something this simple, I think we should also just have the classes that currently inherit from `computeStapleOps` just textually inherit from `KernelOps<thread_array<int, 4>>` as well. Maybe I'm missing a particular reason for having this reused abstraction, though.

include/targets/generic/kernel_ops.h
```diff
@@ -50,4 +67,109 @@ namespace quda
...
+      return T::shared_mem_size(block, arg...);
+    }
+  };
+  template <> struct sharedMemSizeS<NoKernelOps> {
+    template <typename... Arg> static constexpr unsigned int size(dim3, const Arg &...) { return 0; }
+  };
+  template <typename... T> struct sharedMemSizeS<KernelOps<T...>> {
+    template <typename... Arg> static constexpr unsigned int size(dim3 block, const Arg &...arg)
+    {
+      return std::max({sharedMemSizeS<T>::size(block, arg...)...});
```
weinbe2, line 200: I'm a little lost but there's a lot of template meta-programming going on here---why exactly is this a `max` and not a `sum` or something of the sort?

include/kernels/gauge_wilson_flow.cuh
```diff
@@ -56,48 +56,65 @@ namespace quda
     }
   };
 
-  template <typename Arg>
-  __host__ __device__ inline auto computeStaple(const Arg &arg, const int *x, int parity, int dir)
+  template <typename Arg> struct computeStapleOpsWF {
```
weinbe2, line 59: This is a really nice example of the power of the composability of this abstraction, but it did leave me very confused for a while. For example, I wasn't exactly sure (and I'm still not exactly sure) why `using SymanzikOps = combineOps<computeStapleRectangleOps, KernelOps<StapOp, RectOp>>;` couldn't be `using SymanzikOps = combineOps<computeStapleRectangleOps, StapOp, RectOp>;`.

There may also be a typo where `using StapOp = ThreadLocalCache<Link, 0, computeStapleRectangeOps>;` should actually be `using StapOp = ThreadLocalCache<Link, 0, computeStapleOps>;` and it happens to work out because `computeStapleOps` and `computeStapleRectangeOps` are both `KernelOps<thread_array<int, 4>>;`... so maybe that would address my confusion.

In any case, I think overly-verbose documentation in this file and/or a wiki page that really digs into what's going on here as an example would be invaluable.

include/kernels/gauge_wilson_flow.cuh
```diff
@@ -56,48 +56,65 @@ namespace quda
...
-    switch (arg.wflow_type) {
-    case QUDA_GAUGE_SMEAR_WILSON_FLOW :
+    static_assert(Arg::wflow_type == QUDA_GAUGE_SMEAR_WILSON_FLOW || Arg::wflow_type == QUDA_GAUGE_SMEAR_SYMANZIK_FLOW);
+    if constexpr (Arg::wflow_type == QUDA_GAUGE_SMEAR_WILSON_FLOW) {
       // This function gets stap = S_{mu,nu} i.e., the staple of length 3,
-      computeStaple(arg, x, arg.E, parity, dir, Z, Arg::wflow_dim);
-      break;
-    case QUDA_GAUGE_SMEAR_SYMANZIK_FLOW :
+      computeStaple(ftor, x, arg.E, parity, dir, Z, Arg::wflow_dim);
+    }
```
weinbe2: Pedantic, but I think it would be better practice to have an `else if constexpr` here. I know it's essentially impossible for the conditions in each of the `if` statements to both be true, this is just a thought on a convention.

include/kernels/hisq_paths_force.cuh
```diff
@@ -669,8 +677,13 @@ namespace quda {
 
     };
 
-    template <typename Param> struct AllFiveAllSevenLink
-    {
+    template <typename Param> struct AllFiveAllSevenLinkOps {
+      static constexpr int cacheLen = Param::sig_positive ? 3 : 2;
```
weinbe2: `cache_len` instead of `cacheLen`. We're working towards the underscore convention instead of camelCase for simpler variables, `nColor` is just a pervasive beast we're afraid to deal with thus far...

weinbe2:
Well, first of all, this is a very impressive design/implementation, and I like it as a whole.

I can see the approach going on here---doing all of the work at compile time---and I appreciate it for what it is. My other point of reference is the way Kokkos hands around its "scratch" memory pointers, putting all of the onus on the user to get it right, which has cost me a lot of time and sanity.

I'm good on all of the plumbing work (sans one `max` vs `sum` question, but since the code isn't breaking I'm sure there's something my brain can't handle right now). Nearly all of my comments were related to my thoughts on conventions and some requests for more detailed documentation. I emphasized conventions so much because they bounced around a bit in the different files---which I understand is just a byproduct of the initial development process---it just made it hard to follow what was going on at first, so I'd want to agree of this type of thing before it gets merged!

include/kernels/clover_deriv.cuh
```diff
@@ -210,7 +214,7 @@ namespace quda
 
       for (int nu = 0; nu < 4; nu++) {
         if (nu == mu) continue;
-        computeForce<Link>(force, arg, x_cb, parity, mu, nu);
+        computeForce<Link>(force, *this, x_cb, parity, mu, nu);
```
weinbe2, line 217: Somewhat useless comment, but for whatever reason I don't "like" how `*this` is being passed in as an argument---purely textually. My only suggestion would be to have a `const auto &ftor = *this;` somewhere higher up. I can also just deal, but maybe this bugs other people too?

include/kernels/gauge_utils.cuh
jcosborn: The reason for the abstraction was to keep the list of kernel operations used by a function near that function, so that if the requirements changed, it only needs to be changed in one place, which is right above the function being modified.  Also, it serves as documentation for the main kernel, explaining where the resources are used (which would otherwise not be obvious if many functions are called from the main kernel).  Since the kernels don't change that often, I would agree that the first reason may not matter much, but I think the documentation effect of using an alias for the operations needed for the function calls a kernel makes may make it worthwhile.

include/targets/generic/kernel_ops.h
jcosborn: I need to add more documentation, I'll address that in an update.  The short answer is that it is assumed that shared memory is used sequentially (or at least not concurrently), so that two different objects can use the same shared memory, unless the offset has been explicitly specified (in which case the offset has already forced the needed memory size for the offset object to include a sum).  So the total size needed among all objects is just the max of the size of each.

include/kernels/gauge_utils.cuh
weinbe2: That makes sense; I agree with your evaluation that it being self-documenting makes it worthwhile here, so I rescind my comment!

include/targets/generic/kernel_ops.h
weinbe2: Ah yeah, more documentation would help. Your rationale makes sense and is consistent with how you engineered this, I like how it's not going to allocate more shared memory than necessary by construction.

I think my big concern is it'll be hard to debug an issue when someone accidentally forgets to use `combineOps`, etc, to accumulate the total needed shared memory... but between extra documentation and a few of us being familiar with the details, it'll be easy enough to help other developers debug.

include/kernels/gauge_wilson_flow.cuh
jcosborn: Right now, combineOps only takes KernelOps or NoKernelOps types as template parameters, and combines them into a single combined KernelOps (or NoKernelOps) type as appropriate.  We could allow other types (like thread_array, etc.) as template arguments, but it adds to the complexity of the template programming, and makes it a bit harder to prevent users from accidentally using an unintended type (we would need to explicitly allow only the valid types, which I was trying to avoid so it is possible to add new ops types with only making minimal changes to the supporting infrastructure).  Yes, this does point out another place that needs better documentation, thanks.

As for using computeStapleOps vs. computeStapleRectangeOps, you caught me being a bit lazy there :) .  One branch uses computeStaple (Arg::wflow_type == QUDA_GAUGE_SMEAR_WILSON_FLOW) and the other computeStapleRectangle (Arg::wflow_type == QUDA_GAUGE_SMEAR_SYMANZIK_FLOW), so it really should choose between the two.  I'll fix that.

weinbe2: The reasoning makes sense, thanks for explaining!

jcosborn: Actually, the `StapOp` alias is only used in the case that `computeStapleRectangle` is called, so I think the code is correct as is (`StapOp` should use the offset of `computeStapleRectangeOps`).

include/kernels/clover_deriv.cuh
jcosborn: I don't have a strong preference either way.  Maybe we can wait and see what others say.

include/kernels/gauge_wilson_flow.cuh
jcosborn: 6a727a700e87b08f30a3874877cd11e927b95f90

include/kernels/hisq_paths_force.cuh
jcosborn: 6a727a700e87b08f30a3874877cd11e927b95f90

jcosborn:
Thanks for taking time to review @weinbe2.  I think I've addressed all the comments except for the issue of whether to make an alias for `*this` or pass it directly.  I can change it to an alias if there is an overall preference to do so.

I added more documentation in the headers.  Let me know if there is anything missing or could be made clearer.

I also merged in the latest develop so it should mostly ready now, pending final review.

include/kernels/clover_deriv.cuh
weinbe2: I'm fine with leaving it as-is for this PR, it's far from the first time I've seen this type of thing done via passing `*this`

weinbe2:
Thanks @jcosborn , I think the documentation you added more than suffices. It's made both the underlying assumptions and proper usage clearer. The PR is good to go in my book.

As for:
> [...] except for the issue of whether to make an alias for *this or pass it directly.

I noted this elsewhere, but to make the thread easier to follow: I'm ultimately fine leaving it as `*this`.

@maddyscientist , do you want to do a last review?

include/kernels/gauge_wilson_flow.cuh
```diff
@@ -56,48 +56,64 @@ namespace quda
...
-  template <typename Arg>
-  __host__ __device__ inline auto computeStaple(const Arg &arg, const int *x, int parity, int dir)
+  template <typename Arg> struct computeStapleOpsWF {
+    using real = typename Arg::real;
+    using Link = Matrix<complex<real>, Arg::nColor>;
+    using WilsonOps = computeStapleOps;                                  // Ops for case of QUDA_GAUGE_SMEAR_WILSON_FLOW
+    using StapOp = ThreadLocalCache<Link, 0, computeStapleRectangleOps>; // offset by computeStapleRectangleOps
+    using RectOp = ThreadLocalCache<Link, 0, StapOp>;                    // offset by StapOp
+    using SymanzikOps = combineOps<computeStapleRectangleOps, KernelOps<StapOp, RectOp>>; // GAUGE_SMEAR_SYMANZIK_FLOW
+    using Ops = std::conditional_t<Arg::wflow_type == QUDA_GAUGE_SMEAR_SYMANZIK_FLOW, SymanzikOps, WilsonOps>;
```
maddyscientist, line 66: the use of `std::conditional_t` will break Jitify.  Can you switch this out or just `quda::conditional_t` (https://en.cppreference.com/w/cpp/types/conditional)

maddyscientist: Never mind, I rescind that comment, we already support `conditional_t` in Jitify.

include/targets/generic/thread_array.h
```diff
@@ -22,20 +22,18 @@ namespace quda
...
-    __device__ __host__ constexpr thread_array(T first, const Ts... other) :
-      array_(sharedMem()[target::thread_idx_linear<3>()])
-    {
-      array_ = array<T, n> {first, other...};
-    }
+    constexpr thread_array(const thread_array<T, n, O> &) = delete;
 
-    __device__ __host__ T &operator[](int i) { return array_[i]; }
-    __device__ __host__ const T &operator[](int i) const { return array_[i]; }
+    __device__ __host__ inline T &operator[](const int i) { return array_[i]; }
```
maddyscientist: Why set these arguments as `const`?  Seems rather redundant for copy by value integers.

maddyscientist:
No objections from me with this PR.  @weinbe2 did you do any run-time sanity checks as part of your review?

include/targets/generic/thread_array.h
jcosborn: No real reason, I've removed them and a few unnecessary `inline`s too (since member functions are inline by default).
3b71bb635343325aa92a9d5e0262303f1f4f7549

weinbe2:
cscs-ci run

weinbe2 (approved):
> No objections from me with this PR. @weinbe2 did you do any run-time sanity checks as part of your review?

All correctness checks pass.

Runtime is good. There's a reproducible sub-sub-percentile regression on H100-80GB-HBM3 with CUDA 12.4 at a 24^4 volume, nothing I see as worth digging into. This has my approval.

@maddyscientist what are your thoughts on the cscs completing-in-time woes?

maddyscientist:
cscs-ci run

maddyscientist (approved):
Approved.  I've re-kicked off the CSCS CI, hopefully it squeaks through this time and then we can merge.

PR #1429, opened 2024-01-09
TimeProfile fixes

maddyscientist:
This PR improves the robustness of nesting calls to `TimeProfile` which were somewhat broken.  Notable this restores SVD deflation.
* Overhaul of linear solvers and eigen solvers to use `getProfile()` instead of manually tracking a `TimeProfile` reference
* `TimeProfile` stack must be a class instance and not static to prevent badness when multiple `TimeProfile` instances are being used
* Fix precision setting for `eigensolveQuda` which was previously ignoring the user selected precision
* Removal of unnecessary synchronization when nesting calls to the same kind
* Misc cleanup and fixes


lib/multigrid.cpp
```diff
@@ -15,15 +15,13 @@ namespace quda
...
-    Solver(*param.matResidual, *param.matSmooth, *param.matSmoothSloppy, *param.matSmoothSloppy, param, profile),
+  MG::MG(MGParam &param) :
+    Solver(*param.matResidual, *param.matSmooth, *param.matSmoothSloppy, *param.matSmoothSloppy, param),
     param(param),
     transfer(0),
     resetTransfer(false),
     presmoother(nullptr),
     postsmoother(nullptr),
-    profile_global(profile_global),
-    profile("MG level " + std::to_string(param.level), false),
```
weinbe2, line 26: Does this mean there are no longer separate profiler instances for each MG level? Or am I missing something? The breakdowns for each level were in principle nice (though sometimes they seemed a bit quirky).

maddyscientist: Yes, they are gone.  They were indeed very quirky, and didn't report the coarse level properly previously.

I introduced them when I first created the MG solver in QUDA to get a handle of coarse grid time, but their use is somewhat redundant now (given the kernel profile for example).  I might re-introduce them in another form at some point, once more thought has gone into evolving the `TimeProfile`.

weinbe2:
@maddyscientist looking at the `CSCS` output it looks like the eigensolver unit tests are failing due to legitimate convergence issues, can you give this a look?

weinbe2:
This is looking great. I uncovered an issue where when you save vectors with SVD solves, it'll save the `2N` right _and_ left singular vectors, but then on the load looks for `N` values, and as such errors out. The fix is to only have it save the `N` right singular vectors.

This is a reasonable solution: the left singular vectors can be trivially computed from the right ones, _and_ the process of doing so happens regardless during the re-computation of the singular _values_ on the load. (It also allows a workflow where someone calculates the _eigen_vectors of MdagM separately, stores them, and reloads them in an SVD deflation workflow... doing this in reverse is how I discovered this issue in the first place.)

weinbe2:
For reference, a working deflation command:

```
 mpirun -np 1 ./invert_test --verbosity verbose --mass-normalization mass --niter 10000 --tol 1e-5 --mass -0.97 \
                            --solve-type direct --solution-type mat --inv-type bicgstab \
                            --inv-deflate true --eig-use-poly-acc false --eig-spectrum SR \
                            --eig-n-conv 128 --eig-n-ev 130 --eig-n-kr 256 --eig-tol 1e-6 \
                            --eig-compute-svd true
```

And an MG with deflation command:

```
mpirun -np 1 ./invert_test \
 --prec double --prec-sloppy single --prec-null single --prec-precondition single \
 --recon 12 --recon-sloppy 8 --recon-precondition 8 \
 --dim 16 16 16 16 --gridsize 1 1 1 1 --load-gauge l16t16b7p0 --partition 0 \
 --kappa 0.155 --mu 0.00072 --clover-coeff 0.001 \
 --dslash-type wilson --compute-clover true --tol 1e-10 \
 --verbosity verbose --solve-type direct --solution-type mat --inv-type gcr \
 --inv-multigrid true --mg-levels 3 \
 --mg-block-size 0 4 4 4 4 --mg-nvec 0 24 \
 --mg-block-size 1 2 2 2 2 --mg-nvec 1 32 \
 --mg-setup-tol 0 5e-7 --mg-setup-tol 1 5e-7 --mg-setup-inv 0 cgnr --mg-setup-inv 1 cgnr \
 --mg-mu-factor 2 70.0 \
 --nsrc 1 --niter 250 \
 --mg-smoother 0 ca-gcr --mg-smoother-solve-type 0 direct --mg-nu-pre 0 0 --mg-nu-post 0 4 \
 --mg-smoother 1 ca-gcr --mg-smoother-solve-type 1 direct --mg-nu-pre 1 0 --mg-nu-post 1 4 \
 --mg-coarse-solver 1 gcr --mg-coarse-solve-type 1 direct --mg-coarse-solver-tol 1 0.35 --mg-coarse-solver-maxiter 1 10 \
 --mg-coarse-solver 2 gcr --mg-coarse-solve-type 2 direct-pc --mg-coarse-solver-tol 2 0.01 --mg-coarse-solver-maxiter 2 20 \
 --mg-verbosity 0 verbose --mg-verbosity 1 verbose --mg-verbosity 2 verbose \
 --mg-eig 2 true --mg-eig-type 2 trlm --mg-nvec 2 64 --mg-eig-n-ev 2 66 --mg-eig-n-kr 2 198 \
 --mg-eig-tol 2 1e-4 --mg-eig-use-poly-acc 2 false --mg-eig-use-dagger 2 false --mg-eig-use-normop 2 true --mg-eig-spectrum 2 SR
```

Command to generate the config used for the above MG run:
```
mpirun -np 1 ./heatbath_test --dim 16 16 16 16 --save-gauge l16t16b7p0   --heatbath-beta 7.0 --heatbath-coldstart true --heatbath-num-steps 10 --heatbath-warmup-steps 1000
```

weinbe2:
@cpviolator can you give a quick look at what I did in https://github.com/lattice/quda/pull/1429/commits/79c6dee2fd81524fe9cdb9f9ede59ea3368d794b related to saving singular vectors and give it a thumbs up/down? I don't want to blanket approve and merge a PR that I've contributed to :)

I described the issue it's supposed to fix above, but the tl;dr is when saving singular vectors it'd save both the right- and left-singular vectors, while on the load it only expected the right-singular vectors, causing it to error out. My narrow fix is to only have it save the right singular vectors, which fixes both the error and I believe is safe because the eigensolver populates the left singular vectors when it recomputes the singular _values_ anyway. (It also worked empirically---but I wanted to make sure there isn't some subtlety in some workflow I may have missed.)

cpviolator:
Yeah, this looks good. I never really used the SVD much after writing it so I'm glad this error has been picked up in a production case. Totally correct to save/load the minimal amount of data and reconstruct what is needed on the fly.

I'll give this a once over with Arnoldo and TRLM to make doubly sure, but I think this fix will be fine with the other solvers.

weinbe2:
> Yeah, this looks good. I never really used the SVD much after writing it so I'm glad this error has been picked up in a production case. Totally correct to save/load the minimal amount of data and reconstruct what is needed on the fly.
> 
> I'll give this a once over with Arnoldo and TRLM to make doubly sure, but I think this fix will be fine with the other solvers.

I did check with TRLM but not with Arnoldi, so that you for covering that---and regardless more independent checks are always good. Let me know if you hit any issues, otherwise I think this PR will be good to go.

weinbe2 (approved):
This is great, thanks for this cleanup @maddyscientist !

(Note: merging tentative on @cpviolator 's independent review of my bugfix commit)

weinbe2:
I'm going to give this a merge since I did at least test Lanczos so I can confirm that's good, and we can address Arnoldi separately if need be. @cpviolator --- if you find an issue post-merge, let me know and we'll get it fixed asap.

PR #1441, opened 2024-02-26
First attempt at porting 3DLaplace projection kernel.  Multi-node spl…

maddyscientist:
This PR adds the ability to handle the sink projection part of the Laph workflow:
* take a set of quark fields, a set of Laplace eigenvectors, and compute the 3-d spatial projection between these
* presently expects the quark fields are Wilson-like (nSpin=4) and the eigenvectors are 3-d Laplace (nSpin=1)
* rudimentary tiling optimization has been implemented, where we tile both the host->device vector download, and also tile at the kernel level
* a full OMP parallel test code has been included, and multiple tests predefined
* some changes have been to the `colorspinor::FloatNOrder` accessor to better support multi-RHS, like the fine-grained accessor, we now have the ability to enable the accessor without ghost-zone support to reduce the kernel argument footprint
* a bug was found with the block size settings with `TunableMultiReduction` where an invalid grid size could be chosen if the number of items being reduced was too large


maddyscientist:
This PR is ready for review.  @walkloud and @andrewhanlon, can you confirm that this code is working as expected from the calling code?  @weinbe2 can you review from the QUDA point of view?

tests/laph_test.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  // initialize EVs
+  cs_param.nSpin = 1;
+  std::vector<ColorSpinorField> evList(nEv);
+  for (auto &e : evList) {
+    e = ColorSpinorField(cs_param);
+    spinorNoise(e, rng, QUDA_NOISE_GAUSS);
+  }
+
+  // host reference [nSink][nEv][Lt][nSpin][complexity]
```
weinbe2, line 76: I don't necessarily like how this host verification is inlined within the test file as opposed to being placed in a separate file in the host reference directory, but it's not a hill I'm going to die on.

lib/slaph_interface.cpp
```diff
@@ -0,0 +1,106 @@
+#include <quda.h>
```
weinbe2: This should all go in `interface_quda.cpp`. I'm not opposed to refactoring the interface file at some point, it's very messy, but that's a separate discussion and we shouldn't change conventions here.

maddyscientist: [ec81a2e](https://github.com/lattice/quda/pull/1441/commits/ec81a2e9251f248f1abbb25d49e671f3266355ee)

andrewhanlon (approved):
It's been verified that using the new projection kernel implemented here gives correct results. The test was done on a 24^3 x 48 ensemble using 164 eigenvectors. The projections were performed on each of the solutions from 64 inversions (all dilutions on a single source time). The test used 4 of these solutions at a time and set the quark and evec tile sizes equal to 4 and 164, respectively. The result was compared with a cpu version of the projection.

Edit: minor correction and some more info

weinbe2 (approved):
This looks great. Thanks @maddyscientist !

PR #1442, opened 2024-03-01
fix chroma regression 

mathiaswagner:
Small notify that fixes an increased iteration count observed in Chroma


maddyscientist:
This change doesn't make sense to me.  We never use the MR solver as a preconditioner for PCG.

maddyscientist:
I've taken care of this issue in `feature/mrhs-solvers`, so closing this PR.

PR #1437, opened 2024-02-06
Feature/mrhs prep

maddyscientist:
This is primarily a cleanup PR, paving the way for multi-RHS solvers:
* Added `virtual Solver::operator(cvector<ColorSpinorField> &x, cvector_ref<const ColorSpinorField> &b)` method in preparation for multi-RHS
  * For now, this just does a trivial loop of RHS, but will be specialized for various solvers in subsequent PR
* Add copy and move constructors, as well as copy and move assignment operators for `CloverField`
* Modernize CG, CG3 and SD solvers
* Added CG3 and SD solvers to the automated testing
  * CG3 is in an incredibly unstable solver, and does not play well with reliable updates: hence CG3 is only tested in pure fp64
  * We do not test the SD solver for staggered, as due to the operator's spectral properties, it takes many iterations to converge
  * Some cleanup and removal of boiler plate

PR #1443, opened 2024-03-02
Feature/mrhs prep part deux

maddyscientist:
This PR continues the multi-RHS scaffolding from #1437:
* All BLAS functions are now multi-RHS aware (albeit for-loop wrappers around single RHS operations)
  * (Future work will fuse these into single kernels)
* BLAS regression tests are much more robust with most of the tests now tested using elementary equivalent compositions, as opposed to simply comparing CPU and GPU
* Block BLAS functions are now in their own namespace `blas::block` to avoid any collisions with the new multi-RHS aware BLAS functions
* Legacy block BLAS functions (those that take a vector of pointers) are now in the `blas::legacy` namespace.
  * Hopefully these will be deleted soon
* `Dirac::prepare` and `Dirac::reconstruct` functions are now multi-RHS aware
* General clean up of the `Dirac` classes to reduce boiler plate

### DIFF
--- a/include/blas_helper.cuh
+++ b/include/blas_helper.cuh
@@ -552,4 +552,16 @@ namespace quda
 
   } // namespace blas
 
+  template <typename A, typename B> void check_size(const A &a, const B &b)
+  {
+    if (a.size() != b.size()) errorQuda("Mismatched sizes a=%lu b=%lu", a.size(), b.size());
+  }
+
+  template <typename A, typename B, typename... Args> void check_size(const A &a, const B &b, const Args &...args)
+  {
+    check_size(a, b);
+    check_size(a, args...);
+    check_size(b, args...);
+  }
+
 } // namespace quda
--- a/include/blas_helper.cuh
+++ b/include/blas_helper.cuh
@@ -560,7 +560,6 @@ namespace quda
   template <typename A, typename B, typename... Args> void check_size(const A &a, const B &b, const Args &...args)
   {
     check_size(a, b);
-    check_size(a, args...);
     check_size(b, args...);
   }
 
--- a/include/blas_quda.h
+++ b/include/blas_quda.h
@@ -280,7 +280,7 @@ namespace quda {
        @brief Compute the maximum absolute real element of a field
        @param[in] a The field we are reducing
     */
-    double max(const ColorSpinorField &x);
+    cvector<double> max(cvector_ref<const ColorSpinorField> &x);
 
     /**
        @brief Compute the maximum real-valued deviation between two
@@ -288,19 +288,27 @@ namespace quda {
        @param[in] x The field we want to compare
        @param[in] y The reference field to which we are comparing against
     */
-    array<double, 2> max_deviation(const ColorSpinorField &x, const ColorSpinorField &y);
+    cvector<array<double, 2>> max_deviation(cvector_ref<const ColorSpinorField> &x,
+                                            cvector_ref<const ColorSpinorField> &y);
+
+    inline array<double, 2> max_deviation(const ColorSpinorField &x, const ColorSpinorField &y)
+    {
+      return max_deviation(cvector_ref<const ColorSpinorField>(x), cvector_ref<const ColorSpinorField>(y));
+    }
 
     /**
        @brief Compute the L1 norm of a field
        @param[in] x The field we are reducing
     */
-    double norm1(const ColorSpinorField &x);
+    cvector<double> norm1(cvector_ref<const ColorSpinorField> &x);
 
     /**
        @brief Compute the L2 norm (||x||^2) of a field
        @param[in] x The field we are reducing
     */
-    double norm2(const ColorSpinorField &x);
+    cvector<double> norm2(cvector_ref<const ColorSpinorField> &x);
+
+    inline double norm2(const ColorSpinorField &x) { return norm2(cvector_ref<const ColorSpinorField> {x}); }
 
     /**
        @brief Compute y += a * x and then (x, y)
@@ -308,14 +316,20 @@ namespace quda {
        @param[in] x input vector
        @param[in,out] y update vector
     */
-    double axpyReDot(double a, const ColorSpinorField &x, ColorSpinorField &y);
+    cvector<double> axpyReDot(cvector<double> &a, cvector_ref<const ColorSpinorField> &x,
+                              cvector_ref<ColorSpinorField> &y);
 
     /**
        @brief Compute the real-valued inner product (x, y)
        @param[in] x input vector
        @param[in] y input vector
     */
-    double reDotProduct(const ColorSpinorField &x, const ColorSpinorField &y);
+    cvector<double> reDotProduct(cvector_ref<const ColorSpinorField> &x, cvector_ref<const ColorSpinorField> &y);
+
+    inline double reDotProduct(const ColorSpinorField &x, const ColorSpinorField &y)
+    {
+      return reDotProduct(cvector_ref<const ColorSpinorField> {x}, cvector_ref<const ColorSpinorField> {y});
+    }
 
     /**
        @brief Compute z = a * x + b * y and then ||z||^2
@@ -325,7 +339,8 @@ namespace quda {
        @param[in] y input vector
        @param[in,out] z update vector
     */
-    double axpbyzNorm(double a, const ColorSpinorField &x, double b, const ColorSpinorField &y, ColorSpinorField &z);
+    cvector<double> axpbyzNorm(cvector<double> &a, cvector_ref<const ColorSpinorField> &x, cvector<double> &b,
+                               cvector_ref<const ColorSpinorField> &y, cvector_ref<ColorSpinorField> &z);
 
     /**
        @brief Compute y += a * x and then ||y||^2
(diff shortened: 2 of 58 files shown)


include/blas_helper.cuh
```diff
@@ -552,4 +552,16 @@ namespace quda
 
   } // namespace blas
 
+  template <typename A, typename B> void check_size(const A &a, const B &b)
+  {
+    if (a.size() != b.size()) errorQuda("Mismatched sizes a=%lu b=%lu", a.size(), b.size());
+  }
+
+  template <typename A, typename B, typename... Args> void check_size(const A &a, const B &b, const Args &...args)
```
weinbe2, line 560: This gives me pause because I think the number of comparisons grows exponentially and has redundant comparisons... I can't quite think of a sane fix, though. What are your thoughts @maddyscientist ?

include/blas_quda.h
```diff
@@ -28,261 +28,308 @@ namespace quda {
...
        @brief Apply the operation y = a * x
-       @param[in] a scalar multiplier
-       @param[in] x input vector
-       @param[out] y output vector
+       @param[in] a scalar multiplier set
+       @param[in] x input vector set
+       @param[out] y output vector set
     */
-    void axy(double a, const ColorSpinorField &x, ColorSpinorField &y);
+    void axy(cvector<Complex> &a, cvector_ref<const ColorSpinorField> &x, cvector_ref<ColorSpinorField> &y);
```
weinbe2, line 42: `cvector<double>`?

include/blas_quda.h
```diff
@@ -28,261 +28,308 @@ namespace quda {
...
 
     /**
        @brief Apply the rescale operation x = a * x
-       @param[in] a scalar multiplier
-       @param[in] x input vector
+       @param[in] a scalar multiplier set
+       @param[in] x set of vectors
     */
-    inline void ax(double a, ColorSpinorField &x) { axy(a, x, x); }
+    inline void ax(cvector<Complex> &a, cvector_ref<ColorSpinorField> &x) { axy(a, x, x); }
```
weinbe2, line 49: `cvector<double>` (again)?

include/blas_quda.h
```diff
@@ -28,261 +28,308 @@ namespace quda {
...
-       @param[in] y input vector
-       @param[out] z output vector
+       @param[in] a scalar multiplier set
+       @param[in] x input vector set
+       @param[in] b scalar multiplier set
+       @param[in] y input vector set
+       @param[out] z output vector set
     */
-    void axpbyz(double a, const ColorSpinorField &x, double b, const ColorSpinorField &y, ColorSpinorField &z);
+    void axpbyz(cvector<double> &a, cvector_ref<const ColorSpinorField> &x, cvector<double> &b,
```
weinbe2, line 59: Convention choice, should we make these `const cvector<double> &a` (which I'm sure will have some recursing casting nightmares, we'll have to make things const all the way down)?

include/blas_quda.h
```diff
@@ -28,261 +28,308 @@ namespace quda {
...
 
     /**
        @brief Compute the real-valued inner product (x, y)
        @param[in] x input vector
        @param[in] y input vector
     */
-    double reDotProduct(const ColorSpinorField &x, const ColorSpinorField &y);
+    cvector<double> reDotProduct(cvector_ref<const ColorSpinorField> &x, cvector_ref<const ColorSpinorField> &y);
+
+    inline double reDotProduct(const ColorSpinorField &x, const ColorSpinorField &y)
```
weinbe2, line 329: What's the reason why we needed an explicit non-`cvector` definition here? Is this something that needed to temporarily go in before a separate outer-product-of-inner-product namespace was added?

include/blas_quda.h
```diff
@@ -408,40 +478,58 @@ namespace quda {
...
        @brief Computes ||x||^2, ||r||^2 and the MILC/FNAL heavy quark
        residual norm
        @param[in] x input vector
        @param[in] r input vector (residual vector)
     */
-    double3 HeavyQuarkResidualNorm(const ColorSpinorField &x, const ColorSpinorField &r);
+    cvector<double3> HeavyQuarkResidualNorm(cvector_ref<const ColorSpinorField> &x,
+                                            cvector_ref<const ColorSpinorField> &r);
+
+    inline double3 HeavyQuarkResidualNorm(const ColorSpinorField &x, const ColorSpinorField &r)
```
weinbe2, line 493: Similar comment as above, any reason why we need a separate non-`cvector_ref` version here?

include/dirac_quda.h
```diff
@@ -517,11 +535,11 @@ namespace quda {
     virtual void M(ColorSpinorField &out, const ColorSpinorField &in) const;
     virtual void MdagM(ColorSpinorField &out, const ColorSpinorField &in) const;
 
-    virtual void prepare(ColorSpinorField* &src, ColorSpinorField* &sol,
-			 ColorSpinorField &x, ColorSpinorField &b,
-			 const QudaSolutionType) const;
-    virtual void reconstruct(ColorSpinorField &x, const ColorSpinorField &b,
-			     const QudaSolutionType) const;
+    virtual void prepare(cvector_ref<ColorSpinorField> &out, cvector_ref<ColorSpinorField> &in,
```
weinbe2, line 538: Convention choice, do we want to copy+paste the doxygen from the parent classic for each derived class? I think it'd be the right idea for actually generating `doxygen`, but it'd get pretty verbose pretty quickly. It's not a hill I'll die on one way or the other.

include/reference_wrapper_helper.h
```diff
@@ -312,4 +313,64 @@ namespace quda
     return out;
   }
 
+  /**
+     Derived specializaton of std::vector<T>
+     which allows us to write generic multi-scalar functions.
+   */
+  template <class T> struct vector : public std::vector<T> {
```
weinbe2, line 320: I'm not sure that I love calling this class `vector`, which (namespace aside) directly conflicts with a C++ container (yes, I know we have the same issue with `complex`, that's a whole other can of worms). Another reason I don't love it is because we have `using vector = ...` higher up in this file, which could lead to confusion.

All that said, I don't necessarily have a suggestion for a better name (right now)...

lib/blas_quda.cu
```diff
@@ -134,89 +134,127 @@ namespace quda {
...
+                cvector_ref<const ColorSpinorField> &y, cvector_ref<ColorSpinorField> &z)
     {
-      instantiate<axpbyz_, Blas, true>(a, b, 0.0, x, y, x, x, z);
+      check_size(a, x, b, y, z);
+      for (auto i = 0u; i < x.size(); i++)
+        instantiate<axpbyz_, Blas, true>(a[i], b[i], 0.0, x[i], y[i], x[i], x[i], z[i]);
     }
 
-    void axy(double a, const ColorSpinorField &x, ColorSpinorField &y)
+    void axy(const cvector<Complex> &a, cvector_ref<const ColorSpinorField> &x, cvector_ref<ColorSpinorField> &y)
```
weinbe2, line 145: Similar to above, `cvector<Complex>` to `cvector<double>`?

lib/dirac_clover_hasenbusch_twist.cpp
```diff
@@ -33,25 +33,25 @@ namespace quda
     if (!asymmetric) {
       if (matpcType == QUDA_MATPC_EVEN_EVEN) {
```
weinbe2: Can we formulate this conditional in terms of `this_parity`/`other_parity` or not really?

tests/blas_test.cpp
```diff
@@ -487,43 +487,43 @@ class BlasTest : public ::testing::TestWithParam<::testing::tuple<int, int>>
         break;
 
       case Kernel::axpy_block:
-        for (int i = 0; i < niter; ++i) blas::axpy(Ar, xmD, ymoD);
+        for (int i = 0; i < niter; ++i) blas::block::axpy(Ar, xmD, ymoD);
```
weinbe2, line 490: Only because we're here---is there a reason why we didn't have a corresponding host verify call _here_? I'm guessing there's one somewhere in the depths of this file...

include/blas_quda.h
maddyscientist: I changed this function to support complex-valued argument intentionally, since I needed the ability to scale a vector by a complex number as part of the blas test improvements.

include/blas_quda.h
maddyscientist: As above an intentional change to support complex scalar multiplications.

include/blas_quda.h
maddyscientist: The definition of `cvector` is
```c++
using cvector = const vector;
``` 
We only ever pass these objects around as const references to allow for temporaries to be constructed when needed, e.g., when a double scale is passed to a function that is expecting a `cvector<Complex>`.

include/blas_quda.h
maddyscientist: I added the explicit wrappers to a few functions because there's many cases where the function appears directly in a `printfQuda` statement, and there's no way for the compiler to know to convert it automatically to a scalar, even though we have the cast operator.  Least intrusive thing to do was to just add a few wrappers.

Note that this is actually fixed properly in the reproducible reduction branch that I really should dust off and get in shape for a PR....

include/blas_quda.h
maddyscientist: See comment above.

include/dirac_quda.h
maddyscientist: I considered this, but yeah, verbosity would get out of control I think.

include/reference_wrapper_helper.h
maddyscientist: The definition above is just a class-wide convenience wrapper, but it's something that could easily be changed if it makes you happier 😄 

I did think about other names, but decided `vector` is the correct name.  It is derived from `std::vector` after all, just with some convenience additions.

I would argue that it doesn't conflict with `std::vector` since we never have a `using namespace std;` statement in QUDA.

lib/blas_quda.cu
maddyscientist: Intentional 😄

lib/dirac_clover_hasenbusch_twist.cpp
maddyscientist: Good suggestion c5b1081ad12a91e07dc510fb97519801b8ecdbe8

tests/blas_test.cpp
maddyscientist: Sorry, not following this question.  This is in the `benchmark` test, which doesn't do any correctness testing.

include/blas_helper.cuh
maddyscientist: One thing to do here is to remove the redundant comparison.  I think the number of comparisons will now grow linearly.

9881399fdb04f8e06ab9544dc5321a1b80efeb80

tests/blas_test.cpp
weinbe2: That answers my question, I looked quickly at the change without verifying if it was in the benchmark or verify test :)

include/reference_wrapper_helper.h
weinbe2: Maybe `vector_scalar`, but it's not necessarily a hill I'll die on... I appreciate that `vector` is the best/correct name, it's just unfortunate that it coincides with the name of a `std` container. Such is life.

maddyscientist: If it makes you feel any better, `cvector` is the type that is used in all interfaces, and the only time when `vector` is used in actual user code, per se, is when we need to create an explicit lvalue object.  All other uses are implicit (and `cvector`).

weinbe2 (approved):
This is a great bit of clean-up, all looks good, and thank you for beating on a few staggered MG woes while you were add it. Thanks @maddyscientist !

weinbe2:
Waiting on `clang-format` to actually merge

PR #1444, opened 2024-03-05
Optimizations to smearing interface

agrebe:
I have been computing two-link smearing with QUDA and have made two optimizations to the smearing interface:
1) In `lib/milc_interface.cpp`, skip the call to `loadGaugeQuda` if the `useResidentGauge` flag is set (since in this case, the 2-link field is already computed)
2) Simplify the four linear algebra operations in the loop in `performTwoLinkGaussianSmearNStep` in `lib/interface_quda.cpp` to a single operation
I have tested that this agrees with the previous implementation (using QUDA driven by MILC on a single GPU), and overall these changes increase performance by a factor of about 2 in our project.

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -4898,7 +4898,6 @@ void performTwoLinkGaussianSmearNStep(void *h_in, QudaQuarkSmearParam *smear_par
   cudaParam.setPrecision(inv_param->cuda_prec, inv_param->cuda_prec, true);
   ColorSpinorField in(cudaParam);
   ColorSpinorField out(cudaParam);
-  ColorSpinorField temp1(cudaParam);
 
   // Create the smearing operator
   //------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lib/interface_quda.cpp
```diff
@@ -4940,13 +4939,11 @@ void performTwoLinkGaussianSmearNStep(void *h_in, QudaQuarkSmearParam *smear_par
...
   for (int i = 0; i < smear_param->n_steps; i++) {
     if (i > 0) std::swap(in, out);
-    blas::ax(ftmp, in);
-    blas::axpy(a, in, temp1);
 
     qsmear_op.Expose()->SmearOp(out, in, a, 0.0, smear_param->t0, parity);
-    logQuda(QUDA_DEBUG_VERBOSE, "Step %d, vector norm %e\n", i, blas::norm2(out));
-    blas::xpay(temp1, -1.0, out);
-    blas::zero(temp1);
+    if (getVerbosity() >= QUDA_DEBUG_VERBOSE)
```
maddyscientist: No need for this if statement, the `logQuda` function wraps this in there

maddyscientist:
This is great contribution, thanks @agrebe.  Aside from my comment, please also clang-format the diff, and also feel free to add your name to the contributors in the README.

@Jenkins ok to test

lib/interface_quda.cpp
agrebe: Thanks for pointing this out -- `logQuda` is a more clever macro than I realized.  I've pushed a commit to remove this redundant `if` statement

maddyscientist (approved):
Thanks for the fast turn around.  Approving.  We can merge this once CI finishes.

(And I see you were already contributor, apologies for failing to recall this.)

agrebe:
Thank you for reviewing this so quickly -- I'm glad to be able to contribute given how useful this codebase has been for me!

PR #1445, opened 2024-03-15
Make QUDA understand memoryType in older ROCms

dmcdougall:
Older ROCms (before 5.5) didn't have the memoryType/type union that was included for forwards compatibility, so we're safeguarding in the case where folks are stuck with an older ROCm version.

I'm hoping this will help address #1432.


kostrzewa:
Thanks a lot, this fixes #1432. I can run with force offloading on LUMI-G now. I'm glad this backwards-compatibility introduces only these minimal changes.

maddyscientist (approved):

maddyscientist:
Thanks for confirming this fixes #1432.  I'll merge this PR.

PR #1446, opened 2024-03-18
Feature/reproducible

maddyscientist:
## Summary

This PR adds infrastructure for **extended host scalar precision** and **configurable GPU reduction algorithms** in QUDA.

It introduces separate control over:

- **Host scalar type** (`real_t` / `complex_t`): precision used for host-side coefficients, solver bookkeeping, logging, etc.
- **Reduction type** (`reduction_t`): scalar type used for per-site reduction arithmetic on the device
- **Reduction algorithm** (`device_reduce_t`): summation strategy applied on top of `reduction_t`:
  - `NAIVE`: plain `reduction_t` accumulation
  - `KAHAN`: `kahan_t<reduction_t>` (placeholder, not yet implemented)
  - `REPRODUCIBLE`: `rfa_t<reduction_t>` reproducible floating-point accumulator

Device reductions accumulate in `device_reduce_t`, then convert to `real_t` on the host after the multi-GPU reduction completes.

### Build configurations

**Default — double host, double reduction, naive summation**
```bash
cmake -DQUDA_SCALAR_TYPE=double \
      -DQUDA_REDUCTION_TYPE=double \
      -DQUDA_REDUCTION_ALGORITHM=NAIVE ...
```

**Reproducible reductions — double host, RFA summation**
```bash
cmake -DQUDA_SCALAR_TYPE=double \
      -DQUDA_REDUCTION_TYPE=double \
      -DQUDA_REDUCTION_ALGORITHM=REPRODUCIBLE ...
```

**Quad-precision host — `__float128` host scalars with double-double reduction arithmetic**
```bash
cmake -DQUDA_SCALAR_TYPE=__float128 \
      -DQUDA_REDUCTION_TYPE=doubledouble \
      -DQUDA_REDUCTION_ALGORITHM=NAIVE ...
```

### Scope of changes

- Generalize host code from hardcoded `double` to `real_t` / `complex_t` across solvers, interfaces, deflation, multigrid, and logging
- `printfQuda` / `logQuda` / `warningQuda` / `errorQuda` accept `real_t` directly (auto-narrowing `__float128` for C-style formatting)
- RFA reduction kernels with precomputed bin tables
- `doubledouble` reduction arithmetic support on device
- **Multi-GPU extended-precision reductions**: reductions retain `device_reduce_t` through the multi-process reduction and convert to `real_t` only after completion
- Quad (`__float128`) host scalar support, including Eigen dense solves run in double precision where required
- Communicator support for extended-precision reductions (single, MPI, QMP)
- Rework of Wilson contraction kernels to use a finer-grained parallelization strategy, reducing per-thread register/shared-memory state (needed to fit within resource limits, particularly for RFA builds)
- **`blas_test` improvements**: more robust verification, e.g., proving `caxpy` correctness via multiple `axpy` calls rather than relying on a single comparison path

### Known limitations

- `QUDA_SCALAR_TYPE=doubledouble` (double-double host scalars) is not yet implemented
- `QUDA_REDUCTION_ALGORITHM=KAHAN` (`kahan_t`) is a placeholder, not yet implemented
- RFA + CG reductions not yet supported
- RFA with quad or double-double reduction types not yet supported/tested
- RFA, Quad and double-double not supported on non-CUDA targets (HIP/SYCL): mileage may vary
- Eigen dense linear algebra runs in double precision even in quad builds (Eigen limitation)

### Testing

Tested configurations:
- Double host / double reduction (default)
- Quad host (`__float128`) / double-double reduction
- Double host / RFA reduction

## Test plan

- [x] Double host / double reduction build (MPI and QMP comms backends)
- [x] Quad host / double-double reduction build
- [x] Double host / RFA build
- [x] `blas_test` across field precisions
- [x] Eigensolver ctests (quad and double builds)
- [x] Solver ctests
- [x] Multigrid build


include/kernels/gauge_fix_fft.cuh
```diff
@@ -226,7 +226,7 @@ namespace quda {
 
       //35
       //T=36*gauge_dir+65
-      return operator()(data, value);
```
weinbe2, line 229: Has this been a long-standing bug that got missed in the unit testing?

maddyscientist: I'm not following the question here: what is the bug you're referring to here?

weinbe2: The order of arguments---or did that convention get changed in this PR? I'm only a few files in, so I may have missed a systemic change.

maddyscientist: Ok, gotcha, now I understand the question.  The order didn't matter prior to this PR, as before now, the operands on the left and right are the same, e.g., both doubles, and the summation functor returns a double.  With the reproducible reductions (or Kahan in the future) the type on the left is a non-trivial type (i.e., `rfa_t`) and the type on the right is the simple scalar, with the functor returning an `rfa_t`.  The sum reduction functor is conceptionally `return data += value`, if that helps illustrate it better.

The compiler will complain about this, so the user won't do the wrong thing.

I did think about supporting having the non-trivial type as either the input operand, but I decided I prefer the logical consistency of mandating the non-trivial is the first operand.

leonhostetler:
I tried to compile this branch in the default configuration on DeltaAI (GH200), and it fails with errors:
```
include/dbldbl.h:382:31: error: expected type-specifier before '__float128'
include/register_traits.h:233:33: error: '__float128' was not declared in this scope; did you mean '_Float128'?
```
The root cause (according to Claude) is: `__float128` is a GCC extension type that only exists on x86/PowerPC. On aarch64, long double is already IEEE binary128, so GCC provides only `_Float128` (and does not define `__SIZEOF_FLOAT128__`). The quad-precision machinery is gated correctly behind `QUDA_USE_QUAD_SCALAR`, but two references escaped the gating and are compiled unconditionally in every configuration.

Claude recommended guarding the two offending sites with `#ifdef __SIZEOF_FLOAT128__`. If I do that, the default configuration builds cleanly, so I will proceed with eigensolver/deflation regression testing.

Separately, the quad configuration (`QUDA_SCALAR_TYPE=__float128`) is x86-only as written and on DeltaAI fails with `QUDA_SCALAR_TYPE=__float128 requires compiler support for __float128`.

include/reference_wrapper_helper.h
```diff
@@ -554,6 +554,15 @@ namespace quda
     */
     vector(const T &t) : std::vector<T>(1, t) { }
 
+    /**
+       @brief Construct a length-1 vector from an arithmetic scalar
+    */
+    template <class U,
+              std::enable_if_t<std::is_arithmetic_v<U> && !std::is_same_v<U, T> && !std::is_same_v<std::complex<U>, T>> * = nullptr>
```
leonhostetler: This constructor captures `quda::vector<T> v(n)` whenever n is an integer type other than exactly `uint64_t`: it's an exact match, while `vector(uint64_t size, const T &value = {})` needs an integral conversion. The result is a length-1 vector holding n, not n elements — silently, with no diagnostic.
```
int bs = 2;
quda::vector<std::complex<double>> scales(bs);
// scales.size() == 1, scales[0] == (2,0)
```
This breaks `milc_interface.cpp:1424` (see comment there) resulting in heap corruption plus a silently wrong deflation space.

Claude suggests constraining to floating-point, which preserves the intent:
```
std::enable_if_t<std::is_floating_point_v<U> && !std::is_same_v<U, T> && !std::is_same_v<std::complex<U>, T>> * = nullptr>
```
(`!std::is_integral_v<U>` is equivalent. Making the size ctor explicit would be more robust but touches more call sites.) Claude's suggested fix resolves the heap corruption and silently wrong deflation space. An alternative would be to fix at the call site by casting `bs` to `uint64_t`, but that would leave the hazard(?) in place.

lib/milc_interface.cpp
```diff
@@ -1421,8 +1421,8 @@ void qudaLoadDeflationSpace(int external_precision, int quda_precision, const vo
       dEig->Dslash({temps.begin(), temps.begin() + bs},
                    {other_parity_space->evecs.begin() + lo, other_parity_space->evecs.begin() + hi}, parity);
       auto t_norms = blas::norm2({temps.begin(), temps.begin() + bs});
-      quda::vector<Complex> scales(bs);
-      for (int j = 0; j < bs; j++) scales[j] = 1.0 / sqrt(t_norms[j]);
+      quda::vector<complex_t> scales(bs);
```
leonhostetler: `bs` is `int`, so this now selects the new arithmetic-scalar constructor added in `reference_wrapper_helper.h:558` rather than the size constructor — `scales` ends up length 1 holding `bs`. Line 1425 then writes `bs` elements into it (heap overflow, confirmed by valgrind), and `blas::axy` on 1426 gets a length-1 scale vector, so the reconstructed second-parity eigenvectors are mis-normalized.

leonhostetler (changes requested):
I tested on 1 node on DeltaAI (GH200) using the default build configuration. I tested several workflows involving eigensolves and deflations. The eigensolves look fine, but the deflation breaks (see line-level comments). After implementing Claude's suggested fix, deflation works again, and I reproduce to expected precision the results obtained when I run the same tests with the develop branch.

include/kernels/gauge_fix_fft.cuh
weinbe2: This makes sense. I always assumed that it was strictly interpretted as `return data += value;`, which is why I was confused when the order changed. It's clear why it didn't matter in the past, but it does matter now. I support mandating the order.Thanks @maddyscientist !

lib/eig_block_trlm.cpp
```diff
@@ -99,33 +102,54 @@ namespace quda
...
 
       // Lambda that returns mat_norm for LR and returns the relevant alpha
       // (the corresponding Ritz value) for SR
-      auto check_norm = [&](double sr_norm) -> double {
+      auto check_norm = [&](real_t sr_norm) -> real_t {
         if (eig_param->spectrum == QUDA_SPECTRUM_LR_EIG)
           return mat_norm;
         else
-          return fabs(sr_norm);
+          return sr_norm;
```
hummingtree, line 139: Should this remain `quda::fabs(sr_norm)`? For SR, Ritz values can be negative; returning the signed value makes the locking/convergence threshold negative, so a nonnegative residual can never satisfy the test. This also changes the previous behavior.

lib/deflation.cpp
```diff
@@ -9,9 +9,12 @@ namespace quda
 
   using namespace blas;
   using DynamicStride = Stride<Dynamic, Dynamic>;
+  using MatrixXc = Matrix<complex_t, Dynamic, Dynamic>;
```
hummingtree: In a quad build this alias makes both `SelfAdjointEigenSolver<MatrixXc>` calls below operate on `complex<__float128>`. `eigen_helper.h` explicitly notes that Eigen returns incorrect eigenvectors for this type and introduces `eig_solve_complex_t` to keep projected solves in double. Should these deflation solves use that conversion path as well?

lib/inv_eigcg_quda.cpp
```diff
@@ -26,13 +26,13 @@ namespace quda {
    using namespace blas;
 
    using DynamicStride   = Stride<Dynamic, Dynamic>;
-   using DenseMatrix     = MatrixXcd;
-   using VectorSet       = MatrixXcd;
-   using Vector          = VectorXcd;
-   using RealVector      = VectorXd;
+   using DenseMatrix     = EigMatrixXcd;
```
hummingtree: This makes the EigCG projected eigensolves use `complex<__float128>` in a quad build, despite `eigen_helper.h` documenting that Eigen’s extended-precision eigenvectors are incorrect. Could we convert `Tm`/`H2k` to `eig_solve_complex_t`, solve in double, and convert the rotation coefficients back, as the other eigensolvers now do?

include/dbldbl.h
```diff
@@ -283,67 +341,157 @@ struct doubledouble {
...
+    return true;
+  } else {
+    return false;
+  }
 }
 
-__device__  inline doubledouble operator+(const doubledouble &a, const doubledouble &b) {
+__device__ __host__  inline bool operator>=(const doubledouble &a, const doubledouble &b)
+{
+  if (a.head() >= b.head()) {
```
hummingtree: The first comparison needs to be strict here. As written, equal heads return true without considering the tails, so `(1, -1) >= (1, 0)` is incorrectly true. This should be `a.head() > b.head()`, followed by the equal-head tail comparison. The same issue exists in `operator<=` below.

include/eigen_quad_scalar.h
```diff
@@ -0,0 +1,93 @@
...
+    {
+      return quda::eigen_fp128_pow(static_cast<__float128>(2), static_cast<__float128>(-113));
+    }
+    __host__ __device__ static inline __float128 dummy_precision()
+    {
+      return quda::eigen_fp128_pow(static_cast<__float128>(10), static_cast<__float128>(-32));
+    }
+    __host__ __device__ static inline __float128 highest()
+    {
+      return quda::eigen_fp128_pow(static_cast<__float128>(2), static_cast<__float128>(16384));
```
hummingtree: `2^16384` is outside the finite binary128 range, so `highest()` returns `+inf` and `lowest()` returns `-inf`. Eigen’s `NumTraits` expects the largest/smallest finite values here. Can these return the binary128 finite maximum and its negative instead?

include/targets/cuda/reproducible_floating_accumulator.hpp
```diff
@@ -0,0 +1,944 @@
...
+
+  ///Accumulate a binned fp @p x into the binned fp.
+  __host__ __device__ ReproducibleFloatingAccumulator& operator+=(const ReproducibleFloatingAccumulator &other){
+    binned_dbdbadd(other);
+    return *this;
+  }
+
+  ///Accumulate-subtract a binned fp @p x into the binned fp.
+  ///NOTE: Makes a copy and performs arithmetic; slow.
+  __host__ __device__ ReproducibleFloatingAccumulator& operator-=(const ReproducibleFloatingAccumulator &other){
```
hummingtree, line 624: This overload currently has two issues when instantiated: `other` is const but unary `operator-()` is not const, and the function is missing `return *this;`. Could we make unary minus const and return the accumulator here?

lib/CMakeLists.txt
```diff
@@ -759,6 +790,26 @@ if(QUDA_BACKWARDS)
...
+    CACHE STRING "set the reduction computation type (double, doubledouble)")
+set_property(CACHE QUDA_REDUCTION_TYPE PROPERTY STRINGS double doubledouble)
+if(QUDA_REDUCTION_TYPE STREQUAL "doubledouble")
+  set(QUDA_REDUCTION_IS_DOUBLEDOUBLE 1)
+endif()
+
+set(QUDA_REDUCTION_ALGORITHM
+    NAIVE
+    CACHE STRING "set the reduction algorithm (NAIVE, KAHAN, REPRODUCIBLE)")
+set_property(CACHE QUDA_REDUCTION_ALGORITHM PROPERTY STRINGS NAIVE KAHAN REPRODUCIBLE)
```
hummingtree, line 818: `KAHAN` is exposed as a supported cache value, but selecting it reaches `kahan_t<reduction_t>` in `float_vector.h`, and there is currently no `kahan_t` definition. That makes an advertised configuration fail to compile. Should KAHAN be removed or rejected at configure time until it is implemented?

hummingtree (commented):
Additional correctness and build-configuration findings from the extended-precision / reproducible-reduction review.

hummingtree (approved):
It looks nice and clean! I have left some comments (not really - codex did it) ...

include/kernels/blas_3d.cuh
```diff
@@ -320,11 +319,10 @@ namespace quda
       Vector x = arg.x(idx_cb, parity);
       Vector y = arg.y(idx_cb, parity);
 
-      // Get the inner product
-      complex<double> res = innerProduct(x, y);
+      const complex<real> res = innerProduct(x, y);
```
weinbe2: Should the cast to `compute_t` happen before the call to `innerProduct`? Put differently, case `x` and `y` in some appropriate way (I'm not sure how casts of `Vector` would work, more specifically, if `ColorSpinor<real, Arg::nColor, Arg::nSpin>;` can be cast to `ColorSpinor<compute_t, Arg::nColor, Arg::nSpin>;`...)

include/kernels/multi_reduce_core.cuh
```diff
@@ -177,7 +179,7 @@ namespace quda
       template <typename T> __device__ __host__ inline void operator()(reduce_t &sum, T &x, T &y, T &, T &, int, int) const
       {
 #pragma unroll
-        for (int k=0; k < x.size(); k++) dot_<reduce_t, real>(sum, x[k], y[k]);
+        for (int k=0; k < x.size(); k++) sum = plus<reduce_t>::apply(sum, dot_<compute_t>(x[k], y[k]));
```
weinbe2: This type of cast before sum is what I was considering in my comment in `blas_3d.cuh`

include/targets/cuda/reduce_helper.h
```diff
@@ -164,20 +168,23 @@ namespace quda
...
+      for (int i = 0; i < n_reduce; i++) {
+        const auto offset = static_cast<size_t>(i) * n_element / n_reduce;
+        // Deserialize flat atomic words (e.g. double components of RFA / doubledouble)
+        std::memcpy(static_cast<void*>(&result[i]), static_cast<const void*>(&scalar_result[offset]), sizeof(T));
+      }
 
       if (!reset) {
         consumed = true;
       } else {
-        // reset the atomic counter - this allows multiple calls to complete with ReduceArg construction
```
weinbe2, line 175: Is there a reason this comment was dropped?

include/blas_helper.cuh
```diff
@@ -546,7 +546,7 @@ namespace quda
 
         if constexpr (!is_enabled(QUDA_DOUBLE_PRECISION))
           if (x.Location() == QUDA_CUDA_FIELD_LOCATION)
-            errorQuda("QUDA_PRECISION=%d does not enable double precision", QUDA_PRECISION);
+            errorQuda("QUDA_PRECISION=%d does not enable real_t precision", QUDA_PRECISION);
```
weinbe2: This might be an accidental global find/replace mishap, I think this should still be `double`.

include/float_vector.h
```diff
@@ -106,10 +160,14 @@ namespace quda {
      computing an infinity norm
    */
   template <typename T> struct deviation_t {
+    using value_type = T;
+
     T diff;
```
weinbe2, line 165: Not particularly important, but for cleanliness should we change these `T` -> `value_type`?

lib/momentum.cu
```diff
@@ -57,11 +57,11 @@ namespace quda {
     count++;
   }
 
-  void forceRecord(array<double, 2> &force, double dt, const char *fname)
+  void forceRecord(array<reduction_t, 2> &force, double dt, const char *fname)
   {
     if (comm_rank()==0) {
-      force_stream << fname << "\t" << std::setprecision(5) << force[0] << "\t"
-                   << std::setprecision(5) << force[1] << "\t"
+      force_stream << fname << "\t" << std::setprecision(5) << double(force[0]) << "\t"
```
weinbe2: Technically out of scope for this PR, but I say we add a quality-of-life improvement and update this to `logQuda` or something of that sort.

tests/host_reference/dslash_reference.cpp
```diff
@@ -786,14 +786,15 @@ std::array<double, 2> verifyStaggeredInversion(quda::ColorSpinorField &in,
       mxpy(in.data(), ref.data(), in.Volume() * stag_spinor_site_size, inv_param.cpu_prec);
       double nrm2 = norm_2(ref.data(), ref.Volume() * stag_spinor_site_size, inv_param.cpu_prec);
       double src2 = norm_2(in.data(), in.Volume() * stag_spinor_site_size, inv_param.cpu_prec);
-      double hqr = sqrt(quda::blas::HeavyQuarkResidualNorm(out, ref).z);
+      auto hq_norm = quda::blas::HeavyQuarkResidualNorm(out, ref);
+      auto hqr = quda::sqrt(hq_norm[2]);
       double l2r = sqrt(nrm2 / src2);
```
weinbe2: Should this be `auto` instead of `double`?

tests/blas_test.cpp
```diff
@@ -605,8 +609,8 @@ class BlasTest : public ::testing::TestWithParam<::testing::tuple<int, int>>
       xD = xH;
       yoD = yH;
       blas::axpbyz(a, xD, b, yoD, zoD);
-      blas::axy(a, xH, zH);
-      blas::axpy(b, yH, zH);
+      blas::axy(b, yH, zH);
```
weinbe2, line 612: Was this to exactly match the ordering within the QUDA device-side call? I'm impressed it required this much care, if so...

maddyscientist: That's exactly it.  Without this the rounding order means changes at the order of an ULP (single or double precision), which would wash out any quad precision level reduction tests.

lib/momentum.cu
maddyscientist: I think most (all?) I/O in QUDA is done using C++ file streams and not using C APIs, so the present use here is consistent with what we do elsewhere (e.g., the tsv files).

Or is there something I'm missing here?

weinbe2: Ah, no, I just read it a bit too quickly and didn't realize this was writing to a file since that generally doesn't happen in any of the `*.cu` files iirc.

include/float_vector.h
maddyscientist: Claude tells me the usual convention is to use `T` inside the classes, and save `value_type` for external reference.  So I will leave as is.

include/kernels/blas_3d.cuh
maddyscientist: 21b898156a75d4dfb22aa578d51c0e8691f88b49 does what you want here.  Since the inner products are done at the site level first of all, there's no need to do full reproducible reduction within the site, so we only need to ensure the site reduction is done in appropriate precision.

include/targets/cuda/reduce_helper.h
maddyscientist: c623087a2b28e225f1235e2025a544b1d2e2a7f7

include/blas_helper.cuh
maddyscientist: f6b5ab4cb1487d9e98f5cc8d8c74fd4af4d1c3a8

tests/host_reference/dslash_reference.cpp
maddyscientist: f6b5ab4cb1487d9e98f5cc8d8c74fd4af4d1c3a8

weinbe2:
From our convo offline, this line of code: https://github.com/lattice/quda/blob/feature/reproducible/include/kernels/gauge_utils.cuh#L26

Needs `double` -> `real_t` for the anisotropy argument.

lib/eig_block_trlm.cpp
maddyscientist: I see this issue have already been cleaned up by @leonhostetler with #1646 so I won't touch that here.

lib/inv_eigcg_quda.cpp
maddyscientist: With Eigen now working with `__float128` I've cleaned all this up, so we should good on this front now.  All use of Eigen for the eigensolvers now uses `real_t` / `complex_t` consistently.  0d7fa7523ebd6bb23cc978e74d69c8fdaa15c575 is where this was done.

lib/deflation.cpp
maddyscientist: Cleaned up with 0d7fa7523ebd6bb23cc978e74d69c8fdaa15c575

include/dbldbl.h
maddyscientist: 8c6a7f8c43d6337159d7d9f6406d9ddd16b689f8

include/targets/cuda/reproducible_floating_accumulator.hpp
maddyscientist: 45ea5fced40f1779622199bc9c0c9ef26620cf97

lib/CMakeLists.txt
maddyscientist: f6d94726654ef0f1b8f299976703bdc252e23ca3 rejects at configure time.

include/eigen_quad_scalar.h
maddyscientist: thanks for pointing out this bug.  f6d94726654ef0f1b8f299976703bdc252e23ca3 fixes this.

maddyscientist:
Commenting that 5498137d8f536b875d340c06263f25eb74e4bff9 fixed the fundamental issue with Eigen with quad precision, namely that `isSigned` wasn't being correctly defined leading to Eigen assuming it was an unsigned type, leading to badness.  With that fixed, Eigen is absolutely fine now with quad precision.

maddyscientist:
> From our convo offline, this line of code: https://github.com/lattice/quda/blob/feature/reproducible/include/kernels/gauge_utils.cuh#L26
> 
> Needs `double` -> `real_t` for the anisotropy argument.

2830ba57981f215fd7c1d8e1a0453e1bd523c599 addresses this issue.  It wasn't a case of `double` needing to be switched for `real_t`, rather for the local template type `Arg::real`.  Issue now fixed though.

include/reference_wrapper_helper.h
maddyscientist: Fixed with 7b3bd339170932583c26bde64b258413a97ad978

lib/milc_interface.cpp
maddyscientist: Fixed with 7b3bd339170932583c26bde64b258413a97ad978

maddyscientist:
> I tried to compile this branch in the default configuration on DeltaAI (GH200), and it fails with errors:
> 
> ```
> include/dbldbl.h:382:31: error: expected type-specifier before '__float128'
> include/register_traits.h:233:33: error: '__float128' was not declared in this scope; did you mean '_Float128'?
> ```
> 
> The root cause (according to Claude) is: `__float128` is a GCC extension type that only exists on x86/PowerPC. On aarch64, long double is already IEEE binary128, so GCC provides only `_Float128` (and does not define `__SIZEOF_FLOAT128__`). The quad-precision machinery is gated correctly behind `QUDA_USE_QUAD_SCALAR`, but two references escaped the gating and are compiled unconditionally in every configuration.
> 
> Claude recommended guarding the two offending sites with `#ifdef __SIZEOF_FLOAT128__`. If I do that, the default configuration builds cleanly, so I will proceed with eigensolver/deflation regression testing.
> 
> Separately, the quad configuration (`QUDA_SCALAR_TYPE=__float128`) is x86-only as written and on DeltaAI fails with `QUDA_SCALAR_TYPE=__float128 requires compiler support for __float128`.

Thanks for point this out @leonhostetler.  Somewhat of an oversight on my part to not have tested it on Arm CPUs.  I have addressed this issue now:
* QUDA now uses an abstracted type `quda::float128_t` that is either `__float128` on x86-64 or `_Float128` on ARM
* This requires separate abstraction for the host math library since the `__float128` and `_Float128` have differently named functions, e.g., `sqrtq` vs `__builtin_sqrtf128`
* All use of `float128_t` should now be guarded by the preprocessor and hidden if quad-precision is not enabled
* When configuring QUDA, we can specify either `_Float128`, `__float128` or `float128` regardless of the platform when setting `QUDA_SCALAR_TYPE`
* I have verified that quad precision is working fine on Grace-Hopper, with a full build fully passing ctest

include/kernels/dslash_mdw_fused.cuh
```diff
@@ -117,8 +117,8 @@ namespace quda {
 
         if (b_5[0] != b_5[1] || b_5[0].imag() != 0) { errorQuda("zMobius is NOT supported yet.\n"); }
 
-        b = b_5[0].real();
-        c = c_5[0].real();
+        b = static_cast<real>(b_5[0].real());
+        c = static_cast<real>(c_5[0].real());
         kappa = -(c * (4. + m_5) - 1.) / (b * (4. + m_5) + 1.); // This is actually -kappa in my(Jiqun Tu) notes.
```
weinbe2, line 122: Should this get the `const real four = 4.0;` treatment like the other DWF files, or are we safe because `m_5` and `b` have already been cast? Am I correct in assuming that the problem in the other DWF files is how the `b_5`/`c_5` arrays aren't already cast to `real`?

maddyscientist: Exactly as you say it, there's no issues here since every variable on this line is already a `real`, e.g., `double` or `float`.  C++ doesn't mandate that we cast double precision constants, e.g., the `4.0` to float and things just work.  One might argue that explicitly casting the constants to `real` would be desirable, e.g., if a custom precision type is used in future for field-local computations, but this is a pre-existing condition and not something to worry about right now (else it would be a further wide edit QUDA codebase).

include/blas_helper.cuh
```diff
@@ -584,7 +584,7 @@ namespace quda
       if (x.Precision() == QUDA_DOUBLE_PRECISION) {
         if constexpr (!is_enabled(QUDA_DOUBLE_PRECISION))
           if (x.Location() == QUDA_CUDA_FIELD_LOCATION)
-            errorQuda("QUDA_PRECISION=%d does not enable double precision", QUDA_PRECISION);
+            errorQuda("QUDA_PRECISION=%d does not enable real_t precision", QUDA_PRECISION);
```
weinbe2: Pedantic, but I believe this is an erroneous `double` -> `real_t` replacement (as are the other `diff` in this file)

lib/eigensolve_quda.cpp
```diff
@@ -226,15 +227,15 @@ namespace quda
   void EigenSolver::queryPrec(const QudaPrecision prec)
   {
     switch (prec) {
-    case QUDA_DOUBLE_PRECISION: logQuda(QUDA_VERBOSE, "Running eigensolver in double precision\n"); break;
+    case QUDA_DOUBLE_PRECISION: logQuda(QUDA_VERBOSE, "Running eigensolver in real_t precision\n"); break;
```
weinbe2: `real_t` -> `double` in the `logQuda`

maddyscientist: 12216a8

include/blas_helper.cuh
maddyscientist: 12216a8

lib/inv_gmresdr_quda.cpp
```diff
@@ -267,7 +267,7 @@ namespace quda {
     std::vector<ColorSpinorField *> vm(Vm->Components());
 
     RowMajorDenseMatrix Alpha(Qkp1); // convert Qkp1 to Row-major format first
-    blas::legacy::caxpy(static_cast<Complex *>(Alpha.data()), vm, vkp1);
```
weinbe2, line 270: Should this also be changed to `reinterpret_cast`, like the updates above?

lib/pgauge_det_trace.cu
```diff
@@ -13,11 +13,11 @@ namespace quda {
   template <typename Float, int nColor, QudaReconstructType recon>
   class CalcFunc : TunableReduction2D {
     const GaugeField &u;
-    array<double, 2> &result;
+    complex_t &result;
```
weinbe2, line 16: I think this should remain an `array<double, 2>`; it's two separate values (determinant and trace), not a complex number. I have no issues with the variable renames otherwise, though.

weinbe2: Nevermind I'm actually an idiot, it is indeed a complex, this is just confusing.

weinbe2 (approved):
Pending a few uninteresting comments, I approve of this PR. Thanks for the awesome work @maddyscientist !

lib/inv_gmresdr_quda.cpp
maddyscientist: None of this casting was actually necessary so I've just nuked it (and in general one should always prefer `static_cast` over `reinterpret_cast`).

leonhostetler (approved):
Great addition to QUDA!

PR #1447, opened 2024-03-19
Feature/DD

sbacchio:
This is a first PR towards enabling domain decomposition (DD) features in QUDA.
The goal of this PR is to enable a red-black decomposition for the Dirac operator.

### Remarks

- We first consider the blocks to exactly fit in the local lattice. Generalization is left to a future PR.
- We focus only on the application of the Dirac operator. Optimization of BLAS functions is left to a future PR.

### Design

Under domain decomposition, the Dirac operator assumes a block structure, e.g.

$$
D = 
\begin{bmatrix} D_{rr} & D_{rb} \\
  D_{br} & D_{bb} \end{bmatrix}
$$

Thus we need a simple strategy to implement all possible operators and their application to a field.
Our strategy is to add the specifications about the domain decomposition directly to the vector (spinor) field.
E.g. the application of $D_{rb}$, i.e. $y = D_{rb} x = P_r D P_b x$, can be expressed by setting the input vector as "black", i.e. $x_b = P_b x$, and the output vector as "red", i.e. $y_r = P_r y$. A pseudocode would look like this:
```
x.dd_red_active();
y.dd_black_active();
applyD(y,x);
```

Then, we make the application of D DD-aware and only apply it to the in/out active points.

### Summary of changes:
- DDParam is added as a property of lattice fields (used only for ColoSpinorFields at the moment).
- ... TODO

### TODO list:
- [x] Add more comments and function documentation
- [x] Add checks for parameters, e.g. the block size should divide the local lattice size
- [x] ~Add calculation of first block parity~ Use global coordinates
- [x] Test application of individual pieces, i.e. D_rr, D_rb, D_br, D_bb
- [x] Test all operators
- [x] Test that performance without DD is not affected (i.e. compared to current develop)
- [ ] Test MR solver and usage in MG as smoother
- [x] Properly disable comms when not needed (e.g. block fits local lattice)
- [ ] Improve performance by unrolling threads block-wise
- [x] Disable usage of DD for all PC Mat
- [ ] ...

### Current issues:
- [x] ~Tests are failing if the block size is odd in the x direction for the PC operator, e.g. `dslash_test --xdim 6 --ydim 4 --zdim 4 --tdim 4 --dd-block-size 3 2 2 2 --dd-red-black=true --test 0`. It works if not PC, e.g. `--test 2` or if it is odd in any other direction.~ Now it is not allowed to have odd block size in the x-direction.
- [x] Tests are failing if `export QUDA_REORDER_LOCATION=CPU` is set (#1466).
- [x] ~Tests are failing for MatPC with `Caught signal 8 (Floating point exception: integer divide by zero)` . See e.g. `dslash_test --xdim 8 --ydim 8 --zdim 8 --tdim 8 --dd-red-black=true --test 1`~ Not testing for PC operators


CMakeLists.txt
```diff
@@ -151,7 +151,7 @@ option(QUDA_DIRAC_CLOVER_HASENBUSCH "build clover Hasenbusch twist operators" ${
 option(QUDA_DIRAC_NDEG_TWISTED_MASS "build non-degenerate twisted mass Dirac operators" ${QUDA_DIRAC_DEFAULT})
 option(QUDA_DIRAC_NDEG_TWISTED_CLOVER "build non-degenerate twisted clover Dirac operators" ${QUDA_DIRAC_DEFAULT})
 option(QUDA_DIRAC_LAPLACE "build laplace operator" ${QUDA_DIRAC_DEFAULT})
-
+option(QUDA_DIRAC_DD "build code for domain decomposition of the Dirac operator" OFF)
```
maddyscientist: I think this name should be more explicit, e.g. `QUDA_DIRAC_DOMAIN_DECOMPOSITION`

include/enum_quda.h
```diff
@@ -590,6 +590,8 @@ typedef enum QudaExtLibType_s {
   QUDA_EXTLIB_INVALID = QUDA_INVALID_ENUM
 } QudaExtLibType;
 
+typedef enum QudaDDType_s { QUDA_DD_NO, QUDA_DD_RED_BLACK, QUDA_DD_INVALID = QUDA_INVALID_ENUM } QudaDDType;
+
```
maddyscientist, line 642: Need to add matching Fortran entry in `enum_fortran.h`

include/invert_quda.h
```diff
@@ -400,6 +404,14 @@ namespace quda {
 
     // for incremental eigCG:
     void updateRhsIndex(QudaInvertParam &param) { rhs_idx = param.rhs_idx; }
+
+    constexpr inline bool do_block_schwarz() const
```
maddyscientist: I don’t think this function can ever evaluate as `constexpr`?

maddyscientist (commented):
I’ve done a visual review.  As we’ve discussed, I like the overall strategy being applied here.
- Although the `DDArg` additions have been applied to all Dirac operators, it seems that the necessary additions haven’t been added to all operators at the kernel level.  I guess these will be easy enough to identify with the tests that are in place.
- Can the dslash_ctest and staggered_dslash_ctest tests be augmented to included testing DD?  Ideally we’ll want these tests to be run as new instantiations in the present tests that are called by ctest.  (We can discuss on our next call.)
- In C++, the presence of `constexpr` implies `inline`, so there no need to declare `constexpr inline`
- I think to aid readability, better to use Domain Decomposition (especially in comments) as opposed to simply DD.
- I’ve left a few other comments inline

sbacchio:
> I’ve done a visual review. As we’ve discussed, I like the overall strategy being applied here.
> 
> * Although the `DDArg` additions have been applied to all Dirac operators, it seems that the necessary additions haven’t been added to all operators at the kernel level.  I guess these will be easy enough to identify with the tests that are in place.
> * Can the dslash_ctest and staggered_dslash_ctest tests be augmented to included testing DD?  Ideally we’ll want these tests to be run as new instantiations in the present tests that are called by ctest.  (We can discuss on our next call.)
> * In C++, the presence of `constexpr` implies `inline`, so there no need to declare `constexpr inline`
> * I think to aid readability, better to use Domain Decomposition (especially in comments) as opposed to simply DD.
> * I’ve left a few other comments inline

sbacchio:
Hi @maddyscientist , I worked a bit more on this PR.
I have been performing tests for all operators. Just facing an issue beyond my reach and it would be good to have your input. Namely, `dslash_test` fails for the case of `Mat` matrix and the operators `clover-hasenbusch-twist`, also in the `develop` branch.
Here is a reproducible:
`dslash_test --dslash-type clover-hasenbusch-twist --test Mat`

It fails with a segmentation fault as follows.
<details>
  <summary>
 Output:
   </summary>
Disabling GPU-Direct RDMA access
Enabling peer-to-peer copy engine and direct load/store access
Rank order is column major (t running fastest)
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from DslashTest
QUDA 1.1.0 (git 1.1.0-1029d62d5-sm_80)
CUDA Driver version = 12010
CUDA Runtime version = 11080
Graphic driver version = 530.30.02
Found device 0: NVIDIA A100-SXM-64GB
Using device 0: NVIDIA A100-SXM-64GB
WARNING: Data reordering done on GPU (set with QUDA_REORDER_LOCATION=GPU/CPU)
WARNING: Environment variable QUDA_RESOURCE_PATH is not set.
WARNING: Caching of tuned parameters will be disabled.
cublasCreated successfully
[ RUN      ] DslashTest.benchmark
Sending clover field to GPU
Randomizing fields... Sending gauge field to GPU
Creating cudaSpinor with nParity = 2
Creating cudaSpinorOut with nParity = 2
Sending spinor field to GPU
Source 0: CPU = 2.653920e+06, CUDA = 2.653920e+06
running the following test:
prec    recon   dtest_type     matpc_type   dagger   S_dim         T_dimension   Ls_dimension dslash_type    niter
single   18       Mat              even_even    0     24/ 24/ 24         24             16   clover_hasenbusch_twist   100
Grid partition info:     X  Y  Z  T
                         0  0  0  0
Tuning...
[lrdn3448:783349:0:783349] Caught signal 11 (Segmentation fault: address not mapped to object at address 0xf0)
</details>

maddyscientist:
@sbacchio yep, don't worry about that failing test.  I don't think that operator has a non-preconditioned form properly implemented.

maddyscientist:
This is looking very promising with the full file parallelization of precision, color, etc. through CMake.  As is evidenced by the failing builds in the CI, this is still buggy and needs to fixed.

Some minor comments:
* For file naming, keep to lower case (`double` not `DOUBLE`)
* For color filename encoding, I would prefer `_nc_3`  for 3 colors

include/domain_decomposition_helper.cuh
```diff
@@ -0,0 +1,112 @@
...
+    template <typename Coord> constexpr bool block_parity(const Coord &x) const
+    {
+      int block_parity = 0;
+      for (int i = 0; i < x.size(); i++) {
+        if (block_dim[i] > 0) block_parity += x.gx[i] / block_dim[i];
+      }
+      return block_parity % 2 == 1;
+    }
+
+    template <typename Coord> constexpr bool on_border(const Coord &x, int mu, int dir) const
```
sbacchio: Change `dir` to `dist`, since it represents the distance of the hopping

maddyscientist:
I'm starting to do some testing on this PR, and I see that not all Dirac operators have the file parallelization.  Is this something you can do?  See the trace below, for a full build of QUDA, we can see that these files dominate compilation time, and result in a much longer compilation time.  At the same time, it's clear how much faster the split Dirac operator files compile on a multi-core system 😄 

image

pittlerf:
> I'm starting to do some testing on this PR, and I see that not all Dirac operators have the file parallelization. Is this something you can do? See the trace below, for a full build of QUDA, we can see that these files dominate compilation time, and result in a much longer compilation time. At the same time, it's clear how much faster the split Dirac operator files compile on a multi-core system 😄
> 
> image

Hi, yes, sorry, I will do the remaining ones.

maddyscientist:
I have fixed the failing staggered dslash tests (was due to an issue with the long-link field being created erroneously when using regular (unimproved) staggered fermions.

sbacchio:
Hi Kate, that's an impressive tracing :) Ferenc will work on the others soon. About the tests, I still see some staggered tests failing (invert and eigensolve). I guess due to similar reasons. Maybe it's better if you have a look at those too :)

maddyscientist:
Looking great with latest pushes.  Just `dslash_twisted_mass_preconditioned.cu` left I think

image

sbacchio:
Great! and I see now also all 7 checks passed :) Thanks @pittlerf !
@maddyscientist do you have also a trace of the compilation time of the current develop? Just to appreciate the overall improvement :)

maddyscientist:
Checking where we are with this PR: I believe we still need the DD tests added to the ctest list so that DD is tested if enabled?

pittlerf:
Hi, apologize for the delay here. We have included tests for all the DD supported operators to the ctest, additionally we wrote a small assert function to prevent using DD with not supported operators.

sbacchio:
Hi @maddyscientist, the last commits by Ferenc a few weeks ago solved all the pending issues and put guards for the non-supported cases. Would be nice to get a final review when you have time :)

maddyscientist:
Thanks @sbacchio.  I think we are good to go.

At the moment the CSCS CI is broken, and I want to get that fixed in a separate PR, and then have that merged in to this branch before we merge.  So if you can hang on another few days.

Thanks @pittlerf for all the work getting this over the line.

weinbe2:
@pittlerf and/or @sbacchio , can you re-merge develop and resolve the conflicts? Once you can take care of that and (re-)verify it passes, I'll do one more quick review and get this merged in promptly.

include/kernels/dslash_ndeg_twisted_clover.cuh
```diff
@@ -41,7 +41,7 @@ namespace quda
     using real = typename mapper<typename Arg::Float>::type;
     using Vec = ColorSpinor<real, Arg::nColor, 4>;
     using Cache = SharedMemoryCache<Vec>;
-    using Ops = std::conditional_t<kernel_type == INTERIOR_KERNEL, KernelOps<Cache>, NoKernelOps>;
+    using Ops = KernelOps<Cache>; //std::conditional_t<kernel_type == INTERIOR_KERNEL, KernelOps<Cache>, NoKernelOps>;
```
sbacchio: NOTE, this is a temporal fix because without wouldn't compile. We still have to understand why is the case.

jcosborn: I think the issue is this: https://github.com/lattice/quda/pull/1447/files#diff-db133e457034e56706bf162a3cf559751de5ce8a1a62a504faa3b965395bea1aL82
Dropping the constexpr makes it always compile in the `SharedMemoryCache` which means the KernelOps must always include it.  I think the fix is to keep the `if constexpr (mykernel_type == INTERIOR_KERNEL) {` on the outside and put the `if(arg.dd_x.isZero(coord)) {...} else {...}` on the inside.

sbacchio:
@maddyscientist the staggered tests are failing, independently on the DD. Is it something we did or a known issue?

maddyscientist:
> @maddyscientist the staggered tests are failing, independently on the DD. Is it something we did or a known issue?

The failure was caused by this change in tests/staggered_dslash_test_util.h, where `cpuLong` is created conditionally.  Was this change intentional or required?  The cpu host staggered code always assumes the presence of the long link field,e even when we're doing naive staggered.
```
if (dslash_type == QUDA_ASQTAD_DSLASH) cpuLong = GaugeField(cpuLongParam);
```

maddyscientist:
I've noticed another issue that needs to be taken care of, beyond the failure above.  It seems that conditional compilation is broken now, e.g., all Dirac operators are presently built, despite the settings of `QUDA_DIRAC_WILSON`, etc.  This needs to be fixed before we can merge.

pittlerf:
> I've noticed another issue that needs to be taken care of, beyond the failure above. It seems that conditional compilation is broken now, e.g., all Dirac operators are presently built, despite the settings of `QUDA_DIRAC_WILSON`, etc. This needs to be fixed before we can merge.

Hi @maddyscientist , I have resolved the compilation issue, and attempted to fix the staggered issue. I tried to remove the if condition from the commit a7fe5bd57249d2fc0fc8533daac131af296cdd38, however the staggered tests are still failing with the following error:
ERROR: Halo depth 3 is greater than local lattice x = {4 2 6 8} (rank 0, host jwb0085.juwels, gauge_field.cpp:85 in void quda::GaugeField::create(const quda::GaugeFieldParam&)())

Do you know this issue?

pittlerf:
> > I've noticed another issue that needs to be taken care of, beyond the failure above. It seems that conditional compilation is broken now, e.g., all Dirac operators are presently built, despite the settings of `QUDA_DIRAC_WILSON`, etc. This needs to be fixed before we can merge.
> 
> Hi @maddyscientist , I have resolved the compilation issue, and attempted to fix the staggered issue. I tried to remove the if condition from the commit [a7fe5bd](https://github.com/lattice/quda/commit/a7fe5bd57249d2fc0fc8533daac131af296cdd38), however the staggered tests are still failing with the following error: ERROR: Halo depth 3 is greater than local lattice x = {4 2 6 8} (rank 0, host jwb0085.juwels, gauge_field.cpp:85 in void quda::GaugeField::create(const quda::GaugeFieldParam&)())
> 
> Do you know this issue?

We found this fix. Is it OK @maddyscientist ?

maddyscientist:
Looks like we have passing staggered tests again, so that's good.  The issue remains however that all Dirac operators are still being built, regardless of the `DIRAC` CMake settings.

pittlerf:
> Looks like we have passing staggered tests again, so that's good. The issue remains however that all Dirac operators are still being built, regardless of the `DIRAC` CMake settings.

Hi, thanks for the comment. For the `DIRAC` CMake settings, as I understand  for example the Domain Wall in "lib/dslash_domain_wall_4d_m5inv.hpp" there is an if statement:
    `if constexpr (is_enabled<QUDA_DOMAIN_WALL_4D_DSLASH>()) `, which if its not true the code will return an error:`errorQuda("Domain-wall operator has not been built");`.  Am I understand this correctly?

maddyscientist:
> Hi, thanks for the comment. For the `DIRAC` CMake settings, as I understand for example the Domain Wall in "lib/dslash_domain_wall_4d_m5inv.hpp" there is an if statement: `if constexpr (is_enabled<QUDA_DOMAIN_WALL_4D_DSLASH>()) `, which if its not true the code will return an error:`errorQuda("Domain-wall operator has not been built");`. Am I understand this correctly?

Yes, that's correct.

maddyscientist:
I have addressed the issue with the Dirac operators being compiled when they shouldn't be, and also fixed issues related to NVSHMEM linking.  Things are looking good.  I might clean up the dslash code generation bit in the future, but that's minor work, I can do after merge.

@weinbe2 can you review this PR please?

include/declare_enum.h
```diff
@@ -0,0 +1,60 @@
+/*
```
weinbe2, line 1: We should add these details to the LICENSES file

CMakeLists.txt
```diff
@@ -151,8 +151,8 @@ option(QUDA_DIRAC_CLOVER_HASENBUSCH "build clover Hasenbusch twist operators" ${
 option(QUDA_DIRAC_NDEG_TWISTED_MASS "build non-degenerate twisted mass Dirac operators" ${QUDA_DIRAC_DEFAULT})
 option(QUDA_DIRAC_NDEG_TWISTED_CLOVER "build non-degenerate twisted clover Dirac operators" ${QUDA_DIRAC_DEFAULT})
 option(QUDA_DIRAC_LAPLACE "build laplace operator" ${QUDA_DIRAC_DEFAULT})
-
 option(QUDA_DIRAC_DISTANCE_PRECONDITIONING "build code for distance preconditioned Wilson/clover Dirac operators" OFF)
+set(QUDA_DOMAIN_DECOMPOSITION "0" CACHE STRING "which domain decomposition to instantiate in QUDA (1-bit number - RedBlack)")
```
weinbe2, line 155: Can this be just `ON`/`OFF`, or is there a reason we want this to be a bitfield in the future?

include/domain_decomposition.h
```diff
@@ -0,0 +1,153 @@
...
+      }
+
+      return true;
+    }
+
+    // Checks if this is equal to given DDParam
+    inline bool operator==(const DDParam &dd) const
+    {
+      // if both are not in use we return true
+      if (not *this and not dd) return true;
```
weinbe2, line 133: Stylistic thing, we don't really use `not`, `and`, etc, elsewhere in QUDA, opting for the traditional `!`, `&&`, etc, instead. I think we should stick with the latter. Thoughts, @maddyscientist ? I'm willing to take the time to change the `not`/`and`/etc.

lib/CMakeLists.txt
```diff
@@ -504,6 +643,7 @@ endif()
 # set which precisions to enable
 target_compile_definitions(quda PUBLIC QUDA_PRECISION=${QUDA_PRECISION})
 target_compile_definitions(quda PUBLIC QUDA_RECONSTRUCT=${QUDA_RECONSTRUCT})
+target_compile_definitions(quda PUBLIC QUDA_DSLASH5TYPE=${QUDA_DSLASH5TYPE})
```
weinbe2: Is this used anywhere?

lib/color_spinor_project_dd.cu
```diff
@@ -0,0 +1,143 @@
...
+            for (int c = 0; c < p.Ncolor(); c++) p(parity, x_cb, s, c) = 0;
+        }
+      }
+    }
+  }
+
+  template <typename Float, typename DDArg, int nSpin, int nColor, typename Order>
+  void genericProjectDD(ColorSpinorField &a)
+  {
+    /* Reference CPU implementation
```
weinbe2: Remove this, unless there's a reason to uncomment it

include/kernels/dslash_staggered.cuh
```diff
@@ -221,12 +222,21 @@ namespace quda
...
+      }
+
       applyStaggered<nParity, mykernel_type, n_src_tile>(out, arg, coord, parity, idx, thread_dim, active, src_idx);
 
 #pragma unroll
       for (auto s = 0; s < n_src_tile; s++) out[s] *= arg.dagger_scale;
 
-      if (xpay && mykernel_type == INTERIOR_KERNEL) {
+      if (xpay && mykernel_type == INTERIOR_KERNEL && arg.dd_x.isZero(coord)) {
+        out = -out;
```
weinbe2: I believe this needs the loop over `n_src_tile` values?

weinbe2:
@sbacchio @pittlerf I'm beginning to go through this PR and I'm very impressed with all of the work done here. When you have the chance, can you please make a wiki page that describes the DD work and the current state of it?

The description can just be what you included in the PR description (remarks/design), and then an example command with an explanation---just something from one of the ctests, for ex---would be appreciated.

lib/CMakeLists.txt
```diff
@@ -119,6 +119,145 @@ set(QUDA_DSLASH_OBJS
...
+  list(APPEND QUDA_DSLASH_FILENAME_LIST "dslash_domain_wall_5d")
+endif()
+if(QUDA_DIRAC_LAPLACE)
+  list(APPEND QUDA_DSLASH_FILENAME_LIST "laplace")
+endif()
+
+set(QUDA_DSLASH_PREC_LIST "")
+set(QUDA_DSLASH_NCOLOR_LIST "3")
+set(QUDA_DSLASH_DDARG_LIST "DDNo")
+set(QUDA_DSLASH_RECONI_LIST "")
```
weinbe2, line 171: `QUDA_DSLASH_RECON_LIST`

lib/dirac_domain_wall_4d.cpp
```diff
@@ -61,6 +61,7 @@ namespace quda {
 
   void DiracDomainWall4D::M(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in) const
   {
+    assertNoDD(out, in); // TODO: DD not supported yet
```
weinbe2, line 64: I'm guessing it's not supported here yet because of `ApplyDslash5` and/or the `xpay`? If possible, can you add a quick comment as to the reason? Alternatively, this could just be covered in a QUDA issue.

lib/dirac_mobius.cpp
```diff
@@ -99,6 +99,7 @@ namespace quda {
 
   void DiracMobius::M(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in) const
   {
+    assertNoDD(out, in); // TODO: DD not supported yet
```
weinbe2, line 102: Similar comment as for `domain_wall_4d`

lib/dslash_domain_wall_4d_fused_m5.hpp
```diff
@@ -120,30 +120,38 @@ namespace quda
...
 
-    template <Dslash5Type dslash5_type_impl, Dslash5Type... N>
+    template <bool distance_pc, Dslash5Type dslash5_type_impl, Dslash5Type... N>
     DomainWall4DApplyFusedM5(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in,
                              cvector_ref<const ColorSpinorField> &x, const GaugeField &U,
                              cvector_ref<ColorSpinorField> &y, const Complex *b_5, const Complex *c_5, double a,
                              double m_5, int parity, bool dagger, const int *comm_override, double m_f,
-                             Dslash5TypeList<dslash5_type_impl, N...>, TimeProfile &profile)
+                             DistanceType<distance_pc>, Dslash5TypeList<dslash5_type_impl, N...>, TimeProfile &profile)
+#ifdef SIGNATURE_ONLY
```
weinbe2: What's this for?

weinbe2: Or, more specifically, what's the reason for this design---it feels pretty gross...

lib/dslash_domain_wall_5d.in.cu
```diff
@@ -0,0 +1,20 @@
+#include <dslash_domain_wall_5d.hpp>
+
+namespace quda
+{
+
+  constexpr QudaPrecision precision = QUDA_@QUDA_DSLASH_PREC@_PRECISION;
+  constexpr int nColor = @QUDA_DSLASH_NCOLOR@;
+  constexpr int reconI = @QUDA_DSLASH_RECONI@;
+
+  typedef @QUDA_DSLASH_DDARG@ DDArg;
```
weinbe2: These `typedef` here (and in similar files) should be replaced with `using` since that's more in line with the QUDA style, i.e., `using DDArg = @QUDA_DSLASH_DDARG@;`

weinbe2:
I'd like to better understand what's going on with the `#ifdef SIGNATURE_ONLY` pattern across many of the dslash files; at a high level I can sort of see what it's accomplishing, reducing code duplication, but it's messy from a software engineering perspective and I feel like there must be a more C++-y way to accomplish it. Thoughts @maddyscientist ?

tests/dslash_test_utils.h
```diff
@@ -276,6 +284,36 @@ struct DslashTestWrapper {
     inv_param.verbosity = verbosity;
   }
 
+  void init_domain_decomposition(int value, int color)
+  {
+    if (value == 0) {
```
weinbe2: I think we should replace `value` and `color` with some type of `enum`s as opposed to being just magic numbers to make this and other parts of the test utilities clearer. They can be `enum` that only live in the tests subdirectory, perhaps in `tests/utils/misc.h` (for lack of a better place right now), or they can leverage existing `enum` elsewhere in the code.

tests/dslash_test_utils.h
```diff
@@ -803,6 +850,80 @@ struct DslashTestWrapper {
...
+            dslash_time.cpu_min = std::min(dslash_time.cpu_min, host_timer.last());
+            dslash_time.cpu_max = std::max(dslash_time.cpu_max, host_timer.last());
+          }
+        }
+
+        spinor.DD(DD::reset);
+        out.DD(DD::reset);
+        spinorOut = cudaSpinorOut;
+
+        if (niter < 2) { // HACK: when benchmarking we do not produce reference solution
```
weinbe2, line 898: I wouldn't consider this a hack at all, this is a really smart way to do it :)

weinbe2:
I just completed a visual review, and needless to say this is very impressive work---I left some QUDA style convention comments, and a few questions. I think the most important piece of feedback I have is about `SIGNATURE_ONLY`; I'd really like something neater than that.

My thoughts are all orthogonal to doing actual code testing, so I'll carry on with that next.

include/domain_decomposition.h
maddyscientist: I think I agree with @weinbe2 here, and we should stick to legacy logical operators here.

CMakeLists.txt
sbacchio: Yes, we would like this to remain a bitfield since in the future, we would like to support various custom DD, e.g., multilevel-like, timeslices, etc. A bitfield would allow the user to enable the necessary DD at will.

include/declare_enum.h
pittlerf: I have included it in LICENSES

lib/CMakeLists.txt
pittlerf: I removed it

lib/color_spinor_project_dd.cu
pittlerf: I removed it

lib/dslash_domain_wall_5d.in.cu
pittlerf: I replaced all the necessary `typedef` with `using`

tests/dslash_test_utils.h
pittlerf: I created an `enum` for `value` and `color` in `tests/utils/misc.h`

include/domain_decomposition.h
weinbe2: I'll test and ideally create a PR to automagically handle this today, per @mathiaswagner 's suggestion offline. (Reference for the suggestion here: https://clang.llvm.org/extra/clang-tidy/checks/readability/operators-representation.html)

weinbe2:
FYI @pittlerf re:`SIGNATURE_ONLY`, @maddyscientist is going to look into it briefly and take care of it if there's an easy fix, otherwise we'll merge this and deal with it later---I don't want to postpone merging this PR much longer :)

CMakeLists.txt
weinbe2: Understood, thank you for explaining

tests/dslash_test_utils.h
```diff
@@ -1018,7 +1133,7 @@ struct DslashTestWrapper {
 
   void run_test(int niter, bool = false)
   {
-    {
+    if (getTuning()) {
```
weinbe2: This should be reverted because it's not just for autotuning; it also might up-clock the device, warm up caches, etc.

tests/staggered_dslash_ctest.cpp
```diff
@@ -3,17 +3,18 @@
...
+bool ctest_domain_decomposition = false;
 
 using ::testing::Bool;
 using ::testing::Combine;
 using ::testing::Range;
 using ::testing::TestWithParam;
 using ::testing::Values;
 
-class StaggeredDslashTest : public ::testing::TestWithParam<::testing::tuple<int, int, int>>
+class StaggeredDslashTest : public ::testing::TestWithParam<::testing::tuple<int, int, int, int, int>>
```
weinbe2: The new enums `QudaDomainDecompositionType` and `QudaDomainDecompositionColor` also need to be propagated to the staggered tests

weinbe2:
I just got done with the boolean op replacement, done by hand instead of with `clang-tidy` because I figured out very quickly that it was going to be much more complicated than I initially thought. I also elected to rename `include/kernels/color_spinor_project_dd.cuh` to `*domain_decomp.cuh` because I kept confusing it with the `_dd`, `_ds`, etc files for copying between precisions and I wanted to disambiguate that.

tests/dslash_test_utils.h
maddyscientist: 097ef1d327fa13f878d6e48861f915f48c4f3b33

maddyscientist (approved):
Thanks for all this great work.  Approving this PR.

weinbe2 (approved):
This is awesome work, approved!

PR #1452, opened 2024-04-03
Feature/compute clover force quda ndg

Marcogarofalo:
These are some changes I made to support the fermionic forces for the clover twisted mass non-degenerate fermion doubled. So far I only focused on getting the same result as in `tmLQCD`  so, probably the changes I made are not optimal.


lib/clover_force.cpp
```diff
@@ -44,22 +44,39 @@ namespace quda
...
       if (dagger) dirac->Dagger(QUDA_DAG_YES);
       dirac->Dslash(x[i][other_parity], p[i][parity], other_parity);
       // want to apply \hat Q_{-} = \hat M_{+}^\dagger \gamma_5 to get Y_o
       dirac->M(p[i][parity], p[i][parity]); // this is the odd part of Y
       if (dagger) dirac->Dagger(QUDA_DAG_NO);
 
-      gamma5(x[i][other_parity], x[i][other_parity]);
-      if (detratio) blas::xpy(x0[i][parity], p[i][parity]);
+      if(inv_param.twist_flavor == QUDA_TWIST_NONDEG_DOUBLET){
+        // FIXME: here we used the already existing ca_lambda_max member of inv_param. Maybe it's better to create a new parameter for this purpose
```
maddyscientist: Yes, this should be a new parameter, `since ca_lambda_max` is for something else completely (it sets the maximum eigenvalue for communication avoiding CG when using a Chebyshev basis).

tests/host_reference/clover_force_reference.cpp
```diff
@@ -825,41 +866,46 @@ void CloverSigmaOprod_reference(void *oprod_, quda::ColorSpinorField &inp, quda:
   gFloat *oprod = (gFloat *)oprod_;
   sFloat *x = (sFloat *)inx.data();
   sFloat *p = (sFloat *)inp.data();
+  
+  int flavors = 1;
```
maddyscientist: Rewrite this as
```c++
int flavors = inx.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ? inx.TwistFlavor() : 1;
```

lib/dslash_gamma_helper.cu
```diff
@@ -103,4 +103,44 @@ namespace quda {
...
+    }
+
+    void apply(const qudaStream_t &stream) {
+      TuneParam tp = tuneLaunch(*this, getTuning(), getVerbosity());
+      launch<Tau>(tp, stream, GammaArg<Float, nColor>(out, in, d));
+    }
+
+    void preTune() { out.backup(); }
+    void postTune() { out.restore(); }
+    long long flops() const { return 0; }
```
maddyscientist: This function can be deleted since the parent default return 0;

include/kernels/clover_sigma_outer_product.cuh
```diff
@@ -21,13 +21,18 @@ namespace quda
     using F = typename colorspinor_mapper<Float, nSpin, nColor>::type;
 
     Oprod oprod;
+    const bool doublet;         // whether we applying the operator to a doublet
+    const size_t VolumeCB;
```
maddyscientist: I don't think there's any need to make this a `size_t`, and `unsigned int` or `int` is sufficient.  Also could define it to include the factor of half to avoid needed to divide it in the kernel, e.g.
```
volume_4d_cb(inA[0].VolumeCB() / 2)
```

include/kernels/clover_sigma_outer_product.cuh
```diff
@@ -46,10 +51,13 @@ namespace quda
 
 #pragma unroll
     for (int i = 0; i < Arg::nvector; i++) {
-      const Spinor A = arg.inA[i](x_cb, parity);
-      const Spinor B = arg.inB[i](x_cb, parity);
-      Spinor C = A.sigma(nu, mu); // multiply by sigma_mu_nu
-      result += arg.coeff[i][parity] * outerProdSpinTrace(C, B);
+      for (int flavor=0; flavor<=arg.doublet; ++flavor){
```
maddyscientist: Noting that the compiler cannot unroll this loop since it doesn't know the value of `doublet`.  I don't think this will be a problem, since `flavor` isn't used for any local array indexing and only for global offsets, but if this kernel is underperforming, we may want to specialize this kernel for singlet and doublets.

include/kernels/clover_outer_product.cuh
```diff
@@ -76,31 +84,35 @@ namespace quda {
...
-          D_shift = (D_shift.project(dim,-1)).reconstruct(dim,-1);
-          result += outerProdSpinTrace(D_shift,C);
-
-          Link temp = arg.force(dim, x_cb, arg.parity);
-          Link U = arg.U(dim, x_cb, arg.parity);
-          result = temp + U*result*arg.coeff;
-          arg.force(dim, x_cb, arg.parity) = result;
-        }
-      } // dim
+      for (int flavor=0; flavor<=arg.doublet; ++flavor){
```
maddyscientist: As I noted for the clover sigma outer product kernel, this loop can't be unrolled by the compiler, and we may want to do that here through specializing for singlet and doublet.

include/kernels/clover_outer_product.cuh
```diff
@@ -30,11 +30,16 @@ namespace quda {
     int displacement;
     bool partitioned[4];
     real coeff;
+    const bool doublet;         // whether we applying the operator to a doublet
+    const size_t VolumeCB;
```
maddyscientist: No need to make these `size_t`, and they can be `unsigned int` without issue.

include/kernels/clover_outer_product.cuh
```diff
@@ -76,31 +84,35 @@ namespace quda {
...
-
-          Link temp = arg.force(dim, x_cb, arg.parity);
-          Link U = arg.U(dim, x_cb, arg.parity);
-          result = temp + U*result*arg.coeff;
-          arg.force(dim, x_cb, arg.parity) = result;
-        }
-      } // dim
+      for (int flavor=0; flavor<=arg.doublet; ++flavor){
+
+        const int flavor_offset_idx = flavor * (arg.VolumeCB/2);
```
maddyscientist: Include the division by two in the definition of the arg volume parameter to (slightly) simplify the kernel

include/dirac_quda.h
```diff
@@ -1099,7 +1099,7 @@ namespace quda {
   };
 
   void gamma5(ColorSpinorField &out, const ColorSpinorField &in);
-
+  void ApplyTau(ColorSpinorField &out, const ColorSpinorField &in, int d);
```
maddyscientist, line 1109: Doxygen needs to be added to this function declaration

maddyscientist (changes requested):
Thanks for this contribution @Marcogarofalo.  This looks like an excellent addition.  I've left various comments that need to be addressed.  Beyond those comments, a few other things outstanding:
* Augment CTest to include running the Nondeg clover force when non-degenerate twisted clover is compiled 
* clang-format the diff

How does the performance compare with the reference tmLQCD code?

include/kernels/clover_outer_product.cuh
Marcogarofalo: Thank you for the review. do you mean something like [b9519f0](https://github.com/lattice/quda/pull/1452/commits/b9519f0314b87ece2dd50fc093a8ea58cabc8996)
for specialization?

maddyscientist: I meant something a bit simpler: make the `doublet` member of the Arg class a `static constexpr bool` value that is set as a template argument to the Arg class, with it being set differently for the singlet and doublet cases.  This will allow you to use the same functor code for both, with the only difference being a a value of `1` or `2` in the template when instantiating the Arg.  The compiler will then know the trip count of the loop and will be able to unroll it.

Marcogarofalo: maybe [3232e69](https://github.com/lattice/quda/pull/1452/commits/3232e699beb2711f806fac2843e6c6548982c83f) is what you mean

maddyscientist: Yes this is what I was thinking.  😄  Maybe it would be better to have a `n_flavor` static member to make the for loop look more natural, e.g., instead of `<= Arg::doublet`, have `< Arg::n_flavor` with `n_flavor = 1` for singlet and `n_flavor = 2` for doublet.

Marcogarofalo: yes, I change it.

Marcogarofalo:
I think that the CI actions do not trigger the test for the ndeg doublet. I try locally to run
```
mpirun -n 2 $1 ./tests/clover_force_test  \
        --dslash-type twisted-clover \
        --flavor nondeg-doublet \
        --compute-clover 1 \
        --matpc odd-odd-asym \
        --dagger \
        --solution-type mat-pc-dag-mat-pc \
        --dim  2 4 6 8 --prec ${prec} \
        --gridsize 2 1 1 1\
        --niter 1 \
        --compute-clover-trlog 0\
        --verbosity debug \
        --enable-testing 0\
        --determinant-ratio 0 \
        --nsrc 1
```
and it passes but I have problems if I set ` --enable-testing 1` another Dirac operator is called. From `make test` the test is launched with  ` --enable-testing 1`.

I am also trying to measure the performance in some realistic setup.

maddyscientist:
@Marcogarofalo when you set `enable-testing` the predefined tests set at the bottom of `clover_force_test.cpp` will be run.  What you need to do is add the non-degenerate case to this list.  What I would suggest doing is adding a single/doublet option to testing tuple, e.g., add an extra bool
```
using test_t = ::testing::tuple<QudaPrecision, QudaDslashType, bool, int, bool>;
```
and then instantiate both single and double options in the test outer product
```
INSTANTIATE_TEST_SUITE_P(CloverForceTest, CloverForceTest,
                         Combine(Values(QUDA_SINGLE_PRECISION, QUDA_DOUBLE_PRECISION),
                                 Values(QUDA_CLOVER_WILSON_DSLASH, QUDA_TWISTED_CLOVER_DSLASH), Values(false, true),
                                 Values(1, 8), Values(false, true)),
                         gettestname);
```
This will add a nonsense test for doublet with regular clover fermions: you could deal with this by skipping this test if the clover / doublet options is encountered, or by splitting the instantiation into separate clover cases and twisted clover cases:
```
// regular clover (singlet only)
INSTANTIATE_TEST_SUITE_P(CloverForceTest, CloverForceTest,
                         Combine(Values(QUDA_SINGLE_PRECISION, QUDA_DOUBLE_PRECISION),
                                 Values(QUDA_CLOVER_WILSON_DSLASH), Values(false, true),
                                 Values(1, 8), Values(false)),
                         gettestname);

// twisted clover (singlet and doublet)
INSTANTIATE_TEST_SUITE_P(CloverForceTest, CloverForceTest,
                         Combine(Values(QUDA_SINGLE_PRECISION, QUDA_DOUBLE_PRECISION),
                                 Values(QUDA_TWISTED_CLOVER_DSLASH), Values(false, true),
                                 Values(1, 8), Values(false, true)),
                         gettestname);
```
Hope this helps

Marcogarofalo:
@maddyscientist thank you, I hope I manage to fix the tests with your suggestions. Some parameters were initialised with the command line arguments and not from `INSTANTIATE_TEST_SUITE_P`. Now when the test is called with `enable-testing 1` the parameter in `INSTANTIATE_TEST_SUITE_P` will overwrite their command-line counterpart, at least the one I checked.

Regarding the performance I tried on a 64^3x128 lattice integrating 3 orders of the rational approximations (`--nsrc 3` in a test) on 4 nodes of [jewels-booster ](https://apps.fz-juelich.de/jsc/hps/juwels/booster-overview.html) and I got:
tmLQCD native: 10.2713
tmLQCD + QUDA: 8.560943e-01  
of which half 4.157508e-01 was spent in  `computeTMCloverForceQuda`. The rest is spent in tmLQCD preparing and collecting the result, I think it can be reduced using resident_solution and accumulating the momentum on the GPU but,  so far, it isn't a significant amount in the overall HMC cost.

Marcogarofalo:
I fixed a few problems that were still present in the tests. Now there are three instantiations of the test parameters each of them guarded by the condition that the operator is compiled, hope this is ok

kostrzewa:
@maddyscientist What could be causing the incorrect timing of `computeTMCloverForceQuda` that we seem to see now that the ND force has been offloaded?

```
   computeTMCloverForceQuda Total time =   380.787 secs
                 download     =    95.652 secs ( 25.119%),       with     2582 calls at 3.705e+04 us per call
                   upload     =    83.164 secs ( 21.840%),       with     1033 calls at 8.051e+04 us per call
                     init     =    14.246 secs (  3.741%),       with    26165 calls at 5.445e+02 us per call
                  compute     =  8001.109 secs (2101.201%),      with   292924 calls at 2.731e+04 us per call
                    comms     =    54.024 secs ( 14.187%),       with     6426 calls at 8.407e+03 us per call
                     free     =    21.058 secs (  5.530%),       with   236599 calls at 8.901e+01 us per call
        total accounted       =  8269.254 secs (2171.620%)
        total missing         = -7888.466 secs (-2071.620%)
WARNING: Accounted time  8269.254 secs in computeTMCloverForceQuda is greater than total time   380.787 secs
```

Compared to the correct timing with just the light force offloaded?

```
   computeTMCloverForceQuda Total time =   170.504 secs
                 download     =    30.422 secs ( 17.842%),       with     1033 calls at 2.945e+04 us per call
                   upload     =    51.064 secs ( 29.949%),       with      631 calls at 8.093e+04 us per call
                     init     =     6.695 secs (  3.926%),       with    11704 calls at 5.720e+02 us per call
                  compute     =    36.642 secs ( 21.491%),       with    22085 calls at 1.659e+03 us per call
                    comms     =    32.586 secs ( 19.112%),       with     2524 calls at 1.291e+04 us per call
                     free     =    10.476 secs (  6.144%),       with   129818 calls at 8.070e+01 us per call
        total accounted       =   167.885 secs ( 98.464%)
        total missing         =     2.619 secs (  1.536%)
```

Marcogarofalo:
I think some timers were not stopped when multiple fermions were passed to  `computeTMCloverForceQuda`. This never happens in the light sector in tmLQCD.

@maddyscientist I found a similar problem in `loadCloverQuda` which I hope I solved with [a65cafb](https://github.com/lattice/quda/pull/1452/commits/a65cafbe53c66e8146fb29f510c01cfee1b34eeb)

maddyscientist:
I'm traveling this week, but will get to final testing next week.  Looks like almost all issues have been resolved.

Thanks for catching the missing timer stopping with loadCloverQuda @Marcogarofalo

maddyscientist:
This looks good now, and I think it's just about ready to be merged.  I fixed a couple of issues and pushed the changes.  Namely whenever we specialize kernels, e.g., in this case the doublet specialization, we need to ensure that the different variants of the kernels have different `tuneKey` values to ensure they use separate tunings.  This ensures that the optimal parameters are used for the specializations, but more importantly prevents the possibility of an invalid launch config being used, e.g., the block/grid parameters for singlet may not be valid for doublet.

Anything left from your end @Marcogarofalo?

Marcogarofalo:
@maddyscientist thank you for fixing the issue with `tuneKey`. I do not have any left open from my side.

maddyscientist:
cscs-ci run

include/kernels/clover_outer_product.cuh
```diff
@@ -30,11 +32,17 @@ namespace quda {
     int displacement;
     bool partitioned[4];
     real coeff;
+    const unsigned int volume_4d_cb;
+    const unsigned int GhostFace_4d_cb;
```
weinbe2: Variable name doesn't adhere to our typical convention iirc, but maybe there's precedent for this elsewhere.

include/kernels/clover_outer_product.cuh
```diff
@@ -30,11 +32,17 @@ namespace quda {
...
 
     CloverForceArg(GaugeField &force, const GaugeField &U, const ColorSpinorField &inA, const ColorSpinorField &inB,
                    const ColorSpinorField &inC, const ColorSpinorField &inD, const unsigned int parity,
                    const double coeff) :
-      kernel_param(dim3(dim == -1 ? inA.VolumeCB() : inB.GhostFaceCB()[dim])), // inB since it has a ghost allocated
+      // kernel_param(dim3(dim == -1 ? inA.VolumeCB() : inB.GhostFaceCB()[dim])), // inB since it has a ghost allocated
```
weinbe2: This comment can be deleted

include/kernels/dslash_gamma_helper.cuh
```diff
@@ -84,6 +84,16 @@ namespace quda {
       case 3: arg.out(x_cb, parity) = in.gamma(3); break;
       case 4: arg.out(x_cb, parity) = in.gamma(4); break;
       }
+      if (arg.doublet) {
+        ColorSpinor<typename Arg::real, Arg::nColor, 4> in = arg.in(x_cb + 1 * arg.volumeCB, parity);
```
weinbe2: We can remove the `1 *`, though I imagine any compiler would optimize this out.

lib/dslash_gamma_helper.cu
```diff
@@ -103,4 +103,39 @@ namespace quda {
...
+    }
+
+    void preTune() { out.backup(); }
+    void postTune() { out.restore(); }
+    long long bytes() const { return out.Bytes() + in.Bytes(); }
+  };
+
+  // Apply the tau1 matrix to a doublet colorspinor field
+  // out(x) = tau_1*in
+#ifdef GPU_TWISTED_MASS_DIRAC
```
weinbe2, line 133: @maddyscientist do we have a `constexpr` function for `is_enable_twisted_mass` or w/e it should be that we could stick in an `if constexpr` instead of using `#define`s? And should this be twisted mass (boolean or) twisted clover?

tests/clover_force_test.cpp
```diff
@@ -170,11 +172,23 @@ class CloverForceTest : public ::testing::TestWithParam<test_t>
 protected:
   test_t param;
 
+  bool skip()
+  {
+    if (!quda::is_enabled(::testing::get<0>(param))) return true; // skip if precision is not enbaled
```
weinbe2: `enbaled` -> `enabled`

maddyscientist:
cscs-ci run

tests/dslash_test_utils.h
```diff
@@ -124,7 +124,9 @@ struct DslashTestWrapper {
     static bool first_time = true;
     if (first_time) {
       init_host(argc, argv);
-      first_time = false;
+      // init_host is changing inv_param.dagger which is set up from newQudaInvertParam()
+      // so it need to be called all the time
```
weinbe2: `need` -> `needs`; does this logic still hold up? It seems like there's still a risk for a lot of redundant allocations/setup, maybe there's a better way to handle this, or it should only be done this way for non-degenerate clover? (Might just be I'm confused about what exactly this change does.)

weinbe2:
I finished a quick visual review---I have no fundamental issues with the PR, this is fantastic work. All of my comments are pedantic, stylistic things. Getting these addressed before merging isn't a hill I'll die on.

lib/dslash_gamma_helper.cu
maddyscientist: We don't have one at the moment.  I have a long-term grander plan which will be of the form `template <QudaDslashType> constexpr bool is_enabled()`.  With that in mind, I didn't hack it in here.  Will take another look at this function though to see if there's something cleaner here.

maddyscientist: Marking this as resolved: we'll fix this in a future PR when I do as I suggest above.

include/kernels/clover_outer_product.cuh
maddyscientist: 7acf254ba019063551721afd3969ab77469eb8a0

include/kernels/clover_outer_product.cuh
maddyscientist: 7acf254ba019063551721afd3969ab77469eb8a0

include/kernels/dslash_gamma_helper.cuh
maddyscientist: The compiler does indeed optimize this out.  I have done a bit of cleanup to this function to make it a bit cleaner though e7a54b3c4069b2674d6d423eccfe09ea1d40078d.

tests/clover_force_test.cpp
maddyscientist: e7a54b3c4069b2674d6d423eccfe09ea1d40078d

tests/dslash_test_utils.h
maddyscientist: This change also introduced a memory leak.  I've fixed this properly here e7a54b3c4069b2674d6d423eccfe09ea1d40078d

include/kernels/clover_outer_product.cuh
maddyscientist: [57a6e9b](https://github.com/lattice/quda/pull/1452/commits/57a6e9bc214734096b84061971bac8dd77bfd84c)

include/kernels/clover_outer_product.cuh
maddyscientist: [57a6e9b](https://github.com/lattice/quda/pull/1452/commits/57a6e9bc214734096b84061971bac8dd77bfd84c)

lib/dslash_gamma_helper.cu
maddyscientist: https://github.com/lattice/quda/pull/1452/commits/378e5c24afabe99142a9fd48102492edd57d882c

include/kernels/clover_sigma_outer_product.cuh
maddyscientist: [57a6e9b](https://github.com/lattice/quda/pull/1452/commits/57a6e9bc214734096b84061971bac8dd77bfd84c)

include/kernels/clover_outer_product.cuh
maddyscientist: [7efe64e](https://github.com/lattice/quda/pull/1452/commits/7efe64eb014b34efd613744cf6bcd6107c396292)

lib/clover_force.cpp
maddyscientist: Resolving this.  Not that whenever we add something to `QudaInvertParam` or `QudaGaugeParam` we need to mirror the addition in the Fortran interface (b32949e832ff7dd2c4c957a274cf8683b8202094).

include/kernels/clover_sigma_outer_product.cuh
maddyscientist: https://github.com/lattice/quda/pull/1452/commits/00489269fc7186326ae55c90ef4db6837a74740b

tests/host_reference/clover_force_reference.cpp
maddyscientist: [9d9c6a9](https://github.com/lattice/quda/pull/1452/commits/9d9c6a9a450afdb85c73dd5a3c8b1981e9ff7ccd)

maddyscientist (approved):
Approving from my end.  I have addressed all of @weinbe2's comments and fixed a few additional issues I found during my tests.

weinbe2 (approved):
Looks great to me. Thanks all!

Marcogarofalo:
Thank you @maddyscientist and @weinbe2 for all the support and help.

PR #1454, opened 2024-04-09
Fix a bug in `DiracWilsonPC::prepare`.

SaltyChiang:
`./invert_test --solution-type mat --matpc odd-odd` fails in the recent commit. This PR fixes it.


maddyscientist (approved):

maddyscientist:
Thanks for catching my bug @SaltyChiang.

PR #1456, opened 2024-04-25
Added SM_90

bjoo:
Nothing major, just added sm_89 to the arch list

### DIFF
--- a/lib/targets/cuda/target_cuda.cmake
+++ b/lib/targets/cuda/target_cuda.cmake
@@ -17,8 +17,8 @@ endif()
 
 set(QUDA_GPU_ARCH
     ${QUDA_DEFAULT_GPU_ARCH}
-    CACHE STRING "set the GPU architecture (sm_60, sm_70, sm_80)")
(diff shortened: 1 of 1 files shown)


maddyscientist:
The reason we don't have sm_89 in there is because it's not a major version, we decided sometime ago not to repopulate the list with all possible minor arch versions else this list becomes unwieldy, e.g., it would be at least `sm_52, sm_60, sm_61, sm_70, sm_75, sm_80, sm_86, sm_89, sm_90, sm_100`, from Maxwell through Blackwell.

Note that one can override the list by manually setting it directly with `cmake`, this list just controls what is shown in `ccmake`.

Also note that Ada (sm_89) is binary compatible with Ampere (sm_80), so Ada is covered already.

maddyscientist:
I do see however that Hopper is missing the list.  Can you modify your PR to remove `sm_89` and add `sm_90` 😄 ?

bjoo:
Done. I should really change the branch name now to hotfix/sm_90 but that may be a kerfuffle.

maddyscientist (approved):

PR #1457, opened 2024-04-26
Bugfix for single-parity vector inflation in `VectorIO`

weinbe2:
This hotfix branch fixes a bug where the flag `--eig-io-parity-inflate yes` (more specifically, the underlying variable set in the eigensolver parameters structure) wasn't being respected when saving vectors from single-parity operators.

As part of fixing this bug, I extended the `io_test` unit tests to explicitly check parity inflation for both even and odd vectors. The test to save an inflated single-parity field, load the resulting full field, and make sure the even or odd subset of the loaded full field (as appropriate) matches the saved field.

This bug was originally reported by @leonhostetler, who has independently verified that this fix addresses the issue he was hitting in his workflow --- thank you for catching it!

(FYI @maddyscientist --- I `clang-format`ed this along the way, so there's no need for a subsequent `clang-format` commit.)

### DIFF
--- a/lib/vector_io.cpp
+++ b/lib/vector_io.cpp
@@ -56,8 +56,8 @@ namespace quda
       }
 
       // if we're loading inflated vectors, we need to grab the spinor size from `tmp` instead of `v0`.
-      auto spinor_X = parity_inflate ? tmp[0].X() : v0.X();
-      auto spinor_site_subset = parity_inflate ? tmp[0].SiteSubset() : v0.SiteSubset();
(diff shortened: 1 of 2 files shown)


maddyscientist (approved):
Thanks for the fix @weinbe2, and extending the test coverage.

weinbe2:
FYI: I confirmed that `mpirun -np 2 ./io_test --gridsize 1 1 1 2`, i.e., a multi-GPU run with both staggered and Wilson fermions, successfully passes.

PR #1458, opened 2024-04-29
fix some recent test regressions

jcosborn:


maddyscientist (approved):
Much obliged @jcosborn

PR #1459, opened 2024-04-30
Fix buffer overflow in Unpack

dmcdougall:
This addresses a correctness check failure in various staggered dslash tests.

### DIFF
--- a/include/gauge_field_order.h
+++ b/include/gauge_field_order.h
@@ -1480,7 +1480,7 @@ namespace quda {
             for (int i = 0; i < 9; i++) out[i] = cmul(z, out[i]);
           } else { // stagic phase
 #pragma unroll
-            for (int i = 0; i < 18; i++) { out[i] *= phase; }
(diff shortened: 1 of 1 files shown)


dmcdougall:
@jxy Can you please confirm this fixes the staggered dslash test failure you were observing.

jxy:
Hi Damon, that was an excellent detective work! Thanks a lot!

I believe it fixed all the issues as I no longer see all the r8 failures as before.

I got a single failure with half_r8_partition0, but I could never reproduce this failure after 20 more runs. So I guess that's probably because of some early hardware instabilities.

I tested with rocm 6.0.3 on tioga at LLNL.

dmcdougall:
> Hi Damon, that was an excellent detective work! Thanks a lot!

That's very kind of you, but I can't claim credit for this one.  I thought it was a compiler bug and one of our compiler engineers found it.  I'll pass along your comments to them.

Appreciate you testing.

maddyscientist (approved):
Thanks @dmcdougall

PR #1460, opened 2024-04-30
Turn on domain wall builds

### LINKED ISSUE #1440: Turn on domain wall code builds in ROCm CI
Here is the relevant line: https://github.com/lattice/quda/blob/develop/.github/workflows/rocm-build-ci.yml#L25

dmcdougall:
Fixes #1440.


maddyscientist (approved):

maddyscientist:
@dmcdougall I see domain wall fails to build with ROCm 5.7.2.  What do you suggest to do here?

dmcdougall:
I expected that.  Someone (I think it might have been Balint) reported this problem to me in an email, and the reason we didn't catch it is because it was in the domain wall piece, which I didn't have enabled.

The summary is that I already know about this failure and I have a ticket open internally to fix it.

If you don't want noisy failures showing up in other PRs, it's probably a good idea to leave this unmerged until it's fixed in the ROCm stack.

dmcdougall:
Er, wait, my memory sucks.  This failure is with 5.7.2, and I believe this was fixed in 6.0.2.  I'll replace the 5.7.2 workflow with the 6.0.2 workflow and see what happens.  If it's all green there then I'll probably just drop the 5.7.2 from the workflow.

maddyscientist:
@dmcdougall looks like v6 works fine, so I guess we can merge this and move on?

dmcdougall:
@maddyscientist Agree

PR #1464, opened 2024-05-08
Eigensolvers: check and make sure seeding vectors don't have nan or info norms

weinbe2:
By default, the routine that prepared initial guesses for QUDA's eigensolvers only checked to make sure that the initial vector had a non-zero norm. This extends it to checking for `nan` and `inf` as well. Reported by @leonhostetler .

Don't merge until I get confirmation from Leon that this works, the initial "fix" I gave him modified `interface_quda.cpp` instead of checking deeper down in the eigensolver routines.

### DIFF
--- a/lib/eigensolve_quda.cpp
+++ b/lib/eigensolve_quda.cpp
@@ -143,9 +143,10 @@ namespace quda
     // Use 0th vector to extract meta data for the RNG.
     RNG rng(kSpace[0], 1234);
     for (int b = 0; b < block_size; b++) {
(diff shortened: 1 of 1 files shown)


maddyscientist (approved):
Tentatively approving subject to confirmation from @leonhostetler that this fixes the issue.

weinbe2:
> Tentatively approving subject to confirmation from @leonhostetler that this fixes the issue.

He confirmed via our e-mail chain, so this should be good to go.

PR #1465, opened 2024-05-15
Add multi-RHS support throughout QUDA

maddyscientist:
A long-time coming, this PR adds support for multi-RHS to all stencil kernels in QUDA:
* Add multi-RHS to all dslash kernels and derivative operators that utilie the dslash policies
* All member functions of the `Dirac` classes have been promoted to being MRHS aware
* Various member functions additions to `cvector_ref<ColorSpinorField>` for convenience
* The regular bias and reduction kernels are now all batch aware, and can applied to batches of vectors.  This also fixes a bug reported by @SaltyChiang. 
* Cleanup and optimization of the clover outer product kernel
* Extends multi-RHS support to a clover force computation, to provide acceleration of RHMC force computation
* Fix some prior issues with the staggered two link quark smearing tests and apply cleanup
* Fixes some issues with `eigensolver_test` and `invert_test` that led to test failures if we had a pure double precision build (reported by @dmcdougall)

With respect to performance, the performance gain from multi-RHS is dependent on the operator in question, but 2-2.5x is observed for Wilson-like operators at large volume, in line with roofline projections.  For smaller volumes, the gain can exceed roofline expectations due to additional gain from parallelism.

Subsequent work will focus on the addition of multi-RHS solvers.  This PR is focused on getting the kernels available.  Worth noting that the block Lanczos solver works out of the box with this PR though.

This PR is ready for review, but there are few things outstanding:
- [x]  Present default maximum NRHS limit is set to 12, else some of the dslash kernels fail to compile without large kernel arg support.  Need to work out how much we can squeeze in the default 4KiB limit
- [x]  Test NVSHMEM is still working
- [x]  Robustness fixes if dslash is called on large number of RHS: at present the code will error, but it should divide and conquer like is done with block blas kernels


maddyscientist:
@Marcogarofalo @kostrzewa could you test tmLQCD with this branch please if you get a chance, including the force calculations?  You should see a nice performance boost for the clover force, but good to verify correctness 😄

Marcogarofalo:
@maddyscientist thank you for the update. I check with tmLQCD the correctness and everything seems ok.
I see a performance boost on a 64^3x128  with 4 nodes of jewels-booster (4 × NVIDIA A100)
`computeTMCloverForceQuda` with `nvector=5`
- mrhs branch      5.378337e-01 s
- develop branch 5.885901e-01 s

with `nvector=3`
- mrhs branch 3.881676e-01 s
- develop branch 4.161852e-01 s

maddyscientist:
Thanks for confirming at your end @Marcogarofalo.

Another ask: could you send me the output from the QUDA profile for the HMC runs for tmLQCD (both the stdout printed at the end of the run and the tsk files)?  I suspect that the main limiter for the clover force at present is the communication, but I would like to confirm this.  There is 2x more communication being done than is needed in this function, and I'm wondering how much of an overhead this is.  Thx

Marcogarofalo:
@maddyscientist, here is the stdout output of the QUDA profile. I did not find any tsk file, could you please tell me how to generate them?
```

               initQuda Total time =     0.433 secs
                     init     =     0.433 secs ( 99.996%),	 with        2 calls at 2.163e+05 us per call
        total accounted       =     0.433 secs ( 99.996%)
        total missing         =     0.000 secs (  0.004%)

          loadGaugeQuda Total time =     2.002 secs
                 download     =     1.867 secs ( 93.249%),	 with       13 calls at 1.436e+05 us per call
                     init     =     0.093 secs (  4.645%),	 with       91 calls at 1.022e+03 us per call
                  compute     =     0.040 secs (  1.989%),	 with       65 calls at 6.126e+02 us per call
                     free     =     0.000 secs (  0.002%),	 with      364 calls at 1.016e-01 us per call
        total accounted       =     2.000 secs ( 99.885%)
        total missing         =     0.002 secs (  0.115%)

         loadCloverQuda Total time =     1.027 secs
                     init     =     0.082 secs (  8.004%),	 with       65 calls at 1.265e+03 us per call
                  compute     =     0.153 secs ( 14.848%),	 with       26 calls at 5.866e+03 us per call
                    comms     =     0.456 secs ( 44.357%),	 with       13 calls at 3.505e+04 us per call
                     free     =     0.096 secs (  9.300%),	 with      410 calls at 2.330e+02 us per call
        total accounted       =     0.786 secs ( 76.509%)
        total missing         =     0.241 secs ( 23.491%)

             invertQuda Total time =   461.892 secs
                 download     =     2.089 secs (  0.452%),	 with       30 calls at 6.962e+04 us per call
                   upload     =     0.812 secs (  0.176%),	 with       30 calls at 2.708e+04 us per call
                     init     =    13.152 secs (  2.848%),	 with   344740 calls at 3.815e+01 us per call
                 preamble     =    58.827 secs ( 12.736%),	 with   409205 calls at 1.438e+02 us per call
                  compute     =   342.744 secs ( 74.205%),	 with  7419610 calls at 4.619e+01 us per call
                 epilogue     =     2.413 secs (  0.523%),	 with    10412 calls at 2.318e+02 us per call
                     free     =    12.484 secs (  2.703%),	 with  6999292 calls at 1.784e+00 us per call
                    eigen     =     0.100 secs (  0.022%),	 with    63079 calls at 1.593e+00 us per call
        total accounted       =   432.623 secs ( 93.663%)
        total missing         =    29.269 secs (  6.337%)

   invertMultiShiftQuda Total time =   184.085 secs
                 download     =     1.069 secs (  0.581%),	 with       26 calls at 4.112e+04 us per call
                   upload     =    10.385 secs (  5.642%),	 with      140 calls at 7.418e+04 us per call
                     init     =     2.766 secs (  1.502%),	 with    11928 calls at 2.319e+02 us per call
                 preamble     =     0.750 secs (  0.408%),	 with      192 calls at 3.907e+03 us per call
                  compute     =   166.006 secs ( 90.179%),	 with   741854 calls at 2.238e+02 us per call
                 epilogue     =     1.892 secs (  1.028%),	 with     5092 calls at 3.716e+02 us per call
                     free     =     1.052 secs (  0.572%),	 with   757778 calls at 1.389e+00 us per call
        total accounted       =   183.920 secs ( 99.910%)
        total missing         =     0.165 secs (  0.090%)

   computeGaugeForceQuda Total time =     1.798 secs
                 download     =     0.694 secs ( 38.570%),	 with        7 calls at 9.908e+04 us per call
                   upload     =     0.553 secs ( 30.747%),	 with        7 calls at 7.898e+04 us per call
                     init     =     0.041 secs (  2.258%),	 with       77 calls at 5.273e+02 us per call
                  compute     =     0.269 secs ( 14.963%),	 with        7 calls at 3.844e+04 us per call
                    comms     =     0.183 secs ( 10.202%),	 with        7 calls at 2.621e+04 us per call
                     free     =     0.032 secs (  1.779%),	 with     1106 calls at 2.892e+01 us per call
        total accounted       =     1.772 secs ( 98.518%)
        total missing         =     0.027 secs (  1.482%)

   computeTMCloverForceQuda Total time =    10.860 secs
                 download     =     2.321 secs ( 21.369%),	 with       60 calls at 3.868e+04 us per call
                   upload     =     2.000 secs ( 18.417%),	 with       24 calls at 8.334e+04 us per call
                     init     =     0.353 secs (  3.247%),	 with      631 calls at 5.588e+02 us per call
                  compute     =     1.923 secs ( 17.705%),	 with     1614 calls at 1.191e+03 us per call
                    comms     =     3.290 secs ( 30.289%),	 with      200 calls at 1.645e+04 us per call
                     free     =     0.552 secs (  5.087%),	 with     6215 calls at 8.890e+01 us per call
        total accounted       =    10.438 secs ( 96.115%)
        total missing         =     0.422 secs (  3.885%)

                endQuda Total time =     0.909 secs
                     free     =     0.033 secs (  3.603%),	 with       92 calls at 3.560e+02 us per call
        total accounted       =     0.033 secs (  3.603%)
        total missing         =     0.876 secs ( 96.397%)

       initQuda-endQuda Total time =   818.633 secs

                   QUDA Total time =   663.006 secs
                 download     =     8.039 secs (  1.213%),	 with      136 calls at 5.911e+04 us per call
                   upload     =    13.751 secs (  2.074%),	 with      201 calls at 6.841e+04 us per call
                     init     =    16.924 secs (  2.553%),	 with   357534 calls at 4.734e+01 us per call
                 preamble     =    59.582 secs (  8.987%),	 with   409397 calls at 1.455e+02 us per call
                  compute     =   511.220 secs ( 77.106%),	 with  8163176 calls at 6.263e+01 us per call
                    comms     =     3.929 secs (  0.593%),	 with      220 calls at 1.786e+04 us per call
                 epilogue     =     4.306 secs (  0.649%),	 with    15504 calls at 2.777e+02 us per call
                     free     =    14.310 secs (  2.158%),	 with  7765257 calls at 1.843e+00 us per call
                    eigen     =     0.101 secs (  0.015%),	 with    63079 calls at 1.601e+00 us per call
        total accounted       =   632.162 secs ( 95.348%)
        total missing         =    30.845 secs (  4.652%)

Device memory used = 32039.1 MiB
Pinned device memory used = 3024.0 MiB
Managed memory used = 0.0 MiB
Shmem memory used = 0.0 MiB
Page-locked host memory used = 3040.4 MiB
Total host memory used >= 3623.8 MiB
```

lib/blas_quda.cu
```diff
@@ -75,7 +79,7 @@ namespace quda {
           constexpr int N = n_vector<device_store_t, true, nSpin, site_unroll>();
           constexpr int Ny = n_vector<device_y_store_t, true, nSpin, site_unroll>();
           constexpr int M = site_unroll ? (nSpin == 4 ? 24 : 6) : N; // real numbers per thread
-          const int threads = x.Length() / (nParity * M);
+          const int threads = x[0].Length() / (nParity * M);
```
weinbe2: Should this change (and others like it) be unnecessary because of the `Length()`, etc, calls added to the definition of `vector_ref` in `reference_wrapper_helper.h`?

kostrzewa:
@maddyscientist Apologies for the delay here. Please find below two profiles.

### QUDA develop, 8 trajectories, 64c128 at phys. point

- profile : [QUDA_tmLQCD_HMC_8trajectories_64c128_phys_point_develop.txt](https://github.com/lattice/quda/files/15388070/QUDA_tmLQCD_HMC_8trajectories_64c128_phys_point_develop.txt)

- tunecaches : [quda-develop-d15791.zip](https://github.com/lattice/quda/files/15388100/quda-develop-d15791.zip)

### QUDA feature/mrhs, 4 trajectories, 64c128 at phys. point

- profile: [QUDA_tmLQCD_HMC_4trajectories_64c128_phys_point_mrhs.txt](https://github.com/lattice/quda/files/15388107/QUDA_tmLQCD_HMC_4trajectories_64c128_phys_point_mrhs.txt)

- tunecaches: [quda-mrhs-fc92e.zip](https://github.com/lattice/quda/files/15388103/quda-mrhs-fc92e.zip)

include/color_spinor_field_order.h
```diff
@@ -1156,6 +1157,22 @@ namespace quda
...
+
+      /**
+         @brief This accessor routine returns a const
+         colorspinor_ghost_wrapper to this object, allowing us to
+         overload various operators for manipulating at the site
+         level interms of matrix operations.
+         @param[in] dim Dimensions of the ghost we are requesting
+         @param[in] ghost_idx Checkerboarded space-time ghost index we are requesting
+         @param[in] parity Parity we are requesting
+         @return Instance of a colorspinor_ghost+wrapper that curries in access to
```
weinbe2: `colorspinor_ghost+wrapper` -> `colorspinor_ghost_wrapper`

weinbe2: command+F for the same typo, it's a few other places!

include/dslash_helper.cuh
```diff
@@ -349,26 +351,26 @@ namespace quda
       retcount_inter(dslash::get_shmem_retcount_inter())
 #endif
     {
-      if (in.data() == out.data()) errorQuda("Aliasing pointers");
+      for (auto i = 0u; i < in.size(); i++)
```
weinbe2, line 355: Not that I want to open a quadratic scaling can of worms, but wouldn't it be a problem if anything in `in` aliased with anything in `out`? (...or anything in `out` with anything else in `out`)

weinbe2:
Somewhat orthogonal comment to this PR, but while we're here can I request that we rename `covDev.cu[h]` to `covariant_derivative.cu[h]` or something of the sort? It's the only outlier file in terms of the naming convention, and iirc `git` should note that it's just a "rename" so we won't lose the commit history.

weinbe2:
Ghost accessor question---right now, source index offsets to ghost accessors are performed via explicit pointer arithmetic, a representative case below:
```
const Vector in = arg.halo.Ghost(d, 1, ghost_idx + src_idx * arg.dc.ghostFaceCB[d], their_spinor_parity);
```
Where the pointer arithmetic I'm referring to is `ghost_idx + src_idx * arg.dc.ghostFaceCB[d]`. Can we make `src_idx` a separate argument so it looks something more like ` arg.halo.Ghost(d, 1, ghost_idx, src_idx, their_spinor_parity)`? Perhaps `ghostFaceCB` could be a field of the halo accessor.

The reason I'm asking this question is because I see this explicit pointer offsetting as antithetical to an accessor---implementation details should be under the hood. It would also make it more challenging to quickly test some type of hypothetical multi-rhs packed halo format in the future. (I guess this concern is also true with the arrays of "local" accessors, but one thought at a time.)

include/kernels/dslash_clover_helper.cuh
```diff
@@ -42,18 +43,22 @@ namespace quda {
...
 
-    CloverArg(ColorSpinorField &out, const ColorSpinorField &in, const CloverField &clover,
-	      int parity, real kappa=0.0, real mu=0.0, real epsilon = 0.0,
+    CloverArg(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in, const CloverField &clover,
+	      int parity, real kappa = 0.0, real mu = 0.0, real epsilon = 0.0,
 	      bool dagger = false, QudaTwistGamma5Type twist = QUDA_TWIST_GAMMA5_INVALID) :
-      kernel_param(dim3(in.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ? in.VolumeCB()/2 : in.VolumeCB(),
-                        in.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ? 2 : 1, in.SiteSubset())),
-      out(out), in(in),
+      kernel_param(dim3(in.TwistFlavor() == QUDA_TWIST_NONDEG_DOUBLET ? in[0].VolumeCB() / 2 : in[0].VolumeCB(),
```
weinbe2: This is applicable in multiple files, but: should we add a specialized `VolumeCB()` function to `vector_ref` and get rid of explicit indexing?

include/kernels/dslash_domain_wall_4d.cuh
```diff
@@ -8,14 +8,15 @@ namespace quda
...
-    DomainWall4DArg(ColorSpinorField &out, const ColorSpinorField &in, const GaugeField &U, double a, double m_5,
-                    const Complex *b_5, const Complex *c_5, bool xpay, const ColorSpinorField &x, int parity,
+    DomainWall4DArg(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in,
+                    const ColorSpinorField &halo, const GaugeField &U, double a, double m_5,
+                    const Complex *b_5, const Complex *c_5, bool xpay, cvector_ref<const ColorSpinorField> &x, int parity,
                     bool dagger, const int *comm_override) :
-      WilsonArg<Float, nColor, nDim, reconstruct_>(out, in, U, xpay ? a : 0.0, x, parity, dagger, comm_override),
-      Ls(in.X(4))
+      WilsonArg<Float, nColor, nDim, reconstruct_>(out, in, halo, U, xpay ? a : 0.0, x, parity, dagger, comm_override),
+      Ls(in[0].X(4))
```
weinbe2: Similar to comments I've made elsewhere, should we make `.X()` a specialized function in the `vector_ref` class definition?

include/kernels/laplace.cuh
```diff
@@ -29,27 +32,36 @@ namespace quda
...
+               int parity, bool dagger, const int *comm_override) :
+      DslashArg<Float, nDim>(out, in, halo, U, x, parity, dagger, a != 0.0 ? true : false, 1, false, comm_override),
+      halo_pack(halo),
+      halo(halo),
       U(U),
       a(a),
       b(b),
       dir(dir)
     {
+      if (out.size() > max_n_src) errorQuda("vector set size %lu greater than max size %d", out.size(), max_n_src);
```
weinbe2: This check should be in more files (or we should eliminate it from here and guarantee it elsewhere)

maddyscientist: This aspect will be addressed with this commit https://github.com/lattice/quda/pull/1465/commits/314468707963b1b30cbe80204957e62f23d725c4, where the `instantiate` function used for the Dslash functions will deal with auto splitting the fields as needed.  With this done, we can safely remove all these checks.  I'll leave a single check in the `DslashArg` constructor though, since that will guarantee it is called by every operator.

include/kernels/dslash_clover_helper.cuh
maddyscientist: e679e8216

include/color_spinor_field_order.h
maddyscientist: e679e8216

include/kernels/laplace.cuh
maddyscientist: e679e8216 puts the check in `DslashArg` and removes it from all child operators.

include/kernels/dslash_domain_wall_4d.cuh
maddyscientist: 3c9dd1609

lib/blas_quda.cu
maddyscientist: e679e82

include/color_spinor_field_order.h
```diff
@@ -1184,31 +1201,39 @@ namespace quda
       using AllocInt = typename AllocType<huge_alloc>::type;
       using norm_type = float;
       Float *field = nullptr;
+#if 0
```
weinbe2: Marking this just so it doesn't get lost---I'm guessing this can reduce the size of the accessor, with the trade-off being extra runtime work?

weinbe2:
Not the highest priority, but it should be easy to apply the multi-rhs optimizations to the KD operator (`lib/staggered_kd_apply_xinv.cu` and `include/kernels/staggered_kd_apply_xinv_kernel.cuh`). Testing it is another story... I need to get it exposed directly through `staggered_invert_test`, etc. Feel free to punt this back to me for the future, just file an issue about it.

include/dslash.h
```diff
@@ -327,8 +329,9 @@ namespace quda
 
     Arg &dslashParam; // temporary addition for policy compatibility
 
-    Dslash(Arg &arg, const ColorSpinorField &out, const ColorSpinorField &in, const std::string &app_base = "") :
-      TunableKernel3D(in, 1, arg.nParity), arg(arg), out(out), in(in), nDimComms(4), dslashParam(arg)
+    Dslash(Arg &arg, cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in,
+           const ColorSpinorField &halo, const std::string &app_base = "") :
+      TunableKernel3D(in[0], halo.X(4), arg.nParity), arg(arg), out(out), in(in), halo(halo), nDimComms(4), dslashParam(arg)
```
weinbe2, line 334: Should we add a version of `TunableKernel3D` (and others) that can take in `vector_ref`? I can see there being a case for avoiding this one, just because (iirc) the tunable kernel classes take in a `LatticeField&`, so this would open up a whole other can of worms.

include/dslash.h
```diff
@@ -433,7 +436,7 @@ namespace quda
       auto aux_ = (arg.pack_blocks > 0 && (arg.kernel_type == INTERIOR_KERNEL || arg.kernel_type == UBER_KERNEL)) ?
         aux_pack :
         ((arg.shmem > 0 && arg.kernel_type == EXTERIOR_KERNEL_ALL) ? aux_barrier : aux[arg.kernel_type]);
-      return TuneKey(in.VolString().c_str(), typeid(*this).name(), aux_);
+      return TuneKey(in[0].VolString().c_str(), typeid(*this).name(), aux_);
```
weinbe2: Could add `VolString()` to `vector_ref`

include/kernels/dslash_domain_wall_5d.cuh
```diff
@@ -10,16 +10,22 @@ namespace quda
...
+    DomainWall5DArg(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in, const ColorSpinorField &halo,
+                    const GaugeField &U, double a, double m_f, bool xpay, cvector_ref<const ColorSpinorField> &x, int parity,
+                    bool dagger, const int *comm_override) :
+      WilsonArg<Float, nColor, nDim, reconstruct_>(out, in, halo, U, xpay ? a : 0.0, x, parity, dagger, comm_override),
       Ls(in.X(4)),
       a(a),
-      m_f(m_f) {}
+      m_f(m_f)
+    {
+      // FIXME this is a hack to use remove the batch dimension from the dslash constants
```
weinbe2: A reminder to fix this :)

include/kernels/dslash_mobius_eofa.cuh
```diff
@@ -70,6 +68,12 @@ namespace quda
         inv(inv_),
         sherman_morrison(sherman_morrison_)
       {
+        if (out.size() > max_n_src) errorQuda("vector set size %lu greater than max size %d", out.size(), max_n_src);
```
weinbe2: I think you can get rid of this check now that you moved the recursive decomposition into the instantiate routines

include/kernels/laplace.cuh
```diff
@@ -29,27 +32,35 @@ namespace quda
 
     typedef typename mapper<Float>::type real;
 
-    F out;        /** output vector field */
-    const F in;   /** input vector field */
-    const F in_pack; /** input vector field used in packing to be able to independently resetGhost */
-    const F x;    /** input vector field for xpay*/
+    static constexpr unsigned int max_n_src = MAX_MULTI_RHS;
+    const int_fastdiv n_src;
```
weinbe2: I think we can get rid of this, it's not assigned to or used anywhere.

include/kernels/staggered_quark_smearing.cuh
```diff
@@ -29,9 +33,12 @@ namespace quda
 
     using real = typename mapper<Float>::type;
 
-    F out;           /** output vector field */
-    const F in;      /** input vector field */
-    const F in_pack; /** input vector field used in packing to be able to independently resetGhost */
+    static constexpr unsigned int max_n_src = MAX_MULTI_RHS;
+    const int_fastdiv n_src;
```
weinbe2: This doesn't seem to be set/used anywhere

include/color_spinor_field_order.h
maddyscientist: Exactly, this reduces the size of the accessor, allowing us to fit 16 RHS without large args.  I tested the performance with and without this: conclusion was that at both extremes (`n_rhs=1` and `n_rhs = 16`) performance was identical.  This basically swaps some LDCs for a bit of math (which is hoisted out by the compiler and only performed once).  Doesn't change the register count, and the new local `norm` is plopped in a uniform register.

maddyscientist: I was thinking to leave the code in there, in case for some reason it ever needs to come back in the future.

I'll mark this as resolved.

include/kernels/staggered_quark_smearing.cuh
maddyscientist: [9a98546](https://github.com/lattice/quda/pull/1465/commits/9a98546a30b343a906a101d89215ab3c8322775c)

include/kernels/laplace.cuh
maddyscientist: [9a98546](https://github.com/lattice/quda/pull/1465/commits/9a98546a30b343a906a101d89215ab3c8322775c)

include/kernels/dslash_mobius_eofa.cuh
maddyscientist: [9a98546](https://github.com/lattice/quda/pull/1465/commits/9a98546a30b343a906a101d89215ab3c8322775c)

include/kernels/dslash_domain_wall_5d.cuh
maddyscientist: I've clarified the comment there: it's not so much of a hack, and is the right way to do this.

Looking at this made me realize there was a bug with the flops counter for the 5-d DWF kernel, so I've fixed that (23c6ba0e668c1b4ab205eb72ac4f6fa752223c6a)

include/dslash.h
maddyscientist: a805ab08ae34c093a76f3b68efd636bafd4403cc

include/dslash.h
maddyscientist: Yeah, I would skip that for now.

maddyscientist:
> Not the highest priority, but it should be easy to apply the multi-rhs optimizations to the KD operator (`lib/staggered_kd_apply_xinv.cu` and `include/kernels/staggered_kd_apply_xinv_kernel.cuh`). Testing it is another story... I need to get it exposed directly through `staggered_invert_test`, etc. Feel free to punt this back to me for the future, just file an issue about it.

This is an easy ask, so I've just gone and done it (071ccf717b61673c9fec664ddab3e59801a46772).

include/kernels/staggered_quark_smearing.cuh
```diff
@@ -43,18 +49,23 @@ namespace quda
...
-    StaggeredQSmearArg(ColorSpinorField &out, const ColorSpinorField &in, const GaugeField &U, int t0,
-                       bool is_t0_kernel, int parity, int dir, bool dagger, const int *comm_override) :
-      DslashArg<Float, nDim>(out, in, U, in, parity, dagger, false, 3, false, comm_override),
-      out(out, 3),
-      in(in, 3),
-      in_pack(in, 3),
+    StaggeredQSmearArg(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in,
+                       const ColorSpinorField &halo, const GaugeField &U, int t0, bool is_t0_kernel, int parity,
+                       int dir, bool dagger, const int *comm_override) :
+      DslashArg<Float, nDim>(out, in, halo, U, in, parity, dagger, false, improved ? 3 : 1, false, comm_override),
```
weinbe2: I think the 3 in all cases is intentional---staggered smearing requires a shift-by-2 independent of it being naive or improved staggered, so a halo depth of 1 won't cut it.

What this might mean is the staggered quark smearing code may not work as-is for naive staggered, I'm not sure if that's been properly tested? There may be more hard-coded 3s lurking in some exchange routines in the interface that would make it work.

maddyscientist: Yeah, agree with your comments here, good catch.  I added a bit too much copy paste here with the implied non-improved, halo-1 additions.  Will remove those.

weinbe2: I don't think you meant to push it to `build/include/kernels/staggered_quark_smearing.cuh` in https://github.com/lattice/quda/pull/1465/commits/57913f3db473d35ac96c47b6a8dcac23180f1035 :)

maddyscientist: 🤣

lib/dirac_domain_wall_4d.cpp
```diff
@@ -68,10 +69,10 @@ namespace quda {
     blas::xpay(in, -kappa5, out);
   }
 
-  void DiracDomainWall4D::MdagM(ColorSpinorField &out, const ColorSpinorField &in) const
+  void DiracDomainWall4D::MdagM(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in) const
   {
     checkFullSpinor(out, in);
-    auto tmp = getFieldTmp(in);
+    auto tmp = getFieldTmp(out);
```
weinbe2, line 75: What was the reason for switching `in` -> `out` for the `tmp` field? Is it related to the `const`-ness of `in`?

maddyscientist: exactly, otherwise it creates a constant tmp which is less useful...

lib/dirac_domain_wall_4d.cpp
```diff
@@ -178,7 +179,7 @@ namespace quda {
     }
 
     // we desire solution to full system
-    auto tmp = getFieldTmp(b[0].Even());
+    auto tmp = getFieldTmp(x[0].Even());
     for (auto i = 0u; i < b.size(); i++) {
```
weinbe2, line 183: Is this being intentionally left as a loop over each separate 5-d spinor? All good if so, just wanted to make sure this wasn't an accidental oversight.

maddyscientist: This one is an oversight, thanks for the catch.

Aside: I have the `prepare` and `reconstruct` Dirac member functions as loops intentionally, rather than "vectorize" them, since we don't have multi-RHS solvers in place which would be needed for testing that functionality.  Once we do have those, I'll vectorize all those functions.

lib/dirac_mobius.cpp
```diff
@@ -383,7 +387,7 @@ namespace quda {
     }
 
     // we desire solution to full system
-    auto tmp = getFieldTmp(b[0].Even());
+    auto tmp = getFieldTmp(x[0].Even());
```
weinbe2, line 375: Similar question to 4-d preconditioned DWF, should this be multi-rhs-ified?

lib/dirac_domain_wall.cpp
```diff
@@ -33,46 +33,43 @@ namespace quda {
     return *this;
   }
 
-  void DiracDomainWall::checkDWF(const ColorSpinorField &out, const ColorSpinorField &in) const
+  void DiracDomainWall::checkDWF(cvector_ref<const ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in) const
```
weinbe2, line 36: Slightly orthogonal to the PR, but I've noticed over the course of this review that there are several _nearly_ identical `checkDWF` functions across the various flavors of DWF/Mobius. The one semi-outlier is Mobius having a separate check for `zMobius`, but even that seems sufficiently redundant of the non-zMobius path. 

I think we should write one `checkDWF` that can be inserted in the base `Dirac` class or something of the sort and reused across all of the 5-d operators.

lib/dirac_mobius.cpp
```diff
@@ -450,25 +454,27 @@ namespace quda {
...
-                                             MdwfFusedDslashType::D4_D5INV_D5INVDAG);
-
-      mobius_tensor_core::apply_fused_dslash(extended_tmp2, extended_tmp1, *extended_gauge, extended_tmp2,
-                                             extended_tmp1, mass, m5, b_5, c_5, dagger, other_parity, shift1.data,
-                                             shift1.data, MdwfFusedDslashType::D4DAG_D5PREDAG_D5INVDAG);
-
-      mobius_tensor_core::apply_fused_dslash(out, extended_tmp2, *extended_gauge, out, unextended_tmp1, mass, m5, b_5,
-                                             c_5, dagger, this_parity, shift2.data, shift2.data,
-                                             MdwfFusedDslashType::D4DAG_D5PREDAG);
+      for (auto i = 0u; i < in.size(); i++) {
```
weinbe2, line 442: Similar comment about being multi-rhs-ified, but I'm fully aware the tensor core routines are a separate beast that need not be touched in this PR.

maddyscientist: Yep, this one is intentional 😄

lib/dirac_mobius.cpp
```diff
@@ -536,14 +542,16 @@ namespace quda {
...
+    for (auto i = 0u; i < in.size(); i++) {
+      if (in[i].X(4) != Ls) errorQuda("Expected Ls = %d, not %d", Ls, in[i].X(4));
+      if (out[i].X(4) != Ls) errorQuda("Expected Ls = %d, not %d", Ls, out[i].X(4));
+    }
   }
 
-  void DiracMobiusEofa::m5_eofa(ColorSpinorField &out, const ColorSpinorField &in) const
+  void DiracMobiusEofa::m5_eofa(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in) const
   {
     if (in.Ndim() != 5 || out.Ndim() != 5) errorQuda("Wrong number of dimensions");
```
weinbe2, line 532: Again, semi-orthogonal, but this could/should be replaced with a `checkDWF` call, which `DiracMobiusEofa` doesn't have as-is but could be pulled from a base class.

lib/dirac_twisted_clover.cpp
```diff
@@ -175,35 +179,41 @@ namespace quda {
...
-  void DiracTwistedCloverPC::WilsonDslash(ColorSpinorField &out, const ColorSpinorField &in, QudaParity parity) const
+  void DiracTwistedCloverPC::WilsonDslash(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in,
+                                          QudaParity parity) const
   {
     if (in.TwistFlavor() == QUDA_TWIST_SINGLET) {
       DiracWilson::DslashXpay(out, in, parity, in, 0.0);
     } else {
       // we need an Ls=plain 2 Wilson dslash, which is exactly what the 4-d preconditioned DWF operator is
-      ApplyDomainWall4D(out, in, *gauge, 0.0, 0.0, nullptr, nullptr, in, parity, dagger, commDim.data, profile);
+      for (auto i = 0u; i < in.size(); i++)
```
weinbe2: `ApplyDomainWall4D` has been multi-rhs-ified, so we can get rid of the `for` loop here and in the other function below.

lib/dirac_twisted_clover.cpp
```diff
@@ -296,7 +306,7 @@ namespace quda {
     }
 
     // we desire solution to full system
-    auto tmp = getFieldTmp(b[0].Even());
+    auto tmp = getFieldTmp(x[0].Even());
```
weinbe2, line 306: Can the prepare/construct be multi-rhs-ified, or does the `create_alias` logic cause issues I haven't thought of?

maddyscientist: See here: https://github.com/lattice/quda/pull/1465#discussion_r1612742057

lib/dirac_domain_wall_4d.cpp
maddyscientist: GitHub formatting confused me for a minute there: confirming this one was intentional for the reason stated above: I'll update `prepare` / `reconstruct` in a subsequent PR.

lib/dirac_mobius.cpp
maddyscientist: And like dwf-4d: this will be in a followup PR...

include/kernels/staggered_quark_smearing.cuh
maddyscientist: 7f4694c579199a2c247f1269283b7fcc4ae5f68b

lib/dirac_twisted_clover.cpp
maddyscientist: 142ffbe7711caefa9f978bc2405adf3353adb265

kostrzewa (commented):
@maddyscientist I can't pinpoint exactly yet what is happening but it seems that this is causing either some tuning issues or even some kind of lockup. I've tested this latest commit 142ff on a quad-A100 node doing 2+1+1 twisted clover HMC where it takes a very long time to tune (not sure which kernel, would need to run with higher debug levels). I'm also testing an an octo-A40 node (also 2+1+1 HMC) where it tunes for a while and then appears to get stuck, the last output being:

```
WARNING: Unexpected regression when tuning candidates for N4quda31NdegTwistedCloverPreconditionedINS_34NdegTwistedCloverPreconditionedArgIdLi3ELi4EL21QudaReconstructType_s18EEEEE: (0.00920678 > 1.1 * 0.00643123)
```

The GPUs on the node are currently fully idling:

```
Fri May 24 12:54:12 2024
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.104.05             Driver Version: 535.104.05   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A40                     On  | 00000000:01:00.0 Off |                    0 |
|  0%   53C    P0              83W / 300W |   1535MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A40                     On  | 00000000:25:00.0 Off |                    0 |
|  0%   52C    P0              84W / 300W |   1535MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   2  NVIDIA A40                     On  | 00000000:41:00.0 Off |                    0 |
|  0%   51C    P0              86W / 300W |   1535MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   3  NVIDIA A40                     On  | 00000000:61:00.0 Off |                    0 |
|  0%   47C    P0              79W / 300W |   1535MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   4  NVIDIA A40                     On  | 00000000:81:00.0 Off |                    0 |
|  0%   49C    P0              84W / 300W |   1535MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   5  NVIDIA A40                     On  | 00000000:A1:00.0 Off |                    0 |
|  0%   47C    P0              83W / 300W |   1535MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   6  NVIDIA A40                     On  | 00000000:C1:00.0 Off |                    0 |
|  0%   49C    P0              81W / 300W |   1534MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   7  NVIDIA A40                     On  | 00000000:E1:00.0 Off |                    0 |
|  0%   43C    P0              77W / 300W |   1534MiB / 46068MiB |      0%	Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage	|
|=======================================================================================|
|    0   N/A  N/A    404983	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
|    1   N/A  N/A    404984	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
|    2   N/A  N/A    404985	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
|    3   N/A  N/A    404986	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
|    4   N/A  N/A    404987	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
|    5   N/A  N/A    404988	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
|    6   N/A  N/A    404989	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
|    7   N/A  N/A    404990	 C   ...ster-950a3a1-quda-mrhs-142ff/hmc_tm     1522MiB |
+---------------------------------------------------------------------------------------+
```

kostrzewa:
> @maddyscientist I can't pinpoint exactly yet what is happening but it seems that this is causing either some tuning issues or even some kind of lockup. I've tested this latest commit 142ff on a quad-A100 node doing 2+1+1 twisted clover HMC where it takes a very long time to tune (not sure which kernel, would need to run with higher debug levels). I'm also testing an an octo-A40 node (also 2+1+1 HMC) where it tunes for a while and then appears to get stuck, the last output being.

On the A100 I believe the slow tuning was because the device memory was almost full as the job was killed by cudaMalloc shortly thereafter. This also means that with feature/mrhs in place, our HMC runs will require more device memory (and thus more nodes in some cases). Are any mechanisms in place to get the "original" behaviour back (in order to save device memory) for those cases where this would be more efficient or enable one to run at all (in cases where the number of nodes on a system is limited, for example)?

maddyscientist:
> > @maddyscientist I can't pinpoint exactly yet what is happening but it seems that this is causing either some tuning issues or even some kind of lockup. I've tested this latest commit 142ff on a quad-A100 node doing 2+1+1 twisted clover HMC where it takes a very long time to tune (not sure which kernel, would need to run with higher debug levels). I'm also testing an an octo-A40 node (also 2+1+1 HMC) where it tunes for a while and then appears to get stuck, the last output being.
> 
> 
> 
> On the A100 I believe the slow tuning was because the device memory was almost full as the job was killed by cudaMalloc shortly thereafter. This also means that with feature/mrhs in place, our HMC runs will require more device memory (and thus more nodes in some cases). Are any mechanisms in place to get the "original" behaviour back (in order to save device memory) for those cases where this would be more efficient or enable one to run at all (in cases where the number of nodes on a system is limited, for example)?

There is a cmake parameter that sets the maximum rhs allowable, and beyond that the code will split and recurse until the maximum value is met.  I suppose if we set that to 1 the original behavior would be restored.

- [ ] How about I add an env which at runtime can be used to set the maximum value before this recursion takes place to allow for more fine-grained control?

Also highlighting the env `QUDA_ENABLE_DEVICE_MEMORY_POOL`, if you disable the pool setting it 0, this should noticeably the reduce memory footprint used by QUDA, which may help(it does come at the cost of exposing malloc/free overheads which are likely a small single digit overhead.

kostrzewa:
> Also highlighting the env QUDA_ENABLE_DEVICE_MEMORY_POOL, if you disable the pool setting it 0, this should noticeably the reduce memory footprint used by QUDA, which may help(it does come at the cost of exposing malloc/free overheads which are likely a small single digit overhead.

In tmLQCD we even set `QUDA_ENABLE_DEVICE_MEMORY_POOL = 0` by default using `setenv` for exactly this reason with the option for the user to enable it via our input file.

https://github.com/etmc/tmLQCD/blob/950a3a161c7a80e4ffd45ef4b773ea33adf1f4be/quda_interface.c#L403

> There is a cmake parameter that sets the maximum rhs allowable, and beyond that the code will split and recurse until the maximum value is met. I suppose if we set that to 1 the original behavior would be restored.

Oh I hadn't seen that. I think that's already good enough for our case. Having an additional env variable for runtime control would of course be nice.

maddyscientist:
The split/recurse pattern will only work to limit the autotuning overhead and not the memory used.  But I shall work on adding that, so we can control the footprint of the clover force.

lib/dslash_domain_wall_4d_fused_m5.hpp
```diff
@@ -111,19 +103,17 @@ namespace quda
...
 
-      return flops_ + Dslash::flops();
+      return in.size() * flops_ + Dslash::flops();
     }
 
     long long bytes() const override
     {
       if (Arg::dslash5_type == Dslash5Type::M5_INV_MOBIUS_M5_INV_DAG) {
-        return arg.y.Bytes() + Dslash::bytes();
+        return in.size() * arg.y[0].Bytes() + Dslash::bytes();
```
weinbe2: Remove the `[0]`

maddyscientist: [2a5acd0](https://github.com/lattice/quda/pull/1465/commits/2a5acd0c72bb6aa8de0932f4cf4c7abae8661791)

lib/dirac_domain_wall.cpp
maddyscientist: [053bfaa](https://github.com/lattice/quda/pull/1465/commits/053bfaa286655ac83c057583e3c70ee3688c9f82)

lib/dirac_mobius.cpp
maddyscientist: [053bfaa](https://github.com/lattice/quda/pull/1465/commits/053bfaa286655ac83c057583e3c70ee3688c9f82)

maddyscientist:
> Somewhat orthogonal comment to this PR, but while we're here can I request that we rename `covDev.cu[h]` to `covariant_derivative.cu[h]` or something of the sort? It's the only outlier file in terms of the naming convention, and iirc `git` should note that it's just a "rename" so we won't lose the commit history.

Done with [c475655](https://github.com/lattice/quda/pull/1465/commits/c47565502250a0e1b4715888a761d59ac87e93e6)

maddyscientist:
> > Also highlighting the env QUDA_ENABLE_DEVICE_MEMORY_POOL, if you disable the pool setting it 0, this should noticeably the reduce memory footprint used by QUDA, which may help(it does come at the cost of exposing malloc/free overheads which are likely a small single digit overhead.
> 
> In tmLQCD we even set `QUDA_ENABLE_DEVICE_MEMORY_POOL = 0` by default using `setenv` for exactly this reason with the option for the user to enable it via our input file.
> 
> https://github.com/etmc/tmLQCD/blob/950a3a161c7a80e4ffd45ef4b773ea33adf1f4be/quda_interface.c#L403
> 
> > There is a cmake parameter that sets the maximum rhs allowable, and beyond that the code will split and recurse until the maximum value is met. I suppose if we set that to 1 the original behavior would be restored.
> 
> Oh I hadn't seen that. I think that's already good enough for our case. Having an additional env variable for runtime control would of course be nice.

@kostrzewa I have added support for the env `QUDA_MAX_MULTI_RHS` that can be used to cap at run-time them maximum number of rhs per kernel (https://github.com/lattice/quda/pull/1465/commits/04fd7ea9c3314e0d48b8d5c252fd9b9f0bff3e98), and all multi-RHS kernels will split and recurse respectively to respect this limit.  This can be used to control the autotuning time if that is ever an issue.  It is the multi-RHS dslash kernels that take a long time to tune, specifically the main time taken comes from specific dslash policies, which are only really relevant on systems with fast host-device interconnect.  The best solution here is likely to disable these policies by default which fixes this headache.  I'll do that shortly...

With respect to memory footprint, I've tweaked the clover force code a bit to reduce the memory footprint ([552957e](https://github.com/lattice/quda/pull/1465/commits/552957eaec9dbcbf4ce6a276fc93169b39a6349f)).  This reduces the footprint of the clover force by about 15% on the tests I was doing locally.  There's more scope for reduction if this is needed.

include/instantiate.h
```diff
@@ -286,6 +287,55 @@ namespace quda
...
+     @brief instantiate_recurse function is used to instantiate the
+     precision and number of colors for a class that operates on
+     batches.  If necessary the batches are split up if the set size
+     exceeds the maximum.  This specific variant is for when we have
+     two vector sets.
+     @param[out] out Output ColorSpinorField set
+     @param[in] in Input ColorSpinorField set
+     @param[in,out] args Any additional arguments required for the computation at hand
+  */
+  template <template <typename, int> class Apply, typename O, typename I, typename... Args>
```
weinbe2, line 300: I recommend changing the function name to `instantiate_recurse2` to be consistent with `instantiate_recurse3` below.

include/instantiate_dslash.h
```diff
@@ -68,34 +67,30 @@ namespace quda
...
-#else
-      errorQuda("QUDA_PRECISION=%d does not enable half precision", QUDA_PRECISION);
-#endif
-    } else if (U.Precision() == QUDA_QUARTER_PRECISION) {
-#if QUDA_PRECISION & 1
-      instantiate<Apply, Recon, int8_t>(out, in, U, args...);
-#else
-      errorQuda("QUDA_PRECISION=%d does not enable quarter precision", QUDA_PRECISION);
-#endif
+    if (in.size() > get_max_multi_rhs()) {
```
weinbe2, line 73: I think we can get rid of this part of `instantiate` because we have the recursive instantiation functions now. We obviously still need the precision checks below it.

lib/dslash_coarse.in.cu
```diff
@@ -14,6 +14,16 @@ namespace quda {
                                         cvector_ref<const ColorSpinorField> &inB, const GaugeField &Y, const GaugeField &X,
                                         double kappa, int parity, bool dslash, bool clover, const int *commDim, QudaPrecision halo_precision)
   {
+    if (inA.size() > get_max_multi_rhs()) {
```
weinbe2, line 17: We can utilize `instantiate_recurse3` here, I believe, with the overhead of needing an instantiate function? I understand if it becomes too complicated because it's a coarse operator, not fine operator, being instantiated here.

lib/dslash_domain_wall_4d_fused_m5.hpp
```diff
@@ -132,23 +123,44 @@ namespace quda
...
-                                                     in.getDslashConstant().ghostFaceCB, profile);
+      Arg arg(out, in, halo, U, a, m_5, b_5, c_5, a != 0.0, x, y, parity, dagger, comm_override, m_f);
+      DomainWall4DFusedM5<Arg> dwf(arg, out, in, halo, y);
+      dslash::DslashPolicyTune<decltype(dwf)> policy(dwf, in, halo, profile);
 #endif
     }
   };
 
+  // use custom instantiate to deal with field splitting if needed
+  template <template <typename, int, QudaReconstructType> class Apply, typename Recon = WilsonReconstruct, typename... Args>
```
weinbe2: You could create an `instantiate_recurse4` function for this one---I know we can't chase recursion counts forever, but I think it's worth it since this is still a dslash, and it's best to have a single file where all of the dslash recursive instantiations live.

maddyscientist: The issue with doing that is that with it being a dslash, it uses the instantiate functions in `instantiate_dslash.h` not `instantiate.h`.

lib/dslash_coarse.in.cu
maddyscientist: We can't, because the instantiation for the coarse dslash is quite different than that the general instantiation: coarse colors, no double precision, specialization for halo precision etc.

include/instantiate_dslash.h
maddyscientist: We still need this here, as the `instantiate_recurse` variants in `instantiate.h` are not called in the dslash instantiations.

Aside: eventually, I'd like to remove the need for the separate `instantiate.h` and `instantiate_dslash.h` headers, but that's beyond the scope of this PR.

include/instantiate.h
maddyscientist: c1f060fc2

lib/dslash_domain_wall_4d_fused_m5.hpp
maddyscientist: Noting this can be done however in the future when I delete the separate `instantiate_dslash.h` header.

maddyscientist:
@kostrzewa @weinbe2 I have now disabled all zero-copy policies by default, this dramatically accelerates the autotuning time, and avoid us having to pointlessly run dslash variants which are unnecessary for most systems.

I believe this PR is now functionally complete.

kostrzewa:
@maddyscientist @Marcogarofalo I'm afraid that between fc92ee628 and a9714ba25  something was done which appears to trigger something which looks like an integrator instability from "way back when" in our HMC.

Log files from Juwels Booster below:

[tmLQCD_nf211_hmc_64c128_n4_fc92ee628.zip](https://github.com/user-attachments/files/15511931/tmLQCD_nf211_hmc_64c128_n4_fc92ee628.zip)

[tmLQCD_nf211_hmc_64c128_n4_a9714ba25.zip](https://github.com/user-attachments/files/15511930/tmLQCD_nf211_hmc_64c128_n4_a9714ba25.zip) 

showing that in the second log CG from a determinant monomial diverges.

Not quite sure how to catch this kind of thing with QUDA tests but I'm guessing the issue originates in the mrhs improvement of the fermion force for the partial fractions (perhaps the outer product?).

maddyscientist:
@kostrzewa could you run with tuning disabled?  One of the drawbacks of our current testing is the run to run variance possible with different tuning configs.  If you run with disabled and the issue reproduces this would be helpful information for debugging.  Thx

include/color_spinor_field_order.h
hummingtree: Does it make sense to add a new cmake option for these?

include/dslash_helper.cuh
```diff
@@ -265,11 +265,11 @@ namespace quda
     bool remote_write;      // used by the autotuner to switch on/off remote writing vs using copy engines
 
     int_fastdiv threads; // number of threads in x-thread dimension
-    int_fastdiv exterior_threads; //  number of threads in x-thread dimension for fused exterior dslash
+    int_fastdiv exterior_threads; // number of threads in x-thread dimension for fused exterior dslash
     int threadDimMapLower[4];
     int threadDimMapUpper[4];
 
-    const bool spin_project; // whether to spin project nSpin=4 fields (generally true, except for, e.g., covariant derivative)
+    int_fastdiv Ls;
```
hummingtree, line 272: Just a comment: overloading the name `Ls` might add another item the long list of terminology overloading in lattice QCD.

kostrzewa:
> @kostrzewa could you run with tuning disabled? One of the drawbacks of our current testing is the run to run variance possible with different tuning configs. If you run with disabled and the issue reproduces this would be helpful information for debugging. Thx

I think I have to take back what I said in https://github.com/lattice/quda/pull/1465#issuecomment-2141701489. While I did observe solvers diverging twice in a row with https://github.com/lattice/quda/commit/a9714ba25087e77c08c08ea28cabe531b249bc31, I cannot reproduce this any more. I think it must have been a transient machine fluke. I'll run a few more test runs but for now I can say that the mrhs branch works correctly within tmLQCD's HMC and shaves off another 10-15% of the time per trajectory in a simulation of a 64c128 ensemble on 4 Juwels Booster nodes.

kostrzewa:
I'll have to retest on the A40-based machine.

maddyscientist:
> I think I have to take back what I said in [#1465 (comment)](https://github.com/lattice/quda/pull/1465#issuecomment-2141701489). While I did observe solvers diverging twice in a row with [a9714ba](https://github.com/lattice/quda/commit/a9714ba25087e77c08c08ea28cabe531b249bc31), I cannot reproduce this any more. I think it must have been a transient machine fluke. I'll run a few more test runs but for now I can say that the mrhs branch works correctly within tmLQCD's HMC and shaves off another 10-15% of the time per trajectory in a simulation of a 64c128 ensemble on 4 Juwels Booster nodes.

Thanks for the follow up @kostrzewa and restoring my sanity. 😅  I have been doing exhaustive testing since your prior report, and couldn't find anything bad.  Great to hear of the performance improvement.  Things will only get better from here when we add MRHS to the MG setup, which will hopefully be not too far away now....

include/color_spinor_field_order.h
maddyscientist: As discuss on slack, I've put the legacy code path behind a macro that can be enabled if ever need be.  [2bcff04](https://github.com/lattice/quda/pull/1465/commits/2bcff0487888b8b296b598a55c95048c9b441855)

include/dslash_helper.cuh
maddyscientist: 🤷‍♀️

include/kernels/dslash_clover_helper.cuh
```diff
@@ -214,12 +220,12 @@ namespace quda {
...
+        if (chirality == 0) cache.save_y(x[1], target::thread_idx().y);
+        else                cache.save_y(x[0], target::thread_idx().y);
         cache.sync();
-        if (chirality == 0) x[1] = cache.load_y(1 - dir);
-        else                x[0] = cache.load_y(dir);
+        if (chirality == 0) x[1] = cache.load_y(target::thread_idx().y + 1);
+        else                x[0] = cache.load_y(target::thread_idx().y - 1);
       };
 
       swizzle(in_chi, chirality, FORWARDS); // apply the flavor-chirality swizzle between threads
```
bjoo: Random question: `swizzle_direction` is now an unnamed argument and is not used. Why not drop the direction from the arg list ? Forward/backwards seem to depend now only on chirality and the thread_idx.y

weinbe2 (approved):
This gets my conditional approval. It's passed a visual review and a correctness review. @maddyscientist is investigating some performance regressions, so no one else should hit merge until those have been addressed (please)!

bjoo (approved):

maddyscientist:
> This gets my conditional approval. It's passed a visual review and a correctness review. @maddyscientist is investigating some performance regressions, so no one else should hit merge until those have been addressed (please)!

The performance regression was caused by an excess of calls to `qudaDeviceSynchronize` being made.  This is addressed in d01194619344a5934877ef59dace76176ace2e10 (and to a lesser degree 07f142c70d31236f7e0b1537e052e4e1423c85a5).

include/kernels/dslash_clover_helper.cuh
maddyscientist: [768af11](https://github.com/lattice/quda/pull/1465/commits/768af1141de96e4bfe2c2d09ecfc8926907a775e)

PR #1468, opened 2024-05-30
Split Grid modifications to work with Chroma

bjoo:
# Outline
This PR makes split-grid work with additions to Chroma including with QDP-JIT. 
The main list of modifications to the existing split grid setup are:
* Mechanism for saving/restoring gauge fields and clover fields before and after the split
   +  NB: this takes gauge and clover fields out of the interface to the `invertMultiSrc....()` functions. The existing loaded gauge fields are used (in internal native format)
* ColorSpinorFields are 'downloaded' and converted to native format prior to splitting. This adds an extra 'download' since the underlying call to `invertQuda` will also then 're-download' (now from device-to-device) and also this uses some extra memory. This could be avoided by refactoring `invertQuda` to take existing `ColorSpinorField`-s and provide an overloaded version as a wrapper which takes the host pointers to satisfy the C-intarface. The wrapper could do the downloads and call the newly refactored invertQuda. That is a potentially more involved refactor -- because of things like resident solutions etc -- so save for a later (next?) PR.
* A PR to exploit this in Chroma is ready, but I don't want to raise it until this PR is merged, since it needs this PR to work.


include/clover_backup.h
```diff
@@ -0,0 +1,144 @@
...
+
+  struct CloverBundleBackup {
+    CloverField *precise;
+    CloverField *sloppy;
+    CloverField *precondition;
+    CloverField *refinement;
+    CloverField *eigensolver;
+
+    CloverBundleBackup() :
+      precise(nullptr), sloppy(nullptr), precondition(nullptr), refinement(nullptr), eigensolver(nullptr)
```
maddyscientist: Could just set these definitions inline and then use `CloverBundleBackup() = default`.

include/gauge_backup.h
```diff
@@ -0,0 +1,157 @@
...
+  struct GaugeBundleBackup {
+    GaugeField *precise;
+    GaugeField *sloppy;
+    GaugeField *precondition;
+    GaugeField *refinement;
+    GaugeField *eigensolver;
+    GaugeField *extended;
+
+    GaugeBundleBackup() :
+      precise(nullptr), sloppy(nullptr), precondition(nullptr), refinement(nullptr), eigensolver(nullptr), extended(nullptr)
```
maddyscientist: As noted for above, set these values inline above and use the compiler-generated default constructor

include/quda.h
```diff
@@ -1214,37 +1214,26 @@ extern "C" {
    * @param _hp_x       Array of solution spinor fields
    * @param _hp_b       Array of source spinor fields
    * @param param       Contains all metadata regarding host and device storage and solver parameters
-   * @param h_gauge     Base pointer to host gauge field (regardless of dimensionality)
-   * @param gauge_param Contains all metadata regarding host and device storage for gauge field
    */
-  void invertMultiSrcQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param, void *h_gauge, QudaGaugeParam *gauge_param);
+  void invertMultiSrcQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param);
```
maddyscientist, line 1218: @hummingtree I guess this will necessitate changes to CPS: I assume CPS has split-grid plumbed in with QUDA?

include/quda.h
```diff
@@ -1214,37 +1214,26 @@ extern "C" {
...
    * @param _hp_x       Array of solution spinor fields
    * @param _hp_b       Array of source spinor fields
    * @param param       Contains all metadata regarding host and device storage and solver parameters
-   * @param milc_fatlinks     Base pointer to host **fat** gauge field (regardless of dimensionality)
-   * @param milc_longlinks    Base pointer to host **long** gauge field (regardless of dimensionality)
-   * @param gauge_param Contains all metadata regarding host and device storage for gauge field
    */
-  void invertMultiSrcStaggeredQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param, void *milc_fatlinks,
-                                   void *milc_longlinks, QudaGaugeParam *gauge_param);
+  void invertMultiSrcStaggeredQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param);
```
maddyscientist: Do we even need separate interface nows since we can use the resident gauge / fat / long / clover fields?  Can't we delete all these?

include/color_spinor_field_order.h
```diff
@@ -476,12 +476,12 @@ namespace quda
       complex<storeFloat> *v;        /**< Field memory address this wrapper encompasses */
       const int idx;                 /**< Index into field */
     private:
-      const Float scale;             /**< Float to fixed-point scale factor */
-      const Float scale_inv;         /**< Fixed-point to float scale factor */
+      const Float scale;     /**< Float to fixed-point scale factor */
+      const Float scale_inv; /**< Fixed-point to float scale factor */
```
maddyscientist: There seem to be formatting changes to lines here that haven't been edited.  This can cause unnecessary conflicts.

lib/copy_color_spinor_offset.cu
```diff
@@ -22,27 +22,63 @@ namespace quda
     }
   }
 
-  template <class Float, int nColor, int nSpin>
-  void copy_color_spinor_offset(ColorSpinorField &out, const ColorSpinorField &in, CommKey offset, QudaPCType pc_type)
+  template <class Float, int nColor>
```
maddyscientist: I'm not sure I follow the changes in this file.  Why was it necessary to make separate spin-dependent functions here, as opposed to the previous single templated function?

lib/copy_color_spinor_offset.cu
```diff
@@ -22,27 +22,63 @@ namespace quda
     }
```
maddyscientist: It looks like you've clang formatted this entire file as opposed to the diff.  Can you revert this?  Else we end up with a lot of conflicts with other ongoing PRs.

maddyscientist (commented):
Looks like a good step forward for the split grid, thanks @bjoo.  I've left a few comments to be addressed, but the main ones are:
* Can we just delete the staggered and clover specific solver interfaces now?
* clang format to all of interface_quda.cpp will lead to some pain, can you undo that and apply to the diff only?

include/quda.h
bjoo: Possibly...

lib/copy_color_spinor_offset.cu
bjoo: So this may be just me not knowing the code. But originally it supported `QUDA_SPACE_COLOR_SPIN_FIELD_ORDER` and `QUDA_SPACE_SPIN_COLOR_FIELD_ORDER` and I didn't know if it supported specifically the internal orderings of 'QUDA_FLOAT_FIELD_ORDER` or `QUDA_FLOAT_2_FIELD_ORDER` etc. So I wanted to add those. But Staggered only supports spin 1 which is a subset of these. So I decided to make an Ns4 and an Ns1 specialization. 

But then I did more faffing so it may be that I can revert this and it will still work... I will check

lib/copy_color_spinor_offset.cu
bjoo: Done.

include/clover_backup.h
bjoo: Done in latest commit

include/gauge_backup.h
bjoo: Done in latest commit

include/quda.h
bjoo: I have gone and removed the `invertMultiSrcStaggeredQuda` and `invertMultiSrcCloverQuda` as well as `dslashMultiSrcStaggeredQuda` and`dslashMultiSrcCloverQuda`

include/color_spinor_field_order.h
bjoo: I have hopefully undone that now.

lib/copy_color_spinor_offset.cu
bjoo: The original file worked. So I reverted to that. I guess this was when I was messing around with mappers and I did something odd.

bjoo:
Hi @maddyscientist  -- Hopefully I managed to resolve all your comments, I left open the conversation about CPS changing. I also noted that cscs builds failed their Ctest, so that is something I will dig into -- I have been testing solely through Chroma for these changes so far.

lib/interface_quda.cpp
```diff
@@ -3099,85 +3094,85 @@ void callMultiSrcQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param, // col
...
-      quda::split_field(collected_gauge, {in}, split_key);
+      collected_gauge = new quda::GaugeField(gf_param);
+      quda::split_field(*collected_gauge, {*(thin_links_bkup.precise)}, split_key);
     } else {
       std::vector<quda::GaugeField *> v_g(1);
 
       milc_fatlink_param.create = QUDA_NULL_FIELD_CREATE;
-      collected_milc_fatlink_field = GaugeField(milc_fatlink_param);
-      quda::split_field(collected_milc_fatlink_field, {milc_fatlink_field}, split_key);
+      collected_milc_fatlink_field = new GaugeField(milc_fatlink_param);
```
hummingtree, line 3111:
```suggestion
      GaugeField collected_milc_fatlink_field(milc_fatlink_param);
```

lib/interface_quda.cpp
```diff
@@ -3099,85 +3094,85 @@ void callMultiSrcQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param, // col
...
-    quda::GaugeField collected_milc_longlink_field;
+    quda::GaugeField *collected_gauge = nullptr;
+    quda::GaugeField *collected_milc_fatlink_field = nullptr;
+    quda::GaugeField *collected_milc_longlink_field = nullptr;
 
     if (!is_staggered) {
       gf_param.create = QUDA_NULL_FIELD_CREATE;
-      collected_gauge = quda::GaugeField(gf_param);
-      quda::split_field(collected_gauge, {in}, split_key);
+      collected_gauge = new quda::GaugeField(gf_param);
```
hummingtree, line 3105: I think we don't need to use pointers to the gauge fields anymore - instead we can just use the constructor themselves.

include/quda.h
hummingtree: It has, but I don't think it is actually used.

lib/interface_quda.cpp
hummingtree: I remember we have some policy in QUDA that we should not use raw pointers anymore so might make sense to avoid all the raw pointers in the changes here - should be straight forward to do so I believe.

bjoo: We discussed this during our debug sesh. The reason here was that I use some pointer-swapping tricks to back-up the original unsplit fields, to set up the split fields and to then restore the originals. It seemed to make sense to use pointers here. However, I can try and think more carefully and not use those. I could in principle drop the pointers I have into a `std::unique_ptr` but the way we hold `gaugePrecise` and `cloverPrecise` etc. is also not done with smart pointers either... So I think in our conversation we ended up with saying pointer use is OK here.

hummingtree (approved):
Nice changes @bjoo ! Approved given a clang-format is applied on the diffs.

PR #1470, opened 2024-06-06
add beta CUDA actions build based Ubuntu 24.04

mathiaswagner:
This add builds based on the Github Ubuntu 24.04 runner  which comes with gcc-13 and llvm-17.
It uses CUDA 12.5 which does not officially support Ubuntu 24.04 yet but gcc 13 and llvm 17 are supported host compilers. 
GPU_ARCH for this build is sm_90.


.github/workflows/cuda_githubactions_build_beta.yml
```diff
@@ -0,0 +1,66 @@
...
+      steps:
+      - name: Install software
+        run: |
+          wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/cuda-keyring_1.0-1_all.deb
+          sudo dpkg -i cuda-keyring_1.0-1_all.deb
+          sudo apt-get update -y
+          # Note CUDA 12.5 does not officially support Ubnutu 24.04
+          sudo apt-get install -y --no-install-recommends  ninja-build cmake libopenmpi-dev gfortran cuda-compiler-12-5 cuda-libraries-dev-12-5 cuda-nvml-dev-12-5
+      
+      # currently deactivated as it cuased build errors - maybe put in again later
```
maddyscientist, line 31: What are the build errors that it introduces?

.github/workflows/cuda_githubactions_build_beta.yml
```diff
@@ -0,0 +1,66 @@
...
+          max-size: 2000M
+
+      - name: Configure CMake
+        run: >
+          cmake 
+          -DCMAKE_CUDA_COMPILER=/usr/local/cuda-12.5/bin/nvcc
+          -DCMAKE_CXX_COMPILER=${{matrix.compiler}}
+          -DCMAKE_CUDA_COMPILER_LAUNCHER=ccache
+          -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+          -DQUDA_GPU_ARCH=sm_90 -DQUDA_GPU_ARCH_SUFFIX=virtual -DQUDA_JITIFY=ON
```
maddyscientist: Wondering what the overhead is if we switch off Jitify here?  Thinking that the latest CUDA version used in CI should probably do real builds.....

maddyscientist (approved):
LGTM.  A couple of questions, but nothing blocking.

.github/workflows/cuda_githubactions_build_beta.yml
maddyscientist: This looks to be working.  Builds in ~60-70 mins but doesn't seem to be timing out.  I'll merge this when the CI completes.

.github/workflows/cuda_githubactions_build_beta.yml
maddyscientist: (discussed offline)

PR #1416, opened 2023-11-10
staggered contraction code

alexstrel:
This PR primarily adds an extension to the QUDA contraction code for staggered fermions. To utilize this, a new parameter, `QUDA_CONTRACT_TYPE_STAGGERED_FT_T`, has been introduced. 
The main application for testing this is contractFT_test.cpp.

PR #1479, opened 2024-07-09
Feature/power

maddyscientist:
This PR is the initial step towards more power awareness in QUDA, as well as adding OMP threading for host kernels
* Adds power, temperature and clock monitoring
  * Monitoring is enabled with `QUDA_ENABLE_MONITOR=1` (default is off)
  * Monitoring is performed on a spawned thread, maintaining the history in a linked list
  * Default monitoring period is `QUDA_ENABLE_MONITOR_PERIOD=1`
  * Monitor info, together with derived energy usage is dumped to a `monitor_*****.tsv` file, where the **** encodes the rank id, and the date_time of the dump.  All ranks have identical times by construction.   
* Add OpenMP threading support for all CPU kernels
  * Most kernels seem to get reasonable scaling
  * `QUDA_OPENMP` CMake parameter is no longer marked as advanced
* Fixes compiler warning introduced in #1416
* Fixes bug introduced in #1416
* Fixed an issue with `endQuda` if memory leaks were detected when running multi-GPU: `printfQuda` would fail since `comm_rank()` would be called after the comms have been torn down

### DIFF
--- a/include/targets/generic/block_reduction_kernel_host.h
+++ b/include/targets/generic/block_reduction_kernel_host.h
@@ -5,6 +5,7 @@ namespace quda
   {
     Functor<Arg> t(arg);
     dim3 block(0, 0, 0);
+#pragma omp parallel for
     for (block.y = 0; block.y < arg.grid_dim.y; block.y++) {
       for (block.x = 0; block.x < arg.grid_dim.x; block.x++) { t(block, dim3(0, 0, 0)); }
     }
--- a/include/targets/generic/kernel_host.h
+++ b/include/targets/generic/kernel_host.h
@@ -6,12 +6,14 @@ namespace quda
   template <template <typename> class Functor, typename Arg> void Kernel1D_host(const Arg &arg)
   {
     Functor<Arg> f(const_cast<Arg &>(arg));
+#pragma omp parallel for
     for (int i = 0; i < static_cast<int>(arg.threads.x); i++) { f(i); }
   }
 
   template <template <typename> class Functor, typename Arg> void Kernel2D_host(const Arg &arg)
   {
     Functor<Arg> f(const_cast<Arg &>(arg));
+#pragma omp parallel for
     for (int i = 0; i < static_cast<int>(arg.threads.x); i++) {
       for (int j = 0; j < static_cast<int>(arg.threads.y); j++) { f(i, j); }
     }
@@ -20,6 +22,7 @@ namespace quda
   template <template <typename> class Functor, typename Arg> void Kernel3D_host(const Arg &arg)
   {
     Functor<Arg> f(const_cast<Arg &>(arg));
+#pragma omp parallel for
     for (int i = 0; i < static_cast<int>(arg.threads.x); i++) {
       for (int j = 0; j < static_cast<int>(arg.threads.y); j++) {
         for (int k = 0; k < static_cast<int>(arg.threads.z); k++) { f(i, j, k); }
--- a/include/targets/generic/reduction_kernel_host.h
+++ b/include/targets/generic/reduction_kernel_host.h
@@ -11,7 +11,7 @@ namespace quda
     Functor<Arg> t(arg);
 
     reduce_t value = t.init();
-
+#pragma omp parallel for collapse(2) reduction(Functor <Arg>::apply : value)
     for (int j = 0; j < static_cast<int>(arg.threads.y); j++) {
       for (int i = 0; i < static_cast<int>(arg.threads.x); i++) { value = t(value, i, j); }
     }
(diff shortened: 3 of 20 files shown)


include/targets/generic/kernel_host.h
```diff
@@ -6,12 +6,14 @@ namespace quda
...
   {
     Functor<Arg> f(const_cast<Arg &>(arg));
+#pragma omp parallel for
     for (int i = 0; i < static_cast<int>(arg.threads.x); i++) { f(i); }
   }
 
   template <template <typename> class Functor, typename Arg> void Kernel2D_host(const Arg &arg)
   {
     Functor<Arg> f(const_cast<Arg &>(arg));
+#pragma omp parallel for
```
bjoo, line 16: AFAIK this will only parallelize the outer loop. To parallelize both loops you can add the clause `collapse(2)` i.e.
```
#pragma omp parallel for collapse(2)
for(int i=0, ... ) {
 for(int j=0, ...) {
 }
 }
 ```

File: include/targets/generic/block_reduction_kernel_host.h [hunk 2]
```diff
@@ -5,6 +5,7 @@ namespace quda
   {
     Functor<Arg> t(arg);
     dim3 block(0, 0, 0);
+#pragma omp parallel for
```
bjoo, line 8: This will only parallelize the outer loop. If you want to parallelize both loops add the `collapse(2)` clause e.g.:
```
#pragma omp parallel for collapse(2)
for (block.y = 0; ... ) {
   for( block.x = 0; ...) {
    ...
    }
 }
 ```

File: include/targets/generic/kernel_host.h [hunk 3]
```diff
@@ -20,6 +22,7 @@ namespace quda
   template <template <typename> class Functor, typename Arg> void Kernel3D_host(const Arg &arg)
   {
     Functor<Arg> f(const_cast<Arg &>(arg));
+#pragma omp parallel for
```
bjoo, line 25: Add `collapse(3)` here?

lib/monitor.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "tune_quda.h" // hash, version, resource path
+
+namespace quda
+{
+
+  namespace monitor
+  {
+
+    static std::thread monitor_thread;
+    static std::atomic<int> check(0);
```
hummingtree: Some comments on what these two variables are might be useful. I guess `check` could use some more intuitive name like `is_running`?

lib/targets/cuda/device.cpp
```diff
@@ -13,6 +13,19 @@ static const int Nstream = 9;
...
+    if (ret == NVML_ERROR_NOT_SUPPORTED) {				\
+      static int i=0;							\
+      if (i==0) printf("%s not supported on this GPU\n", #func);	\
+      i++;								\
+    } else if (ret != NVML_SUCCESS) {					\
+      printf("Error %s returns %s\n", #func, nvmlErrorString(ret));	\
+      exit(-1);								\
+    }									\
+  }
+
```
hummingtree: Needs a clang-format. Also should we just turn off monitor entirely in the `NVML_ERROR_NOT_SUPPORTED` case?

include/targets/generic/reduction_kernel_host.h
```diff
@@ -11,7 +11,7 @@ namespace quda
     Functor<Arg> t(arg);
 
     reduce_t value = t.init();
-
+#pragma omp parallel for collapse(2) reduction(Functor <Arg>::apply : value)
```
bjoo, line 14: Is this legal?  I thought for reduction one would need either one of the predefined reductions (+,-,*,&,|,^,&& and II), or it has to be declared as a custom reduction? A custom reduction here would be awkward because we'd need a name for the new reduction, which would vary with whatever Arg::apply is? 
Or is this cool as per the standard?

bjoo (approved):
After a quick scan this looks fine to me, I had only a few OpenMP comments I left as single comments regarding the use of  `collapse()` I saw you use it in some cases and not in others, but I was just wondering if it could be used in one or two more. The only thing I worry about is dropping a functor  ( Arg::apply ) or some such into a reduction clause. I wonder if it ought to be a custom reduction (like done for Multi-Reductions).
This could be just me not being up to spec with my OpenMP though. Modulo these I approve.

include/targets/generic/kernel_host.h
maddyscientist: Yep, that's indeed the case.  This is very much a first pass at host parallelization and there's a bunch of things one can do to improve this, such as collapsing as you note.  We also don't know whether it's best to have the `i` or `j` dimension as the slowest running one.  The reason I left it without collapse(2) is that the `j` dimension is more often mapped to a dimension which has temporal locality, so naively may exhibit good L1 reuse.

What I eventually plan to do here is to add auto tuner support for the host kernels which would autotune the order of the loops, e.g., which dimension runs outer most, and the degree of parallelization.

So for now, I would suggest leaving it as it is.

include/targets/generic/reduction_kernel_host.h
maddyscientist: This method is standard: writing this made me realize why @jcosborn @jxy wanted to define the reduction operators as static methods for the SYCL/OMP targets, which is required for this to work.

lib/targets/cuda/device.cpp
maddyscientist: I've updated the code to issue a `warningQuda`, but I think we can just let the monitoring continue, as this may only affect a subset of the measurements.  In practice, I don't think this will ever be a real issue as I think this only affect older GeForce cards, and all modern GPUs support all these queries.

include/targets/generic/reduction_kernel_host.h
maddyscientist: [c52fac3](https://github.com/lattice/quda/pull/1479/commits/c52fac3e6e75dc3479d0c01935e714cc04c91802)

lib/targets/cuda/device.cpp
maddyscientist: [c52fac3](https://github.com/lattice/quda/pull/1479/commits/c52fac3e6e75dc3479d0c01935e714cc04c91802)

lib/monitor.cpp
maddyscientist: [3f7939e](https://github.com/lattice/quda/pull/1479/commits/3f7939e18e29334c59c7fb9d753549658c2b9104)

include/targets/generic/kernel_host.h
maddyscientist: Marking as resolved: see conversation [above](https://github.com/lattice/quda/pull/1479#discussion_r1671042901)

include/targets/generic/block_reduction_kernel_host.h
maddyscientist: Marking as resolved: see conversation [above](https://github.com/lattice/quda/pull/1479#discussion_r1671042901).

hummingtree (approved):

PR #1480, opened 2024-07-09
Interface for Yang-Mills gradient flow

aniketsen:
Implemented an interface for Yang-Mills gradient flow (arXiv:1302.5246, Appendix D). Uses existing kernels for Wilson flow and Laplace operator. Added a function `GFlowStep` to perform individual sub-steps of the Wilson flow. For this purpose, moved `WFlowStepType` to `enum_quda` to make it globally accessible.


maddyscientist:
@Jenkins ok to test

include/gauge_tools.h
```diff
@@ -145,6 +145,21 @@ namespace quda
   */
   void WFlowStep(GaugeField &out, GaugeField &temp, GaugeField &in, double epsilon, QudaGaugeSmearType smear_type);
 
+  /**
+     @brief Apply intermediary Wilson Flow steps W1, W2 or Vt to the gauge field.
+     This routine assumes that the input and output fields are
+     extended, with the input field being exchanged prior to calling
+     this function.  On exit from this routine, the output field will
+     have been exchanged.
+     @param[out] dataDs Output smeared field
```
maddyscientist: I see there are typos in the doxygen here, inherited from the typos in `WFlowStep` above.  Can you this these?  E.g., 
```
     @param[out] dataDs Output smeared field
```
should be
```
     @param[out] out Output smeared field
```

lib/interface_quda.cpp
```diff
@@ -5574,6 +5578,167 @@ void performWFlowQuda(QudaGaugeSmearParam *smear_param, QudaGaugeObservableParam
...
+
+  GaugeFieldParam gParam(*gaugePrecise);
+  gParam.reconstruct = QUDA_RECONSTRUCT_NO; // temporary field is not on manifold so cannot use reconstruct
+  GaugeField gaugeTemp(gParam);
+
+  GaugeField &gin = *gaugeSmeared;
+  GaugeField &gout = gaugeAux;
+
+  // helper gauge field for Laplace operator
+  GaugeField *precise = nullptr;
```
maddyscientist: Now that the `GaugeField` class has copy and move constructors and assignment operators, there's no need to make this heap allocated (a pointer).  Can you make this an object?

While there's plenty of legacy QUDA that uses pointers, going forward, new code should use stack based allocation where possible.

lib/interface_quda.cpp
```diff
@@ -5574,6 +5578,167 @@ void performWFlowQuda(QudaGaugeSmearParam *smear_param, QudaGaugeObservableParam
...
+  GaugeFieldParam gParam_helper(*gaugePrecise);
+  gParam_helper.create = QUDA_NULL_FIELD_CREATE;
+  precise = new GaugeField(gParam_helper);
+
+  // spinor fields
+  ColorSpinorParam cpuParam(h_in, *inv_param, gaugePrecise->X(), false, inv_param->input_location);
+  ColorSpinorField fin_h(cpuParam);
+
+  ColorSpinorParam cudaParam(cpuParam, *inv_param, QUDA_CUDA_FIELD_LOCATION);
```
maddyscientist: Probably I prefer `deviceParam` or something else that doesn't state `cuda` in its name.  Now that QUDA is portable, we should be saving the use of the CUDA word for target-specific things.

maddyscientist (changes requested):
Thanks for this contribution @aniketsen, an excellent addition to QUDA.  I've left a few review comments, and a few things need fixed, but all should be easy to do.

Can you also apply clang-format to the diff please, as described [here](https://github.com/lattice/quda/wiki/Coding-Conventions-and-Style)?

Do you have any thoughts on how we can correctness tests for this (as well as the regular Wilson Flow function) in QUDA?  Short of fully implementing Wilson flow in a separate test code, is there anything that makes sense for testing?

aniketsen:
@maddyscientist sorry for taking so long to get back. I have made the changes that you suggested.

For testing, we have a native cpu implementation which both the interfaces for Wilson flow and Gradient flow have been tested against. For a correctness test for Wilson flow, I don't see any other way than implementing the individual steps. For the Gradient flow, if the Wilson flow steps and ApplyLaplace are tested, the rest is just algebra, which I suppose is also separately tested.

cpviolator:
@maddyscientist @aniketsen That reminds me to implement some CPU code for (at least) the 3-staple action Stout and APE routines. I can get to that relatively soon if I find time away from my day job. Please leave it with me and I'll either make a separate PR or attach to this if it's not merged before I'm done.

maddyscientist:
cscs-ci run

maddyscientist (approved):

PR #1487, opened 2024-08-29
Add "hisq" as a test executable `--dslash-type` argument

weinbe2:
This PR adds support to a custom argument `hisq` that can be passed to the command-line flag to `--dslash-type` which is equivalent to `--dslash-type asqtad --compute-fat-long true`. In other words---this gives you a HISQ that is really HISQ.

This doesn't make any other changes under the hood, it does not add a new `enum` type, it is purely for convenience in the test executables.

### DIFF
--- a/include/enum_quda.h
+++ b/include/enum_quda.h
@@ -89,6 +89,8 @@ typedef enum QudaGaugeFixed_s {
 // Types used in QudaInvertParam
 //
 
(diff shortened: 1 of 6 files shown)


weinbe2:
Hm, well, clearly something's going awry. That's what I get for assuming something's "obviously" going to be fine...

maddyscientist (approved):
LGTM

Issue #1098, opened 2021-01-11
Replace `isStaggered()`, etc, in the `Dirac` class with a virtual function implemented in each derived class


weinbe2:
Described in title

PR #1488, opened 2024-08-30
Cleaned up isWilson/isStaggered/isDwf/getStencilSteps routines

weinbe2:
This PR addresses issue #1098 and extends it to the `getStencilSteps` function as well, replacing various functions that look like this:

```
bool isWilsonType() const
    {
      return (Type() == typeid(DiracWilson).name() || Type() == typeid(DiracWilsonPC).name()
              || Type() == typeid(DiracClover).name() || Type() == typeid(DiracCloverPC).name()
              || Type() == typeid(DiracCloverHasenbuschTwist).name()
              || Type() == typeid(DiracCloverHasenbuschTwistPC).name() || Type() == typeid(DiracTwistedMass).name()
              || Type() == typeid(DiracTwistedMassPC).name() || Type() == typeid(DiracTwistedClover).name()
              || Type() == typeid(DiracTwistedCloverPC).name()) ?
        true :
        false;
    }
```

~And turning them into `Dirac` abstract functions so each individual class is "forced" to define it so functions like `isWilsonType`, etc, don't need to be manually updated each time (or, more realistically, don't accidentally get forgotten).~

~There are some other functions that could this be applied to, `hermitian()` and `hasSpecialMG()`, but it seemed reasonable to leave those as false unless proven otherwise.~

This PR now replaces those above functions with one-stop shop static `Dirac::[...]` functions that can query if various `QudaDslashType` or `QudaDiracType` enums correspond to Wilson, dwf, or staggered operators. The various `Dirac*` classes then query these `static` functions as appropriate. There is scope to apply these functions elsewhere in QUDA.

This PR also fixes a small bug in split grid for staggered vs asqtad type fermions.

### DIFF
--- a/include/dirac_quda.h
+++ b/include/dirac_quda.h
@@ -201,9 +201,9 @@ namespace quda {
     }
 
     /**
-      @brief static function that returns if a Dirac type is staggered-type depending on a QudaDiracType
-     */
-    static bool is_staggered(QudaDiracType);
+      @brief Whether the Dirac object is the DiracCoarse.
+    */
+    virtual bool isCoarse() const { return false; }
 
     /**
       @brief static function that returns if a Dirac type is staggered-type depending on a QudaDiracType
@@ -211,14 +211,9 @@ namespace quda {
     static bool is_wilson_type(QudaDiracType);
 
     /**
-      @brief static function that returns if a Dirac type is a domain wall operator (5-dimensional) depending on a QudaDiracType
+      @brief static function that returns if a Dslash type is staggered-type depending on a QudaDslashType
      */
-    static bool is_dwf(QudaDiracType);
-
-    /**
-      @brief Whether the Dirac object is the DiracCoarse.
-    */
-    virtual bool isCoarse() const { return false; }
+    static bool is_wilson_type(QudaDslashType);
 
     /**
       @brief return if the operator is a Wilson-type 4-d operator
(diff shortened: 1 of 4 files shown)


include/dirac_quda.h
```diff
@@ -368,9 +383,9 @@ namespace quda {
     QudaMatPCType getMatPCType() const { return matpcType; }
 
     /**
-       @brief  I have no idea what this does
+       @brief returns the number of stencil applications per dslash application; 1 for fused operators, generally 2 otherwise
```
maddyscientist: I think this description is wrong.  It's 1 for operator that consist of a single hopping term, generally full operators, and 2 for composite operators that consist of two hopping terms.

include/dirac_quda.h
```diff
@@ -205,6 +205,21 @@ namespace quda {
     */
     virtual bool isCoarse() const { return false; }
 
+    /**
+      @brief return if the operator is a Wilson-type 4-d operator
+     */
+    virtual bool isWilsonType() const = 0;
```
maddyscientist: Having every operator need to specify this seems a bit cumbersome.  Why not default all of these operators to return false, and then specialize where needed?  `DiracWilson` can return `true` and all children will then inherit that.  Similarly for staggered and DWF.

include/dirac_quda.h
weinbe2: It's now true in practice, however there is an exception with the clover hasenbush operator where the "full" operator is still a separate D_eo and D_oe. Getting this counting correct was necessary for the worker abstraction in various operators.

include/dirac_quda.h
weinbe2: It was cumbersome for me, but it's a marginal effort for anyone going forward. The reason why I made it abstract and not have it return false by default is to force folks to think about it, but again only once.

I do like the idea of setting it to true for sufficiently high-up "base" classes such as `DiracWilson`, so I'll pop that in.

include/dirac_quda.h
weinbe2: I'll update the wording to make that clearer, though

weinbe2: https://github.com/lattice/quda/pull/1488/commits/83d603f91d50b2c48462dc1132aa9a2b100e436e

include/dirac_quda.h
weinbe2: https://github.com/lattice/quda/pull/1488/commits/b4b9e91ad6c2a222f440795e7c1db2212c6a7aa5

maddyscientist (approved):
Aside from trivial comment, which can likely be ignored, this is good to go.

lib/dirac.cpp
```diff
@@ -254,55 +254,274 @@ namespace quda {
...
+    case QUDA_WILSON_DIRAC:
+    case QUDA_WILSONPC_DIRAC:
+    case QUDA_CLOVER_DIRAC:
+    case QUDA_CLOVERPC_DIRAC:
+    case QUDA_CLOVER_HASENBUSCH_TWIST_DIRAC:
+    case QUDA_CLOVER_HASENBUSCH_TWISTPC_DIRAC:
+    case QUDA_TWISTED_CLOVER_DIRAC:
+    case QUDA_TWISTED_CLOVERPC_DIRAC:
+    case QUDA_TWISTED_MASS_DIRAC:
+    case QUDA_TWISTED_MASSPC_DIRAC: return true; break;
```
maddyscientist: the `break` after `return here is redundant (and similar below).  Not that it's a big deal.

weinbe2: Discussed offline, since this "does no harm" I'll put a fix into one of the other branches I'm working on.

weinbe2: jk, done: https://github.com/lattice/quda/pull/1488/commits/919a6eff37660a7614fcc8b1e8f44cfcb3918779

maddyscientist:
cscs-ci run

PR #1489, opened 2024-09-05
Feature/mrhs solvers

maddyscientist:
This is PR is a biggie:
* Adds MRHS solvers to QUDA (at present CG, MR, SD, GCR, CA-GCR) are all implemented
  * Reliable updates performed using 0th RHS
  * Do not flag convergence until all RHS are converged
* Adds MRHS support to multigrid
  * Batched null space finding implemented, exposed with new parameter `QudaMultigridParam::n_vec_batch`
  * MRHS supported in actual solves as well
* Expose MRHS solvers using `invertMultiSrcQuda` interface
  * This is compatible with the split-grid interface, batching is used across the number of sources per sub grid
* Explicit breaking of the interface
  * `QudaInvertParam::true_res` and `QudaInvertParam::true_res_hq` are now arrays
* Batched deflation implemented
  * Batch eigenvalue deflation confirmed working using staggered fermions
  * Batch singular-value deflation confirmed working as a coarse grid deflator
* Eigenvalue computation is now batch, to take advantage of MRHS
* All `Dirac::prepare` and `Dirac::reconstruct` functions are now MRHS optimized
* Chronological solver is now MRHS optimized
* Cast from `cvector<T>` to `T` is now explicit instead of implicit
  * This makes it much easier to catch bugs when updating code to be MRHS
* Tensor-core 3xFP16 `DslashCoarse` is now robust to underflow 
* Miscellaneous cleanup and additions to aid all of the above
* Fixes some earlier bugs introduced in prior MRHS PRs
* Add MMA instantiations for 32 -> 64 coarsening
* Since QUDA is now threaded, update to using `MPI_THREAD_FUNNELED`
* Augmentations to the power monitoring

Things left to do
- [x] Add real MRHS support to all solvers (or at least to a few outstanding ones: CGNE, CGNR, BiCGStab, etc.)
- [x] Verify NVSHMEM MRHS operators are all working as expected
- [ ] Update QUDA version number due to breakage of interface (QUDA 2.0?)
- [ ] Fix reporting of true residual per RHS when running split grid


weinbe2:
While I remember: I support this being a new major version number, but mayhaps it should be a 2.0-rc (release candidate) and before we go full 2.0 I can dust off this PR: https://github.com/lattice/quda/pull/1283 ; it can leverage the multirhs work _and_ it breaks the interface, so I think it would be good to bring it along for the 2.0 ride.

include/blas_quda.h
```diff
@@ -293,7 +294,7 @@ namespace quda {
 
     inline array<double, 2> max_deviation(const ColorSpinorField &x, const ColorSpinorField &y)
```
weinbe2, line 295: ~Design idea, thoughts on adding a third optional argument `int n = 0` or something that will return element `[n]` instead of the hard-coded `[0]`? Whether or not it's ever needed is another story... but may as well have that support in there for convenience?~

weinbe2: Oh, nevermind, these are the versions taking in a _single_ `ColorSpinorField`. Please disregard.

maddyscientist:
> While I remember: I support this being a new major version number, but mayhaps it should be a 2.0-rc (release candidate) and before we go full 2.0 I can dust off this PR: #1283 ; it can leverage the multirhs work _and_ it breaks the interface, so I think it would be good to bring it along for the 2.0 ride.

Absolutely.  It would be excellent to dust this sucker off.

Noting also, that there's a difference between updating the version number in the header and tagging a version number.  I was just proposing that we do the former and not the latter.  Regardless, we can do something > 1.2.x and < 2.0.0 for this PR.

include/accelerator.h
```diff
@@ -46,7 +46,12 @@ namespace quda
      * @param out Solution vector.
      * @param in Right-hand side.
      */
-    virtual void operator()(ColorSpinorField &out, ColorSpinorField &in)
+    virtual void operator()(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in) override
```
weinbe2, line 49: Meta-comment, should we start filing issues for things that haven't been natively mrhs'd yet (edit: won't be mrhs'd in the scope of this specific PR) to keep track of outstanding work?

maddyscientist: Yes, that seems like a good idea.

maddyscientist: https://github.com/lattice/quda/wiki/MRHS-to-do-list

include/quda.h
```diff
@@ -550,6 +554,8 @@ extern "C" {
     int batched_rotate;
     /** For block method solvers, the block size **/
     int block_size;
+    /** The block size when computing eigenvalues **/
```
weinbe2: I'd make this a bit clearer since on first (and second, and third) read it's hard to grok why this is different from `block_size`. Maybe call the variable `eval_compute_batch_size` or something like that? Clarifying that (a) it's for the eigenvalue computation and (b) that it's just batching work, not a formal block algorithm.

include/reference_wrapper_helper.h
```diff
@@ -490,6 +490,17 @@ namespace quda
...
 
+    /**
+       Constructor from pair of iterators
+       @param[in] first Begin iterator
+       @param[in] last End iterator
+     */
+    template <class U, std::enable_if_t<is_iterator_v<U>> * = nullptr> vector(U first, U last)
+    {
+      std::vector<T>::reserve(last - first);
+      for (auto it = first; it != last; it++) std::vector<T>::push_back(*it);
```
weinbe2: `std::vector` supports using a begin and end iterator in the constructor (number 5 here: https://en.cppreference.com/w/cpp/container/vector/vector ); is there a reason that doesn't work here?

lib/dirac_coarse.cpp
```diff
@@ -585,68 +577,56 @@ namespace quda {
...
-#if 0
-      // x_o = A_oo^-1 (b_o - D_oe x_e)
-      DiracCoarse::Dslash(tmp, x.Even(), QUDA_ODD_PARITY);
-      blas::xpay(b.Odd(), -1.0, tmp);
-      CloverInv(x.Odd(), tmp, QUDA_ODD_PARITY);
+    if (solType == QUDA_MATPC_SOLUTION || solType == QUDA_MATPCDAG_MATPC_SOLUTION) return;
+
+    checkFullSpinor(x, b);
+    auto tmp = getFieldTmp(x.Even());
+#if 1
```
weinbe2: This needs to be switched back to `#if 0`, I'm guessing this was for some intermediate debugging purposes. The origin of this was when we were (more) memory constrained with HISQ MG and we wanted to keep the non-preconditioned links fully on the host in the managed-memory mode.

lib/dirac_clover_hasenbusch_twist.cpp
```diff
@@ -29,14 +29,14 @@ namespace quda
   void DiracCloverHasenbuschTwist::M(cvector_ref<ColorSpinorField> &out, cvector_ref<const ColorSpinorField> &in) const
   {
     if (symmetric) {
-      ApplyWilsonCloverHasenbuschTwist(out[this_parity], in[other_parity], *gauge, *clover, -kappa, mu, in[this_parity],
```
weinbe2, line 32: Outside of the scope of this PR, but while we're here... I see that `DiracCloverHasenbuschTwistPC` below didn't get any of the alias/etc treatment. I'm guessing there was a reason for that (unused codepath?), but if it's quick... would it be worth applying it?

lib/interface_quda.cpp
```diff
@@ -297,6 +302,13 @@ static void profilerStop(const char *f) {
 
 namespace quda {
   void printLaunchTimer();
+
+  void massRescale(cvector_ref<ColorSpinorField> &b, QudaInvertParam &param, bool for_multishift);
```
weinbe2, line 303: Is there an `extern` or something of the sort missing on this declaration? I see how these routines are defined in the new `lib/solve.cpp` file... but I don't see a corresponding header definition?

lib/inv_cg_quda.cpp
```diff
@@ -244,23 +252,29 @@ namespace quda {
...
-      printfQuda("Warning: inverting on zero-field source\n");
-      x = b;
-      param.true_res = 0.0;
-      param.true_res_hq = 0.0;
-      return;
+    if (param.compute_null_vector == QUDA_COMPUTE_NULL_VECTOR_NO) {
+      bool zero_src = true;
+      for (auto i = 0u; i < b.size(); i++) {
+        if (b2[i] == 0) {
+          warningQuda("inverting on zero-field source");
```
weinbe2: I'd add the index of the zero-field source, so more like `warningQuda("inverting on zero-field source %d", i);` or something of that ilk.

lib/inv_gcr_quda.cpp
```diff
@@ -240,20 +255,25 @@ namespace quda {
     }
 
     // Check to see that we're not trying to invert on a zero-field source
-    if (b2 == 0) {
-      if (param.compute_null_vector == QUDA_COMPUTE_NULL_VECTOR_NO) {
+    if (param.compute_null_vector == QUDA_COMPUTE_NULL_VECTOR_NO) {
+      bool zero_src = true;
+      for (auto i = 0u; i < b.size(); i++) {
+        if (b2[i] == 0) {
+          warningQuda("inverting on zero-field source");
```
weinbe2: Similar comment to above, printing the index of the source that's zero

lib/inv_gcr_quda.cpp
```diff
@@ -264,8 +284,9 @@ namespace quda {
     const int maxResIncrease = param.max_res_increase; // check if we reached the limit of our tolerance
     const int maxResIncreaseTotal = param.max_res_increase_total;
 
-    double heavy_quark_res = 0.0; // heavy quark residual
-    if(use_heavy_quark_res) heavy_quark_res = sqrt(blas::HeavyQuarkResidualNorm(x,r).z);
+    std::vector<double> heavy_quark_res(b.size()); // heavy quark residual
+    if (use_heavy_quark_res)
+      for (auto i = 0u; i < b.size(); i++) heavy_quark_res[i] = sqrt(blas::HeavyQuarkResidualNorm(x, r)[i].z);
```
weinbe2: Not that I'm stressed about how efficiently we compute the HQRes, but this recomputes the HQR multiple times, I think this should be more like:

```
std::vector<double> heavy_quark_res(b.size()); // heavy quark residual
if (use_heavy_quark_res) {
  auto hqr = blas::HeavyQuarkResidualNorm(x, r);
  for (auto i = 0u; i < b.size(); i++) heavy_quark_res[i] = sqrt(hqr[i].z);
}
```

lib/milc_interface.cpp
```diff
@@ -1426,8 +1426,9 @@ void qudaInvertMsrc(int external_precision, int quda_precision, double mass, Qud
 
   // return the number of iterations taken by the inverter
   *num_iters = invertParam.iter;
```
weinbe2, line 1435: I believe MILC expects the aggregate number of iterations as if it wasn't a batch/block solve, but I forget. I'll also scope out how it expects the residuals as part of the work I'm doing on the MILC side and sort this out

lib/multigrid.cpp
```diff
@@ -660,7 +664,7 @@ namespace quda
       param_coarse_solver->verbosity_precondition = param.mg_global.verbosity[param.level+1];
 
       // preconditioned solver wrapper is uniform precision
-      param_coarse_solver->precision = r_coarse.Precision();
+      param_coarse_solver->precision = r_coarse[0].Precision();
```
weinbe2, line 666: iirc we added support for directly calling `*.Precision()` on the ColorSpinorVector wrapper?

tests/invert_test.cpp
```diff
@@ -47,6 +48,7 @@ void display_test_info()
     printfQuda(" - number of levels %d\n", mg_levels);
     for (int i = 0; i < mg_levels - 1; i++) {
       printfQuda(" - level %d number of null-space vectors %d\n", i + 1, nvec[i]);
+      printfQuda(" - level %d null-space vector batch size %d\n", i + 1, nvec_batch[i]);
```
weinbe2, line 52: This needs to be propagated into `staggered_invert_test.cpp`

tests/utils/command_line_params.cpp
```diff
@@ -788,6 +791,7 @@ void add_eigen_option_group(std::shared_ptr<QUDAApp> quda_app)
   opgroup->add_option("--eig-ortho-block-size", eig_ortho_block_size,
                       "The block size to use when orthonormalising vectors in hybrid modified Gram-Schmidt"
                       "0 for always Classical, 1 for Modified, n > 1 for Hybrid)");
+  opgroup->add_option("--eig-eval-block-size", eig_eval_block_size, "The block size used when computing eigenvalues in the eigensolver");
```
weinbe2: `block size` -> `batching size`? This connects to my comment on documenting the `eval_block_size` variable earlier.

tests/utils/host_utils.cpp
```diff
@@ -421,7 +421,15 @@ void initComms(int, char **, int *const commDims)
     QMP_declare_logical_topology_map(commDims, 4, map, 4);
   }
 #elif defined(MPI_COMMS)
-  MPI_Init(&argc, &argv);
+  int provided = 0;
+  int required = MPI_THREAD_FUNNELED;
+  int flag = MPI_Init_thread(&argc, &argv, required, &provided);
+
+  if (provided != required) {
```
weinbe2, line 413: Is there a specific reason why we want this to error out if it's not the case? Can we do a cleanup and re-initialize with vanilla `MPI_Init` if we can't get funneled?

tests/utils/set_params.cpp
```diff
@@ -142,6 +142,9 @@ void setInvertParam(QudaInvertParam &inv_param)
   // Use 3D or 4D laplace
   inv_param.laplace3D = laplace3D;
 
+  if (Nsrc < Nsrc_tile || Nsrc % Nsrc_tile != 0)
```
weinbe2, line 145: I think we should relax this sooner as opposed to later and put in the logic to let the last "tile" of sources be incomplete in the scope of this PR

lib/interface_quda.cpp
maddyscientist: All function declarations are implicitly `extern` in C++.

lib/dirac_coarse.cpp
maddyscientist: e89be7d340ad72d37cec7a83c4d1fc25d8f85a75

include/reference_wrapper_helper.h
maddyscientist: 44d8a2a1b17bda0bc4fac1b071d50ac225359d3c

lib/interface_quda.cpp
weinbe2: Today I learned!

include/quda.h
maddyscientist: fc0762d6152ff9b3f61b4c5d1787273c3d64ca02

lib/dirac_clover_hasenbusch_twist.cpp
maddyscientist: I'm not sure what you mean by the alias treatment here?  Your comment here refers to the PC class, but the line number referred to is the non PC class.

I did do some cleanup, but I don't think this is what you mean.

dcd0d439a99c31c89a6d2ec8370129df614be2af

weinbe2: I have no idea what I meant either... but that clean-up looks good to me, so I'll consider today-me happy.

include/invert_quda.h
```diff
@@ -546,7 +543,12 @@ namespace quda {
...
-    static double stopping(double tol, double b2, QudaResidualType residual_type);
+    static vector<double> stopping(double tol, cvector<double> &b2, QudaResidualType residual_type);
+
+    static inline double stopping(double tol, double b2, QudaResidualType residual_type)
+    {
+      return stopping(tol, cvector<double>(b2), residual_type)[0];
+    }
 
     /**
        @briefTest for solver convergence
```
weinbe2: Randomly noticed this: `@briefTest` -> `@brief Test`

tests/utils/set_params.cpp
maddyscientist: This is outside the scope of this PR.  To have different tiles sizes on each subset will very quickly break the autotuner, hence why we require this condition.  Fixing this is dependent on allowing sub-communicators to perform autotuning locally.

tests/utils/host_utils.cpp
maddyscientist: Every MPI supports this now, and if we use `MPI_Init` we are skating on thin ice of UB given that QUDA is now threaded on the CPU.  So erroring out is the correct thing to do here.

tests/utils/command_line_params.cpp
maddyscientist: https://github.com/lattice/quda/commit/fc0762d6152ff9b3f61b4c5d1787273c3d64ca02

lib/multigrid.cpp
maddyscientist: Indeed we did, but the vectors contained in multigrid.cpp are not `quda::vector_ref<ColorSpinorField>`, and instead they are `std::vector<ColorSpinorField>`.

Eventually we could replace all use of `std::vector` with the QUDA derived flavor, `quda::vector`, but that's quite a bit of surgery, that is beyond the scope of this PR.

lib/inv_gcr_quda.cpp
maddyscientist: f0d9f3c8dc79bdb4ed751b34193262c517a6b287

lib/inv_cg_quda.cpp
maddyscientist: f0d9f3c8dc79bdb4ed751b34193262c517a6b287

lib/inv_gcr_quda.cpp
maddyscientist: f0d9f3c8dc79bdb4ed751b34193262c517a6b287

tests/invert_test.cpp
maddyscientist: f0d9f3c8dc79bdb4ed751b34193262c517a6b287

include/invert_quda.h
maddyscientist: f0d9f3c8dc79bdb4ed751b34193262c517a6b287

lib/inv_ca_cg.cpp
```diff
@@ -458,34 +505,37 @@ namespace quda
...
+        }
 
         // break-out check if we have reached the limit of the precision
         if (r2 > r2_old) {
           resIncrease++;
           resIncreaseTotal++;
           warningQuda(
             "CA-CG: new reliable residual norm %e is greater than previous reliable residual norm %e (total #inc %i)",
-            sqrt(r2), sqrt(r2_old), resIncreaseTotal);
+            sqrt(r2[0]), sqrt(r2_old[9]), resIncreaseTotal);
```
weinbe2: `r2_old[0]` not `r2_old[9]` :)

lib/milc_interface.cpp
weinbe2: Note to self, can confirm that is the case: https://github.com/lattice/milc_qcd/blob/develop/generic_ks/d_congrad5_fn_milc.c#L409-L417

lib/solver.cpp
```diff
@@ -368,108 +370,125 @@ namespace quda {
...
+        if (std::isnan(r2[i]) || std::isinf(r2[i])) errorQuda("Solver appears to have diverged with residual %9.6e", r2[i]);
+        if (r2[i] > r2_tol[i]) return false;
+      }
     }
-
     return true;
   }
 
-  bool Solver::convergenceHQ(double, double hq2, double, double hq_tol)
+  bool Solver::convergenceHQ(cvector<double> &hq2, cvector<double> &hq_tol)
```
weinbe2, line 424: There's a big "gotcha" that happens here: in, for example, `inv_cg_quda.cpp`, this routine is called as:

```
convergenceHQ(hq_res, param.tol_hq)
```

Where `param.tol_hq` is just a _double_. Within `convergenceHQ` (and elsewhere), this is blindly converted to a `cvector<double>` where the 0th element is `param.tol_hq` and then the rest are uninitialized. This gives garbage for the other values, messing up convergence.

I think the only way to catch this is to make a conversion from a type `T` to `cvector<T>` require an explicit cast, which I think you've imposed on other types as well.

lib/inv_cg_quda.cpp
```diff
@@ -688,20 +586,22 @@ namespace quda {
...
+      }
+      rNorm = sqrt(r2[0]);
 
       // If the iterated norm has dropped by more than a factor of delta, trigger
       // an update. The baseline we check against differs depending on if
       // we're still checking the L2 norm, or if that has converged/broken down and we're
       // now looking at the HQ residual.
 
-      if (!L2breakdown && (L2_required || convergenceL2(r2, hq_res, stop, param.tol_hq))) {
+      if (!L2breakdown && (L2_required || convergenceL2(hq_res, param.tol_hq))) {
```
weinbe2: This should be `convergenceL2(r2, stop)` not `convergenceL2(hq_res, param.tol_hq)`

lib/inv_ca_cg.cpp
maddyscientist: 7523f3a5749840f0ff26ce1130713043cde2a656

lib/solver.cpp
maddyscientist: 50e78799c7b9e5babb2ced06bcf2f27e35f036e3 addresses this issue: the convergence and printing helper functions now check the sizes of the arguments and will call a sane error if the sizes don't match.

lib/inv_cg_quda.cpp
maddyscientist: 50e78799c7b9e5babb2ced06bcf2f27e35f036e3

maddyscientist:
Noting all solvers have been made MRHS aware now, with the exception of the legacy GMRES-DR and EigCG solvers, which are overdue for a complete cleanup which is outside the solve of this PR.

maddyscientist:
This PR is now functionally complete, and all tests are passing.  This is ready for final review (@weinbe2 @hummingtree @mathiaswagner @bjoo).

lib/milc_interface.cpp
weinbe2: Resolved by 36ba139856a8a7e25a9734a509a5084457aecb6c

lib/inv_bicgstab_quda.cpp
```diff
@@ -371,10 +370,12 @@ namespace quda {
 
     if (!param.is_preconditioner) { // do not do the below if we this is an inner solver
```
weinbe2: In the spirit of typo fixing, `if we this is` -> `if this is`

weinbe2:
I have tested the batch CG solver with a modified version of MILC that properly utilizes the multisource MILC interface function. This is available here: https://github.com/lattice/milc_qcd/tree/feature/quda-block-solver-interface ; current commit is https://github.com/lattice/milc_qcd/commit/f0404fe841712b63837711e2252d08d1491e0502 . This PR works perfectly fine with the current `develop` version of MILC.

I will note that this has only tested vanilla CG. I have not yet plumbed in multi-rhs support for the MG solver; I consider that within the scope of a second QUDA PR.

weinbe2:
When using coarsest-level deflation (perhaps just with staggered operators?) it looks like we need to change the default values corresponding to the flag `--mg-eig-evals-batch-size 2 [###]`. I uncovered this with coarsest-level deflation for staggered operators, coarse Nc = 64 or 96, on `sm_80`. I hit the following error when trying to converge 16 eigenvalues:

```
[...]
MG level 2 (GPU): RitzValue[0015]: (+2.6378149309623524e-03, +0.0000000000000000e+00) residual 1.4461873963747640e-05
MG level 2 (GPU): ERROR: nVec = 8 not instantiated
 (rank 0, host ipp1-1780.nvidia.com, block_transpose.cu:116 in void quda::launch_span_nVec(v_t&, quda::cvector_ref<O>&, quda::IntList<nVec, N ...>) [with v_t = quda::ColorSpinorField; b_t = const quda::ColorSpinorField; vFloat = float; bFloat = float; int nSpin = 2; int nColor = 64; int nVec = 16; int ...N = {}; quda::cvector_ref<O> = const quda::vector_ref<const quda::ColorSpinorField>]())
MG level 2 (GPU):        last kernel called was (name=cudaMemsetAsync,volume=bytes=8192,aux=zero,color_spinor_field.cpp,406)
```

I compiled with MG MRHS support for 16 and 32; I can't think of anywhere that I've imposed a multiple of 8. The solution was explicitly setting `--mg-eig-evals-batch-size 2 16` (where 2 is the lowest level).

My (relatively reduced) command is:

```
mpirun -np 1 ./staggered_invert_test \
  --prec double --prec-sloppy single --prec-null half --prec-precondition half \
  --mass 0.1 --recon 13 --recon-sloppy 9 --recon-precondition 9 \
  --dim 8 8 8 8 --gridsize 1 1 1 1 \
  --dslash-type staggered --compute-fat-long true --tadpole-coeff 0.905160183 --tol 1e-10 \
  --verbosity verbose --solve-type direct --solution-type mat --inv-type gcr \
  --inv-multigrid true --mg-levels 3 --mg-coarse-solve-type 0 direct --mg-staggered-coarsen-type kd-optimized \
  --mg-block-size 0 1 1 1 1 --mg-nvec 0 3 \
  --mg-block-size 1 4 4 4 4 --mg-nvec 1 64 \
  --mg-setup-tol 1 1e-5 --mg-setup-inv 1 cgnr \
  --nsrc 1 --niter 25 \
  --mg-setup-use-mma 0 true --mg-setup-use-mma 1 true --mg-setup-use-mma 2 true \
  --mg-dslash-use-mma 0 true --mg-dslash-use-mma 1 true --mg-dslash-use-mma 2 true \
  --mg-smoother 0 ca-gcr --mg-smoother-solve-type 0 direct  --mg-nu-pre 0 0 --mg-nu-post 0 4 \
  --mg-smoother 1 ca-gcr --mg-smoother-solve-type 1 direct --mg-nu-pre 1 0 --mg-nu-post 1 4 \
  --mg-smoother 2 ca-gcr --mg-smoother-solve-type 2 direct-pc  --mg-nu-pre 2 0 --mg-nu-post 2 4 \
  --mg-coarse-solver 1 gcr --mg-coarse-solve-type 1 direct --mg-coarse-solver-tol 1 0.25 --mg-coarse-solver-maxiter 1 16 \
  --mg-coarse-solver 2 ca-gcr --mg-coarse-solve-type 2 direct-pc --mg-coarse-solver-tol 2 0.25 --mg-coarse-solver-maxiter 2 16 --mg-coarse-solver-ca-basis-size 2 16 \
  --mg-verbosity 0 verbose --mg-verbosity 1 verbose --mg-verbosity 2 verbose --mg-verbosity 3 verbose \
  --mg-eig 2 true --mg-eig-type 2 trlm --mg-eig-use-dagger 2 false --mg-eig-use-normop 2 true \
  --mg-nvec 2 16 --mg-eig-n-ev 2 16 --mg-eig-n-kr 2 128 --mg-eig-tol 2 1e-1 \
  --mg-eig-use-poly-acc 2 false --mg-eig-poly-deg 2 100 --mg-eig-amin 2 1e-1 \
  --mg-eig-max-restarts 2 1000
```

Neither toggling `--mg-setup-use-mma 2 false` nor `--mg-dslash-use-mma 2 false` works around this.

I can't quite think of a good way to address this (yet), but I'm also not clear on the details in the weeds. Maybe you know exactly where the fix is @maddyscientist ?

maddyscientist:
> When using coarsest-level deflation (perhaps just with staggered operators?) it looks like we need to change the default values corresponding to the flag `--mg-eig-evals-batch-size 2 [###]`. I uncovered this with coarsest-level deflation for staggered operators, coarse Nc = 64 or 96, on `sm_80`. I hit the following error when trying to converge 16 eigenvalues:
> 
> ```
> [...]
> MG level 2 (GPU): RitzValue[0015]: (+2.6378149309623524e-03, +0.0000000000000000e+00) residual 1.4461873963747640e-05
> MG level 2 (GPU): ERROR: nVec = 8 not instantiated
>  (rank 0, host ipp1-1780.nvidia.com, block_transpose.cu:116 in void quda::launch_span_nVec(v_t&, quda::cvector_ref<O>&, quda::IntList<nVec, N ...>) [with v_t = quda::ColorSpinorField; b_t = const quda::ColorSpinorField; vFloat = float; bFloat = float; int nSpin = 2; int nColor = 64; int nVec = 16; int ...N = {}; quda::cvector_ref<O> = const quda::vector_ref<const quda::ColorSpinorField>]())
> MG level 2 (GPU):        last kernel called was (name=cudaMemsetAsync,volume=bytes=8192,aux=zero,color_spinor_field.cpp,406)
> ```
> 
> I compiled with MG MRHS support for 16 and 32; I can't think of anywhere that I've imposed a multiple of 8. The solution was explicitly setting `--mg-eig-evals-batch-size 2 16` (where 2 is the lowest level).
> 
> My (relatively reduced) command is:
> 
> ```
> mpirun -np 1 ./staggered_invert_test \
>   --prec double --prec-sloppy single --prec-null half --prec-precondition half \
>   --mass 0.1 --recon 13 --recon-sloppy 9 --recon-precondition 9 \
>   --dim 8 8 8 8 --gridsize 1 1 1 1 \
>   --dslash-type staggered --compute-fat-long true --tadpole-coeff 0.905160183 --tol 1e-10 \
>   --verbosity verbose --solve-type direct --solution-type mat --inv-type gcr \
>   --inv-multigrid true --mg-levels 3 --mg-coarse-solve-type 0 direct --mg-staggered-coarsen-type kd-optimized \
>   --mg-block-size 0 1 1 1 1 --mg-nvec 0 3 \
>   --mg-block-size 1 4 4 4 4 --mg-nvec 1 64 \
>   --mg-setup-tol 1 1e-5 --mg-setup-inv 1 cgnr \
>   --nsrc 1 --niter 25 \
>   --mg-setup-use-mma 0 true --mg-setup-use-mma 1 true --mg-setup-use-mma 2 true \
>   --mg-dslash-use-mma 0 true --mg-dslash-use-mma 1 true --mg-dslash-use-mma 2 true \
>   --mg-smoother 0 ca-gcr --mg-smoother-solve-type 0 direct  --mg-nu-pre 0 0 --mg-nu-post 0 4 \
>   --mg-smoother 1 ca-gcr --mg-smoother-solve-type 1 direct --mg-nu-pre 1 0 --mg-nu-post 1 4 \
>   --mg-smoother 2 ca-gcr --mg-smoother-solve-type 2 direct-pc  --mg-nu-pre 2 0 --mg-nu-post 2 4 \
>   --mg-coarse-solver 1 gcr --mg-coarse-solve-type 1 direct --mg-coarse-solver-tol 1 0.25 --mg-coarse-solver-maxiter 1 16 \
>   --mg-coarse-solver 2 ca-gcr --mg-coarse-solve-type 2 direct-pc --mg-coarse-solver-tol 2 0.25 --mg-coarse-solver-maxiter 2 16 --mg-coarse-solver-ca-basis-size 2 16 \
>   --mg-verbosity 0 verbose --mg-verbosity 1 verbose --mg-verbosity 2 verbose --mg-verbosity 3 verbose \
>   --mg-eig 2 true --mg-eig-type 2 trlm --mg-eig-use-dagger 2 false --mg-eig-use-normop 2 true \
>   --mg-nvec 2 16 --mg-eig-n-ev 2 16 --mg-eig-n-kr 2 128 --mg-eig-tol 2 1e-1 \
>   --mg-eig-use-poly-acc 2 false --mg-eig-poly-deg 2 100 --mg-eig-amin 2 1e-1 \
>   --mg-eig-max-restarts 2 1000
> ```
> 
> Neither toggling `--mg-setup-use-mma 2 false` nor `--mg-dslash-use-mma 2 false` works around this.
> 
> I can't quite think of a good way to address this (yet), but I'm also not clear on the details in the weeds. Maybe you know exactly where the fix is @maddyscientist ?

Ok, I understand this issue.  There's two things at play here:
* For whatever reason `--mg-dslash-use-mma i` acts on the `i + 1` level, so you should set --mg-dslash-use-mma 1 false`, somewhat counter intuitively.  This was likely an oversight from when the MMA dslash was added.  I can fix this.
* If the evec batch size isn't set at the command line, it will use a default value of 8, which is what you've found.  Perhaps 16 would be a better value for this, since that's the default MMA MRHS sizes in CMake?

Perhaps it would also be a good idea to have fall back to non-MMA dslash if the requested size isn't available?  That would make things more bullet proof?  Perhaps with a warning on first call?

tests/utils/set_params.cpp
```diff
@@ -782,13 +787,16 @@ void setMultigridEigParam(QudaEigParam &mg_eig_param, int level)
...
   mg_eig_param.ortho_block_size = mg_eig_ortho_block_size[level];
+  mg_eig_param.compute_evals_batch_size = mg_eig_evals_batch_size[level];
   mg_eig_param.block_size
     = (mg_eig_param.eig_type == QUDA_EIG_TR_LANCZOS || mg_eig_param.eig_type == QUDA_EIG_IR_ARNOLDI) ?
     1 :
     mg_eig_block_size[level];
   mg_eig_param.n_ev = mg_eig_n_ev[level];
   mg_eig_param.n_kr = mg_eig_n_kr[level];
   mg_eig_param.n_conv = nvec[level];
+  mg_eig_param.compute_evals_batch_size
```
weinbe2, line 797: This variable is being set twice

weinbe2 (approved):
This has passed my visual review and my tests with MILC. Awesome PR @maddyscientist !

include/invert_x_update.h
```diff
@@ -43,6 +43,11 @@ namespace quda
       }
     }
 
+    XUpdateBatch(const XUpdateBatch &other) = default;
+    XUpdateBatch(XUpdateBatch &&other) = default;
+    XUpdateBatch &operator=(const XUpdateBatch &other) = default;
+    XUpdateBatch &operator=(XUpdateBatch &&other) = default;
+
```
hummingtree, line 50: Is this just for making sure that we use the default versions of these functions or it has a real impact?

include/kernels/dslash_coarse_mma.cuh
```diff
@@ -162,152 +164,464 @@ namespace quda
       backward_idx[d] = linkIndexHop(coord, arg.dim, d, -arg.nFace);
     }
 
-    auto dslash_forward_producer = [&](int d, float &scale_inv_a, float &scale_inv_b, int k_offset) {
-      const int fwd_idx = forward_idx[d];
+    if constexpr (do_rescale) {
```
hummingtree, line 167: The rescaling changes here can be simplified a lot - I will definitely cover these in my prolongator PR.

include/targets/cuda/reduce_helper.h
```diff
@@ -55,7 +55,8 @@ namespace quda
     bool reset = false; /** reset the counter post completion (required for multiple calls with the same arg instance */
     using system_atomic_t = typename atomic_type<T>::type; /** heterogeneous atomics must use lock-free atomics -> operate on scalars */
     static constexpr int n_item = sizeof(T) / sizeof(system_atomic_t); /** number of words per reduction variable */
-    cuda::atomic<T, cuda::thread_scope_device> *partial; /** device atomic buffer */
+    // FIXME on Hopper this could be a 128-bit type
+    cuda::atomic<system_atomic_t, cuda::thread_scope_device> *partial;  /** device atomic buffer */
```
hummingtree, line 59: Would a 128-bit type break the code?

include/dirac_quda.h
```diff
@@ -542,7 +542,6 @@ namespace quda {
     DiracWilson(const DiracWilson &dirac);
     DiracWilson(const DiracParam &param, const int nDims); // to correctly adjust face for DW and non-deg twisted mass
 
-    virtual ~DiracWilson();
```
bjoo, line 545: Just out of interest, why are we removing the virtual destructors? Is there a lower base class? Or did we make them protected or something?

maddyscientist: Since this was a trivial destructor, we're deferring to the compiler to generate a default destructor.  I think this is best practice.

lib/inv_gcr_quda.cpp
```diff
@@ -323,34 +345,39 @@ namespace quda {
 
       // update since n_krylov or maxiter reached, converged or reliable update required
       // note that the heavy quark residual will by definition only be checked every n_krylov steps
-      if (k == n_krylov || total_iter == param.maxiter || (r2 < stop && !l2_converge) || sqrt(r2 / r2_old) < param.delta) {
+      if (k == n_krylov || total_iter == param.maxiter || (r2[0] < stop[0] && !l2_converge)
+          || sqrt(r2[0] / r2_old[0]) < param.delta) {
 
         // update the solution vector
-        updateSolution(x, alpha, beta, gamma, k, p);
+        for (auto i = 0u; i < b.size(); i++) updateSolution(x[i], alpha[i], beta[i], gamma[i], k, get_i(p, i));
```
bjoo, line 352: Why loop instead of an MRHS udpateSolution here?

lib/solver.hpp
```diff
@@ -51,18 +51,18 @@ namespace quda
   /**
      @brief Generate a Krylov space in a given basis
      @param[in] diracm Dirac matrix used to generate the Krylov space
-     @param[out] Ap dirac matrix times the Krylov basis vectors
-     @param[in,out] p Krylov basis vectors; assumes p[0] is in place
+     @param[out] Ap dirac matrix times the Krylov basis vector sets
+     @param[in,out] p Krylov basis vector sest; assumes p[0] is in place
```
bjoo: 'test' is misspelled as 'sest' here.

bjoo (approved):
I had a very cursory look. I haven't had a chance to test the very latest with Chroma, but one or two commits out I think we fixed all the Chroma issues.  So I am happy to approve. This is a great change.

lib/inv_gcr_quda.cpp
maddyscientist: This is because the multi-blas routines are not yet batch aware (which are used in `updateSolution`).  This is left as a future optimization for some PR in the future.

lib/solver.hpp
maddyscientist: [05b2bc6](https://github.com/lattice/quda/pull/1489/commits/05b2bc6e1d1f2012b19b13da9a72fa5cb21f8b00)

tests/utils/set_params.cpp
maddyscientist: [7903288](https://github.com/lattice/quda/pull/1489/commits/7903288629f0fcc474989fec5a1393ecc17a4b42)

hummingtree (approved):

include/kernels/dslash_coarse_mma.cuh
maddyscientist: Great to hear 😄

include/invert_x_update.h
maddyscientist: Needed to add these functions to ensure I could create an empty object and then assign it in a loop (over RHS).

include/targets/cuda/reduce_helper.h
maddyscientist: Hopper introduced 128-bit atomic support.  On prior architectures you cannot write 128-bit words atomically, which could lead to UB.

PR #1490, opened 2024-09-13
print b5 and c5 MDWF params with QUDA_DEBUG_VERBOSE printing of params

walkloud:
We cleaned up how the b5 and c5 params are printed.  They are now printed with the printing of the invert params instead of separately, and still printed only if verbosity = QUDA_DEBUG_VERBOSE


maddyscientist:
cscs-ci run

maddyscientist (approved):

PR #1493, opened 2024-09-19
Remove VLAs and replace with std::vector

dmcdougall:
VLAs were rejected by WG21 and, in light of this, recent-ish Clangs updated the warning set to include their use in C++ code.  See [here](https://github.com/llvm/llvm-project/commit/7339c0f782d5c70e0928f8991b0c05338a90c84c) for the relevant LLVM commit.

STRICT builds therefore fail to compile them.  This patch replaces their use with std::vector where appropriate.

### DIFF
--- a/tests/host_reference/contract_ft_reference.h
+++ b/tests/host_reference/contract_ft_reference.h
@@ -117,10 +117,7 @@ void contractFTHost(Float **h_prop_array_flavor_1, Float **h_prop_array_flavor_2
       for (int dir = 0; dir < 4; ++dir) {
         double theta = 2. * M_PI / L[dir];
         theta *= (sink[dir] - source_position[dir]) * mom_modes[4 * mom_idx + dir];
-        double z[2];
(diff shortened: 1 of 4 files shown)


dmcdougall:
`contract_ft_test` is failing... investigating.

maddyscientist (approved):
Thanks @dmcdougall

dmcdougall:
Oh, the first argument to `FourierPhase` [isn't a pointer to `const`](https://github.com/lattice/quda/blob/c4856e72e73ab08b56ff955ed2c34464bdc9cca9/tests/host_reference/contract_ft_reference.h#L41).  You're probably expecting this method to change the stuff pointed to by `phase`, and I just put it on the stack before calling the method.  That's probably why the test is failing; changes aren't propagated back to the caller.

Give me a moment to fix it.

dmcdougall:
`contract_ft_test` appears to be passing now.

PR #1495, opened 2024-09-26
Speed up rocm builds using fast compile macros

dmcdougall:


dmcdougall:
This is just for the github action workflow, not some general compiler performance improvement.

maddyscientist (approved):

PR #1506, opened 2024-10-22
Feature/trlm 3d

maddyscientist:
This PR adds support for finding eigenvectors of the 3-d Laplace operator.
* Add new Eigensolver algorithm `TRLM3D` which is a batched 3-d eigensolver
* Fixes support for autotuning with split grid enabled
   * When we return to the global grid, the tunecaches from each sub grid are merged with all processes having a copy of this uniform cache
   * This enables support for sub-grid tuning as opposed to world-wide tuning only as previously.
* Add support for using the resident smeared gauge field for the eigensolver operator (Laplace operator only supported at present)
* Add support for host verification of Laplace 3-d and 4-d operators when running the Eigensolver
* Fixes a bug with candidate tuning (uninitialized variable)

A couple of things are left to do:
- [x] Only broadcast and merge the split grid tunecaches if we need to
- [x] Add ctest for Laplace 3-d and 4-d eigensolves


include/quda.h
```diff
@@ -506,6 +505,10 @@ extern "C" {
         false, but preserve_deflation would be true */
     QudaBoolean preserve_evals;
 
+    /** Whether to use the smeared gauge field for the Dirac operator
+        for whose eigenvalues are are computing. */
+    bool use_smeared_gauge;
```
hummingtree, line 510: Why ... don't we use `QudaBoolean`?

lib/communicator_stack.cpp
```diff
@@ -49,8 +50,16 @@ namespace quda
   void push_communicator(const CommKey &split_key)
   {
     if (comm_nvshmem_enabled())
-      errorQuda(
-        "Split-grid is currently not supported with NVSHMEM. Please set QUDA_ENABLE_NVSHMEM=0 to disable NVSHMEM.");
+      errorQuda("Split-grid is currently not supported with NVSHMEM. Set QUDA_ENABLE_NVSHMEM=0 to disable NVSHMEM.");
```
hummingtree, line 53: Just ... curious, why removing the 'Please'?

hummingtree (approved):
The issue regarding having tuning with split grid is solved cleanly! I just have two trivial questions :)

lib/communicator_stack.cpp
maddyscientist: So clang-format would fit it on one line 😉

include/quda.h
maddyscientist: Eventually I want to deprecate and remove `QudaBoolean` from QUDA completely.  `bool` type was introduced with C99, and it's safe to assume that it is always present which wasn't necessarily the case 15 years ago when we started QUDA.

maddyscientist:
I have pushed a few more updates:
* split grid tests are now always run (previously only run if tuning disabled)
* fixed a bug with MRHS fused Mobius Dslash I found while doing additional testing of this branch
* fixed an uninitialized variable (`QudaEigParam::use_smeared_gauge` added to check_param.h)

include/kernels/blas_3d.cuh
```diff
@@ -0,0 +1,341 @@
...
+      x(x),
+      y(y)
+    {
+      for (int i = 0; i < 4; i++) Xh[i] = x.SiteSubset() == 2 && i == 0 ? x.X()[i] / 2 : x.X()[i];
+    }
+
+    __device__ __host__ double init() const { return double(); }
+  };
+
+  template <int reduction_dim, class T> __device__ int idx_from_t_xyz(int t, int xyz, T X[4])
```
weinbe2: We should put this into `index_helper.cuh`

lib/communicator_qmp.cpp
```diff
@@ -166,6 +166,9 @@ void Communicator::comm_init(int ndim, const int *dims, QudaCommsMap rank_from_c
               grid_size, QMP_comm_get_number_of_nodes(QMP_COMM_HANDLE));
   }
 
+  // defer handling MPI errors to QMP
```
weinbe2, line 169: I'm a bit confused, in `lib/communicator_mpi.cpp` an identical function call defers handling MPI errors to QUDA, why does this one defer it to QMP? Is this just a statement of "who called MPI_Init"?

tests/host_reference/staggered_dslash_reference.cpp
```diff
@@ -29,12 +29,15 @@
  * @param[in] oddBit The odd/even bit for the site index
  * @param[in] daggerBit Perform the ordinary dslash (0) or Hermitian conjugate (1)
  * @param[in] dslash_type The type of Dslash operation
+ * @param[in] laplace3D Whether we applying the 3-d laplace operator
+ * (in the case of dslash_type being QUDA_LAPLACE_DSLASH
```
weinbe2: missing a closing `)`

tests/host_reference/staggered_dslash_reference.cpp
```diff
@@ -169,24 +171,21 @@ void stag_mat(ColorSpinorField &out, const GaugeField &fat_link, const GaugeFiel
...
   // have the minus sign. Inverting the expected dagger convention
   // solves this discrepancy.
-  stag_dslash(out.Even(), fat_link, long_link, in.Odd(), QUDA_EVEN_PARITY, 1 - daggerBit, dslash_type);
-  stag_dslash(out.Odd(), fat_link, long_link, in.Even(), QUDA_ODD_PARITY, 1 - daggerBit, dslash_type);
+  stag_dslash(out.Even(), fat_link, long_link, in.Odd(), QUDA_EVEN_PARITY, 1 - daggerBit, dslash_type, laplace3D);
+  stag_dslash(out.Odd(), fat_link, long_link, in.Even(), QUDA_ODD_PARITY, 1 - daggerBit, dslash_type, laplace3D);
 
   if (dslash_type == QUDA_LAPLACE_DSLASH) {
-    double kappa = 1.0 / (8 + mass);
+    double kappa = 1.0 / ((laplace3D == 3 ? 6 : 8) + mass);
```
weinbe2: iirc `laplace3D` is an integer equal to the dimension being partitioned, so in the off-chance someone is partitioning the `z` direction instead of `t` direction, `laplace3D` would be `2`, and this would be wrong.

tests/host_reference/staggered_dslash_reference.h
```diff
@@ -23,7 +23,7 @@ void setDims(int *);
  * @param[in] dslash_type Dslash type
  */
 void stag_dslash(ColorSpinorField &out, const GaugeField &fat_link, const GaugeField &long_link,
```
weinbe2, line 29: Needs updated doxygen

weinbe2: This applies to multiple functions in this file

tests/staggered_eigensolve_test.cpp
```diff
@@ -192,24 +192,75 @@ std::vector<double> eigensolve(test_t test_param)
 
   if (!enable_testing || (enable_testing && getVerbosity() >= QUDA_VERBOSE)) display_test_info(eig_param);
 
+  // Gauge Smearing Routines
+  if (gauge_smear) {
+    quda::host_timer_t host_timer;
+    host_timer.start(); // start the timer
+
+    QudaGaugeObservableParam *obs_param = new QudaGaugeObservableParam[gauge_smear_steps / measurement_interval + 1];
```
weinbe2: This needs a corresponding `delete[]`

tests/staggered_eigensolve_test.cpp
```diff
@@ -192,24 +192,75 @@ std::vector<double> eigensolve(test_t test_param)
...
+      obs_param[i] = newQudaGaugeObservableParam();
+      obs_param[i].compute_plaquette = QUDA_BOOLEAN_TRUE;
+      obs_param[i].compute_qcharge = QUDA_BOOLEAN_TRUE;
+      obs_param[i].su_project = su_project ? QUDA_BOOLEAN_TRUE : QUDA_BOOLEAN_FALSE;
+    }
+
+    // We here set all the problem parameters for all possible smearing types.
+    QudaGaugeSmearParam smear_param = newQudaGaugeSmearParam();
+    setGaugeSmearParam(smear_param);
+    smear_param.smear_type = gauge_smear_type;
```
weinbe2: Can (some of) these parameters be set in `setGaugeSmearParam`?

tests/staggered_eigensolve_test.cpp
weinbe2: Or we could use a `std::vector<QudaGaugeObservableParam>` instead, I believe?

weinbe2:
I know that historically `laplace3D` has been an integer that in principle can be set to `0`, `1`, `2`, or `3` to specify the orthogonal direction, but iirc only the `t == 3` direction is only ever used (and implemented, in most cases). I also can't think of a real need for it to be set to anything but `3`. Should we just turn `laplace3D` into a `bool` and be done with it?

If someone has a need for any of the spatial dimensions to be the orthogonal direction, we can revisit the decision then. Getting the spatial dimensions working will be far more complicated of an enterprise than just flipping a switch anyway.

weinbe2:
Can confirm the MILC `develop` safely compiles with the updated `quda.h`. I verified visually that it includes `-std=c99` in its default flags for compiling `ks_imp_rhmd` and `ks_spectrum`.

include/kernels/blas_3d.cuh
```diff
@@ -0,0 +1,341 @@
+#pragma once
+
+#include <color_spinor_field_order.h>
+#include <index_helper.cuh>
+#include <fast_intdiv.h>
+#include <quda_matrix.h>
+#include <matrix_field.h>
+#include <kernel.h>
+//#include <kernels/contraction_helper.cuh>
```
weinbe2: This can be deleted

include/kernels/blas_3d.cuh
```diff
@@ -0,0 +1,341 @@
...
+    F x;
+    const int slice;
+
+    copy3dArg(ColorSpinorField &y, ColorSpinorField &x, int slice) :
+      baseArg(dim3(y.VolumeCB(), y.SiteSubset(), 1), y), y(y), x(x), slice(slice)
+    {
+    }
+  };
+
+  template <typename Arg> struct copyTo3d {
```
weinbe2, line 45: Should we be explicitly enforcing `x.X(3) == 1` within `copyTo3d` and `copyFrom3d`? I know there's a check in [lib/blas_3d.cu](https://github.com/lattice/quda/pull/1506/files#diff-1c272c738b087a6e48b33efe392437636acdfe9fabad2e4a1be03198a16ab4f4) , but maybe there's a straightforward way to do it here as well.

include/blas_3d.h
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+       @brief Swap the slice in two given fields
+       @param[in] slice The slice we wish to swap in the fields
+       @param[in,out] x Field whose slice we wish to swap
+       @param[in,out] y Field whose slice we wish to swap
+     */
+    void swap(int slice, ColorSpinorField &x, ColorSpinorField &y);
+
+    // Reductions
```
weinbe2: doxygen

lib/gauge_laplace.cpp
```diff
@@ -79,7 +79,10 @@ namespace quda {
     // do nothing
   }
 
-  GaugeLaplacePC::GaugeLaplacePC(const DiracParam &param) : GaugeLaplace(param) { }
+  GaugeLaplacePC::GaugeLaplacePC(const DiracParam &param) : GaugeLaplace(param)
+  {
+    for (auto i = 0; i < 4; i++) if (laplace3D == i) errorQuda("Cannot use 3-d operator with with e/o preconditioning");
```
weinbe2: Supurfluous `with`

lib/solve.cpp
```diff
@@ -394,18 +395,6 @@ namespace quda
       x[0] = ColorSpinorField(cudaParam);
     }
 
-    if (param.use_init_guess == QUDA_USE_INIT_GUESS_YES && !param.chrono_use_resident) { // download initial guess
```
weinbe2: Why are we deleting this chunk of code?

include/eigensolve_quda.h
```diff
@@ -273,7 +276,7 @@ namespace quda
        @brief Compute eigenvalues and their residiua.  This variant compute the number of converged eigenvalues.
        @param[in] mat Matrix operator
        @param[in] evecs The eigenvectors
-       @param[in] evals The eigenvalues
+       @param[in,out] evals The eigenvalues
```
weinbe2: Now that we have a 3-d eigensolver, we should be make the `computeEvals` routine virtual and in the case of the 3-d eigensolvers make a custom override that `errorQuda`s out. This applies to many other routines as well---essentially anything that has a `3D` version in `TRLM3D`. We may also want to create a virtual function `virtual bool is_3d_eigensolver()` that by default returns `false` and for the 3-d eigensolvers returns `true`.

lib/eig_trlm_3d.cpp
```diff
@@ -0,0 +1,701 @@
...
+        for (int i = vecs_t.size(); i < dim; i++) vecs_t.push_back(csParamClone);
+        for (int i = kSpace_t.size(); i < keep; i++) kSpace_t.push_back(csParamClone);
+
+	blas::zero(kSpace_t);
+
+        // Copy to data to 3D array, zero out workspace, make pointers
+        for (int i = 0; i < dim; i++) blas3d::copy(t, blas3d::COPY_TO_3D, vecs_t[i], vecs_locked[i]);
+
+        // Compute the axpy
+        blas::block::axpy(ritz_mat_keep[t], {vecs_t.begin(), vecs_t.begin() + dim},
```
weinbe2, line 530: Just confirming this is correctly `blas::` and shouldn't be `blas3d::`?

lib/eig_trlm_3d.cpp
```diff
@@ -0,0 +1,701 @@
...
+    // load the vectors
+    VectorIO io(eig_param->vec_infile, eig_param->io_parity_inflate == QUDA_BOOLEAN_TRUE);
+    io.load({kSpace.begin(), kSpace.begin() + n_conv});
+
+    // Error estimates (residua) given by ||A*vec - lambda*vec||
+    computeEvals3D(mat, kSpace, evals);
+  }
+
+  template <extremumType min_max, typename T>
+  T TRLM3D::getArrayMinMax3D(const std::vector<T> &array)
```
weinbe2: I think it could be worth splitting up this routine and making it a bona-fide member of `comm_quda.h`, have something like `comm_allreduce_array_min` and `comm_allreduce_array_max` or something of the sort. This could have far more value than just this context here (though I can't think of exactly what for right now).

lib/tune.cpp
```diff
@@ -320,30 +347,26 @@ namespace quda
     }
   }
 
-  /**
-   * @brief Distribute the tunecache from a given rank to all other nodes.
-   * @param[in] root_rank From which global rank to do the broadcast
-   */
-  static void broadcastTuneCache(int32_t root_rank = 0)
+  static void broadcastTuneCache(int32_t root_rank, map &tc_recv)
```
weinbe2, line 351: re-add doxygen?

weinbe2: nvm

lib/tune.cpp
```diff
@@ -152,9 +152,36 @@ namespace quda
...
+    std::vector<map> split_tc(global_tune_rank.size());
+    // broadcast each tunecache to every process
+    for (auto i = 0u; i < global_tune_rank.size(); i++) {
+      broadcastTuneCache(global_tune_rank[i], split_tc[i]);
+      if (comm_rank() == global_tune_rank[i]) split_tc[i] = tunecache;
+      logQuda(QUDA_DEBUG_VERBOSE, "i = %d tune_rank = %d tc size = %lu\n", i, global_tune_rank[i], split_tc[i].size());
+    }
+
+    // now merge the maps
+    for (auto i = 0u; i < global_tune_rank.size(); i++) { tunecache.merge(split_tc[i]); }
```
weinbe2: This needs to be modified to make the merge deterministic across ranks

weinbe2: After digging into the documentaiton on `std::map::merge`, it looks like this may not do what we want. Based on https://en.cppreference.com/w/cpp/container/map/merge , the `merge` routine doesn't do explicit copies: "No elements are copied or moved, only the internal pointers of the container nodes are repointed." This means that if/when `split_tc[i]` falls out of scope, its contents are invalidated, and as such the pointers to objects from `split_tc[i]` now in `tunecache` are also invalid.

A more concrete code example of my thought:

```
void add_things(std::map<...> &base) {
  std::map<...> local_thing;
  // populate local_thing
  base.merge(local_thing);
  // local_thing falls out of scope
}

...

std::map<...> blah = { ... };
// I want to add things to blah
add_things(blah);

// "blah" now, in theory, contains elements that point to elements of "local_thing", but since "local_thing" fell out of scope those pointers are now invalid
```

weinbe2:
Looking at the host verify... at this point I think we should be splitting the Laplace verification routines out from the staggered ones at this point, if anything for code cleanliness/maintainability reasons. Alternatively, there needs to be many more checks for invalid combinations: `matpc` and `matdagmat` don't make sense for the 3-d Laplace operator, `laplace3D` doesn't make sense for the staggered operators... I'm sure I could find more cases of inconsistencies if I wanted to.

lib/solve.cpp
maddyscientist: Accident as part of the merge of conflicted branches.  Thanks for catching this.

maddyscientist: cbe9845685fc4fe72b878d68591377d2bc660f7a

lib/blas_3d.cu
```diff
@@ -0,0 +1,251 @@
...
+        TunableKernel2D(y, y.SiteSubset()), y(y), x(x), t_slice(t_slice), type(type)
+      {
+        // Check spins
+        if (x.Nspin() != y.Nspin()) errorQuda("Unexpected number of spins x=%d y=%d", x.Nspin(), y.Nspin());
+
+        // Check colors
+        if (x.Ncolor() != y.Ncolor()) errorQuda("Unexpected number of colors x=%d y=%d", x.Ncolor(), y.Ncolor());
+
+        // Check slice value
+        if (t_slice >= y.X()[3]) errorQuda("Unexpected slice %d", t_slice);
```
weinbe2: Extend to `< 0` as well

lib/blas_3d.cu
```diff
@@ -0,0 +1,251 @@
...
+      long long flops() const override { return 14 * x.Volume() * x.Nspin() * x.Ncolor(); }
+      long long bytes() const override { return x.Bytes() + 2 * y.Bytes(); }
+    };
+
+    void caxpby(std::vector<Complex> &a, ColorSpinorField &x, std::vector<Complex> &b, ColorSpinorField &y)
+    {
+      checkPrecision(x, y);
+
+      // Check spins
+      if (x.Nspin() != y.Nspin()) errorQuda("Unexpected number of spins x=%d y=%d", x.Nspin(), y.Nspin());
```
weinbe2: Maybe out of the scope of this PR, but either here or in the future (I can create an issue, let me know) we should create a set of routines `checkSpin`, `checkColor`, `checkSpinColor` that acts similarly to `checkPrecision`, etc.

lib/eig_trlm_3d.cpp
```diff
@@ -0,0 +1,701 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
+#include <iostream>
+#include <vector>
+#include <algorithm>
+
+#include <quda_internal.h>
```
weinbe2, line 8: Should we start implementing using `#include "..."` for headers inside QUDA, `#include <...>` for external headers with this PR? Or should we just do a mega-change at some point in the future?

lib/eig_trlm_3d.cpp
```diff
@@ -0,0 +1,701 @@
...
+      for (int t = 0; t < ortho_dim_size; t++) {
+        if (num_converged_3D[t] >= n_conv) {
+          converged_3D[t] = true; 
+        } else {
+          all_converged = false;
+        }
+      }
+
+      if (getVerbosity() >= QUDA_VERBOSE && comm_coord(0) == 0 && comm_coord(1) == 0 && comm_coord(2) == 0) {
+        printf("iter = %d rank = %d converged = ", restart_iter + 1, comm_rank_global());
```
weinbe2: I see what's going on here, do we want to make a custom `printfQuda` (`printfQuda3d`?) that handles this type of conditional on the print instead of using `printf`? Or should we avoid over-specializing?

In any case---it might be good to modify this to build the entire string before printing so the outputs from different ranks don't get interleaved... or at least minimize the risk.

lib/eig_trlm_3d.cpp
```diff
@@ -0,0 +1,701 @@
...
+    }
+  }
+
+
+  void TRLM3D::eigensolveFromArrowMat3D()
+  {
+    getProfile().TPSTART(QUDA_PROFILE_EIGEN);
+
+    // Loop over the 3D problems
+#ifdef _OPENMP
```
weinbe2: Is there a particular need for this `#ifdef` or can we just get rid of it?

weinbe2: To be clear I understand the purpose of OpenMP parallelism here, my only question is about the guard around it

lib/eig_trlm_3d.cpp
```diff
@@ -0,0 +1,701 @@
...
+      }
+
+      // Block orthogonalise
+      blas3d::caxpby(s_t, vecs[i], unit_new, rvecs[0]);
+    }
+  }
+
+  void TRLM3D::prepareInitialGuess3D(std::vector<ColorSpinorField> &kSpace, int ortho_dim_size)
+  {
+    if (sqrt(blas::norm2(kSpace[0])) == 0.0) {
```
weinbe2: I learned the hard way through interfacing with MILC that we should also check for +/- inf and nan in the context of completely uninitialized data, see here: https://github.com/lattice/quda/blob/develop/lib/eigensolve_quda.cpp#L145-L150

cjmorningstar10:
In  lib/interface_quda.cpp   on line 2577
         GaugeField *cudaGauge = checkGauge(inv_param);
For the 3d laph ev calculation, if use_smeared_gauge is TRUE,  the original gauge configuration is not needed, only the smeared gauge is needed, so the above check can fail if prior tasks do not load the original config.   In this case, only a check of the smeared gauge field is needed.

lib/communicator_qmp.cpp
maddyscientist: I don't believe we have the capability to defer the QMP error handling to QUDA, though that would be desirable.  It's something that I will look into adding to QMP and then propagate enabling it into QUDA.

The motivation in the first place for this is to allow us to do post-error analysis, e.g., print a stack trace if `QUDA_BACKWARDS` is enabled, or anything else we might want to do if an MPI error is thrown.  I found this very helpful when encountering fatal MPI errors.

include/kernels/blas_3d.cuh
maddyscientist: I don't think a check here is merited.  As you note there's already a check in the code that calls this in blas_3d.cu: one can get stuck ingoing deeper into wonderland if you don't trust anything at a given level.....

lib/eig_trlm_3d.cpp
maddyscientist: Yes, confirming this is correct.  We are using multi-blas routines on a pure 3-d grid here, which just works.  It's suboptimal, in that it would be better to have true batched 3-d multi-blas without the copy back and forward, but that will come in a subsequent PR.

lib/blas_3d.cu
maddyscientist: Yeah, I would say that's out of scope, but certainly good to have.  Please create an issue 😉

lib/eig_trlm_3d.cpp
maddyscientist: Let's defer that joy for another day....

lib/eig_trlm_3d.cpp
maddyscientist: Indeed the guard isn't needed, will be nuked.

lib/eig_trlm_3d.cpp
maddyscientist: Good call.  Will use `std::isfinite` here.

maddyscientist: c5b568171

lib/eig_trlm_3d.cpp
maddyscientist: c5b568171

lib/blas_3d.cu
maddyscientist: c5b568171

lib/gauge_laplace.cpp
maddyscientist: c5b568171

include/blas_3d.h
maddyscientist: 1d66a556b

tests/staggered_eigensolve_test.cpp
maddyscientist: 196cef438

tests/staggered_eigensolve_test.cpp
maddyscientist: 196cef438

tests/host_reference/staggered_dslash_reference.h
maddyscientist: 196cef438

tests/host_reference/staggered_dslash_reference.cpp
maddyscientist: 196cef438

tests/host_reference/staggered_dslash_reference.cpp
maddyscientist: 196cef438

include/kernels/blas_3d.cuh
maddyscientist: 3767371511982bbe8a6ec4cb23a09dda501b6272

include/kernels/blas_3d.cuh
maddyscientist: 3767371511982bbe8a6ec4cb23a09dda501b6272

include/eigensolve_quda.h
maddyscientist: [1e2c0c7](https://github.com/lattice/quda/pull/1506/commits/1e2c0c7cf2da19395b92e18dcf4996a38087afd0) does something related: remove code replication and make `computeEvals3D` a specialization of the regular `computeEvals` which is what I was planning on doing anyway.

Marking as resolved, but we can discuss this more if needed.

lib/eig_trlm_3d.cpp
maddyscientist: We already have array min / max comm reduction functions, but that's not what's being used here.  What's here is simpler, and is a scalar reduction, not array reduction.  However, `getArrayMinMax3D` is completely unnecessary, and it's better to just delete it.  1ddda7c54 does this.

lib/eig_trlm_3d.cpp
maddyscientist: I'd steer away from `printfQuda3d` but your point about building the entire string first is a good one.

maddyscientist: 030775807

lib/blas_3d.cu
maddyscientist: I just went and did it dfd36e754

lib/tune.cpp
maddyscientist: I've confirmed that `std::map::merge` does do what we want.  When `split_tc[I]` goes out scope there is no issue.  The text regarding `..the internal pointers of the container nodes are repointed...` concern the underlying container of the respective maps.  While a `std::map` doesn't use a (doubly) linked list (will use a red-black tree instead), it's instructive to think in these terms: we splice in various entries by updating the appropriate next and prev pointers of adjacent entries in the list we're inserting an entry into.  No data is moved, only the pointers updated.

maddyscientist:
> In lib/interface_quda.cpp on line 2577 GaugeField *cudaGauge = checkGauge(inv_param); For the 3d laph ev calculation, if use_smeared_gauge is TRUE, the original gauge configuration is not needed, only the smeared gauge is needed, so the above check can fail if prior tasks do not load the original config. In this case, only a check of the smeared gauge field is needed.

Thanks @cjmorningstar10, 3ae79cc61 takes care of this.

maddyscientist:
> Looking at the host verify... at this point I think we should be splitting the Laplace verification routines out from the staggered ones at this point, if anything for code cleanliness/maintainability reasons. Alternatively, there needs to be many more checks for invalid combinations: `matpc` and `matdagmat` don't make sense for the 3-d Laplace operator, `laplace3D` doesn't make sense for the staggered operators... I'm sure I could find more cases of inconsistencies if I wanted to.

Given the similarity between the staggered and Laplace kernels (they are essentially identical), I would push back on this, at least in the scope of this PR.  I have, however, added the consistency checks you suggest (0e906a6d89c4bb73659ac99c9e9cd002f93dc1df).

maddyscientist:
@weinbe2 I believe your points are addressed, and I have added the final few things I wanted to get done: mainly adding Laplace TRLM-3d to the ctest list.

include/blas_3d.h
```diff
@@ -0,0 +1,80 @@
...
+#include <color_spinor_field.h>
+
+namespace quda
+{
+
+  namespace blas3d
+  {
+
+    // Local enum for the 3D copy type
+    enum copy3dType { COPY_TO_3D, COPY_FROM_3D, SWAP_3D };
```
weinbe2: I've been trying to use `enum class` for internal (i.e., non-`quda_enum.h`) enums, I think we should apply that here.

maddyscientist: [6383489](https://github.com/lattice/quda/pull/1506/commits/63834893048dbee7acee91f788bbc14003d933cc)

include/color_spinor_field.h
```diff
@@ -1054,6 +1056,38 @@ namespace quda
   */
   void spinorDistanceReweight(ColorSpinorField &src, double alpha0, int t0);
 
+  /**
+     @brief Helper function for determining if the spin of the fields is the same.
+     @param[in] a Input field
+     @param[in] b Input field
+     @return If spin is unique return the number of spins
+   */
+  inline int Spin_(const char *func, const char *file, int line, const ColorSpinorField &a, const ColorSpinorField &b)
```
weinbe2, line 1065: Should we add a version of this for vector_ref/cvector_ref/etc of this as well?

maddyscientist: I would say no.  The `Nspin()` methods of `cvector_ref<ColorSpinorField>` themselves already test internal for consistency.  I don't think we need to go the length of comparisons between arbitrary combinations of fields and cvector_refs of fields.

lib/communicator_stack.cpp
```diff
@@ -49,8 +50,16 @@ namespace quda
...
+
+    // if reverting to global, we will need to join the tunecaches
+    bool join_tune_cache = split_key == default_comm_key;
```
weinbe2, line 56: Shouldn't we theoretically need to do this each time we go to a `split_key` with a smaller `[key].product()`?

maddyscientist: Sorry, not following.  What do you mean?

weinbe2: If there are recursive split-grid calls, i.e., further splitting an already split grid.

maddyscientist: Ah, gotcha now.  That is precluded and recursive split grids are not supported.

weinbe2:
I think we should add a test of the 3-d reduction routines to `blas_test`, but in the interest of getting this PR merged I'm okay with that being added to a later PR, just file an issue about it :)

lib/communicator_stack.cpp
maddyscientist: The way to do it would be with a `std::stack` I guess.

maddyscientist:
> I think we should add a test of the 3-d reduction routines to `blas_test`, but in the interest of getting this PR merged I'm okay with that being added to a later PR, just file an issue about it :)

That sounds good.  Eventually I'm going to nuke these routines entirely with them to be replaced by the domain-decomposed bias that I've threatened to do on the Domain Decomposition calls.

tests/utils/host_blas.cpp
```diff
@@ -63,20 +63,20 @@ void mxpy(void *x, void *y, int len, QudaPrecision precision)
 }
 
 // returns the square of the L2 norm of the vector
-template <typename real_t> inline double norm2(real_t *v, int len)
+template <typename real_t> inline double norm2(real_t *v, int len, bool global)
 {
   double sum = 0.0;
   for (int i = 0; i < len; i++) sum += v[i] * v[i];
-  quda::comm_allreduce_sum(sum);
+  if (global) quda::comm_allreduce_sum(sum);
```
weinbe2, line 70: I need to think about it a bit, but I think that for the context this is used in (3-d laplace eigensolvers) you don't want to outright disable the full reduction, you only want to reduce in the `x/y/z` plane?

tests/utils/set_params.cpp
```diff
@@ -130,11 +134,11 @@ void setInvertParam(QudaInvertParam &inv_param)
   if (kappa == -1.0) {
     inv_param.mass = mass;
     inv_param.kappa = 1.0 / (2.0 * (1 + 3 / anisotropy + mass));
-    if (dslash_type == QUDA_LAPLACE_DSLASH) inv_param.kappa = 1.0 / (8 + mass);
+    if (dslash_type == QUDA_LAPLACE_DSLASH) inv_param.kappa = 1.0 / ((laplace3D == 3 ? 6 : 8) + mass);
```
weinbe2: I saw elsewhere in the code that `(laplace3D == 3)` was replaced with something more like `(laplace3D < 4)` or `(laplace3D != -1)` or something like that, I forget the convention; I preferred that to the hard compare against `3`, and it seems safer.

maddyscientist: The less than convention was there to give the option for future X/Y/Z orthogonal dimension.  Good to make it consistent though.

tests/utils/host_blas.cpp
maddyscientist: The disablement of the global reduction was intentional here, as the point was to defer the global reduction until later and do it once in a padded structure to allow for a single 4-d all reduce to get the correct answer.  Looks like I didn't actually do this properly so I'll fix this.

weinbe2 (approved):
Compiling against MILC looks good, as does testing MG. This is good to go (pending reconciling the testing failures with the latest changes). Awesome work @maddyscientist and @cpviolator

tests/utils/set_params.cpp
maddyscientist: 8fb71814ce06ce0a626b6a7c3ec7deae024e054a

tests/utils/host_blas.cpp
maddyscientist: 38b1610d52b2503ce467ed1a706e394eda8d8fa1

maddyscientist:
We should be good now for this PR.  We can merge this once CI completes 🥳

maddyscientist:
cscs-ci run

PR #1507, opened 2024-10-23
hotfix/solver-normalization

SaltyChiang:
The solver normalization is incorrect as we don't actually change values in `nb_inv` and `nb`. Using the reference of the vector element instead.


maddyscientist (approved):
Thanks for this fix @SaltyChiang

maddyscientist:
@Jenkins test this please

PR #1509, opened 2024-11-01
Fix MG verify issue in some specific cases. 

SaltyChiang:
`MG::verify` will be called in the `updateMultigridQuda` function if `run_verify` and `thin_update_only` are true. There is an `M` call here, and the input parameter `r_coarse` will not be a length 1 vector if the MRHS solver has been used before, which causes a segfault because of the mismatch of input and output lengths.

The PR also contains some minor updates:
- Don't try to ensure the input gauge field order is `QUDA_MILC_GAUGE_ORDER` in the `computeHISQForceQuda` function. We have already made an app to use the function with `QUDA_QDP_GAUGE_ORDER`.
- Add the hash for the Eigen tarball for machines without network connection.
- A typo.


maddyscientist (approved):
Thanks @SaltyChiang

PR #1515, opened 2024-11-08
Feature/mrhs misc

maddyscientist:
This PR is a bit of a catch all
* Adds register talking for MRHS staggered dslash
   * Level of register tiling is controlled by a CMake parameter `QUDA_MAX_MULTI_RHS_TILE`, with the default left at size 1 for now.
   * This feature will be further developed in subsequent PRs
   * (Although not included in this PR, it's straightforward to this support to other stencils)
* Update to the latest version of Eigen on the 3.4 branch
   * This has improved support for nvc++ allowing us to remove some prior WARs
* Fixes performance regressions of the MMA dslash when the memory pool is switched off
   * The `FieldTmp` now supports creating temporaries using parameters as opposed another field instance
   * We use this to create the temporary used for the reordered quark fields
 * Adds WAR for performance regressions with ROCm
   * This improves performance on ROCm 5.3 by 30% for the Laplace 3-d operator, though it's still off by integer factors 
 * Various fixes for nvc++ compilation
 * Add alternative sentinel for heterogeneous reductions in the case that the compiler optimizes away non-finite math (enabled with `QUDA_HETEROGENEOUS_ATOMIC_INF_INIT=OFF`).  Not a problem by default, but is with latest clang with `-Ofast`.
 * Fix various compiler warnings with more recent compilers, e.g., gcc-15
 * Fixes a hang caused by process divergence when calling `printGenericMatrix`


include/kernels/laplace.cuh
```diff
@@ -143,8 +143,9 @@ namespace quda
     static constexpr const char *filename() { return KERNEL_FILE; } // this file name - used for run-time compilation
 
     template <KernelType mykernel_type = kernel_type>
-    __device__ __host__ __forceinline__ void operator()(int idx, int src_idx, int parity)
+    __device__ __host__ __forceinline__ void operator()(int idx, int src_idx_, int parity)
     {
+      int src_idx = MAX_MULTI_RHS == 1 ? 0 : src_idx_;
```
hummingtree, line 148: Why does the `MAX_MULTI_RHS == 1` case need special treatment? When `MAX_MULTI_RHS == 1`, wouldn't `src_idx` always be equal to `0`?

lib/inv_mr_quda.cpp
```diff
@@ -82,10 +82,10 @@ namespace quda
 
     int iter = 0;
     int step = 0;
-    bool converged = false;
+    bool is_done = false;
```
hummingtree, line 85: Is there a reason for this name change other than ... making it clear?

include/kernels/dslash_staggered.cuh
```diff
@@ -104,13 +103,18 @@ namespace quda
         if (doHalo<kernel_type>(d) && ghost) {
           const int ghost_idx = ghostFaceIndexStaggered<1>(coord, arg.dim, d, 1);
           const Link U = arg.improved ? arg.U(d, coord.x_cb, parity) : arg.U(d, coord.x_cb, parity, StaggeredPhase(coord, d, +1, arg));
-          Vector in = arg.halo.Ghost(d, 1, ghost_idx + src_idx * arg.nFace * arg.dc.ghostFaceCB[d], their_spinor_parity);
-          out = mv_add(U, in, out);
+          for (auto s = 0; s < n_src_tile; s++) {
```
hummingtree, line 107: Should `#pragma unroll` be added here? Although I would assume the compiler does this by itself already.

lib/inv_mr_quda.cpp
maddyscientist: `converged` is a misnomer, since it can terminate if it runs out of iterations and not be converged.  `is_done` seems more appropriate.

include/kernels/laplace.cuh
maddyscientist: It's telling the compiler that the index is guaranteed to be 0, as opposed to use knowing at run-time the value is 0.  Helps with less good compilers 😉

include/kernels/dslash_staggered.cuh
maddyscientist: Probably a good idea.  I'll add that on next push.

maddyscientist: [22393c5](https://github.com/lattice/quda/pull/1515/commits/22393c5a893b20a99b85b1f92dc86e32ef1c1d40)

lib/targets/cuda/target_cuda.cmake
```diff
@@ -103,8 +103,11 @@ if(CMAKE_CUDA_COMPILER_ID MATCHES "NVIDIA" OR CMAKE_CUDA_COMPILER_ID MATCHES "NV
 endif()
 cmake_dependent_option(QUDA_HETEROGENEOUS_ATOMIC "enable heterogeneous atomic support ?" ON
                        "QUDA_HETEROGENEOUS_ATOMIC_SUPPORT" OFF)
+cmake_dependent_option(QUDA_HETEROGENEOUS_ATOMIC_INF_INIT "use infinity as the sentinal for heterogeneous atomic support?" ON
```
weinbe2: `sential` -> `signal`?

maddyscientist: no, it should be `sentinel` which it seems I can't spell

CMakeLists.txt
```diff
@@ -216,6 +216,11 @@ if(QUDA_MAX_MULTI_BLAS_N GREATER 32)
   message(SEND_ERROR "Maximum QUDA_MAX_MULTI_BLAS_N is 32.")
 endif()
 
+set(QUDA_MAX_MULTI_RHS_TILE "1" CACHE STRING "maximum tile size for MRHS kernels")
```
weinbe2: I'd add a comment that this currently only applies to staggered-type dslash kernels, just for sanity's sake

include/kernels/dslash_pack.cuh
```diff
@@ -217,12 +226,16 @@ namespace quda
 
     if (face_num == 0) { // backwards
       int idx = indexFromFaceIndexStaggered<4, QUDA_4D_PC, dim, nFace, 0>(ghost_idx, parity, arg);
-      Vector f = arg.in[src_idx](idx, spinor_parity);
-      arg.halo_pack.Ghost(dim, 0, ghost_idx + src_idx * nFace * arg.dc.ghostFaceCB[dim], spinor_parity) = f;
+      for (auto s = 0; s < n_src_tile; s++) {
```
weinbe2: Is there a specific reason why you opted for `for` loop from `0 -> n_src_tile` here as opposed to `src -> src + n_src_tile` in `pack`? I think the convention should be consistent between the routines; I have no particular preference for which convention gets used.

weinbe2 (approved):
Pending a few cosmetic requests this looks good. Thanks @maddyscientist !

lib/targets/cuda/target_cuda.cmake
maddyscientist: a860b1cfe86ed079cdaef3352f867d35b8dcc86f

CMakeLists.txt
maddyscientist: c3fbf502e6f9fbcbd52d86da68e8497ada169839

include/kernels/dslash_pack.cuh
maddyscientist: c3fbf502e6f9fbcbd52d86da68e8497ada169839

PR #1516, opened 2024-11-17
add printing gauge field for debugging

jcosborn:
add printing gauge field for debugging
initialize `dslash_type_precondition`
fix CMake to use `QUDA_DIRAC_LAPLACE` instead of old `QUDA_LAPLACE`

### DIFF
--- a/lib/color_spinor_util.in.cu
+++ b/lib/color_spinor_util.in.cu
@@ -329,7 +329,7 @@ namespace quda {
       printf("rank = %d x = %u, s = %d, { ", comm_rank(), x_cb, s);
       for (int c = 0; c < o.Ncolor(); c++) {
         auto value = complex<double>(o(parity, x_cb, s, c));
-        printf("(%f,%f) ", value.real(), value.imag());
+        printf("(%g,%g) ", value.real(), value.imag());
       }
       printf("}\n");
     }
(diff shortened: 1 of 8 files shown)


lib/color_spinor_util.in.cu
```diff
@@ -340,7 +340,7 @@ namespace quda {
   {
     if (a.isNative()) {
       constexpr auto order = colorspinor::getNative<Float>(nSpin);
-      print_vector(FieldOrderCB<double, nSpin, nColor, 1, order, Float, Float, false, true>(a), parity, x_cb);
+      print_vector(FieldOrderCB<double, nSpin, nColor, 1, order, Float, Float, false, false>(a), parity, x_cb);
```
maddyscientist: So this last template parameter determines whether we're doing block float or not.  Setting this to false will mean we can print half precision globally normalized fields but not block normalized fields.  What was the use case for setting this to `false`: I think it's only relevant for multigrid null space storage.

maddyscientist (approved):
Thanks @jcosborn.  Just one query regarding `print_vector` template instantiation, but apart from that this is good to merge.

maddyscientist:
cscs-ci run

lib/color_spinor_util.in.cu
jcosborn: I was just printing vectors in CG and needed to change it to get correct output.

maddyscientist: For staggered fermions I presume?

maddyscientist: and in half precision or double/single?

jcosborn: Yes, staggered.  I was looking at both half and single vectors and don't recall the output before the change, but my main focus was on half precision.

maddyscientist: Ok, I've reproduced the garbage with half precision staggered.  Other precisions and Wilson is working fine.  Investigating.....

maddyscientist: Bug located, and it was specifically for the case of half-precision staggered.  I'll push the fix to this PR and revert the change to color_spinor_util.in.cu.

maddyscientist: [bff3830](https://github.com/lattice/quda/pull/1516/commits/bff3830b2efa1d2a67567855179c00cb11640f3a) fixes the issue.  With that now done, we can merge this PR once CI completes.

PR #1517, opened 2024-11-18
Update inv_mr_quda.cpp

pittlerf:
The convergence check of the MR solver was recently updated (9002399d2ba37a1ac6d8bbe5d08723ff5836cfc9) and in the new version, the solver does not go out when the max. number of iterations is exceeded.

This resulted in our MG solver getting stucked in the smoother.


maddyscientist (approved):
Thanks for the fix @pittlerf!

PR #1518, opened 2024-11-18
Fix HIP debug flags to use -G only for NVIDIA compilers

jakurzak:
- Updated CMake configuration to set `-g -G` only when the HIP compiler is NVIDIA (`nvcc`).
- For Clang-based HIP compilers, use `-g` alone as it supports both host and device debugging.
- Improved flag descriptions in CMake cache to reflect the specific behavior of each compiler.
- This ensures compatibility and avoids unsupported flag errors with Clang.


maddyscientist (approved):
Thanks @jakurzak

PR #1522, opened 2024-12-10
Modifications to add use_smeared_gauge to InvertParam

cjmorningstar10:
Added use_smeared_gauge to QudaInvertParam  to get rid of awkward extra bool in some routines.  In setDiracEigParam(...), added code to allow this to run without needing gaugePrecise having already been set.   Some first tests have been done, but would like feedback on how this was done!   These changes facilitate computation of the LapH eigenvectors using the smeared gauge field.

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -464,8 +464,8 @@ extern "C" {
     /** The t0 parameter for distance preconditioning, the timeslice where the source is located */
     int distance_pc_t0;
 
-    /** Whether to use the smeared gauge field for the Dirac operator
-        for whose eigenvalues are are computing. */
+    /** Whether to use the smeared gauge field for the Dirac operator, usually
+        when defined as a spatial Laplacian: mainly used in computing Laplacian eigenvectors */
     QudaBoolean use_smeared_gauge;
 
   } QudaInvertParam;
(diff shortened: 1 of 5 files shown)


include/quda.h
```diff
@@ -507,7 +511,7 @@ extern "C" {
 
     /** Whether to use the smeared gauge field for the Dirac operator
         for whose eigenvalues are are computing. */
-    bool use_smeared_gauge;
+    QudaBoolean use_smeared_gauge;
```
maddyscientist: Why the change to `QudaBoolean` from `bool`?  I intentionally used `bool` with this variable as we intend to deprecate `QudaBoolean`.

maddyscientist: Also I think with this addition of the `use_smeared_gauge` parameter to `QudaInvertParm` the `QudaEigParam` variant is never used now?  Can we just delete it?

maddyscientist (commented):
@cjmorningstar10 do these changes fix the issues that were being seen?

maddyscientist:
cscs-ci run

include/quda.h
cjmorningstar10: I used QudaBoolean because all the other boolean members did...it looked weird to see a difference.  But bool is fine.  The use_smeared_gauge is needed in the InvertParam  because only that gets sent to the setDiracParam functions, where it is needed.  It is ok to keep it in the EigParam too.  I think I put in a check that these two bools have to be the same.

cjmorningstar10:
> @cjmorningstar10 do these changes fix the issues that were being seen?

No.   The main reason for this PR is to enable computing the LapH eigenvectors if you read the smeared gauge and NOT the original gauge.   Once the perambulators are computed, the original gauge field is never needed any more.   When we start evaluating correlator functions,  we will just read the smeared gauge, compute the Laph evs on the fly, then start computing baryon triplets/meson doublets/ then the correlators.   Even displacements in the operators need only the smeared gauge field. No need to read the original gauge, smear it, then ....    Just cuts out one step.  No need to waste device mem on the original gauge.    This is a minor issue,  but .... why not support this choice?

PR #1524, opened 2024-12-11
replace use of std math in kernels with quda versions

jcosborn:
replace use of std math in kernels with quda versions
allow building without cufft
change some casting

### DIFF
--- a/include/targets/cuda/math_helper.cuh
+++ b/include/targets/cuda/math_helper.cuh
@@ -2,7 +2,6 @@
 
 #include <cmath>
 #include <target_device.h>
-#include <quda_cuda_api.h>
 
 #if (CUDA_VERSION >= 11070) && !defined(_NVHPC_CUDA)
 #define BUILTIN_ASSUME(x) \
@@ -27,6 +26,7 @@ namespace quda {
   template <typename T> inline __host__ __device__ T acos(const T a) { return ::acos(a); }
   template <typename T> inline __host__ __device__ T pow(const T a, const int b) { return ::pow(a, b); }
   template <typename T> inline __host__ __device__ T pow(const T a, const T b) { return ::pow(a, b); }
+  template <typename T> inline __host__ __device__ T fmod(const T a, const T b) { return ::fmod(a, b); }
 
   /**
    * @brief Maximum of two numbers
--- a/include/targets/cuda/math_helper.cuh
+++ b/include/targets/cuda/math_helper.cuh
@@ -2,6 +2,7 @@
 
 #include <cmath>
 #include <target_device.h>
+#include <quda_cuda_api.h>
 
 #if (CUDA_VERSION >= 11070) && !defined(_NVHPC_CUDA)
 #define BUILTIN_ASSUME(x) \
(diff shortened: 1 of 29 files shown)


include/targets/hip/shared_memory_helper.h
```diff
@@ -72,8 +72,11 @@ namespace quda
     /**
        @brief Constructor for SharedMemory object.
     */
-    constexpr SharedMemory() : data(cache(get_offset(target::block_dim()))) { }
+    __device__ __host__ constexpr SharedMemory() : data(cache(get_offset(target::block_dim()))) { }
```
maddyscientist: Why is this needed?

include/targets/cuda/shared_memory_helper.h
```diff
@@ -72,8 +72,11 @@ namespace quda
     /**
        @brief Constructor for SharedMemory object.
     */
-    constexpr SharedMemory() : data(cache(get_offset(target::block_dim()))) { }
+    __device__ __host__ constexpr SharedMemory() : data(cache(get_offset(target::block_dim()))) { }
```
maddyscientist: Why is this needed?

include/targets/generic/mrg32k3a.h
```diff
@@ -191,14 +192,14 @@ namespace quda
       constexpr double uniform(MRG32k3a &prn)
       {
         double p1 = a12 * (double)(prn.s1[1]) - a13n * (double)(prn.s1[0]);
-        p1 = std::fmod(p1, m1);
+        p1 = fmod(p1, m1);
```
maddyscientist, line 195: Confused here.  There is no `quda::fmod` function in the math_helper, so why making this change?

include/targets/cuda/math_helper.cuh
```diff
@@ -2,6 +2,7 @@
 
 #include <cmath>
 #include <target_device.h>
+#include <quda_cuda_api.h>
```
maddyscientist: Why did this header need to be added here?

maddyscientist (commented):
Looks good James.  I've left a few questions.

Perhaps I forget the context, but what is the goal to allow building without cuFFT on CUDA?  Does the same need to be enabled for the HIP target?

include/targets/hip/shared_memory_helper.h
jcosborn: Not sure why I added that.  I've removed it now (21d2cadf0cd2a4c23e4131b6bad0a3e92c9f7dc3).

include/targets/cuda/shared_memory_helper.h
jcosborn: Not sure why I added that.  I've removed it now (21d2cadf0cd2a4c23e4131b6bad0a3e92c9f7dc3).

include/targets/generic/mrg32k3a.h
jcosborn: You're right.  I think SYCL on CUDA was still working without the quda version.  For consistency though we should have the quda version so I've add that in (21d2cadf0cd2a4c23e4131b6bad0a3e92c9f7dc3).

include/targets/cuda/math_helper.cuh
jcosborn: Not sure why, but there probably was a reason.  I've reverted for now and will revisit if I encounter it again.
21d2cadf0cd2a4c23e4131b6bad0a3e92c9f7dc3

jcosborn:
> Perhaps I forget the context, but what is the goal to allow building without cuFFT on CUDA? Does the same need to be enabled for the HIP target?

This was for testing with LLVM on Polaris.  By default, it didn't find the libraries, and it seemed easier to just disable them than try to point it in the right direction.  Disabling worked fine for blas, but wasn't fully implemented for fft.
I don't think there is a similar need for HIP at the moment.  It could be attempted now for completeness, or just wait until there is a test case.

maddyscientist (approved):
Approved, but one final question: I see you've restore the `constexpr` but you still have the `__device__ __host__` qualifiers on the shared memory constructor.  Were there needed for some compiler?

jcosborn:
I think you mean in the `sharedMem` accessor?  I just reverted that too.
f9a551aac4397ef6975f59d64d55bc4087398ea1

maddyscientist:
cscs-ci run

PR #1526, opened 2024-12-15
Feature/adjoint flow

rkarur:
Implemented the adjoint fermion flow  algorithm (both the serial and hierarchical versions detailed in Appendix E of https://arxiv.org/pdf/1302.5246) and wrote a test that tests both the adjoint and regular (forward) gradient flow, in which the equivalence of the serial and hierarchical methods is demonstrated, as well as a numerical verification that the forward and adjoint fermionic flows are indeed adjoint methods of one another (Identity E.4 in https://arxiv.org/pdf/1302.5246)


maddyscientist:
@cpviolator can you review this when you get the chance?

lib/check_params.h
```diff
@@ -1159,6 +1159,18 @@ void printQudaGaugeSmearParam(QudaGaugeSmearParam *param)
 #if defined CHECK_PARAM
   if (param->struct_size != (size_t)INVALID_INT && param->struct_size != sizeof(*param))
     errorQuda("Unexpected QudaGaugeSmearParam struct size %lu, expected %lu", param->struct_size, sizeof(*param));
+    
+  if (param->n_steps <= param->adj_n_save ) {
+      
+      logQuda(QUDA_SUMMARIZE,"Not good practice to adj_n_save (%d) >= n_steps (%d); adj_n_save manually altered: \n",param->n_steps,param->adj_n_save);
```
cpviolator: This is a neat idea, but it can cause issues further down production pipelines. It would be better to issue a `warningQuda` and leave the values unaltered.

rkarur: got it, I issued an `errorQuda` instead just to be sure, lmk if that is better.

cpviolator:
@rkarur This is a great PR! I took a look over the file diff and now I'm spinning up to see how it goes. FYI, there is a PR parallel to this [here](https://github.com/lattice/quda/pull/1555) that significantly changes the underlying code for gauge smearing, without changing the final result. For example, the 3 step Runge-Kutta integrator can now be 6 step if desired, keeping the 3 step functionality in place. I would like to use your test routines in `su3_fermion_update.cpp` to check that everything is working as it should after the 6 step upgrade by running tests in this PR, merging in new code, then repeating tests. This will require a bit of a rewrite as the current W1, W2, VT enums used in develop will be replaced by more general F1...F6 enums. I'm more than happy to do the rewrite given the effort already put in by both authors of both PRs, and neither need be burdened with extra effort for my convenience.

cpviolator:
@rkarur When you get time, can you please cook up a typical command line execution of your `su3_fermion_test`? There's just one or two things that need cleaning up in there and I want to make sure I'm using it as intended.

rkarur:
hi @cpviolator , I would do something like (the dirac op parameters are not necessary anyhow but Im copying and pasting):

```
EXEC=$SOME_PATH/quda/bin/su3_fermion_test

srun $EXEC  --dim 8 8 8 8 --Lsdim 1 --gridsize 1 1 1 2 \
  --dslash-type clover  --mass 0 --mass-normalization mass --clover-coeff 1.0  \
  --prec double --prec-sloppy double \
  --inv-type cgne --solution-type mat --solve-type normop-pc \
  --tol 1.0e-14 --verbosity verbose --su3-smear-type wilson --su3-smear-steps 100 --su3-adj-gauge-nsave 10 --su3-hier-threshold 6 --su3-smear-epsilon 0.1
```

lib/interface_quda.cpp
```diff
@@ -5424,6 +5430,464 @@ void performGFlowQuda(void **h_out, void **h_in, QudaInvertParam *inv_param, Qud
...
+  gParam.reconstruct = QUDA_RECONSTRUCT_NO; // temporary field is not on manifold so cannot use reconstruct
+  GaugeField gaugeTemp(gParam);
+
+  const GaugeField gin = *gaugeSmeared;
+  GaugeField &g_W0 = *gaugeSmeared;
+  GaugeField &g_W1 = gaugeW1;
+  GaugeField &g_W2 = gaugeW2;
+  GaugeField &g_VT = gaugeVT;
+
+  //necessary?
```
cpviolator: Did you get an answer to this?

lib/interface_quda.cpp
```diff
@@ -5424,6 +5430,464 @@ void performGFlowQuda(void **h_out, void **h_in, QudaInvertParam *inv_param, Qud
...
+  pushOutputPrefix("performAdjGFlowQudaHier: ");
+  checkGaugeSmearParam(smear_param);
+
+  if (smear_param->n_steps <= smear_param->adj_n_save ) {
+      
+      errorQuda("Not good practice to have adj_n_save (%d) >= n_steps (%d); adj_n_save should be manually altered to min(nsteps, %d): \n",smear_param->n_steps,smear_param->adj_n_save,smear_param->n_steps - 1);
+      
+  }
+    
+  // pushVerbosity(inv_param->verbosity);
```
cpviolator: delete this if not needed

lib/interface_quda.cpp
```diff
@@ -5424,6 +5430,464 @@ void performGFlowQuda(void **h_out, void **h_in, QudaInvertParam *inv_param, Qud
...
+      for (int j = 0; j < (int) hier_list.size(); j++ ){
+          logQuda(QUDA_DEBUG_VERBOSE,"%d \n",(int) hier_list[j]);
+      }
+      logQuda(QUDA_DEBUG_VERBOSE,"\n");
+      
+      for (unsigned int j = 0; j < (unsigned int) hier_list[ret_idx]; j++){
+          if (j > 0) std::swap(g_2,g_1);
+          WFlowStep(g_2, gaugeTemp, g_1, smear_param->epsilon, smear_param->smear_type);
+      }
+      // break;
```
cpviolator: Still needed?

lib/interface_quda.cpp
```diff
@@ -5424,6 +5430,464 @@ void performGFlowQuda(void **h_out, void **h_in, QudaInvertParam *inv_param, Qud
...
+
+  // initialize a and b for Laplace operator
+  double a = 1.;
+  double b = -8.;  
+  
+  int comm_dim[4] = {};
+  // only switch on comms needed for directions with a derivative
+  for (int i = 0; i < 4; i++) { comm_dim[i] = comm_dim_partitioned(i); }
+    
+  // f_temp3 = fin;
```
cpviolator: Still need this?

lib/interface_quda.cpp
```diff
@@ -5424,6 +5430,464 @@ void performGFlowQuda(void **h_out, void **h_in, QudaInvertParam *inv_param, Qud
...
+    copyExtendedGauge(precise, g_W0, QUDA_CUDA_FIELD_LOCATION);
+    precise.exchangeGhost();  
+    ApplyLaplace(f_temp0, f_temp4, precise, 4, a, b, f_temp4, parity, comm_dim, profile); 
+    
+    // [0] <- 1/4 eps [0]; [0] <- [2] + [0]; [0] <- [1] + [0]
+    blas::ax(smear_param->epsilon * 1. / 4., f_temp0);
+    blas::axpy(1.,f_temp2, f_temp0);
+    blas::axpy(1.,f_temp1, f_temp0);
+    
+    // fout = f_temp0;
```
cpviolator: And this?

lib/interface_quda.cpp
```diff
@@ -5424,6 +5430,464 @@ void performGFlowQuda(void **h_out, void **h_in, QudaInvertParam *inv_param, Qud
...
+    
+    int diff = n_save - current_size;
+    
+    
+    for (int i=current_size - 1; i >= 0; --i){
+
+        if (hier_list[i] > threshold){
+            
+            temp_list = get_hier_list(hier_list[i], n_b, diff+1,false);
+            // for (int ii = 0; ii< temp_list.size();ii++) printf("tempf %d \n",temp_list[ii]);
```
cpviolator: If you want to keep debug prints like this, best to put them inside a boolean on debug verbosity

tests/CMakeLists.txt
```diff
@@ -199,6 +199,11 @@ target_link_libraries(su3_test ${TEST_LIBS})
 quda_checkbuildtest(su3_test QUDA_BUILD_ALL_TESTS)
 install(TARGETS su3_test ${QUDA_EXCLUDE_FROM_INSTALL} DESTINATION ${CMAKE_INSTALL_BINDIR})
 
+add_executable(su3_fermion_test su3_fermion_test.cpp)
```
cpviolator: If you're test explicitly requires the Laplace operator, you should guard this around the LAPLACE build definition like [here](https://github.com/lattice/quda/blob/fcd462cc1811f6338d0518771c254918b14b38d3/tests/CMakeLists.txt#L93)

tests/su3_fermion_test.cpp
```diff
@@ -0,0 +1,338 @@
...
+  smear_param.epsilon = gauge_smear_epsilon;
+  smear_param.alpha1 = gauge_smear_alpha1;
+  smear_param.alpha2 = gauge_smear_alpha2;
+  smear_param.alpha3 = gauge_smear_alpha3;
+  smear_param.dir_ignore = gauge_smear_dir_ignore;
+
+
+  quda::ColorSpinorField check,check_safe,check_hier,check_fwd;  
+  QudaInvertParam invParam = newQudaInvertParam();
+  invParam.cpu_prec = QUDA_DOUBLE_PRECISION;
```
cpviolator, line 108: Can these precision options be exposed to the user? Also, you use the `inv_param` structure to set verbosity levels in the interface code, so it's defaulting to `SILENT` regardless of the command line input.

rkarur: Notes: cpu_prec hard coded, the cuda prec is set to `prec`

rkarur:
Need to sort out correct profile timings

lib/interface_quda.cpp
rkarur: i chose to keep it in

rkarur:
refined errors tests

cpviolator:
@maddyscientist This looks pretty goof from my perspective, if you give it the once over for @rkarur we can merge and I can refactor into classes as a separate PR.

tests/CMakeLists.txt
```diff
@@ -199,6 +199,13 @@ target_link_libraries(su3_test ${TEST_LIBS})
 quda_checkbuildtest(su3_test QUDA_BUILD_ALL_TESTS)
 install(TARGETS su3_test ${QUDA_EXCLUDE_FROM_INSTALL} DESTINATION ${CMAKE_INSTALL_BINDIR})
 
+if(QUDA_DIRAC_LAPLACE)
```
weinbe2, line 202: I believe this also needs to be guarded around making sure some Wilson-type (nSpin = 4) dslash:
```
if(QUDA_DIRAC_WILSON
   OR QUDA_DIRAC_CLOVER
   OR QUDA_DIRAC_TWISTED_MASS
   OR QUDA_DIRAC_TWISTED_CLOVER
   OR QUDA_DIRAC_NDEG_TWISTED_MASS
   OR QUDA_DIRAC_NDEG_TWISTED_CLOVER
   OR QUDA_DIRAC_DOMAIN_WALL)
   ```

Because this test assumes an nSpin=4 field. I can pop that in, just noting it here for now.

**Comment by weinbe2** [comment]
The test executable `./su3_fermion_test` doesn't run out of the box with no arguments passed in:

```
> mpirun -np 1 ./su3_fermion_test
[...]
cublasCreated successfully
ERROR: Smear type 1 not supported - only Wilson or Symanzik supported (rank 0, host ipp1-2318.nvidia.com, su3_fermion_test.cpp:147 in int main(int, char**)())
       last kernel called was (name=cudaMemcpyDeviceToHost,volume=bytes=47775744,aux=copy,gauge_field.cpp,1024)
[...]
```

Smear type "1" is stout smearing, which is the default specified in the command_line_params files. Should we change the default there, or manually override it in `./su3_fermion_test`?

**Comment by weinbe2** [comment]
This also fails an error check after manually specifying `wilson`-type smearing:

```
> mpirun -np 1 ./su3_fermion_test --su3-smear-type wilson --verbosity verbose
[...]
Fractional error of (<check,adj_check> - <check,fwd_check>.conj()) = 2.82863e-14
adjoint safe/hier match precision passed!
ERROR: fractional error precision failed
 (rank 0, host ipp1-2318.nvidia.com, su3_fermion_test.cpp:279 in int main(int, char**)())
       last kernel called was (name=cudaMemcpyDefault,volume=bytes=63700992,aux=copy,color_spinor_field.cpp,498)
```

Symanzik has the same issue.

**Comment by weinbe2** [comment]
Further, it looks like setting the precision with `--prec double` vs `--prec single` doesn't do anything---it seems like this is intentional?

Lines 75 to 80 of the test executable:
```
  QudaGaugeParam gauge_param = newQudaGaugeParam();
  if (prec_sloppy == QUDA_INVALID_PRECISION) {
    prec = QUDA_DOUBLE_PRECISION;
    prec_sloppy = prec;
  }
  if (link_recon_sloppy == QUDA_RECONSTRUCT_INVALID) link_recon_sloppy = link_recon;
```

If there's a reason for this that's fine, but there should be a `warningQuda` noting so, and maybe some other warning or error deeper down if there's a fundamental concern about running this in single precision.

maddyscientist:
> Further, it looks like setting the precision with `--prec double` vs `--prec single` doesn't do anything---it seems like this is intentional?
> 
> Lines 75 to 80 of the test executable:
> 
> ```
>   QudaGaugeParam gauge_param = newQudaGaugeParam();
>   if (prec_sloppy == QUDA_INVALID_PRECISION) {
>     prec = QUDA_DOUBLE_PRECISION;
>     prec_sloppy = prec;
>   }
>   if (link_recon_sloppy == QUDA_RECONSTRUCT_INVALID) link_recon_sloppy = link_recon;
> ```
> 
> If there's a reason for this that's fine, but there should be a `warningQuda` noting so, and maybe some other warning or error deeper down if there's a fundamental concern about running this in single precision.

This was a bug: fixed in my cleanup.

maddyscientist:
> This also fails an error check after manually specifying `wilson`-type smearing:
> 
> ```
> > mpirun -np 1 ./su3_fermion_test --su3-smear-type wilson --verbosity verbose
> [...]
> Fractional error of (<check,adj_check> - <check,fwd_check>.conj()) = 2.82863e-14
> adjoint safe/hier match precision passed!
> ERROR: fractional error precision failed
>  (rank 0, host ipp1-2318.nvidia.com, su3_fermion_test.cpp:279 in int main(int, char**)())
>        last kernel called was (name=cudaMemcpyDefault,volume=bytes=63700992,aux=copy,color_spinor_field.cpp,498)
> ```
> 
> Symanzik has the same issue.

This was due to requesting too tight a tolerance in double precision.  Fixed.

maddyscientist:
> The test executable `./su3_fermion_test` doesn't run out of the box with no arguments passed in:
> 
> ```
> > mpirun -np 1 ./su3_fermion_test
> [...]
> WARNING: Using device memory pool allocator
> WARNING: Using pinned memory pool allocator
> cublasCreated successfully
> ERROR: Smear type 1 not supported - only Wilson or Symanzik supported (rank 0, host ipp1-2318.nvidia.com, su3_fermion_test.cpp:147 in int main(int, char**)())
>        last kernel called was (name=cudaMemcpyDeviceToHost,volume=bytes=47775744,aux=copy,gauge_field.cpp,1024)
> [...]
> ```
> 
> Smear type "1" is stout smearing, which is the default specified in the command_line_params files. Should we change the default there, or manually override it in `./su3_fermion_test`?

Fixed in my cleanup: the default smear type is Stout (which is inherited from pre-existing state).  If the user doesn't specify, then Wilson-flow is now set (and a message printed to this effect).

tests/CMakeLists.txt
maddyscientist: I have guarded the ctest with the WILSON condition, and also added a check in the test for enablement of nSpin 4.

maddyscientist:
This is great addition @rkarur, thank you.

I have cleaned up the test a bit, and it now uses google test when the `--enable-testing true` flag is passed to the exe.  Also, I've added the test to the ctest list, so it will be run in our CI.  Can you check what I've done to make sure I've not done anything untoward?

I'm approving this PR now, but before we merge we should wait for @weinbe2's sign off.

maddyscientist (approved):

weinbe2 (approved):
This all looks good to me. Thanks @rkarur and @maddyscientist !

PR #1529, opened 2024-12-22
MILC batched deflation

leonhostetler:
This pull request implements batched deflation for MILC. Most of the work was limited to the QUDA/MILC interface, however, there is one change (the first one listed below) that goes beyond that scope, so please check that one in particular to ensure that my solution is okay.

1. Changed `errorQuda()` in `constructDeflationSpace()` to `warningQuda()`. See below for more description
2. Added a function `qudaCleanUpDeflationSpace()`. See below for more description
3. Added some basic stuff to `invertQuda()` to preserve the deflation space. The preservation of the eigenvectors and eigenvalues is then controlled from the MILC side by the appropriate parameters in the `QudaEigParam` struct.
4. Added ability to control `tol_restart` and `cuda_prec_eigensolver` from the MILC side
5. Changes to `invertQuda()` were copied to `invertQudaMsrc()`.

Note there is a companion MILC pull request at https://github.com/milc-qcd/milc_qcd/pull/76

Note that this implementation only performs deflation for even parity solves. This seems to be fine when the UML solver is selected for MILC. With UML, the odd parity solution is reconstructed from the even parity solution and then polished with a few CG iterations. Usually this is a small number and might not benefit from deflation anyway. On the other hand, if the CG or CGZ solvers are preferred then odd parity deflation should also be implemented in the future.

More details:

1. I changed `errorQuda()` in `constructDeflationSpace()` to `warningQuda()`. Here's why:

When loading eigenvectors from file, `loadFromFile()` -> `computeEvals()` is called, which extends the deflation space by the amount of the batch size:
```
if (size + batch_size > static_cast<int>(evecs.size())) resize(evecs, size + batch_size, QUDA_NULL_FIELD_CREATE);
```
After the first deflated solve, the deflation space is preserved, but it is now of size `evecs.size() + batch_size`.

This is fine for the first deflated solve, but on the second deflated solve, when `constructDeflationSpace()` is called and the preserved deflation space is attempted to be loaded, the following check fails:
```
if ((!space->svd && param.eig_param.n_conv != (int)space->evecs.size())
    || (space->svd && 2 * param.eig_param.n_conv != (int)space->evecs.size()))
  errorQuda("Preserved deflation space size %lu does not match expected %d", space->evecs.size(),
	    param.eig_param.n_conv);
```
It expects
```
param.eig_param.n_conv == space->evecs.size()
```
However, it is now actually
```
param.eig_param.n_conv + param.eig_param.compute_evals_batch_size == space->evecs.size()
```
Replacing the `errorQuda()` with `warningQuda()` allows it to run, and it runs fine. However, if there are cases where we really need the warning to be an error, then this is probably not the right way to fix it.

2. Added a function `qudaCleanUpDeflationSpace()`, which can be called from MILC to clean up the deflation space. An alternative approach would be to set `preserve_deflation_space` to false on the last solve. However, this would limit how much we could amortize the cost of the eigenvector loading. The parameter input file for a large MILC calculation is chunked into "readin sets". MILC reads one of these readin sets, performs all calculations therein, and then reads the next readin set. So during a job, MILC does not know when it is doing the last solve. At best, it knows it's the last solve in the current readin set. So it seemed to me the best way to ensure that the deflation space is preserved over all readin sets is to add a cleanup function that can be called from MILC's `finalize_quda()` function in `milc_qcd/generic/milc_to_quda_utilities.c`.


maddyscientist (approved):
Thanks for this PR @leonhostetler.  This will add some critical functionality between MILC and QUDA.
* I agree with your change from error to warning for the size of the basis.  to be honest I'm not even sure if we need it to be a warning at all.

weinbe2:
> Thanks for this PR @leonhostetler. This will add some critical functionality between MILC and QUDA.
> 
> * I agree with your change from error to warning for the size of the basis.  to be honest I'm not even sure if we need it to be a warning at all.

I think we need to demand `space->evecs.size() >= param.eig_param.n_conv`, as in, make sure there at least _enough_ eigenvectors. I agree there's no need for a warning otherwise, if there are extra vectors, whatever.

That line would then become:

```
if ((!space->svd && (int)space->evecs.size() < param.eig_param.n_conv)
    || (space->svd && (int)space->evecs.size() < (2 * param.eig_param.n_conv)))
  errorQuda("Preserved deflation space size %lu is smaller than the necessary %d", space->evecs.size(),
	    param.eig_param.n_conv);
```

If this seems reasonable can you please put this change in @leonhostetler ? I don't think I can push directly to your branch.

leonhostetler:
@weinbe2 I have committed the change you suggested after re-running tests on my end.

weinbe2:
> @weinbe2 I have committed the change you suggested after re-running tests on my end.

Thanks @leonhostetler! Per our offline convo I'll do some due-diligence last tests, but ideally we'll get this all merged in tomorrow.

weinbe2:
As an update here: after some testing, I realized this PR breaks backwards compatibility of new QUDA + old MILC. This is because it modifies structs; C-style function signature/argument resolution doesn't encode this and because the _function_ signature didn't change it'll happily link and then give undefined behavior at runtime. @leonhostetler and I are discussing how to safely resolve this offline.

include/quda_milc_interface.h
```diff
@@ -52,7 +52,37 @@ extern "C" {
...
+    int block_size; /** For block method solvers, the block size **/
+    char vec_infile[256];  /** Filename prefix where to load the null-space vectors */
+    char vec_outfile[256]; /** Filename prefix for where to save the null-space vectors */
+    QudaPrecision save_prec; /** The precision with which to save the vectors */
+    QudaBoolean partfile; /** Whether to save eigenvectors in QIO singlefile or partfile format */
+    QudaBoolean io_parity_inflate; /** Whether to inflate single-parity eigen-vector I/O **/
+    QudaBoolean use_norm_op;
+    QudaBoolean use_pc;
+    
+  } QudaEigensolverArgs_t;
```
weinbe2: This broadly looks good---to be more future-proof (in particular to if folks want to use this structure for other fermion types for whatever reason), we should add more of the fields that are currently getting default values in `qudaInvertDeflatable`.

```
QudaEigType eig_type = ( qep.block_size > 1 ) ? QUDA_EIG_BLK_TR_LANCZOS : QUDA_EIG_TR_LANCZOS;  /* or QUDA_EIG_IR_ARNOLDI, QUDA_EIG_BLK_IR_ARNOLDI */
QudaEigSpectrumType spectrum = QUDA_SPECTRUM_SR_EIG; /* Smallest Real. Other options: LM, SM, LR, SR, LI, SI */
QudaBoolean require_convergence = QUDA_BOOLEAN_TRUE;
QudaBoolean check_interval = 10;
QudaBoolean use_dagger = QUDA_BOOLEAN_FALSE;
QudaBoolean compute_gamma5 = QUDA_BOOLEAN_FALSE;
QudaBoolean compute_svd = QUDA_BOOLEAN_FALSE;
QudaBoolean use_eigen_qr = QUDA_BOOLEAN_TRUE;
QudaBoolean use_poly_acc = QUDA_BOOLEAN_TRUE;
QudaBoolean compute_evals_batch_size = 16;
```

The defaults can be then set on the MILC side of things instead of on the QUDA side of things.

---

Re: loading eigenvectors computed for the odd operator then applying them to the even operator (or any combination), you could add a field `QudaParity vec_in_parity;` or something of the sort. For the scope of this PR you could just require that the input parity be even otherwise it errors, but in a next PR we could expand it to handling constructing the even eigenvectors from input odd ones (or vice-versa).

lib/milc_interface.cpp
```diff
@@ -1137,14 +1153,72 @@ void qudaMultishiftInvert(int external_precision, int quda_precision, int num_of
...
                 double target_residual, double target_fermilab_residual, const void *const fatlink,
                 const void *const longlink, void *source, void *solution, double *const final_residual,
                 double *const final_fermilab_residual, int *num_iters)
 {
+  
+  // If this function is called then QUDA is not doing deflation
+  // Create dummy QudaEigensolverArgs_t that requests 0 eigenvalues
+  QudaEigensolverArgs_t eig_args;
+  eig_args.n_ev_deflate=0;
+  eig_args.prec_eigensolver = QUDA_DOUBLE_PRECISION; /* Not used in the undeflated case but needs to be initialized */
```
weinbe2: As discussed briefly offline, this needs to be set to something reasonable b/c of how QUDA handles creating copies of fields (aliasing as appropriate) "under the hood". A "safer" value to set this to is `eig_args.prec_eigensolver = (quda_precision == 2) ? QUDA_DOUBLE_PRECISON : QUDA_SINGLE_PRECISION;`, which should ~guarantee the eigensolver field exactly matches something that's already going to be there.

lib/milc_interface.cpp
```diff
@@ -1137,14 +1153,72 @@ void qudaMultishiftInvert(int external_precision, int quda_precision, int num_of
...
+                double target_residual, double target_fermilab_residual, const void *const fatlink,
+                const void *const longlink, void *source, void *solution, double *const final_residual,
+                double *const final_fermilab_residual, int *num_iters)
+{
   static const QudaVerbosity verbosity = getVerbosity();
   qudamilc_called<true>(__func__, verbosity);
 
+  QudaParity local_parity = inv_args.evenodd;
+  
+  // parameters for the eigensolve/deflation
```
weinbe2, line 1223: For sanity/cleanliness/reuse between this and the msrc variety, this initialization should be moved to a function `setEigensolverParams` in the same spirit of `setInvertParams` and `setGaugeParams`.

lib/milc_interface.cpp
```diff
@@ -1182,12 +1256,27 @@ void qudaInvert(int external_precision, int quda_precision, double mass, QudaInv
...
 
-  QudaParity local_parity = inv_args.evenodd;
   const double reliable_delta = 1e-1;
 
   setInvertParams(host_precision, device_precision, device_precision_sloppy, mass, target_residual,
                   target_fermilab_residual, inv_args.max_iter, reliable_delta, local_parity, verbosity,
                   QUDA_CG_INVERTER, &invertParam);
+ 
+  // Deflation for even parity solves when desired 
+  invertParam.eig_param = (local_parity == QUDA_EVEN_PARITY)&&(qep.n_ev_deflate>0) ? &qep : nullptr;
```
weinbe2: Given the comments on supporting even and odd solves above, I think this should just be:

```
invertParam.eig_param =(qep.n_ev_deflate>0) ? &qep : nullptr;
```

And then, until we get even/odd eigenvector reconstruct going, you can put in something like

```
if (qep.n_ev_deflate > 0 && local_parity != QUDA_EVEN_PARITY)
  errorQuda("MILC interface deflation currently only supports even parity solves.");
```

lib/milc_interface.cpp
```diff
@@ -1362,13 +1456,70 @@ void qudaSpinTaste(int external_precision, int quda_precision, const void *const
...
+  eig_args.prec_eigensolver = QUDA_DOUBLE_PRECISION; /* Not used in the undeflated case but needs to be initialized */
+
+  qudaInvertMsrcDeflatable(external_precision, quda_precision, mass, inv_args, eig_args,
+                    target_residual, target_fermilab_residual, fatlink,
+                    longlink, sourceArray, solutionArray, final_residual,
+                    final_fermilab_residual, num_iters, num_src);
+
+} // qudaInvertMsrc
+
+void qudaInvertMsrcDeflatable(int external_precision, int quda_precision, double mass, QudaInvertArgs_t inv_args, QudaEigensolverArgs_t eig_args,
```
weinbe2: All of my comments for `qudaInvertDeflatable` also apply here (unsurprisingly)

lib/milc_interface.cpp
leonhostetler: This should be fixed now with cdedc07. Let me know if I should change anything.

include/quda_milc_interface.h
leonhostetler: The first part should be fixed now with cdedc07. The second part, RE loading eigenvectors, should be fixed now with 02eb6f5 and 2598582. Let me know if anything should be changed.

lib/milc_interface.cpp
leonhostetler: This should be fixed now with 2e14fda. Let me know if it needs changing.

lib/milc_interface.cpp
leonhostetler: This should be fixed now with f50b9ae. Let me know if anything needs changing.

lib/milc_interface.cpp
leonhostetler: I think these are all fixed now.

include/quda_milc_interface.h
weinbe2: This is great! One thing I forgot: you should add a field `struct_size` that gets set to `sizeof(QudaEigensolverArgs_t)` on the MILC side, and then on the QUDA side you separately evaluate that and compare it with what the size of the struct gets set to in MILC. It should be the first field in the structure.

The reason to do this is to make error checking more explicit---if QUDA and MILC disagree on the size of the structure due to some mixed version issues, we'll get an error. Let me know if this makes sense/if you'd like me to better sketch it out.

lib/milc_interface.cpp
```diff
@@ -1180,14 +1258,35 @@ void qudaInvert(int external_precision, int quda_precision, double mass, QudaInv
...
+  if (eig_args.vec_in_parity != QUDA_EVEN_PARITY)
+    errorQuda("MILC interface deflation currently only supports even parity eigenvectors.");
+    
+  invertParam.tol_restart = eig_args.tol_restart;
+
+  // Eigensolver precision
+  invertParam.cuda_prec_eigensolver = eig_args.prec_eigensolver;
+
+  // Preserve deflation space
+  static bool deflation_init = false;
```
weinbe2: This variable should be moved into the global scope next to `preserved_deflation_space` so `qudaInitDeflatable` and `qudaInvertMsrcDeflatable` see consistent information

leonhostetler: Good catch! This should be taken care of by [8106f58](https://github.com/lattice/quda/pull/1529/commits/8106f5892471242de439e8742d7e5be0213490e4).

include/quda_milc_interface.h
leonhostetler: This should be fixed by b3f4e8d and c87e4dd

maddyscientist:
Before we merge, this needs to be clang formatted.

weinbe2:
@leonhostetler instructions here: https://github.com/lattice/quda/wiki/Coding-Conventions-and-Style#applying-it-to-lines-changes-in-a-pull-request ; sorry I forgot to mention that earlier

leonhostetler:
I think I did the clang-format correctly. I used v11, which is the version noted in @weinbe2's link above

PR #1530, opened 2025-01-03
Hotfix/pool

maddyscientist:
Minor hotfix PR
* Expose new interface function `flushPoolQuda(QudaMemoryType)` to facilitate user control over flushing QUDA's memory pools
* Remove legacy option `QudaMultigridParam::setup_minimize_memory` which was broken, and redundant in the era of unified memory
* Fix BiCGStab and BiCGStab(l) for use as null-space generators

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -404,6 +404,9 @@ endif()
 
 find_package(Threads REQUIRED)
 if(QUDA_OPENMP)
+  if(${CMAKE_CXX_COMPILER_ID} MATCHES "NVHPC")
+    message(FATAL_ERROR "Host compiler (nvc++) not supported with QUDA_OPENMP=ON")
+  endif()
   find_package(OpenMP REQUIRED)
 endif()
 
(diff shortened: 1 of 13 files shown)


CMakeLists.txt
```diff
@@ -404,6 +404,9 @@ endif()
 
 find_package(Threads REQUIRED)
 if(QUDA_OPENMP)
+  if(${CMAKE_CXX_COMPILER_ID} MATCHES "NVHPC")
```
bjoo, line 407: Question here: Rather than making this bomb with an error. Would it be nicer to just say 'we are not enabling'. I can see arguments both ways. Bombing ensures we turn the option off and so a 'dockerfile' or a cmake command line is a more faithful record of what actually is enabled. Carrying on disabling the option may be more user friendly (?) but may also be a little more work as we'd need to make sure we are also not looking for the OpenMP dependency in the Config.cmake file as well.

bjoo (approved):
This looks nice and clean to me - only the question about the strategy to bomb on OpenMP with the NVHPC++ compiler remains and it is not very important. I am happy to approve this.

CMakeLists.txt
maddyscientist: The bomb here is intentional, as
1.) it gives the user the obvious work around explicitly
2.) encourages the user to reach out to us if this is an important issue that needs to be rectified
Since QUDA is typically run by semi-developers and not "monkey users", I prefer the explicit approach.

FWIW: it's the lack of user-defined reductions in nvc++ that causes this problem.

weinbe2 (approved):

PR #1531, opened 2025-01-04
pass KernelOps into SharedMemoryCache constructor

jcosborn:
This will allow the SYCL target to allocate shared memory at kernel launch and pass the pointer in through the KernelOps structure.


include/gauge_fix_ovr_hit_devf.cuh
```diff
@@ -40,20 +40,29 @@ namespace quda {
...
+                                               const Ftor &ftor)
   {
     auto blockSize = target::block_dim().x;
     auto tid = target::thread_idx().x;
 
     //Container for the four real parameters of SU(2) subgroup in shared memory
-    SharedMemoryCache<Float> cache;
-    auto elems = cache.data();
+    SharedMemoryCache<Float, GaugeFixHitDims<4>> cache(ftor);
+    Float *elems = cache.data();
```
maddyscientist: Why can't this be `auto` here?

include/kernels/coarse_op_kernel.cuh
```diff
@@ -303,7 +304,7 @@ namespace quda {
 #pragma unroll
     for (int s = 0; s < uvSpin; s++) {
       if constexpr (Arg::compute_max) {
-        uv_max = fmax(UV[s].abs_max(), uv_max);
+        uv_max = max(UV[s].abs_max(), uv_max);
```
maddyscientist, line 308: This file uses a mixed of `quda::max` and `max`.  Can you make it consistent?

include/kernels/gauge_fix_ovr.cuh
```diff
@@ -205,16 +221,19 @@ namespace quda {
         int idx1 = (((x[3] * X[2] + x[2]) * X[1] + x[1]) * X[0] + x[0]) >> 1;
         Link link1 = arg.u(mu, idx1, 1 - parity);
 
-        switch (Arg::type) {
+        if constexpr (Arg::type == 3) {
```
maddyscientist, line 224: Why the change here from `switch` to `if` ?  The former is more logical.

include/gauge_fix_ovr_hit_devf.cuh
```diff
@@ -138,17 +147,20 @@ namespace quda {
...
+                                                 const Ftor &ftor)
   {
     auto blockSize = target::block_dim().x;
     auto tid = target::thread_idx().x;
 
     //Container for the four real parameters of SU(2) subgroup in shared memory
-    SharedMemoryCache<Float> cache;
-    auto elems = cache.data();
+    SharedMemoryCache<array<Float, 4>> cache(ftor);
+    Float *elems = &(*cache.data())[0];
```
maddyscientist: I am not liking this change.  Why is this pathology necessary?

include/kernels/block_transpose.cuh
```diff
@@ -55,6 +51,19 @@ namespace quda
         return block;
       }
     };
+    using color_spinor_t = ColorSpinor<typename Arg::real, 1, Arg::nSpin>;
```
maddyscientist, line 56: I don't like how we have to redeclare `color_spinor_t` here after it is already defined on line 79.

maddyscientist (changes requested):
I've left some comments on the PR, but the main issue I have is with
1.) The lack of documentation regarding using `KernelOps`.  E.g., how does someone who is writing a new kernel that uses shared memory know what to do.  We need a extensive wiki page describing this.
2.) How `KernelOps` will be tested in CI without any SYCL target.  How can we ensure that `KernelOps` are developed correctly?

We can discuss in our call.

jcosborn:
> I've left some comments on the PR, but the main issue I have is with 1.) The lack of documentation regarding using `KernelOps`. E.g., how does someone who is writing a new kernel that uses shared memory know what to do. We need a extensive wiki page describing this. 2.) How `KernelOps` will be tested in CI without any SYCL target. How can we ensure that `KernelOps` are developed correctly?
> 
> We can discuss in our call.

I've added some documentation here
https://github.com/lattice/quda/wiki/KernelOps-usage
It will need to be updated as more pieces are merged in from the SYCL branch.
Let me know if anything isn't clear so far.

As we discussed on the call, the `KernelOps` handling should be fairly robust (once fully merged), so that instantiating tracked operations will require passing in a kernel functor that inherited from the appropriate `KernelOps` type.  This type checking happens at compile time.

include/gauge_fix_ovr_hit_devf.cuh
jcosborn: I think it is fine.
4c190e390f85014044a1f6a2610cd3dce54912dd

maddyscientist: Thanks 😄

include/kernels/block_transpose.cuh
maddyscientist: Marking as resolved, my comment here was one in ignorance.

maddyscientist:
> I've added some documentation here https://github.com/lattice/quda/wiki/KernelOps-usage
It will need to be updated as more pieces are merged in from the SYCL branch.  Let me know if anything isn't clear so far.

Documentation is exactly what I wanted, thanks @jcosborn.

include/gauge_fix_ovr_hit_devf.cuh
jcosborn: Yes, there was no need to use an inner array.  I've made it similar to the others now.
99632a5870fe340914fb2ce1cc4800ce535d6762

include/kernels/gauge_fix_ovr.cuh
jcosborn: This has to be `constexpr` now since the different routines being called in each case have different variants of the `SharedMemoryCache`.  This corresponding `KernelOps` are selected here https://github.com/lattice/quda/pull/1531/files#diff-aa538a7e33ac26c68858653fe4c25222aae7dc96f95459c62bf86ea5acd254b4R135 at compile time.  Due to the compile-time consistency checking of the `SharedMemoryCache` instantiation, we can't compile the branches that don't have the correct `KernelOps` even if they are never called.

include/kernels/coarse_op_kernel.cuh
jcosborn: `quda::max` was needed in some places to avoid conflict with local variables named `max`.
I've renamed the local variables and replaced `quda::max` with just `max`.
768f37b3b845e99d796119e5c1acf06b99133509

include/kernels/block_transpose.cuh
jcosborn: Ok, I've made the one on line 79 reuse the definition in `BlockTransposeKernelOps`.
4cc79647e8ca7d512e541dde830b44a10bdab151

weinbe2:
This design looks good to me given the requirements, thanks @jcosborn. The wiki page is greatly appreciated.

I don't "like" the `decltype(*this)` that needed to be used in one of the coarse-related file, but since it's a one-off in a single file I'll let it go. I understand why template deduction couldn't be used without a bunch of unnecessary function argument rearranging.

My only request is for the wiki page; I think the links to files + line-numbers should be modified to be permalinks to a specific commit as opposed to just `develop`. This isn't a blocking requirement for this PR; if anything it can be updated with the commit id of the merge.

jcosborn:
Thanks for the review.  Making the wiki use permalinks is a good idea.  Once this is merged I'll update it with permalinks to develop.

weinbe2 (approved):

jcosborn:
I've fixed the CI, and I think everything has been addressed (except for updating the docs after the merge).
Let me know if there is anything else.

maddyscientist (approved):
Loos good now.  Thanks for the work on this @jcosborn.

maddyscientist:
@jcosborn can you merge in the latest version of develop?  This will fix the CSCS pipeline which would be good to run before we merge this in.

weinbe2:
@jcosborn merge in `develop` one last time to get the csci running, and assuming there are no issues I'll hit merge tomorrow, I promise!

Issue #1532, opened 2025-01-16
Generating null-vector using bicgstab broken


chaoos:
Hello,

I found that using the newest `develop` branch, the setup for multigrid doesn't work anymore. When creating null-vectors with bicgstab, the first batch of `param.n_vec_batch` vectors are solved properly, but all following ones diverge. There seems to be some internal state that get carried on from one invocation of the solver to the next. I was able to pin down the problem to this line that swaps `x` and `y`.

https://github.com/lattice/quda/blob/97658476712228b0e637469d8f2c21af39adb786/lib/inv_bicgstab_quda.cpp#L130-L131

If I change it back to what it was before the mRHS stuff (it was changed here https://github.com/lattice/quda/commit/a3186a0fd88d400e519cf8f875355ab201eeddc4), i.e.

https://github.com/lattice/quda/blob/57ba15ef38abb016b791c0362bb876e20dda058c/lib/inv_bicgstab_quda.cpp#L104

I get correct results and the multigrid works again as intended. Notice I need that `param.n_vec_batch < param.n_vec` for this to happen.

Best,
Roman

maddyscientist:
Thanks for filing this issue @chaoos.  I've fixed the bug in #1534

weinbe2:
Closed by #1534 ! Extra thanks @chaoos because this uncovered similar issues in a few other solvers when they were used for near-null vector generation.

PR #1534, opened 2025-01-16
Hotfix/bicgstab null

maddyscientist:
Hotfix PR that fixes the issue raised in #1532.
* Avoid use of std::swap between `x` and `y` vectors as this is buggy
  * the owning vector containers still point to the original objects and so become invalid 🤮 )
* When using initial guesses, the solution vector needs to be zeroed after residual has been computed
* Remove redundant zeroing when fields are first allocated (any required zeroing is handled when solver is applied)

I've cherry picked the changes that allow GitHub actions to continue to work from #1497, as I assume this PR may be merged first.


weinbe2 (approved):
This looks great to me, awesome work @maddyscientist 

Of note, as part of testing this I uncovered a few issues triggered by `ctest` that are unrelated to this PR; I'll file a separate issue for it momentarily.

Issue #1535, opened 2025-01-23
Schwarz preconditioned CG gives `nan`s in develop


weinbe2:
As described in the title.

`cmake` command:
```
cmake -DCMAKE_BUILD_TYPE=RELEASE -DQUDA_DIRAC_DEFAULT_OFF=ON -DQUDA_DIRAC_DOMAIN_WALL=ON  -DQUDA_GPU_ARCH=sm_80 -DQUDA_DOWNLOAD_USQCD=ON -DQUDA_QIO=ON -DQUDA_QMP=ON -DQUDA_FAST_COMPILE_DSLASH=ON -DQUDA_FAST_COMPILE_REDUCE=ON ../quda
```

Test executable command (~same as on the wiki page: https://github.com/lattice/quda/wiki/The-Multi-Splitting-Preconditioned-Conjugate-Gradient-(MSPCG),-an-application-of-the-additive-Schwarz-Method )
```
mpirun -np 1 ./invert_test \
  --matpc even-even \
  --dim 12 12 12 16 \
  --Lsdim 8 \
  --gridsize 1 1 1 1 \
  --dslash-type mobius \
  --b5 2.5 --c5 1.5 \
  --inv-type pcg \
  --precon-type cg \
  --precon-schwarz-type additive \
  --tol 1e-6 \
  --tol-precondition 1e-6 \
  --niter 860 \
  --maxiter-precondition 6 \
  --mass 0.01 \
  --prec single \
  --prec-sloppy half \
  --prec-precondition half
```

Output (with `--verbosity verbose` added):
```
[...]
Computed plaquette is 1.226087e-01 (spatial = 1.221845e-01, temporal = 1.230330e-01)
Solution = mat, Solve = normop_pc, Solver = pcg, Precision = single, Sloppy precision = half
DiracMobius: Detected fixed real cofficients: using regular Mobius
DiracMobius: Detected fixed real cofficients: using regular Mobius
DiracMobius: Detected fixed real cofficients: using regular Mobius
DiracMobius: Detected fixed real cofficients: using regular Mobius
Source: 2.65451e+06
Prepared: source = 2.1658e+06, solution = 0
Creating a PCG solver
CG: Convergence at 6 iterations, L2 relative residual: iterated = 4.735067e-02 (requested = 1.000000e-06)
PCG:     0 iterations, <r,r> = 2.841162e+06, |r|/|b| = 1.000000e+00
[...]

PR #1536, opened 2025-02-10
Fixed bug causing NaN residuals in PCG solves

agrebe:
In previous version, `x_update_batch` was initialized before `minvr_sloppy`, so the first vector in `x_update_batch` was initialized to 0.  The result then had zero norm, and this is in the denominator of a computed result that was then NaN, crashing the code.  Moving the initialization of `x_update_batch` later fixes this and lets `invert_test` run to completion.


hummingtree:
Closes #1535.

hummingtree (approved):
I have verified that this fixes #1535.

maddyscientist (approved):

PR #1537, opened 2025-02-13
Feature/mg misc

maddyscientist:
Some fixes and cleanup for multigrid:
* Remove legacy separate host and device versions of the null space vectors to make it more uniform.  (Host-side computation will take very different design methodology in future work).
* Fix some bugs with using specific combinations of e/o preconditioning / direct solving for the smoother and outer solver (Thanks @agrebe for the bug report).
* Small optimization of MG operator to remove some unnecessary blas kernels.
* Remove residual vector from `MG` object and only allocate this when needed (when using pre smoothing).  This reduces memory when only using a post smoother.


maddyscientist:
@agrebe marking you down as a reviewer: would you mind approving this please?

agrebe (approved):
I have tested the new code in this PR and have confirmed that this fixes the bug in `invert_test` that I had seen earlier.  In particular,
`./invert_test --dim 16 16 16 16 --inv-multigrid 1 --solve-type direct --inv-type gcr`
now successfully inverts.

weinbe2:
I'm going to perform a narrow test with staggered MG just to make sure nothing went awry with the KD preconditioner, then we can hit merge.

lib/multigrid.cpp
```diff
@@ -780,13 +761,14 @@ namespace quda
 
     // temporary fields used for verification
     std::vector<ColorSpinorField> fine_tmp(param.Nvec);
-    ColorSpinorParam fine_param(r[0]);
-    fine_param.create = QUDA_NULL_FIELD_CREATE;
+    ColorSpinorParam fine_param(param.B[0]);
+    fine_param.setPrecision(param.mg_global.invert_param->cuda_prec_sloppy, QUDA_INVALID_PRECISION,
+                            fine_param.location == QUDA_CUDA_FIELD_LOCATION ? true : false);
+    fine_param.gammaBasis = param.level > 0 ? QUDA_DEGRAND_ROSSI_GAMMA_BASIS : QUDA_UKQCD_GAMMA_BASIS;
```
weinbe2: This line breaks staggered --- if `param.B[0].Nspin() == 1`, it needs to be `QUDA_DEGRAND_ROSSI_GAMMA_BASIS` always.

weinbe2: The exact line that fixes it: `fine_param.gammaBasis = (param.level > 0 || param.B[0].Nspin() == 1) ? QUDA_DEGRAND_ROSSI_GAMMA_BASIS : QUDA_UKQCD_GAMMA_BASIS;`

weinbe2:
I uncovered a long-standing bug with the pre-smoother that effects KD preconditioning if you use mixed `direct` solves and `direct-pc` smoothers, where in reality I think you need a sequence that looks similar to what's going on here: https://github.com/lattice/quda/blob/develop/include/invert_quda.h#L1378-L1388

I've probably never hit this before because I haven't mixed `direct` solves and `direct-pc` smoothers. Furthermore, even if I did, I missed it because (a) my convention has been to not pre-smooth and (b) the post-smoother doesn't require a `prepare` + the KD `reconstructSpecialMG` happens to be a no-op.

I don't think it's worth fixing this right now (the `[prepare/reconstruct]SpecialMG` has always been a train-wreck design).

weinbe2 (approved):
Pending the staggered fix and an outstanding issue we've discussed offline, this is good to merge. Thanks @maddyscientist , as well as @agrebe for reporting the MG issue and testing this PR.

hummingtree (approved):
Thanks @maddyscientist for the fix (especially the one for the MMA types)!

PR #1539, opened 2025-03-11
Hotfix/ci

maddyscientist:
A PR to restore passing status for the CI, which is now failing due to the increase in number of GPUs being run on the CSCS target
* Add global volume dependence factor for the deviation tolerable between CPU and GPU
  * As we increase the number of nodes, the global volume increases, and so will the error in any global summations
  * We need to account for this
* Fix some bugs in the eigensolve_test
  * Benign valgrind error about uninitialized memory
  * Tolerance not being set correctly for Arnoldi tests
* Improve robustness of Arnoldi ctests: Krylov space should be 3x the number of requested e-vectors


maddyscientist:
cscs-ci run

weinbe2 (approved):
This is great, I appreciate the robustness fixes, thanks @maddyscientist !

PR #1540, opened 2025-03-11
MMA-optimized batched MRHS MG

maddyscientist:
This PR focuses on improving the use of MMA with batched multigrid, resulting in the optimal MRHS multigrid performance:
* Make the mma-ordered version of the null space vectors persistent to avoid needing to reorder on the fly
* Full coarse-grid correction can now be done in MMA order, minimizing the reordering overhead
  * This is enabled on a per-level basis using `QudaMultigridParam::collapse_mrhs` (`--mg-collapse-mrhs` from the command line)
  * Doing so collapse the MRHS solve to a single solve on the coarse level in question
  * Enabling this requires that both the transfer operation into that grid and the dslash operator for that grid are deployed using MMA
* Improve memory utilization: if pre- and post- smoother are identical, they can alias, avoiding needing to be allocated separately
* Various boiler-plate improvements to facilitate the above, e.g.,
  * input and/or output for transfer operators can be collapsed
  * new method `vector_ref<ColorSpinorField>::size_actual()` to allow us to query the number of RHS in a collapsed system
* Improve usability of `FieldTmp`
  * Temporaries can now be references, which helps with composability
* `BlockTranspose` input field set can now be in half precision


lib/multigrid.in.hpp
```diff
@@ -32,29 +32,59 @@ namespace quda
   template <class F> auto create_color_spinor_copy(cvector_ref<F> &fs, int nVec, QudaFieldOrder order)
```
hummingtree: Can we have some documentation on what `create_color_spinor_copy` and `expand_color_spinor` does? When we come back at some point they would be useful.

hummingtree (approved):
Nice clean up to make MMA/MG more accessible! Just have a small request for comments.

lib/multigrid.in.hpp
maddyscientist: d123db48dcca82ceaf57447fe0d7594ff56e49d8

PR #1541, opened 2025-03-11
Fix CSCS CI: create ctest dir on /dev/shm

havogt:
ctest might get stuck when trying to write to the same file in the container from multiple ranks. Workaround is to create the `Testing` directory on `/dev/shm`. See https://github.com/eth-cscs/DLA-Future/issues/1305.


maddyscientist (approved):
Thanks @havogt!

PR #1543, opened 2025-03-19
Feature/split gauge reuse

aroadant:
These modifications will allow the reuse of the redistributed gauge buffers. 

`QudaGaugeParam.use_split_gauge_bkup` is added to determine whether to delete the buffer after use or not. Its default value is `1`. The internal flag `update_split_gauge` is used to determine whether an update of buffer is needed or not.

The initial test with `staggered_invert_test.cpp` gives the same results with `gauge_param.use_split_gauge_bkup = 0` or `1`. Also inversions can be done correctly with `inv_param.split_grid[i] = 1` after split gauge inversions. Results from various default tests are good with default value `use_split_gauge_bkup=0/1`.

The test `staggered_invert_test` is updated with the flag `--use_split_gauge_bkup`. Results from the following input looks good

 staggered_invert_test --dim 6 6 6 6 --gridsize 2 2 2 2 --grid-partition 2 2 2 2
  -dslash-type staggered --compute-fat-long false --unit-gauge 0 --tadpole-coeff 1.0  --verbosity-precondition debug  --recon 13
  --mass 0.02  --solve-type direct-pc --solution-type mat --inv-type cg --matpc even-even
  --prec double --prec-sloppy single
  --nsrc 64 --nsrc-tile 16 --niter 10000 --tol 1e-100
  --inv-multigrid false  --mass-normalization mass --fermion-t-boundary periodic --verify 1 --use_split_gauge_bkup 1


maddyscientist:
@bjoo and @hummingtree can you review this PR?  @bjoo is this helpful for Chroma at all?

include/quda.h
```diff
@@ -70,6 +70,7 @@ extern "C" {
     int staple_pad;   /**< Used by link fattening */
     int llfat_ga_pad; /**< Used by link fattening */
     int mom_ga_pad;   /**< Used by the gauge and fermion forces */
+    int use_split_gauge_bkup; /**< Used by gauge split buffers (default=1 keep split gauge after usage)*/
```
maddyscientist: Perhaps this should be a `bool`?

aroadant: Yes, it's better to use bool. I have updated to QudaBoolean just now.

maddyscientist:
Just a note regarding the the CI failing on CSCS at the moment: the Alps machine is being used for Gordon Bell runs at the moment.  We expect the CI to be restored again next week.

lib/interface_quda.cpp
```diff
@@ -108,6 +108,27 @@ CloverField *cloverEigensolver = nullptr;
...
+quda::GaugeField   *collected_milc_longlink_field = nullptr;
+quda::CloverField *collected_clover = nullptr;
+GaugeBundleBackup* thin_links_bkup = nullptr;
+GaugeBundleBackup* fat_links_bkup  = nullptr;
+GaugeBundleBackup* long_links_bkup = nullptr;
+CloverBundleBackup* clov_bkup = nullptr;
+lat_dim_t X_bkup;
+int is_clover_bkup = -1;
+int is_asqtad_bkup = -1;
+
```
hummingtree: Better to use Enum to replace the `-1`/`0`/`1` so it's more readable.

hummingtree: Same for `is_clover_bkup`/`is_asqtad_bkup`.

lib/interface_quda.cpp
```diff
@@ -138,6 +159,7 @@ static TimeProfile profileInvert("invertQuda");
 
 //!< Profiler for invertMultiSrcQuda
 static TimeProfile profileInvertMultiSrc("invertMultiSrcQuda");
+static TimeProfile profileUpdate_split_gauge("Update_split_gauge");
```
hummingtree: Let's use CamelCase consistently.

lib/interface_quda.cpp
```diff
@@ -568,6 +590,15 @@ void freeUniqueSloppyGaugeUtility(GaugeField *&precise, GaugeField *&sloppy, Gau
 void freeUniqueGaugeUtility(GaugeField *&precise, GaugeField *&sloppy, GaugeField *&precondition, GaugeField *&refinement,
                             GaugeField *&eigensolver, GaugeField *&extended, bool preserve_precise);
 
+// Generate or swap to re-distributed gauge links 
+void Update_split_gauge(QudaInvertParam *param, const int is_asqtad, const bool is_clover, CommKey& split_key);
```
hummingtree: Let's use CamelCase consistently.

lib/interface_quda.cpp
```diff
@@ -568,6 +590,15 @@ void freeUniqueSloppyGaugeUtility(GaugeField *&precise, GaugeField *&sloppy, Gau
...
+
+// free current gauges if keep_buffer == 0, restore previous gauge links, could be used to swap back to split gauge
+void swapGaugeSplit(const int keep_buffer );
```
hummingtree: `keep_buffer` can be a `bool`? Also I am not too sure about what this function does. Could probably use some more documentations.

lib/interface_quda.cpp
```diff
@@ -3112,6 +3157,237 @@ void loadFatLongGaugeQuda(QudaInvertParam *inv_param, QudaGaugeParam *gauge_para
   }
 }
 
+void swapGaugeSplit(const int keep_buffer = 0)
```
hummingtree: Could use some documentation on what this does.

lib/interface_quda.cpp
```diff
@@ -3112,6 +3157,237 @@ void loadFatLongGaugeQuda(QudaInvertParam *inv_param, QudaGaugeParam *gauge_para
...
+void freeGaugeSplit()
+{
+  // swap current gauge with the splitted ones if splits are not null
+  swapGaugeSplit(1);
+
+  // free the splitted gauge and swap to loaded gauge, if splits are not null
+  swapGaugeSplit(0);
+}
+
+void Update_split_gauge(QudaInvertParam *param, const int is_asqtad, const bool is_clover, CommKey& split_key)
```
hummingtree: Could use some documentation here.

tests/utils/command_line_params.cpp
```diff
@@ -570,6 +571,7 @@ std::shared_ptr<QUDAApp> make_app(std::string app_description, std::string app_n
     "--laplace3D", laplace3D,
     "Restrict laplace operator to omit the t dimension (n=3), or include all dims (n=4) (default 4)");
   quda_app->add_option("--load-gauge", latfile, "Load gauge field \" file \" for the test (requires QIO)");
+  quda_app->add_option("--use_split_gauge_bkup", use_split_gauge_bkup, "Use gauge split buff or not");
```
hummingtree: Let's do `--use-split-gauge-bkup`

hummingtree (changes requested):
Thanks @aroadant for this - I have left some comments, and I feel we could use some better documentation for the newly added functions.

lib/interface_quda.cpp
aroadant: I have removed the unnecessary parameters and changed to Enum.

aroadant:
@hummingtree Thanks for the suggestions. I have updated the function documentation and made some changes. There is one issue about clover which may need to be resolved. The check of changes of clover is not good enough and will cause trouble if one needs to switch different clover parameters. https://github.com/lattice/quda/blob/d1c3eb0a99324696fab12f0ea9d5db3bf09f31fe/lib/interface_quda.cpp#L3262

include/enum_quda.h
```diff
@@ -648,6 +648,13 @@ typedef enum QudaWFlowStepType_s {
   WFLOW_FOURTH_ORDER_STEP_6,
 } QudaWFlowStepType;
 
+//Used by update_split_gauge
+typedef enum QudaUpdateSpliteGauge_s {
```
hummingtree:
```suggestion
typedef enum QudaUpdateSplitGauge_s {
```

include/enum_quda.h
```diff
@@ -648,6 +648,13 @@ typedef enum QudaWFlowStepType_s {
   WFLOW_FOURTH_ORDER_STEP_6,
 } QudaWFlowStepType;
 
+//Used by update_split_gauge
```
hummingtree: Could use a more descriptive comment by explaining what these 3 options mean.

include/enum_quda.h
```diff
@@ -648,6 +648,13 @@ typedef enum QudaWFlowStepType_s {
...
+typedef enum QudaUpdateSpliteGauge_s {
+  QUDA_UPDATE_SPLITE_GAUGE_OFF,
+  QUDA_UPDATE_SPLITE_GAUGE_TRUE  = 1,
+  QUDA_UPDATE_SPLITE_GAUGE_FALSE = 0,
+} QudaUpdateSpliteGauge;
```
hummingtree:
```suggestion
  QUDA_UPDATE_SPLIT_GAUGE_OFF,
  QUDA_UPDATE_SPLIT_GAUGE_TRUE  = 1,
  QUDA_UPDATE_SPLIT_GAUGE_FALSE = 0,
} QudaUpdateSplitGauge;
```

lib/interface_quda.cpp
```diff
@@ -108,6 +108,30 @@ CloverField *cloverEigensolver = nullptr;
 GaugeField momResident;
 GaugeField *extendedGaugeResident = nullptr;
 
+/** 
+ * callMultiSrcQuda related gauge split
+ * update_split_gauge : 
+ *   QUDA_UPDATE_SPLITE_GAUGE_OFF delete buffer after usage
```
hummingtree: Let's use split instead of splite consistently.

lib/interface_quda.cpp
```diff
@@ -108,6 +108,30 @@ CloverField *cloverEigensolver = nullptr;
...
+ *   QUDA_UPDATE_SPLITE_GAUGE_FALSE split gauge in buf
```
hummingtree: "in buf" probably means reusing the split gauge that has been backed up previously?

lib/interface_quda.cpp
```diff
@@ -568,6 +593,27 @@ void freeUniqueSloppyGaugeUtility(GaugeField *&precise, GaugeField *&sloppy, Gau
...
+/**
+ * Generate the re-distributed gauge links based on is_asqtad, is_clover, split_key and current gauges
+ * Swap the current gauges into buffers
+ * If the split_key/gauges is the same as previous ones, the already splitted buffers will swap in
+ */
+void UpdateSplitGauge(QudaInvertParam *param, const int is_asqtad, const bool is_clover, CommKey& split_key);
+
+/**
+ * If keep_buffer == true :
+ *   Swap between current gauges with the buffered ones (possibly splitted gauges or original gauges)
```
hummingtree:
```suggestion
 *   Swap the current gauges with the buffered ones (possibly split gauges or original gauges)
```

lib/interface_quda.cpp
```diff
@@ -568,6 +593,27 @@ void freeUniqueSloppyGaugeUtility(GaugeField *&precise, GaugeField *&sloppy, Gau
...
+ * If keep_buffer == false:
+ *   Delete current splitted gauges and then swap the gauges in buffers to Quda gauges
```
hummingtree:
```suggestion
 *   Delete the current split gauges and then swap the gauges in buffers with the current(?) gauges
```

lib/interface_quda.cpp
```diff
@@ -1386,13 +1449,16 @@ void flushPoolQuda(QudaMemoryType type)
...
 
+
 void endQuda(void)
 {
   if (!initialized) return;
 
   {
     auto profile = pushProfile(profileEnd);
 
+    freeGaugeSplit();// free split Gauge and restore original Gauge
```
hummingtree:
```suggestion
    freeGaugeSplit(); // free the split gauges and restore the original gauges 
```

lib/interface_quda.cpp
```diff
@@ -3112,6 +3178,239 @@ void loadFatLongGaugeQuda(QudaInvertParam *inv_param, QudaGaugeParam *gauge_para
...
+  }
+
+  if(!keep_buffer and update_split_gauge != QUDA_UPDATE_SPLITE_GAUGE_OFF){
+    update_split_gauge = QUDA_UPDATE_SPLITE_GAUGE_TRUE;
+  }
+}
+
+void freeGaugeSplit()
+{
+  // swap current gauge with the splitted ones if splits are not null
```
hummingtree:
```suggestion
  // swap current gauges with the split ones if split ones are not null
```

lib/interface_quda.cpp
```diff
@@ -3112,6 +3178,239 @@ void loadFatLongGaugeQuda(QudaInvertParam *inv_param, QudaGaugeParam *gauge_para
...
+  swapGaugeSplit(true);
+
+  // free the splitted gauge and swap to loaded gauge, if splits are not null
```
hummingtree:
```suggestion
  // free the split gauges and swap the split gauges with the buffered gauges, if split gauges are not null
```

lib/interface_quda.cpp
```diff
@@ -3379,47 +3569,16 @@ void callMultiSrcQuda(void **_hp_x, void **_hp_b, QudaInvertParam *param, // col
...
-      gaugeFatExtended = fat_links_bkup.extended;
-
-      freeUniqueGaugeQuda(QUDA_ASQTAD_LONG_LINKS);
-      gaugeLongPrecise = long_links_bkup.precise;
-      gaugeLongSloppy = long_links_bkup.sloppy;
-      gaugeLongPrecondition = long_links_bkup.precondition;
-      gaugeLongRefinement = long_links_bkup.refinement;
-      gaugeLongEigensolver = long_links_bkup.eigensolver;
-      gaugeLongExtended = long_links_bkup.extended;
+    // swap back to original links, detete split gauge if update_split_gauge == QUDA_UPDATE_SPLITE_GAUGE_OFF
```
hummingtree:
```suggestion
    // switch back to the original links, detete split gauge if update_split_gauge == QUDA_UPDATE_SPLITE_GAUGE_OFF
```

include/quda.h
```diff
@@ -70,6 +70,7 @@ extern "C" {
     int staple_pad;   /**< Used by link fattening */
     int llfat_ga_pad; /**< Used by link fattening */
     int mom_ga_pad;   /**< Used by the gauge and fermion forces */
+    QudaBoolean use_split_gauge_bkup; /**< Used by gauge split buffers (default=true keep split gauge after usage)*/
```
hummingtree: We can use bool directly - C99 supports boolean types now.

lib/interface_quda.cpp
```diff
@@ -108,6 +108,30 @@ CloverField *cloverEigensolver = nullptr;
...
+ * callMultiSrcQuda related gauge split
+ * update_split_gauge : 
+ *   QUDA_UPDATE_SPLITE_GAUGE_OFF delete buffer after usage
+ *   QUDA_UPDATE_SPLITE_GAUGE_FALSE split gauge in buf
+ *   QUDA_UPDATE_SPLITE_GAUGE_TRUE  split gauge not in buf, need to call split
+ * update_split_gauge set to QUDA_UPDATE_SPLITE_GAUGE_OFF if QudaGaugeParam.use_split_gauge_bkup == QUDA_BOOLEAN_FALSE, QUDA_UPDATE_SPLITE_GAUGE_TRUE if QudaGaugeParam.use_split_gauge_bkup == QUDA_BOOLEAN_TRUE
+ * update_split_gauge != QUDA_UPDATE_SPLITE_GAUGE_OFF and split gauge in buf, the original links will swap with *_bkup 
+ * split_grid_bkup will be used to check whether split layout is changed or not
+ * change in gauge precsions will need loadGaugeQuda which results in re-distribute
+ * assumed no change in clover parameters
```
hummingtree: I modified the comment here to make it clearer - but please correct me if I say anything wrong.
```suggestion
 * - QUDA_UPDATE_SPLITE_GAUGE_TRUE: the input gauge fields will be split and the buffered (split) gauges will be updated accordingly;
 * - QUDA_UPDATE_SPLITE_GAUGE_FALSE: the input gauge fields will not be split and the buffered (split) gauges will be used for split grid solves;
 * - QUDA_UPDATE_SPLITE_GAUGE_OFF: nothing will be done.
 
 * split_grid_bkup will be used to check whether split layout is changed or not
 * change in gauge precsions will need loadGaugeQuda which results in re-distribute
 * assumed no change in clover parameters
```

hummingtree (commented):
@bjoo When you get a chance, can you check and see if you are comfortable with the changes as it changes many of the codes you added. It would be even better if you could see if there is anything that you can use in Chroma.

aroadant:
@hummingtree Thanks! It's very helpful to make the comments to be clear!

bjoo:
> @bjoo When you get a chance, can you check and see if you are comfortable with the changes as it changes many of the codes you added. It would be even better if you could see if there is anything that you can use in Chroma.

I tested a propagator code with this and it didn't break, but then I didn't try to update the gauge field. In Chroma, solvers are objects which keep the field, and only call `loadGaugeQuda` once on  solver object construction. It could be useful if someone calls the solver more than once in a split grid setting. Currently I don't know if this happens or not (E.g. right now a 12 component propagator, using a 4 way split grid, would do a single 3-component MRHS solve in each split grid and then the solver object will be destroyed, when the propagator measurement method exits.

The next solver object will call a `loadGaugeQuda()` again.
 
Now there are measurements that could benefit from this. So I am happy with it. I think what would be ideal tho is if a loadGaugeQuda() would trigger a redistribution - I am not 100% sure the PR does this unless all gauge fields are checksummed (not just  the BQCD ones)

aroadant:
> > @bjoo When you get a chance, can you check and see if you are comfortable with the changes as it changes many of the codes you added. It would be even better if you could see if there is anything that you can use in Chroma.
> 
> I tested a propagator code with this and it didn't break, but then I didn't try to update the gauge field. In Chroma, solvers are objects which keep the field, and only call `loadGaugeQuda` once on solver object construction. It could be useful if someone calls the solver more than once in a split grid setting. Currently I don't know if this happens or not (E.g. right now a 12 component propagator, using a 4 way split grid, would do a single 3-component MRHS solve in each split grid and then the solver object will be destroyed, when the propagator measurement method exits.
> 
> The next solver object will call a `loadGaugeQuda()` again.
> 
> Now there are measurements that could benefit from this. So I am happy with it. I think what would be ideal tho is if a loadGaugeQuda() would trigger a redistribution - I am not 100% sure the PR does this unless all gauge fields are checksummed (not just the BQCD ones)

@bjoo Currently line https://github.com/lattice/quda/blob/0017931eab3da529053742eebe91a56e0ee87917/lib/interface_quda.cpp#L695C1-L696C1 is added to trigger a redistribution by default when calling loadGaugeQuda(). And if the gauge is BQCD so that the gauge is not reloaded by checking the checksum, the buffered split gauge will be reused. Do you think we could add more flags so that Chroma could reuse the buffer as often as possible?

bjoo:
> @bjoo Currently line https://github.com/lattice/quda/blob/0017931eab3da529053742eebe91a56e0ee87917/lib/interface_quda.cpp#L695C1-L696C1 is added to trigger a redistribution by default when calling loadGaugeQuda(). And if the gauge is BQCD so that the gauge is not reloaded with checksum, the buffered split gauge will be reused. Do you think we could add more flags so that Chroma could reuse the buffer as often as possible?

No I am fine with just redistributing with every `loadGaugeQuda`. I think that should make things work for Chroma.

aroadant:
> > @bjoo Currently line https://github.com/lattice/quda/blob/0017931eab3da529053742eebe91a56e0ee87917/lib/interface_quda.cpp#L695C1-L696C1 is added to trigger a redistribution by default when calling loadGaugeQuda(). And if the gauge is BQCD so that the gauge is not reloaded with checksum, the buffered split gauge will be reused. Do you think we could add more flags so that Chroma could reuse the buffer as often as possible?
> 
> No I am fine with just redistributing with every `loadGaugeQuda`. I think that should make things work for Chroma.

I see. I am thinking about adding a flag to force reuse the split gauge even after `loadGaugeQuda` is called, but it seems to be not that safe unless one is really sure about what going on. It may be better to add a more general checksum for all the gauge field orders. I am not sure about this part, is it difficult?

hummingtree (approved):
Thanks @aroadant for the changes!

maddyscientist (approved):
Approving this PR.  Thanks for your work on this @aroadant.

The PR needs a clang format, so I'll just do that now and push.  Then we can merge.

aroadant:
> Approving this PR. Thanks for your work on this @aroadant.
> 
> The PR needs a clang format, so I'll just do that now and push. Then we can merge.

@maddyscientist @hummingtree @bjoo Thanks for your review and suggestions! I also learnt a lot from this PR.

PR #1545, opened 2025-03-19
Host gauge field RNG and creation clean-up, plus various routine instantiation clean-up and other things

weinbe2:
This exceptionally painful PR runs the gambit of host cleanups.

For creating and modifying gauge and momentum fields:
- Deterministic, decomposition-agnostic host RNG: creates new routines `random_uniform_host` and `random_gaussian_host` that take in a 4-d checkerboard index and parity and returns fresh random numbers in a form agnostic to the problem decomposition. It is thread-safe by construction and has already been applied to all routines that generate gauge and momentum fields. This is confined to the host file `rng_utils.hpp`.
- The creation of all random host gauge fields has been consolidated into one routine, `createSiteLinkCPU`.
- The set of routines related to creating new gauge fields, applying anisotropy or phases, etc, has been taken out of various different files and has been consolidated into `gauge_utils.[cpp/h]`.
- All gauge momentum field routines have been consolidated into `momentum_utils.[cpp/h]`

For reference routines for dslash, forces, link building, etc:

- An `instantiate` abstraction, similar to the one inside the QUDA library for device functions, has been introduced in `instantiate_host.hpp`. This has been applied to an overwhelming majority of host routines.
- Some, but not all, ways of representing 3x3 matrices on the host have been consolidated into `field_utils.hpp`. Some progress has been made towards homogenizing which form gets used across all host routines, but it isn't complete yet.
- Perhaps most importantly I ripped out the `dsu3_matrix` and `fsu3_matrix`, etc, and consolidated it into a templated `su3_matrix` field, which was a necessary condition for the instantiate abstraction.
- Significant progress was made towards removing `#ifdef MULTI_GPU` across the host verification routines, mostly by removing the single-GPU versions of various routines and making sure the multi-GPU versions play nicely with running on a non-MPI build.
- Redundant indexing routines have been consolidated into `index_utils.hpp`.

And one other miscellaneous thing:

- Forward progress was made towards replacing tiny allocations of arrays of scalars (for ex, `b_5` and `c_5` arrays for mdwf) with `std::vector<...>`.

I'm sure I did other things too, but who knows.

This has been mostly tested, but I still need to do a few extra tests for completeness: namely, a super-build with all dslash types (I've covered most of them at one point or another in my testing) running a full `ctest` with single-GPU and multi-GPU compiles.

There's outstanding host cleanup to be done, but at this point I need to focus on other things and I don't want this progress to be lost to the ether.


tests/heatbath_test.cpp
```diff
@@ -53,101 +52,153 @@ void display_test_info()
              dimPartitioned(3));
 }
 
-void heatbath_test(int argc, char **argv)
+// CURAND random generator initialization
```
maddyscientist: Not necessarily cuRAND (e.g., on non-CUDA platforms).  Also why the need for a `shared_ptr` here?

tests/staggered_invert_test_gtest.hpp
```diff
@@ -107,6 +107,9 @@ TEST_P(StaggeredInvertTest, verify)
   auto ca_basis_tmp = inv_param.ca_basis;
   if (solve_type == QUDA_DIRECT_SOLVE && inverter_type == QUDA_CA_GCR_INVERTER) inv_param.ca_basis = QUDA_POWER_BASIS;
 
+  // Single precision needs a tiny bump due to small host/device precision deviations
+  if (prec == QUDA_SINGLE_PRECISION) verify_tol *= 1.03;
```
maddyscientist, line 111: Is this still needed after my recent PR?

tests/staggered_gsmear_test_utils.h
```diff
@@ -5,17 +5,19 @@
 #include <cstdlib>
 #include <cstring>
 #include <algorithm>
+#include <assert.h>
```
maddyscientist: Mild preference for `cassert` instead of `assert.h` since the former is the C++ variant.  If changing things anyway, can make this change (here and wherever else you are touching).

tests/host_reference/domain_wall_dslash_reference.cpp
```diff
@@ -31,55 +32,93 @@ using namespace quda;
...
-template <QudaPCType type, typename real_t>
-void dslashReference_4d(real_t *out, const real_t *const *gauge, real_t const *const *ghostGauge, const real_t *in,
-                        const real_t *const *fwdSpinor, const real_t *const *backSpinor, int parity, int dagger)
-{
+template <typename real_t, QudaPCType type> struct DslashReference4D {
+  void operator()(void *out_, const void *const *gauge_, const void *const *ghostGauge_, const void *in_,
+                  const void *const *fwdSpinor_, const void *const *backSpinor_, int parity, int dagger)
+  {
+    real_t *out = reinterpret_cast<real_t *>(out_);
+    const real_t *const *gauge = reinterpret_cast<const real_t *const *>(gauge_);
```
maddyscientist: Minor point here: you could mark the lhs for these lines as `auto` to make the lines a easier to read IMHO.

maddyscientist (approved):
Very impressed with the amount of cleanup and rework of the tests done here @weinbe2.  Just left a few minor comments.

tests/heatbath_test.cpp
weinbe2: https://github.com/lattice/quda/commit/2f32193a0015fdff402242a621bebabbb536c98d removes `shared_ptr` via adding a default constructor to `quda::RNG`

tests/staggered_invert_test_gtest.hpp
weinbe2: Yeah, just tried it

tests/staggered_gsmear_test_utils.h
weinbe2: https://github.com/lattice/quda/commit/a45cb662ec707800a2fe86da8600956208666204

tests/host_reference/domain_wall_dslash_reference.cpp
weinbe2: https://github.com/lattice/quda/commit/0d34df8423a88838d511df8120cd143a257c552f

weinbe2:
@maddyscientist I believe I've addressed all your git comments. I also did a bonus refactor of the kernel that applies the clover on the host while I was there!

PR #1547, opened 2025-03-26
Enable double complex support in BatchInvertMatrix

dmcdougall:
I think this was removed way back when (roc|hip)blas didn't have support for double complex types.  The support exists now, so I re-hipified the apropriate branch of the cu target.

This fixes a failing blas_interface_test that complained double precision wasn't supported.


maddyscientist (approved):
Thanks @dmcdougall

PR #1549, opened 2025-03-28
Tweak setting QR tolerances in IRAM

cpviolator:
This PR changes how the QR tolerance is set in `IRAM` when NOT using the EIGEN QR solver between restarts. The default value is set to `0` but can be set to a custom value by the user. `IRAM` will detect a `0` value in the `eig_qr_tol` field when the class is constructed and default to `eig_tol * 1e-2`. Warnings are issued when a QR tolerance greater than `eig_tol * 1e-2` is selected.

### DIFF
--- a/include/eigensolve_quda.h
+++ b/include/eigensolve_quda.h
@@ -28,6 +28,7 @@ namespace quda
     int n_conv = 0;       /** Number of converged eigenvalues requested */
     int n_ev_deflate = 0; /** Number of converged eigenvalues to use in deflation */
     double tol = 0.0;     /** Tolerance on eigenvalues */
+    double qr_tol = 0.0;  /** Tolerance on QR (IRAM) */
     bool reverse = false; /** True if using polynomial acceleration */
     std::string spectrum; /** Part of the spectrum to be computed */
     bool compute_svd; /** Compute the SVD if requested **/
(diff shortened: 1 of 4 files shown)


maddyscientist (approved):

dmcdougall:
Let me give this a spin locally.

dmcdougall:
Before this change, the default `qr_tol` was `1e-11` and I was seeing IRAM choke because it encountered an invariant subspace in `DirectFull/EigensolveTest.verify/double_iram_direct_full_noSVD_LM`.  I was able to get a passing test by tightening up the tolerance to `1e-12`.  After this change, the `qr_tol` is internally set to `1e-14` because the solver requested a tolerance of `1e-12`.

I have a consistently passing test after this change.  Cheers @cpviolator!

dmcdougall:
Actually, I'm seeing that for `DirectFull/EigensolveTest.verify/double_iram_direct_full_noSVD_LI` I need to bump `--eig-qr-tol` from `1e-14` to `1e-15` to get it to converge.

Is it reasonable to modify the logic to be `qr_tol = tol * 1e-3` instead of `qr_tol = tol * 1e-2`?

cpviolator:
@dmcdougall Yeah, I think that's perfectly reasonable to do. Let me do a quick heuristic check to see how low one can go on the QR tol before it borks the routine. I don't want it to get stuck iterating because a user requests `eig_tol = 1e-14` and the QR grinds metal trying to satisfy an impossible `eig_qr_tol = 1e-17`.

dmcdougall:
@cpviolator Thanks.  Yeah, your concern is valid.  Thanks for looking into it.

dmcdougall (approved):

maddyscientist (approved):

PR #1552, opened 2025-04-08
merge rest of KernelOps updates

jcosborn:
This merges in the rest of the KernelOps infrastructure from the SYCL branch.
One question for discussion is whether `CHECK_SHARED_BYTES` in tunable_*.h should be on by default (and not optional).


include/kernel_ops_base.h
```diff
@@ -0,0 +1,165 @@
+#pragma once
```
maddyscientist, line 1: Wondering if this should go in `include/target/generic` ?

lib/llfat_quda.cu
```diff
@@ -15,7 +15,6 @@ namespace quda {
   class LongLink : public TunableKernel3D {
     LinkArg<Float, nColor, recon> arg;
     unsigned int minThreads() const { return arg.threads.x; }
-    unsigned int sharedBytesPerThread() const { return 4 * sizeof(int); } // for thread_array
```
maddyscientist, line 18: I assume this erroneously stated before?

lib/dslash_ndeg_twisted_clover.cu
```diff
@@ -24,7 +24,10 @@ namespace quda
 
       unsigned int sharedBytesPerThread() const
       {
-        return 2 * in.Ncolor() * 4 * sizeof(typename mapper<typename Arg::Float>::type);
+        if (arg.kernel_type == INTERIOR_KERNEL) {
```
maddyscientist: I think this needs to be 
```
if (arg.kernel_type == INTERIOR_KERNEL || arg.kernel_type == UBER_KERNEL) {
```

maddyscientist (approved):
This looks good to me.  A couple of comments, with one fix needed I think (UBER kernel comment for ndeg twisted clover).

include/kernel_ops_base.h
jcosborn: That would make sense, but then that directory would need to be added to the tests cmake.  I can do that if that is fine to do.

lib/llfat_quda.cu
jcosborn: Yes, that seems to be the case.

include/kernel_ops_base.h
maddyscientist: Apologies for being slow here, but why do the tests cmake need to see this?

jcosborn: The testsuite includes `tune_quda.h` which includes `kernel_ops_base.h`.  That's the reason it was split into the target independent and dependent headers.

maddyscientist: Worth asking then before we merge whether we fix the test suite not to include tune_quda.h.  Then we have a single kernel_ops.h file.

thoughts?

jcosborn: That seems not so easy with ones like tune_test.  One possibility is to split the tests into internal and external ones, with the latter not requiring any internal headers.  The internal ones could then be target specific. This is probably beyond the scope of this PR.  Maybe it is best to leave the headers as they are and discuss what to do with tests separately.

jcosborn:
I also fixed an issue with the shared memory tracking in `FusedMobiusDslash`.  I think it is correct now, but still have some questions about the interplay between `SharedMemory` and the `__shared__` in `BlockReduce` which we can discuss on the next Wed call.

include/kernel_ops_base.h
maddyscientist: Ok, agree, we can take this up after this PR is merged.

maddyscientist:
> I also fixed an issue with the shared memory tracking in `FusedMobiusDslash`. I think it is correct now, but still have some questions about the interplay between `SharedMemory` and the `__shared__` in `BlockReduce` which we can discuss on the next Wed call.

Regarding this open question, on the interplay between the dynamic shared memory declared in `SharedMemory` and the static shared memory used by `BlockReduce`, is this something to delay merging for?

Otherwise I'm good to merge now.

maddyscientist:
cscs-ci run

jcosborn:
There's no need to delay if everything else looks fine (invert_test_clover_asym_distance_pc seems to be barely failing 2 tests, so maybe needs tolerance increased a bit?).  The shared memory usage should be no different than before, which was presumably correct.

maddyscientist:
Yeah, the failure isn't caused by your PR.  It's cropped up now in other PRs too.  As you say an epsilon failure, and perhaps caused by a changed in compile stack at CSCS.

Will go ahead and merge this.

Thanks @jcosborn for the contribution.

PR #1555, opened 2025-04-14
Gauge flow update

leonhostetler:
This PR does the following:

1. Adds a fourth-order gauge flow integrator in addition to the existing third-order integrator
  
2. Fixes a bug in the topological charge calculation
  
3. Adds support for *anisotropic* Wilson and Symanzik flow
  
4. Adds support for rectangle calculation
  
5. Now prints spatial and temporal parts of the plaquette separately
  

More details about each are given below.

### Fourth-order Integrator

Implemented 6-stage fourth-order Runga-Kutta integrator in addition to the existing 3-stage third-order integrator. The integrator order is specified via a new parameter `unsigned int rk_order` in the `QudaGaugeSmearParam`. This is passed as a function parameter in ` WFlowStep()`. The algorithms are implemented in `lib/gauge_wilson_flow.cu` with the actual "flow step" functions implemented in `include/kernels/gauge_wilson_flow.cuh`.

Instead of having defining six separate functions for each of the six stages, I implemented a general function `computeWStep()` since the stages differ only in the two coefficients that are involved. This is the approach used in MILC's CPU implementation. Similarly, with appropriate choice of coefficients, one could refactor the 3-stage integrator to replace the three functions `computeW1Step()`, `computeW2Step()`, and `computeVtStep()` with three separate calls to the single function `computeWStep()`:

```cpp
computeW1Step() --> computeWStep(... 0, 1/4)
computeW2Step() --> computeWStep(... 17/32, 8/9)
computeVtStep() --> computeWStep(... 32/27, 3/4)
```

In the end, the result of applying the three function(s or function calls) one after the other is the same, however, the result is not the same at the end of any each step. Thus, if anyone were to call these functions individually and do something with the result between each step, refactoring as described above might mess that up. In fact, this kind of thing seems to be done in the function `performGFlowQuda()`, defined in `lib/interface_quda.cpp`. Since I don't have a ready-made application that runs `performGFlowQuda()` and so cannot test this function before and after making changes, I decided not to refactor the 3-stage algorithm and just leave it as it was.

### Topological Charge Bugfix

This bugfix fixes: [[MILC and QUDA Q charge discrepancy · Issue #959 · lattice/quda · GitHub](https://github.com/lattice/quda/issues/959)](https://github.com/lattice/quda/issues/959)

Issue recap: MILC and QUDA's topological charge differ in the 3rd or 4th significant digit, but converge to the same value at large flow times.

Solution: Make the `F[i]` matrices traceless before calculating the topological charge. Now QUDA and MILC agree to ~13 digits. In other words, replace

```
      for (int i=0; i<3; i++) {
        Qi[i] = getTrace(F[i] * F[5 - i]).real();
      }
```

with:

```
      for (int i=0; i<3; i++) {
         auto tmp_s = F[i] - n_inv * getTrace(F[i]) * iden;
         auto tmp_t = F[5 - i] - n_inv * getTrace(F[5 - i]) * iden;
         Qi[i] = getTrace(tmp_s * tmp_t).real();
       }
```

in `include/kernels/gauge_qcharge.cuh`. Since the matrices should be traceless in infinite precision, making them traceless apparently corrects the numerics. In fact, the `F[i]` *are* currently made traceless before the energy is computed, but the scope is a for loop, and so when the charge gets computed later (outside that loop), it uses `F[i]` which were not made traceless. Thus, why MILC and QUDA agree on the energy but not the charge.

### Anisotropic Flow

Anisotropic Wilson and Symanzik flow is enabled via a new parameter `double smear_anisotropy` in `QudaGaugeSmearParam`. The default is a value of 1.0 (no anisotropy). The flow anisotropy is implemented as in MILC's anisotropic CPU flow and tested/compared against the same.

The anisotropy is passed as a function parameter from `WFlowStep` down to `computeStaple` and `computeStapleRectangle` (with default value 1.0) in `include/kernels/gauge_utils.cuh`, where it multiplies the staples and rectangles in the temporal direction. Here, the type is set via `using real = typename Ftor::real;` since APE and stout classes, which use these functions, seem to compile with different precision. This required slight modifications to the `WFlow` and `APE` classes to expose their "real" types. Maybe there is a better way to set the type of the anisotropy parameter using type traits?

I have not updated any kernel `flops()` functions to account for the anisotropy addition, because the anisotropy only adds some scalar-matrix multiplications, and it looks like all the kernels that use `computeStaple`and`computeStapleRectangle` only count matrix-matrix multiplications when counting flops.

I have only implemented anisotropy for gauge flow, and not for APE or stout smearing. If this is something that would be useful, I could include that.

### Rectangle Observable

Like the plaquette, but 1x2 and 2x1 rectangles instead. Along with the plaquette and clover energy, this can be used to reconstruct improved observables. The rectangle observable is enabled via a new parameter `QudaBoolean compute_rectangle` in the `QudaGaugeObservableParam`. I added a new function `plaquetteRectangle()` modeled after the `plaquette` function, but this one returns a `double4` containing the spatial plaquette, temporal plaquette, spatial rectangle, and temporal rectangle. Since the plaquette can be obtained nearly for free at an intermediate stage of the rectangle calculation, it made sense for the rectangle function to return both. If `compute_rectangle` is false but `compute_plaquette` is true, then the original `plaquette()` function is called.

The rectangle implementation was achieved by creating a new class/kernel `GaugePlaqRect` that is essentially a copy of `GaugePlaq` expanded to compute also the rectangle. One might be able to combine the plaquette and plaquetteRectangle kernels into a single kernel to reduce code size and compilation time, but I wasn't sure how to do this easily in a way that maintains backward compatibility for the plaquette function.

### Misc.

In `performWFlowQuda`, we now print the spatial and temporal parts of the plaquette instead of just the sum. This is especially useful with the new option for anisotropic flow.

Note: Due to different normalization conventions, there is a factor of 3 between MILC and QUDA's plaquette and rectangle values.


maddyscientist:
@cpviolator can you review please?

include/gauge_tools.h
```diff
@@ -28,11 +28,19 @@ namespace quda
 
      @param[in] U The gauge field upon which to compute the plaquette
      @return double3 variable returning (plaquette, spatial plaquette,
-     temporal plaquette) site averages normalized such that each
-     plaquette is in the range [0,1]
```
cpviolator: Has this normalisation convention changed?

include/kernels/gauge_qcharge.cuh
```diff
@@ -70,7 +70,9 @@ namespace quda
       // unroll computation
 #pragma unroll
       for (int i=0; i<3; i++) {
-        Qi[i] = getTrace(F[i] * F[5 - i]).real();
```
cpviolator, line 73: An excellent insight, well done catching that!

lib/gauge_plaqrect.cu
```diff
@@ -0,0 +1,55 @@
...
+    long long flops() const
+    {
+      auto Nc = u.Ncolor();
+      // 10 mat-mat multiplies to compute plaquette and 2 rectangles
+      // Nc * Nc * (8 * Nc - 2) flops per mat-mat multiply
+      // Plus 2 traces ~ 2 * Nc flops
+      // All of the above * 6 (number of planes) * volume
+      return 6ll*u.Volume()*(10 * Nc * Nc * (8 * Nc - 2) + 2 * Nc);
+    }
+    // LEON FIXME: I don't know what I'm doing here
```
cpviolator: This is fine. It just measures the bytes of the gauge field `u` for throughput measurement in the profiler. One gauge goes in, and it's byte count is returned by `u.bytes()`

lib/gauge_wilson_flow.cu
```diff
@@ -108,11 +152,13 @@ namespace quda {
       long long threads = in.LocalVolume() * wflow_dim;
       long long mat_flops = nColor * nColor * (8 * nColor - 2);
       long long mat_muls = 1; // 1 comes from Z * conj(U) term
-      switch (wflow_type) {
+      switch (wflow_type) { // Add mat-muls coming from staple calculation
       case QUDA_GAUGE_SMEAR_WILSON_FLOW: mat_muls += 4 * (wflow_dim - 1); break;
       case QUDA_GAUGE_SMEAR_SYMANZIK_FLOW: mat_muls += 28 * (wflow_dim - 1); break;
       default : errorQuda("Unknown Wilson Flow type");
       }
+      // Leon: This seems to miss the two mat-muls occurring in exponentiate_iQ(Z)
```
cpviolator: This is easily fixable. Shall we bump it up by three as you suggest?

lib/gauge_wilson_flow.cu
```diff
@@ -124,12 +170,17 @@ namespace quda {
       case QUDA_GAUGE_SMEAR_SYMANZIK_FLOW: links = 24; break;
       default : errorQuda("Unknown Wilson Flow type");
       }
-      auto temp_io = step_type == WFLOW_STEP_W2 ? 2 : step_type == WFLOW_STEP_VT ? 1 : 0;
+      // Leon: I am not certain that the byte counting is correct here!
+      // First and last steps have 1 store (retrieve) to (from) temp
+      auto temp_io = 1;
+      // Middle steps have an additional store or retrieve to or from temp
+      if( step_type==WFLOW_STEP_W2 || step_type==WFLOW_4THORDER_STEP_2 || step_type==WFLOW_4THORDER_STEP_3 
+      || step_type==WFLOW_4THORDER_STEP_4 || step_type==WFLOW_4THORDER_STEP_5 ); temp_io += 1;
```
cpviolator: This line is causing the ROCM compiler to bork. Is that semi-colon intentional?

cpviolator:
@leonhostetler What a superb PR, and an excellent addition to QUDA. I did a pass over the file diff and left a couple of comments.

I'm almost certain that the fragmented implementation of the flow in `performGFlowQuda` will be unaffected by replacing the explicit calls to W1, W2, and VT. It only does that to flow the fermion field along with the gauge. @maddyscientist do you know who wrote that? We could implement the change and completely get rid of the W1, W2, VT internal enums and code, and use the much cleaner approach in this PR, then ask them to do a quick test themselves. I can also hack it into the options for `su3_test` and eyeball data to make sure all is well.

@leonhostetler, how much work would be involved implementing anisotropy for APE and Stout? I think that would be an excellent addition too, but not if it means an excessive time sink on your end. I'll tag on some options somewhere to actually run `performGFlowTest` and confirm that the change in function calls does not affect data, then you can delete code and refactor code as you see fit. Sound like a plan?

include/gauge_tools.h
leonhostetler: No, this is my mistake. I had briefly changed the plaquette normalization to match MILC's (which is not normalized to the range [0,1]), but in the end I reverted that change and forgot to revert this description. I have fixed it.

lib/gauge_plaqrect.cu
leonhostetler: Ah, ok. I wasn't sure how if I should multiply by the number of unique plaquettes requested or not. I have removed the comment.

lib/gauge_wilson_flow.cu
leonhostetler: I bumped it up by three

lib/gauge_wilson_flow.cu
leonhostetler: Arg! I'm flabbergasted that my compilations did not bork. I have removed the semi-colon

leonhostetler:
@cpviolator Thanks for looking at this! Implementing anisotropy for APE and Stout should be straightforward. Since it's already in `computeStaple` and `computeStapleRectangle`, it's just a matter of propagating it as a function parameter down to that level. I had hesitated since I did not know if anybody would find it useful and because I do not have a third-party application to verify against like I do for wilson and symanzik. But I can certainly add it.

Re: refactoring... That sounds good to me!

cpviolator:
@leonhostetler @maddyscientist I'm going to add a host call to `performGFlowQuda` to the `su3_test.cpp` file, and add it to the gtest suite code in this PR. That should be enough to actually eyeball data from `performGFlowQuda` and confirm that refactoring has not affected the pipeline. Then we can finish up this PR and I can crack on with some host side verification for the smearing so that both `performGFlowQuda` and `performWFlowQuda` have adequate host side verification.

kostrzewa:
> I'm almost certain that the fragmented implementation of the flow in performGFlowQuda will be unaffected by replacing the explicit calls to W1, W2, and VT. It only does that to flow the fermion field along with the gauge. @maddyscientist do you know who wrote that?

The fermion flow is by @aniketsen.

aniketsen:
> We could implement the change and completely get rid of the W1, W2, VT internal enums and code, and use the much cleaner approach in this PR, then ask them to do a quick test themselves. 

sounds good. This should not change anything in the `performGFlowQuda` as long as the W1, W2 and Vt steps are translated correctly. But I can also test the correctness afterwards.

cpviolator:
In a related PR, the author has created a file [su3_fermion_update.cpp](https://github.com/lattice/quda/pull/1526/files#diff-5c8aa069ba08cafef4eb156c58f768355c649789c9de8093a8fbd9924a3afcac) which would suit very nicely for testing `performGFlowQuda` too. @leonhostetler If it's not too much trouble, may we wait until a couple of iterations on that have been done, then I can modify that file to create a test script. It would work out well if you just went ahead and made the changes you suggested to this branch (removing all the W1, W2, VT stuff) and checking that the pure gauge smearing routines are unaffected, and then I can use the old W1, W2, VT version with a host test `performGFlowQuda` in the above PR, merge in the host test code to this PR and make sure everything is the same. That should keep us both busy ish while not wasting time.

cpviolator:
@aniketsen Thanks for the clarification there and offer of help for code testing. I'll ping you when we have something for you to build against.

cpviolator:
@leonhostetler Also, having taken a proper look, I agree that one could save on LOC by merging `Palquette` and `PalaquetteRectangle` but then one would have a few too many awkward booleans making it less readable. Personally, I like how you wrote `PalquetteRectangle` stating the goal at the top and performing the action with minimal thrashing and spillage, and each `.cuh` file does its own job very well. I may in the near future add a Lucher-Weiss action, and it would be much simpler to write another `.cuh` file.

cpviolator:
As discussed offline with @leonhostetler @maddyscientist @weinbe2 @rkarur and @cpviolator, this PR offers a lot of scope for further work, i.e. #1557 #1556,  but presents no immediate blockers as it stands for merging with develop.

cpviolator:
@leonhostetler When you add yourself to the QUDA contributor list in the README.md file, can you please change
```
Dean Howarth (Lawrence Livermore Lab, Lawrence Berkeley Lab)
```
to
```
Dean Howarth (Cahill Center for Astronomy and Astrophysics, Caltech)
```

cpviolator (approved):
All looks superb to me. An excellent new addition to QUDA.

include/enum_quda.h
```diff
@@ -640,6 +640,12 @@ typedef enum QudaWFlowStepType_s {
   WFLOW_STEP_W1,
   WFLOW_STEP_W2,
   WFLOW_STEP_VT,
+  WFLOW_4THORDER_STEP_1,
```
maddyscientist, line 643: `4THORDER` is causing my brain to break (I'm reading it as 4 thorder).  Can you change to `FOURTH_ORDER` or `4TH_ORDER`?

include/kernels/gauge_wilson_flow.cuh
```diff
@@ -164,6 +190,12 @@ namespace quda
       case WFLOW_STEP_W1: Z = computeW1Step(*this, U, x, parity, x_cb, dir); break;
       case WFLOW_STEP_W2: Z = computeW2Step(*this, U, x, parity, x_cb, dir); break;
       case WFLOW_STEP_VT: Z = computeVtStep(*this, U, x, parity, x_cb, dir); break;
+      case WFLOW_4THORDER_STEP_1: Z = computeWStep(*this, U, x, parity, x_cb, dir, 0.0, 0.032918605146, false, true); break;
```
maddyscientist, line 193: I wonder if writing this as something like the below makes this a bit easier to parse.
```
double coeff_a[] = {0.737101392796, 1.634740794341, ...};
double coeff_b[] = {0.823256998200, 0.381530948900, ...};
...
case WFLOW_4THORDER_STEP_2: Z = computeWStep(*this, U, x, parity, x_cb, dir, coeff_a[0], coeff_b[0], true, true); break;
case WFLOW_4THORDER_STEP_3: Z = computeWStep(*this, U, x, parity, x_cb, dir, coeff_a[1], coeff_b[1], true, true); break;
...
```

weinbe2:
cscs-ci run

lib/gauge_stout.cu
```diff
@@ -26,14 +27,15 @@ namespace quda {
 
   public:
     // (2,3/4): 2 for parity in the y thread dim, 3 or 4 corresponds to mapping direction to the z thread dim
-    GaugeSTOUT(GaugeField &out, const GaugeField &in, bool improved, double rho, double epsilon, int dir_ignore) :
+    GaugeSTOUT(GaugeField &out, const GaugeField &in, bool improved, double rho, double epsilon, int dir_ignore, const double anisotropy) :
```
maddyscientist, line 30: There's no need to make a pass-by-value variable `const` (I know a bunch of code does this already in QUDA, but we're working towards fixing this).

lib/interface_quda.cpp
```diff
@@ -5208,7 +5213,7 @@ void performGaugeSmearQuda(QudaGaugeSmearParam *smear_param, QudaGaugeObservable
 void performWFlowQuda(QudaGaugeSmearParam *smear_param, QudaGaugeObservableParam *obs_param)
 {
   auto profile = pushProfile(profileWFlow);
-  pushOutputPrefix("performWFlowQuda: ");
+  pushOutputPrefix("");
```
maddyscientist, line 5216: Why this change?

weinbe2:
FYI, @leonhostetler , I hit merge too quickly---going to revert b/c @maddyscientist has a few comments

lib/interface_quda.cpp
```diff
@@ -5233,22 +5238,61 @@ void performWFlowQuda(QudaGaugeSmearParam *smear_param, QudaGaugeObservableParam
 
   gaugeObservables(in, obs_param[measurement_n]);
 
-  logQuda(QUDA_SUMMARIZE, "flow t, plaquette, E_tot, E_spatial, E_temporal, Q charge\n");
-  logQuda(QUDA_SUMMARIZE, "%le %.16e %+.16e %+.16e %+.16e %+.16e\n", smear_param->t0, obs_param[0].plaquette[0],
-          obs_param[0].energy[0], obs_param[0].energy[1], obs_param[0].energy[2], obs_param[0].qcharge);
+  QudaBoolean compute_plaq = obs_param[measurement_n].compute_plaquette;
```
maddyscientist, line 5241: We're working towards using `bool` throughout QUDA.  Can you switch these to `auto` so that they automatically pick up this change?

maddyscientist (changes requested):
Mostly good, some style comments.

Also, I see that not all files have been clang-formatted.  Specifically the CUDA ones: you need to do the modifications to the clang format Python script outlined here: https://github.com/lattice/quda/wiki/Coding-Conventions-and-Style#applying-it-to-lines-changes-in-a-pull-request

PR #1565, opened 2025-05-06
Multi-rhs MILC HISQ MG interface

weinbe2:
This PR first and foremost exposes multi-rhs HISQ MG in the MILC interface via a new function `qudaInvertMsrcMG`. This is paired with the MILC branch at https://github.com/lattice/milc_qcd/tree/feature/milc-multirhs-mg ; I'll open a PR for that once this work is done.

This PR does a few other things:

- Re-writes `qudaInvertMG` to be a thin wrapper around `qudaInvertMsrcMG`, reducing code duplication
- Moves many of the internal details of the MILC MG interface into a pair of new files, `milc_interface_internal.[cpp/hpp]`. This mostly comes down to moving the routines and structs for parsing the `mgparams.txt` input file and populating the standard invert params, MG params, etc structs.
- Exposes the option to use block TRLM for coarsest-level deflation through `mgparams.txt` with the new `deflate_block_size` argument; the default is `1` which is sticking with regular TRLM
- Exposes the option to set a batch size for multi-rhs MG solves with the new `block_solver_batch_size` argument. The default is to solve all rhs in one go, but by setting some integer value solves will be performed in batches of that size regardless of the number of rhs MILC passes in
- Tangentially related, this PR makes the non-MG `qudaInvert` a thin wrapper around `qudaInvertMsrc`, reducing code duplication.

Outstanding tasks:

- [ ] Testing RHMC (this PR should be nearly completely orthogonal, but may as well)
- [ ] Update the wiki page on MILC HISQ MG (https://github.com/lattice/quda/wiki/HISQ-MG-for-Measurements) to document the new flags and provide a reference input file that supports block solves


include/quda_milc_interface.h
```diff
@@ -487,6 +487,37 @@ extern "C" {
...
+   * @param[in] mass Fermion mass parameter
+   * @param[in] inv_args Struct setting some solver metadata
+   * @param[in] target_residual Target residual
+   * @param[in] target_relative_residual Target Fermilab residual
+   * @param[in] milc_fatlink Fat-link field on the host
+   * @param[in] milc_longlink Long-link field on the host
+   * @param[in] mg_pack_ptr MG preconditioner structure created by qudaSetupMultigrid
+   * @param[in] mg_rebuild_type whether to do a full (1) or thin (0) MG rebuild
+   * @param[in] sourceArray Array of right-hand side source fields
+   * @param[out] solutionArray Array of solution spinor fields
```
maddyscientist, line 509: Should this be `[in,out]` since the solution can be read in the case of initial guesses?

maddyscientist (approved):
This looks like a nice clean PR, thanks @weinbe2.

Beyond a trivial comment, this passes a visual review.  I defer the MILC testing aspect to @leonhostetler.

include/quda_milc_interface.h
weinbe2: As I currently implemented things it doesn't support an initial guess, for better or for worse: https://github.com/lattice/quda/blob/feature/milc-multirhs-mg/lib/milc_interface.cpp#L1857-L1859

This is downstream of how QUDA and MILC have opposite sign conventions for the full HISQ dslash operator, which I deal with on the solution here: https://github.com/lattice/quda/blob/feature/milc-multirhs-mg/lib/milc_interface.cpp#L1863-L1873

Though I can also pop that onto the initial guess and then pass it in, my reason at the time was self-documented (MG is so fast so why bother). In any case I should pop some host OpenMP on these things, I'll take care of that in a bit.

leonhostetler (approved):
An excellent addition that I think makes staggered MG fully production ready.

I ran a battery of tests using MILC's ks_spectrum application:
- MG solves with multiple right-hand sides on a 96^3x192 lattice
- The above also with block Lanczos (deflate_block_size = 8) for the deflation at the coarse level
- The above also with block_solver_batch_size less than the total number of RHS
- For good measure, I also retested non-MG solves (i.e. qudaInvert and qudaInvertMsrc) using my workflow for ks_spectrum GB baryons

All tests look good!

weinbe2:
Thanks for the thorough testing, @leonhostetler , and for the reviews from you and @maddyscientist . I'll test RHMC today and then assuming that goes through I'll merge. @maddyscientist, I'm going to put off the initial guess fix until a subsequent PR so I can check perf/make sure there are no unintended surprises. I'll file an issue for it later today.

include/quda_milc_interface.h
weinbe2: This is now being tracked in https://github.com/lattice/quda/issues/1566 , I'll take care of it in a subsequent PR.

maddyscientist: Sounds good @weinbe2.  Thanks.

maddyscientist:
Good to merge by me post RHMC verification.  Great work @weinbe2!

weinbe2:
I can confirm RHMC goes through with identical results and iso-perf, so I'm going to merge!

PR #1569, opened 2025-05-20
Add the CMake option QUDA_MAX_KERNEL_ARG_SIZE.

### LINKED ISSUE #1568: Too large kernel argument size reduces the performance after MRHS commits on AMD GPUs
The MRHS code was first introduced in #1465. Kernel argument sizes have been increased since then. Performance regressions (especially in dslash kernels) are found on AMD devices in new commits, and can mostly be fixed by the [patch](https://github.com/SaltyChiang/Build-USQCD-SciDAC/blob/master/patch/quda_dtk_performance.patch) (not widely tested, but works in our use case).

Kernel arguments can be passed by simply setting parameters in hipLaunchKernel, or by copying argument structs into constant device memory. On AMD GPUs, SGPR/VGPR will be used to handle the argument struct in the former method, which should be the optimized situation for kernel functions accessing the arguments. However, if too many registers are occupied, the kernel function will have to use the slower cache for variables, which will slow down the kernel significantly. The above patch forces kernels to use the latter method, which is not the best solution for all kernels, and we still notice some regression with many GPUs.

Recently @leonhostetler found a similar situation on the newer AMD architecture. A CMake option to control the max_kernel_arg_size should be a good workaround for this issue.

SaltyChiang:
As mentioned in #1568, adding the CMake option QUDA_MAX_KERNEL_ARG_SIZE to manually control the return value of `max_kernel_arg_size` might be helpful on AMD devices.

This is implemented by:
- Moving definition of `!use_kernel_arg<Arg>()` version `get_arg()` and `get_constant_buffer()` to target_device.h
- Make declariation of buffer `extern __constant__ char` and define it in constant_kernel_arg.cu
- Remove `if (location == QUDA_CUDA_FIELD_LOCATION && use_constant_memory()) strcat(aux, "cmem,");` line because we always enable the potential constant memory argument

PS:
I noticed that the new mechanism to separately build dslash instantiates always builds kernels with and without distance preconditioning if QUDA_DSLASH_DISTANCE_PRECONDITIONING is enabled. This is not necessary for dslash kernels which are not Wilson-type. I'm trying to find a way to ensure distance PC will only be applied to Wilson and clover dslashes.


CMakeLists.txt
```diff
@@ -222,6 +222,11 @@ if(QUDA_MAX_MULTI_RHS_TILE GREATER QUDA_MAX_MULTI_RHS)
   message(SEND_ERROR "QUDA_MAX_MULTI_RHS_TILE is greater than QUDA_MAX_MULTI_RHS")
 endif()
 
+set(QUDA_MAX_KERNEL_ARG_SIZE "4096" CACHE STRING "maximum static size of the kernel arguments in bytes passed to a kernel on the target architecture")
+if(QUDA_MAX_KERNEL_ARG_SIZE GREATER 8192)
```
maddyscientist: The maximum kernel arg size supported on CUDA is 32,764 bytes.  We should not be hobbling the limit to 8,192 bytes.  By default at present QUDA will support large kernel arguments when running on CUDA 12.1 or above (see the CMake parameter `QUDA_LARGE_KERNEL_ARG`).  This behavior should be retained.

maddyscientist (changes requested):
I'm not sure I fully understand the details of this PR.
* I think this might break NVSHMEM when constant memory is used (NVSHMEM uses RDC and requires only a single definition)
* To avoid the overhead of defining the constant memory in all translation units, there should be no constant memory defined if `QUDA_LARGE_KERNEL_ARG=ON` to avoid any unnecessary overhead for CUDA
* Given that the use of constant memory is threshold defined, we should retain the markup of the tune string if constant memory is used or not.  This allows us to make easier comparisons with a single tunecache when changing the threshold.

I would think a simpler approach is to just include constant_kernel_arg.h in kernel.h, and enable the use of the threshold constant memory with the CMake defined threshold.

SaltyChiang:
> * I think this might break NVSHMEM when constant memory is used (NVSHMEM uses RDC and requires only a single definition)

I declared `buffer` as `extern` in target_device.h, and defined it in constant_kernel_arg.cu, which should be the only definition. This should be similar to the NVSHMEM situation, as it declared `buffer` in const_kernel_arg.h and defined it in dslash_constant_arg.cu (renamed to constant_kernel_arg.cu now)

The reason why I didn't simply include constant_kernel_arg.h in kernel.h is that I was thinking every kernel .cu file will have its own definition of `buffer` if I do so, which looks unnecessary as we never call two kernels at the same time, so a single definition should be enough and reduce the memory usage.

But I might have some misunderstanding about the constant memory. Please correct me if I have any mistakes.



> I would think a simpler approach is to just include constant_kernel_arg.h in kernel.h, and enable the use of the threshold constant memory with the CMake defined threshold.

That makes sense. I will try to add the information in the aux string.

CMakeLists.txt
SaltyChiang: If `QUDA_LARGE_KERNEL_ARG` is defined, `max_kernel_arg_size()` will return 32764 no matter what number is set here, as it will call `max_constant_size` in this case. `QUDA_MAX_KERNEL_ARG_SIZE` changes the return value of `max_kernel_arg_legacy_size()`.

8K is just an arbitrary value. What should be the proper value here? Is 32K OK for both NVIDIA and AMD? Or I should set the maximum value depending on the target?

SaltyChiang:
> * Given that the use of constant memory is threshold defined, we should retain the markup of the tune string if constant memory is used or not.  This allows us to make easier comparisons with a single tunecache when changing the threshold.

I suppose that you want to add `"cmem"` to the tuning string to identify if the kernel is launched with the argument struct or the constant memory. The problem here is, we do not know the argument size until launching the kernel, and it's too late to add the tag to aux string then.

A trivial idea here is adding the `QUDA_MAX_KERNEL_ARG_SIZE` value to the aux string. When `QUDA_LARGE_KERNEL_ARG ` is enabled, it should be something like `large_kernel_arg` instead of a number.

SaltyChiang:
A question:
Why default value of `QUDA_COVDEV` not controlled by `QUDA_DIRAC_DEFAULT`? I think rename it to `QUDA_DIRAC_COVDEV` and set its default value by `QUDA_DIRAC_DEFAULT` just like `QUDA_DIRAC_LAPLACE` is reasonable.

maddyscientist:
> A question:
> 
> Why default value of `QUDA_COVDEV` not controlled by `QUDA_DIRAC_DEFAULT`? I think rename it to `QUDA_DIRAC_COVDEV` and set its default value by `QUDA_DIRAC_DEFAULT` just like `QUDA_DIRAC_LAPLACE` is reasonable.

Yes, this would be a good idea.  Thanks.

maddyscientist:
> > * Given that the use of constant memory is threshold defined, we should retain the markup of the tune string if constant memory is used or not.  This allows us to make easier comparisons with a single tunecache when changing the threshold.
> 
> 
> 
> I suppose that you want to add `"cmem"` to the tuning string to identify if the kernel is launched with the argument struct or the constant memory. The problem here is, we do not know the argument size until launching the kernel, and it's too late to add the tag to aux string then.
> 

That's a good point, I hadn't thought of that. 😄

> A trivial idea here is adding the `QUDA_MAX_KERNEL_ARG_SIZE` value to the aux string. When `QUDA_LARGE_KERNEL_ARG ` is enabled, it should be something like `large_kernel_arg` instead of a number.

That's an easy work around, so yeah  lets go with that.  Something t like `arg_threshold=xxxxx`.

CMakeLists.txt
```diff
@@ -222,6 +221,11 @@ if(QUDA_MAX_MULTI_RHS_TILE GREATER QUDA_MAX_MULTI_RHS)
   message(SEND_ERROR "QUDA_MAX_MULTI_RHS_TILE is greater than QUDA_MAX_MULTI_RHS")
 endif()
 
+set(QUDA_MAX_KERNEL_ARG_SIZE "4096" CACHE STRING "maximum static size of the kernel arguments in bytes passed to a kernel on the target architecture")
```
maddyscientist, line 224: Default for CUDA should be 32,764 bytes not 4096.

CMakeLists.txt
```diff
@@ -222,6 +221,11 @@ if(QUDA_MAX_MULTI_RHS_TILE GREATER QUDA_MAX_MULTI_RHS)
...
+if(QUDA_MAX_KERNEL_ARG_SIZE GREATER 32768)
```
maddyscientist: And CUDA max is 32,764 bytes

include/targets/cuda/tunable_kernel.h
```diff
@@ -23,18 +23,18 @@ namespace quda
...
-  */
-  static bool use_constant_memory()
-  {
-#ifdef QUDA_USE_CONSTANT_MEMORY
-    return true;
-#else
-    return false;
-#endif
-  }
+//   /**
```
maddyscientist: I guess delete this code block rather than comment it.

include/targets/hip/tunable_kernel.h
```diff
@@ -20,18 +20,18 @@ namespace quda
...
-  */
-  static bool use_constant_memory()
-  {
-#ifdef QUDA_USE_CONSTANT_MEMORY
-    return true;
-#else
-    return false;
-#endif
-  }
+//   /**
```
maddyscientist: Delete code block

maddyscientist:
I think this is looking good now.  Beyond the comments I've already made, the only other thing to do I think is that the `__constant__` memory declaration should be hidden if `QUDA_LARGE_KERNEL_ARG=ON` since in this case, we should never use the constant memory, so we should be able to hide it so that we don't waste memory unnecessarily.

CMakeLists.txt
SaltyChiang: I set the default value to 4096 here because I'm trying to make the default value the same as before. It was 4096 for both CUDA and HIP when large kernel arg is not enabled. CUDA will always return 32764 if the large kernel arg is enabled whatever value of `QUDA_MAX_KERNEL_ARG_SIZE ` is set.

SaltyChiang:
> I think this is looking good now. Beyond the comments I've already made, the only other thing to do I think is that the `__constant__` memory declaration should be hidden if `QUDA_LARGE_KERNEL_ARG=ON` since in this case, we should never use the constant memory, so we should be able to hide it so that we don't waste memory unnecessarily.

I believe this is already implemented. When large kernel arg is enabled (which is only possible for CUDA), https://github.com/lattice/quda/blob/237bb0ace58087cdac85574fe43c9450b234be52/include/targets/cuda/constant_kernel_arg.h#L14-L20 ensures no buffer will be allocated.

include/targets/hip/tunable_kernel.h
SaltyChiang: [3a4f46f](https://github.com/lattice/quda/pull/1569/commits/3a4f46f29c5000376812d5c0b6dd30817687d2fa)

include/targets/cuda/tunable_kernel.h
SaltyChiang: [3a4f46f](https://github.com/lattice/quda/pull/1569/commits/3a4f46f29c5000376812d5c0b6dd30817687d2fa)

CMakeLists.txt
SaltyChiang: [ea6f025](https://github.com/lattice/quda/pull/1569/commits/ea6f02508d580748cb33db1b132f9410eae659b9)

maddyscientist:
@leonhostetler can you review this please before we merge?

leonhostetler:
> @leonhostetler can you review this please before we merge?

It looks good to me. I tested this with `staggered_eigensolve_test` on Frontier for a small case (1 node) and intermediate case (27 nodes), and the post-MRHS performance regression detailed in issue #1568 appears to be resolved when compiled with `QUDA_MAX_KERNEL_ARG_SIZE=0`. For good measure, I also tested this branch on Perlmutter. Thanks for this @SaltyChiang!

dmcdougall (approved):
This change ought not to be needed at all.  If there's a performance regression in ROCm we need to investigate where and why it happened to reason about whether or not it was due to an engineering tradeoff or simply due to oversight.

Nevertheless, that's on me to investigate, and this PR gets folks their performance back as a stop-gap until the regression is root-caused.

I'd like to thank @SaltyChiang and @leonhostetler for putting together the PR and for testing it out on Frontier.  I really appreciate that.

maddyscientist (approved):

PR #1570, opened 2025-05-20
added multifermion changes

rkarur:
Added multi-fermion support for adjoint flow algorithms.

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -1701,33 +1701,38 @@ extern "C" {
    * Performs Gradient Flow (gauge + fermion) on gaugePrecise and stores it in gaugeSmeared
    * @param[out] h_out Output fermion field set
    * @param[in] h_in Input fermion field set
+   * @param[in] inv_param Dirac/Laplacian and solver meta data
    * @param[in] smear_param Parameter struct that defines the computation parameters
    * @param[in,out] obs_param Parameter struct that defines which
    * observables we are making and the resulting observables.
    * @param[in] nSpinors Number of spinors in the input and output fields
    */
   void performGFlowQuda(void **h_out, void **h_in, QudaInvertParam *inv_param, QudaGaugeSmearParam *smear_param,
-                        QudaGaugeObservableParam *obs_param, const size_t nSpinors);
+                        QudaGaugeObservableParam *obs_param, size_t nSpinors);
 
   /**
    * Performs Adjoint Gradient Flow (gauge + fermion) the "safe" way on gaugePrecise and stores it in gaugeSmeared
-   * @param[out] h_out Output fermion field
-   * @param[in] h_in Input fermion field
+   * @param[out] h_out Output fermion field set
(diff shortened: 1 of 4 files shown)


include/quda.h
```diff
@@ -1717,7 +1717,7 @@ extern "C" {
    * @param[in,out] obs_param Parameter struct that defines which
    * observables we are making and the resulting observables.
    */
-  void performAdjGFlowSafe(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaGaugeSmearParam *smear_param);
+  void performAdjGFlowSafe(void **h_out, void **h_in, QudaInvertParam *inv_param, QudaGaugeSmearParam *smear_param, const size_t nSpinors);
```
maddyscientist: The doxygen doesn't match the function signature, and the `nSpinors` argument needs to be added to the doxygen.

maddyscientist: Also, `nSpinors` does not need to be `const` here since this is passed by value.

include/quda.h
```diff
@@ -1727,7 +1727,7 @@ extern "C" {
    * @param[in,out] obs_param Parameter struct that defines which
    * observables we are making and the resulting observables.
    */
-  void performAdjGFlowHier(void *h_out, void *h_in, QudaInvertParam *inv_param, QudaGaugeSmearParam *smear_param);
+  void performAdjGFlowHier(void **h_out, void **h_in, QudaInvertParam *inv_param, QudaGaugeSmearParam *smear_param, const size_t nSpinors);
```
maddyscientist: Same as above: need to make the doxygen consistent with the function and remove `const` from `nSpinors`.

lib/interface_quda.cpp
```diff
@@ -5624,11 +5627,11 @@ void adjSafeEvolve(std::vector<std::reference_wrapper<ColorSpinorField>> sf_list
   GaugeField &gaugeTemp = gf_list[4].get();
   GaugeField &precise = gf_list[5].get();
 
-  ColorSpinorField &f_temp0 = sf_list[0].get();
-  ColorSpinorField &f_temp1 = sf_list[1].get();
-  ColorSpinorField &f_temp2 = sf_list[2].get();
-  ColorSpinorField &f_temp3 = sf_list[3].get();
-  ColorSpinorField &f_temp4 = sf_list[4].get();
+  std::vector<ColorSpinorField> &f_temp0 = sf_list[0].get();
```
maddyscientist: I would probably declare these as `auto &`.

maddyscientist (changes requested):
Easy PR to review 😄 

Aside from my comments, just needs a clang-format.

maddyscientist:
cscs-ci run

rkarur:
modified tests to actually take multiple fermions, as per @maddyscientist 's suggestion

maddyscientist (approved):

PR #1571, opened 2025-05-21
Misc hotfix: nvshmem Ls > 1 packer, compiler errors, tunecache, atomic gauge fix policies

### LINKED ISSUE #1558: If tuning is off and resource path is available, the current tunecache.tsv will be overwrited
If `QUDA_ENABLE_TUNE=0` and `QUDA_RESOURCE_PATH=/some/available/path` are set, QUDA will write the tune cache to `tunecache.tsv` with default parameters after #1531. However, it will overwrite the current cache file even though optimized parameters are already stored in it. I lost some multigrid parameters after trying to debug with some eigensolve functions without enabling autotuning.

Is this behavior by design? I think keeping the current optimized parameters is useful.

weinbe2:
This mini, hostfix-ish PR addresses four things:

1. Fixes a bug in the inlined halo packer (nvshmem) relevant for ndeg twisted fermions and dwf
2. Fixes a compile-time error triggered in clang and nvc++ when compiling the dwf host verify routines
3. Saving the tunecache to a different name if the resource path is set but tuning is disabled (closes #1558 )
4. Removes the gauge fixing policies that use shared memory atomics to eliminate a source of non-determinism

The first three were cherry-picked from ongoing work by @maddyscientist and have already been verified to fix the issues; the last was debugged and had correctness ensured as I went.


maddyscientist (approved):
Approving a PR that's mostly my commits feels a little dishonest 😄

weinbe2:
> Approving a PR that's mostly my commits feels a little dishonest 😄

That's... fair. To each point, though:
1. I've confirmed that fix myself, I just happened to cherry-pick it here.
2. I confirmed this for nvc++, and we've addressed this issue in the past (and just accidentally re-introduced it) in the context of clang and just repeated the fix (and I believe was confirmed by @SaltyChiang ?)
3. This was my contribution, which you've approved, and
4. Was reviewed by me and also I believe confirmed to work by @SaltyChiang 

So I think this is all reasonable. That said, for due diligence --- @SaltyChiang , just confirming, removing `complex.h` from the domain wall host verify file worked for you, as did the `tunecache_notune.tsv` fix?

SaltyChiang:
> So I think this is all reasonable. That said, for due diligence --- @SaltyChiang , just confirming, removing `complex.h` from the domain wall host verify file worked for you, as did the `tunecache_notune.tsv` fix?

Yes. I manually removed the `complex.h` line and passed the compilation.

PR #1572, opened 2025-06-04
Hotfix/link

maddyscientist:
Fixes some minor issues
* Link failures caused when enabling clover Hasenbusch Dirac operator
* Remove some deprecated CMake variables causing CMake error related to `clover_force_test`
* Add missing fast_compile str for dslash kernels


weinbe2:
cscs-ci run

maddyscientist:
I have pushed a fix for the failing ctest.  As of #1569, distance preconditioning is now included in the ctest runs, but this was failing.  I tracked this to the cosh * 1/cosh transformation not being exactly unitary, it is an extensive quantity that scales as ~ epsilon * dof.  This factor is now taken into account when testing for solver correctness.

maddyscientist:
cscs-ci run

weinbe2:
This looks good and indeed fixes the linker issues, I did hit an error running `ctest` because I _only_ compiled Wilson and twisted Hasenbusch:

```
9: ERROR: Clover has not been built (rank 0, host viking-dvt-151.nvidia.com, copy_clover.cu:158 in void quda::copyGenericClover(quda::CloverField&, const quda::CloverField&, bool, QudaFieldLocation, void*, const void*)())
```

Which I'm guessing can be fixed by just updating this codepath to use `if constexpr (is_enabled_clover())`, or something of the sort: https://github.com/lattice/quda/blob/develop/lib/copy_clover.cu#L148-L160

(I'd fix this myself, but I want to avoid the conflict of interest, etc)

weinbe2 (approved):
With the clover test fix, this is now good to go. Noting for the record that the cscs failure is unrelated and all other tests are passing. Thanks for the PR @maddyscientist

PR #1573, opened 2025-06-10
Feature/byte array

maddyscientist:
Small feature PR that adds a new class `byte_array` which is an indexable array type for 8-bit types (signed / unsigned chars) of length 8.  This is useful, as it's exactly what we use for the various kernels that do path tracing around the links of the lattice.

With this class, we can use it instead of `thread_array`, the advantages of `byte_array` being that it is lower latency, and does not consume any shared memory.  E.g., saving more L1 cache which can be helpful for performance.  On the kernels I tested, I saw either performance neutrality or a perf uplift.

Also included here:
* Remove use of KernelOps where possible due to removal of shared-memory storage from using `byte_array` instead of `thread_array`
* Fix stack frame induced in `DeGrandRossi` contraction kernel (was due to some run-time indexing)
* Fix stack frame induced in pure-gauge heath bath kernel (was due to run-time indexing in local array)
* Fix stack frame in plaquette rectangle action (run-time indexing and running out of registers)

### DIFF
--- a/include/kernels/contraction.cuh
+++ b/include/kernels/contraction.cuh
@@ -92,6 +92,7 @@ namespace quda
     template <typename U> static inline void comm_reduce(U &) { }
 
     // Final param is unused in the MultiReduce functor in this use case.
+    template <int s1>
     __device__ __host__ inline reduce_t operator()(reduce_t &result, int xyz, int, int t)
     {
       constexpr int nSpin = Arg::nSpin;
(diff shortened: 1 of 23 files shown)


maddyscientist:
@jcosborn FYI, this PR removes the use of `thread_array` (and thus `KernelOps`) from a whole much of kernels.  This approach is faster though, and takes less resources.

include/kernels/contraction.cuh
```diff
@@ -92,6 +92,7 @@ namespace quda
     template <typename U> static inline void comm_reduce(U &) { }
```
weinbe2, line 92: ~Was this file supposed to get pulled in?~

weinbe2: Nevermind, disregard.

weinbe2 (approved):
This looks good and passes my various tests (including correctness) with large (> 255) local dimensions, focusing on fat/long link construction. I'm seeing ~perf parity on H100-80GB-SXM5.

This has my approval conditional on csci tests passing, but (last I checked) it seems like it's having domain resolution issues...

weinbe2:
Update, I see that these issues are identical to the issues being tracked in https://github.com/lattice/quda/pull/1572 and orthogonal to this PR. All other tests are passing. For this reason I'm going to merge this PR.

Issue #1575, opened 2025-06-12
`su3_fermion_test` is failing


dmcdougall:

dmcdougall:
Assign to me please.

dmcdougall:
```
100: [ RUN      ] Flow/SU3FermionTest.verify/double_Wilson_Flow
100: ERROR: QUDA_PRECISION=8 does not enable 4 precision (rank 0, host TheraC58, instantiate.h:480 in void quda::instantiatePrecision2(F &, Args &&...) [Apply = quda::GaugeCopy, T = double, F = quda::GaugeField, Args = <const quda::GaugeField &, QudaFieldLocation_s &, void *&, void *&, void **&, void **&, int &>]())
100:        last kernel called was (name=N4quda9CopyGaugeINS_12CopyGaugeArgIddLi18ELb0ENS_5gauge11FloatNOrderIdLi18ELi2ELi18EL20QudaStaggeredPhase_s0ELb1EL19QudaGhostExchange_sn2147483648ELb0EEES6_EEEE,volume=2x4x6x8,aux=GPU-offline,kernel_arg_threshold=4096,vol=384,stride=288,precision=8,geometry=4,Nc=3,vol=384,stride=288,precision=8,geometry=4,Nc=3,ghost)
100: ERROR: QUDA_PRECISION=8 does not enable 4 precision (rank 1, host TheraC58, instantiate.h:480 in void quda::instantiatePrecision2(F &, Args &&...) [Apply = quda::GaugeCopy, T = double, F = quda::GaugeField, Args = <const quda::GaugeField &, QudaFieldLocation_s &, void *&, void *&, void **&, void **&, int &>]())
100:        last kernel called was (name=N4quda9CopyGaugeINS_12CopyGaugeArgIddLi18ELb0ENS_5gauge11FloatNOrderIdLi18ELi2ELi18EL20QudaStaggeredPhase_s0ELb1EL19QudaGhostExchange_sn2147483648ELb0EEES6_EEEE,volume=2x4x6x8,aux=GPU-offline,kernel_arg_threshold=4096,vol=384,stride=288,precision=8,geometry=4,Nc=3,vol=384,stride=288,precision=8,geometry=4,Nc=3,ghost)
100: Saving 10711 sets of cached parameters to /home/damon/git/quda/build-O3-6.4.1/tunecache_error.tsv
100: [TheraC58:260369] *** Process received signal ***
100: [TheraC58:260369] Signal: Aborted (6)
[...]

PR #1578, opened 2025-06-12
Properly init gauge field in su3_fermion test

dmcdougall:
Fixes an erroneous test failure that complains when QUDA is only built with double precision support, for example.


dmcdougall:
Fixes #1575

dmcdougall:
This has [been a problem before](https://github.com/lattice/quda/commit/9d4b4cfa6ae0d993a9f6ab03f625d9de852bd5e4#diff-b1e7727690970a5e54f6480f6494c23a46680f28d612fd1476832451462ab25dR17-R37).  Maybe a better approach is to have better initialisation defaults for fields.

maddyscientist (approved):
Thanks @dmcdougall

PR #1581, opened 2025-06-14
fixes shared mem checks in MG

jcosborn:


include/targets/generic/shared_memory_cache_helper.h
```diff
@@ -92,13 +92,13 @@ namespace quda
...
     template <typename... U, typename... Arg>
-    constexpr SharedMemoryCache(const KernelOps<U...> &ops, Arg... arg) :
+    constexpr SharedMemoryCache(const KernelOps<U...> &ops, const Arg &...arg) :
       Smem(ops), block(D::dims(target::block_dim(), arg...)), stride(block.x * block.y * block.z)
     {
       checkKernelOps<SharedMemoryCache<T, D, O>>(ops);
       // sanity check
-      static_assert(shared_mem_size(dim3 {32, 16, 8})
-                    == Smem::get_offset(dim3 {32, 16, 8}) + SizeDims<D>::size(dim3 {32, 16, 8}) * sizeof(T));
+      // static_assert(shared_mem_size(dim3 {32, 16, 8}, arg...)
```
maddyscientist: Delete this commented out block?

maddyscientist (approved):
Thanks @jcosborn for the fast turn around on getting this fixed.

I've checked that multigrid is fixed with this PR.  What was the issue before: that you were counting static shared memory erroneously?

include/targets/generic/shared_memory_cache_helper.h
jcosborn: We can either 1) delete it, 2) make it a runtime check (possibly only in a DEBUG build), 3) try to make it static when possible (through SFINAE template trickery) and either ignore it or do the runtime check if it can't be static.  I wasn't sure what to do so left it as a comment for now.

jcosborn:
Yes, it was erroneously counting static shared memory.

include/targets/generic/shared_memory_cache_helper.h
maddyscientist: The overhead here is negligible here, surely, so no harm in having it runtime I think?  I'm happy to go with that, if you are.

jcosborn: Ok, I made it a runtime check 33d595a5e333a53deec5ada8e62cbe4fcb6db213

PR #1584, opened 2025-06-20
start merging early thread exit handling

jcosborn:
This covers all files except the dslashes.  Those will be merged later to keep the complexity of this PR down.


jcosborn:
@maddyscientist As for simplifying the calls to `dslash.template operator()` in `dslash_helper.cuh`, actually only the dslashes that needed the allthreads handling have been updated to take the extra parameters, the others can't currently be called that way.  We could simplify the calling code if we made all dslashes have the same interface, but I haven't done that mainly to reduce the number of changes.  I can easily implement that if you prefer it.

maddyscientist:
@jcosborn for those QUDA developers not on the portability calls (who are reviewing this PR) can you describe why these changes are needed?

jcosborn:
The early thread exit handling changes are to support programming models (like SYCL) which only support block collectives when all threads in a block are active (non-exited).  The changes allow targets to have all threads enter a kernel functor to participate in the block collectives when block collectives are used in a kernel.  For these kernels, instead of exiting when a thread is determined to be out-of-bounds for the kernel, all threads can enter the kernel, and the out-of-bounds ones will be marked inactive with an extra argument.  Only kernels that need this handling need any changes.  The kernel functors that need this handling have an extra template parameter `allthreads` which is true when it is being called with all threads entering (with some possibly out-of-bounds), and when false the functor should behave exactly as before.  There is also an extra functor argument `active` which specifies if the thread is in-bounds (active) or not.  When `allthreads` is true, the modified kernels then need to ensure that no out-of-bounds memory accesses occur from threads that aren't active, and also ensure that all threads (active or not) participate in the collectives.

include/kernels/restrictor.cuh
```diff
@@ -149,50 +150,55 @@ namespace quda {
...
+        }
       }
 
       reduced = BlockReduce_t(*this, thread.z).Sum(reduced);
 
-      if (target::thread_idx().x == 0) {
-        const int parity_coarse = x_coarse >= arg.out[src_idx].VolumeCB() ? 1 : 0;
-        const int x_coarse_cb = x_coarse - parity_coarse*arg.out[src_idx].VolumeCB();
+      if (!allthreads || alive) {
+        if (target::thread_idx().x == 0) {
```
maddyscientist, line 190: Any reason not to merge these two if statements?

include/dslash_helper.cuh
```diff
@@ -500,6 +500,12 @@ namespace quda
     dslash.template operator()<kernel_type>(x_cb, s, parity);
   }
 
+  template <KernelType kernel_type, class D>
```
maddyscientist: Do we need both versions of `apply_dslash` here?  Couldn't we just template the all thread parameter and have a single function definition?

maddyscientist (approved):
Approving this PR.  I've left a few remarks, ideally it would be nice to minimize code changes, but these are nice to have.

Great work @jcosborn!

include/dslash_helper.cuh
jcosborn: Yes, we can push the `allthreads` check into this function which would simplify the call.  I'll do that.

include/kernels/restrictor.cuh
jcosborn: The main reason is just to separate the thread-exit logic from the other logic.  We can merge them if preferred.

maddyscientist: Ok, just leave this one as is then.

include/dslash_helper.cuh
jcosborn: 67c37936642bcf8b761f458663775275dd85107e

weinbe2 (commented):
I just did a visual inspection and everything looks good/seems reasonable. I know @maddyscientist wants to do a little more testing before the final go, so this has my provisional approval depending on her thumbs up there. Thanks @jcosborn !

Issue #1580, opened 2025-06-14
Staggered eigensolve ctest fails with a pure double-precision build


maddyscientist:
@weinbe2 this cropped up after your recent work on this test.

```
100: [==========] Running 64 tests from 4 test suites.
100: [----------] Global test environment set-up.
100: [----------] 12 tests from DirectEvenOdd/StaggeredEigensolveTest
100: [ RUN      ] DirectEvenOdd/StaggeredEigensolveTest.verify/double_trlm_direct_evenodd_noSVD_LR
100: ERROR: QUDA_PRECISION=8 does not enable 4 precision (rank 2, host nvsocal2, instantiate.h:480 in constexpr void quda::instantiatePrecision2(F&, Args&& ...) [with Apply = GaugeCopy; T = double; F = GaugeField; Args = {const GaugeField&, QudaFieldLocation_s&, void*&, void*&, void**&, void**&, int&}]())
100:        last kernel called was (name=cudaMemcpyDefault,volume=bytes=55296,aux=copy,gauge_field.cpp,1087)
100: ERROR: QUDA_PRECISION=8 does not enable 4 precision (rank 1, host nvsocal2, instantiate.h:480 in constexpr void quda::instantiatePrecision2(F&, Args&& ...) [with Apply = GaugeCopy; T = double; F = GaugeField; Args = {const GaugeField&, QudaFieldLocation_s&, void*&, void*&, void**&, void**&, int&}]())
100:        last kernel called was (name=cudaMemcpyDefault,volume=bytes=55296,aux=copy,gauge_field.cpp,1087)
100: Stack trace (most recent call last):
100: Stack trace (most recent call last):
100: ERROR: QUDA_PRECISION=8 does not enable 4 precision (rank 3, host nvsocal2, instantiate.h:480 in constexpr void quda::instantiatePrecision2(F&, Args&& ...) [with Apply = GaugeCopy; T = double; F = GaugeField; Args = {const GaugeField&, QudaFieldLocation_s&, void*&, void*&, void**&, void**&, int&}]())
100:        last kernel called was (name=cudaMemcpyDefault,volume=bytes=55296,aux=copy,gauge_field.cpp,1087)
100: Stack trace (most recent call last):
[...]

PR #1585, opened 2025-06-24
Hotfix/iram monitor covdev

### LINKED ISSUE #1576: `covdev_policytune` test is failing
Assign to me please.

maddyscientist:
Fixes some bugs
* CUDA device monitoring wasn't guaranteed to map to the same device used for compute.  This is now fixed using the PCIe device id to guarantee the correct device is monitored
* Fixes a stability with IRAM: if too low a QR tolerance is used, the eigensolver will fail to converge (this fixes an issue with CI introduced in #1549.
* `covdev_test` was completely broken: gtests were ignored, reference dagger variant was broken, CI would fail for a uni-precision build.  These are all fixed now.  Closes #1576.

### DIFF
--- a/lib/covariant_derivative.cu
+++ b/lib/covariant_derivative.cu
@@ -40,9 +40,10 @@ namespace quda
       if (arg.xpay) errorQuda("Covariant derivative operator only defined without xpay");
       if (arg.nParity != 2) errorQuda("Covariant derivative operator only defined for full field");
 
+      constexpr bool dagger = false;
       constexpr bool xpay = false;
       constexpr int nParity = 2;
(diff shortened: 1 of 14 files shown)


weinbe2 (approved):
I've tested everything that this PR covers and it all looks good, so this PR gets my approval. Awesome work @maddyscientist . Before I hit merge, what should we do about the CSCS-CI test timeout?

maddyscientist:
cscs-ci run

maddyscientist:
I have made some changes to ensure the tests pass in the required time slot on cscs, namely reduced the number of eigenvectors requested.

I should also note I’ve fixed #1580 as it was a trivial fix.

PR #1589, opened 2025-07-02
Enable running fused Mobius kernels with NVSHMEM

hummingtree:
Avoid hangs due to `__syncthreads`/exiting threads for fused Mobius kernel + NVSHMEM.

### DIFF
--- a/include/dslash_helper.cuh
+++ b/include/dslash_helper.cuh
@@ -483,6 +483,10 @@ namespace quda
      classed.
    */
   struct dslash_default {
+
+    // By default the dslash types do not have __syncthreads() in their operator();
+    constexpr static bool use_syncthreads = false;
+
     constexpr QudaPCType pc_type() const { return QUDA_4D_PC; }
     constexpr int twist_pack() const { return 0; }
   };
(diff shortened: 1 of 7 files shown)


include/dslash_helper.cuh
```diff
@@ -722,7 +720,23 @@ namespace quda
...
+#endif
+
+        if (x_cb >= arg.threads) {
+#ifdef NVSHMEM_COMMS
+          // Need to increment the counter even for the exiting threads
+          block_counter.fetch_add(1, cuda::std::memory_order_acq_rel);
+#endif
+          return;
+        }
+        __syncthreads();
```
hummingtree: @maddyscientist I am a bit puzzled why this `__syncthreads()` is needed - without it the code hangs, but I have no idea why

include/dslash_helper.cuh
```diff
@@ -722,7 +720,23 @@ namespace quda
         const int dslash_block_offset
           = ((kernel_type == INTERIOR_KERNEL || kernel_type == UBER_KERNEL) ? arg.pack_blocks : 0);
         int x_cb = (target::block_idx().x - dslash_block_offset) * target::block_dim().x + target::thread_idx().x;
-        if (x_cb >= arg.threads) return;
+#ifdef NVSHMEM_COMMS
+        // Initialize a shared memory counter for the threads int the block
```
maddyscientist: typo: int -> in

include/dslash_helper.cuh
maddyscientist: Will ponder this......

maddyscientist (commented):
Thanks @hummingtree for getting up.
* Is there any change to the regular dslash kernels NVSHMEM performance with these changes?
* Is there any effect on the non-NVSHMEM fused def perf?  I think you said on our call there is some impact, but from looking at this, I can't see how it would be affected.

include/dslash_helper.cuh
maddyscientist: Regardless, should the `__syncthreads()` be guarded here with `#ifdef NVSHMEM_COMMS`?

hummingtree:
> * Is there any change to the regular dslash kernels NVSHMEM performance with these changes?

Will need to benchmark (again).
> * Is there any effect on the non-NVSHMEM fused def perf?  I think you said on our call there is some impact, but from looking at this, I can't see how it would be affected.

With this approach there should be no effect not he non-NVSHMEM fused DWF perf, if we can figure out why we need that `__syncthreads()` ...

include/dslash_helper.cuh
```diff
@@ -722,19 +730,42 @@ namespace quda
...
+        constexpr bool use_nvshmem_comms = true;
 #else
-        switch (parity) {
-        case 0: dslash.template operator()<kernel_type == UBER_KERNEL ? INTERIOR_KERNEL : kernel_type>(x_cb, s, 0); break;
-        case 1: dslash.template operator()<kernel_type == UBER_KERNEL ? INTERIOR_KERNEL : kernel_type>(x_cb, s, 1); break;
-        }
+        constexpr bool use_nvshmem_comms = false;
 #endif
+        if constexpr (use_nvshmem_comms && Arg::D::use_syncthreads) {
 #ifdef NVSHMEM_COMMS
```
maddyscientist, line 740: Is this #ifdef guard needed here?  I think the code will be hidden from the compiler anyway since it's in the `if constexpr (use_nvshmem_comms....)`

lib/dslash_mdw_fused.in.hpp
```diff
@@ -22,6 +22,13 @@ namespace quda
                                  const Complex *b_5, const Complex *c_5, bool dagger, int parity, int shift[4],
                                  int halo_shift[4], MdwfFusedDslashType type);
 
+    inline void apply_fused_dslash_list(ColorSpinorField &, const ColorSpinorField &, const GaugeField &,
```
maddyscientist: Can you make this consistent with the `void inline` on line 61.  This is triggering my OCD 😜

maddyscientist (approved):
Beyond the trivial comments, this is good to go.  Thank @hummingtree.

include/dslash_helper.cuh
hummingtree: Somehow it does not - it complains about not finding `cuda::atomic`.

lib/dslash_mdw_fused.in.hpp
hummingtree: https://github.com/lattice/quda/pull/1589/commits/c3eafc58c5c7a6fa1c178b5ba1872eb4ceb3477f

include/dslash_helper.cuh
maddyscientist: I guess that means the header inclusion is guarded by an `ifdef`, but still maybe a bit surprising by this.  I guess we can just call it a day here and move on....

PR #1590, opened 2025-07-11
MILC interface: Two-parity deflation and more

leonhostetler:
This PR enables two-parity deflation (previously it only supported EVEN parity deflation), and adds a few new functions to enable that as well as prepare for upcoming support for current calculations with low-mode averaging.

New functions added to the MILC interface:

- `qudaLoadDeflationSpace`: Load a specific parity deflation space from QUDA eigensolve, from QUDA file, from the other parity deflation space, or from MILC (e.g. Grid eigenpack)
- `qudaGetDeflationSpace`: To pass eigenvectors and eigenvalues from QUDA's deflation space back to MILC if desired
- `qudaProject`: Given a source vector, this function returns the source vector with the low mode projection subtracted off

The `qudaLoadDeflationSpace` function is quite large and handles multiple major code paths (`COMPUTE`, `FROM_OTHER_PARITY`, and `LOAD`) and with multiple levels of nesting. If advised, I can break this into multiple smaller functions as well as eliminate some redundant parts by adding a couple small helper functions. Let me know what you think.

Feel free to suggest alternate names to use for these functions.


lib/milc_interface.cpp
```diff
@@ -1192,6 +1206,308 @@ void qudaMultishiftInvert(int external_precision, int quda_precision, int num_of
...
+  std::vector<ColorSpinorField> tmp(nvec);
+  for (int i = 0; i < nvec; i++) {
+    sol[i] = ColorSpinorField(gpuParam);
+    sol[i] = sol_h[i];
+    tmp[i] = ColorSpinorField(gpuParam);
+    src[i] = ColorSpinorField(gpuParam);
+    src[i] = src_h[i]; // Copy host sources to device sources
+  }
+
+  // 1. Take block inner product: (V_i)^dag * src = s_i
```
maddyscientist: Just thinking out loud here: could it be advantageous to do more than one step of projection, in case of finite-precision errors?  If we ever do this in single precision, then we might get an improved projection if we do two passes at projection.

maddyscientist: And given the overhead of host-device copying here, I suppose doing additional passes would be essentially free?

maddyscientist (approved):
This looks great @leonhostetler, thank you for this contribution.

One question: could we simplify much of the "even or odd" logic if we have an array of preserved deflation space and array of preserved masses?  Then the parity could be used as an index for selecting the correct space, and it might clean things up a lot and reduce boilerplate.

lib/milc_interface.cpp
leonhostetler: I have added this with commit [b074075](https://github.com/lattice/quda/pull/1590/commits/b0740750c17055cf19445931dc02a90092bc1910).

leonhostetler:
Sorry about the delay with this. I have implemented the suggested changes and retested.

leonhostetler:
Thank you @maddyscientist for the suggested changes! Would this be ready to merge or are there additional changes I should make?

P.S. I'll have another PR coming soon for branch feature/milc-exact-current, which will complete the MILC/QUDA interfacing for our disconnected vector current calculations.

weinbe2:
@leonhostetler orthogonal of your ask of @maddyscientist , I'll give this a cursory visual review (expecting no surprises). Since I know you're already using this in MILC I'll skip my standard tests. (I meant to do this a while ago, sorry about that.)

weinbe2 (approved):
This is a really nice PR (and maybe something we'll pull deeper into QUDA one day, namely the even/odd eigenvector reconstruction for staggered, but not today!), thanks @leonhostetler !

PR #1591, opened 2025-07-11
Decrease ctest overhead by fusing across precisions

maddyscientist:
This PR will hopefully get the CI passing within the time limit again
* Fix some warnings with nvc++
* All ctest instances now use a single instance for different precisions, reducing the test latency

There is a small amount of cleanup that came for free here, but this PR doesn't attempt to clean up all the tests affected.

### DIFF
--- a/tests/hisq_unitarize_force_test.cpp
+++ b/tests/hisq_unitarize_force_test.cpp
@@ -1,35 +1,35 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <sys/time.h>
 
 #include <quda.h>
+#include <instantiate.h>
+#include <gauge_field.h>
+
 #include "host_utils.h"
-#include <command_line_params.h>
-#include "gauge_field.h"
+#include "command_line_params.h"
 #include "misc.h"
 #include "test.h"
 #include "hisq_force_reference.h"
 #include "ks_improved_force.h"
-#include <sys/time.h>
-#include <gtest/gtest.h>
 
-quda::GaugeField *cudaFatLink = NULL;
-quda::GaugeField *cpuFatLink = NULL;
+using test_t = ::testing::tuple<QudaPrecision>;
 
-quda::GaugeField *cudaOprod = NULL;
-quda::GaugeField *cpuOprod = NULL;
-
-quda::GaugeField *cudaResult = NULL;
-quda::GaugeField *cpuResult = NULL;
-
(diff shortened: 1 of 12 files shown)


tests/staggered_gsmear_test.cpp
```diff
@@ -52,41 +60,55 @@ TEST_F(StaggeredGSmearTest, verify)
...
-  ::testing::TestEventListeners &listeners = ::testing::UnitTest::GetInstance()->listeners();
-  if (comm_rank() != 0) { delete listeners.Release(listeners.default_result_printer()); }
+  if (link_recon != QUDA_RECONSTRUCT_NO) errorQuda("Error: link reconstruction is currently not supported");
 
-  if (link_recon != QUDA_RECONSTRUCT_NO) {
-    printfQuda("Error: link reconstruction is currently not supported.\n");
-    exit(0);
-  }
+  int test_rc = 0;
+  if (!enable_testing) {
```
weinbe2, line 108: It looks like something got lost here?

tests/hisq_unitarize_force_test.cpp
```diff
@@ -1,35 +1,34 @@
...
 #include "host_utils.h"
-#include <command_line_params.h>
-#include "gauge_field.h"
+#include "command_line_params.h"
 #include "misc.h"
 #include "test.h"
 #include "hisq_force_reference.h"
 #include "ks_improved_force.h"
-#include <sys/time.h>
-#include <gtest/gtest.h>
```
weinbe2, line 14: Not a hill I'll die on, but I'd prefer leaving in the explicit `#include <gtest/gtest.h>` as opposed to leaving it implicitly in another header. I think it makes it easier to quickly understand what's going on for the testing mechanism in this file. (But, again, not a hill I'll die on)

tests/staggered_gsmear_test.cpp
maddyscientist: Oops, I didn't commit my final bit for that one.  Will do so.

tests/hisq_unitarize_force_test.cpp
maddyscientist: The reason I am replacing the explicit gtest inclusion is to reduce the explicit dependency on gtest.  I'm not saying I plan to move away from gtest, but some abstraction is good.  E.g., in a similar spirit to how we abstract MPI or QMP.  I intend `test.h` to simplify the process of writing tests, and move as much functionality to there as possible.

PR #1592, opened 2025-07-17
Hardcode 64 wavefront size in the HIP backend

dmcdougall:
The `__GFX9__` macros (and friends) are only defined in the device pass.  Since the `warp_size` function is used on the host side, the `#else` branch was returning something different than what was returned on the device.

We need to be able to handle devices with different wavefront sizes, and we need to be able to expose those to the host.  The usual mechanism for host exposure is to probe the HIP runtime for the current device's `warpSize` property, but this approach isn't a constant expression.

In the future, we'll need an approach that either a) doesn't rely on constexpr-ness; or b) or can rely on constexpr-ness but simply isn't used on the host.

**Note**: This is only intended to be a short-term workaround.

@leonhostetler Can you give this a shot and see if it fixes your runtime error?

### DIFF
--- a/include/targets/hip/target_device.h
+++ b/include/targets/hip/target_device.h
@@ -145,11 +145,8 @@ namespace quda
        architecture we are running on.
     */
     constexpr int warp_size() {
-      #if defined(__GFX9__)
+      // FIXME: Need to handle devices with different wavefront sizes
       return 64;
-      #else
-      return 32;
-      #endif
     }
 
     /**


leonhostetler:
@dmcdougall I have tested this on Frontier, and it resolves the runtime error I was seeing. Thanks for the quick resolution!

dmcdougall:
You're very welcome, I'm happy to help.

This is just a short-term workaround.  I'm talking to the other QUDA developers and maintainers at a later date to figure out a plan for support for other devices.

maddyscientist (approved):

PR #1593, opened 2025-07-19
Added missing popVerbosity() to lib/interface_quda.cpp

leonhostetler:
In `lib/interface_quda.cpp`, three functions--`performGFlowQuda`, `performAdjGFlowSafe`, and `performAdjGFlowHier`--were missing the `popVerbosity()` at the end of the function after using `pushVerbosity()` at the beginning. Thus, when calling one of these functions repeatedly, one ends up with an output file containing thousands of warnings like:
```
performGFlowQuda: WARNING: Verbosity stack contains 189 elements.  Is there a missing popVerbosity() somewhere?
```


maddyscientist (approved):
Thank you @leonhostetler

PR #1600, opened 2025-09-19
Autotuning Enhancement

maddyscientist:
This PR seeks to reduce the autotuning overheads when doing shared memory throttling
* The advancement through shared memory over allocation was previously done in a dumb way, assuming that the only limiter for occupancy is dynamic shared memory and the number of threads in a block.
* Using this information, we then over allocate the shared memory to test `1..n` CTAs per SM, where `n` is the max blocks per SM.
* By computing `n` using only the shared memory, we end up testing many more configs than we should, since we may be much more limited by registers.
* By using the CUDA occupancy calculator, we can compute what `n` is exactly and use that to set the tuning space.  To do so, I've created an API wrapper `qudaOccupancyMaxActiveBlocksPerMultiprocessor`.
* For double precision kernels the space reduction factor is more like 4x and for single precision, it's closer to 2x (a statement of being more register limited with double than single).

This PR has a couple of other additions included as well
* Avoid shared carve out tuning on targets that don't support it, e.g., AMD
* WAR for CUDA 13 CUB compilation with MG enabled

### DIFF
--- a/include/dslash.h
+++ b/include/dslash.h
@@ -86,7 +86,7 @@ namespace quda
 
     bool tuneSharedCarveOut() const override
     {
-      static bool tune_shared = true; // default is to do carve out tuning
+      static bool tune_shared = device::shared_carve_out_supported(); // default is to do carve out tuning
       static bool init = false;
 
       if (!init) {
--- a/include/dslash.h
+++ b/include/dslash.h
@@ -86,7 +86,8 @@ namespace quda
 
     bool tuneSharedCarveOut() const override
     {
-      static bool tune_shared = device::shared_carve_out_supported(); // default is to do carve out tuning
+      // default is to do carve out tuning if the architecture supports it
+      static bool tune_shared = device::shared_carve_out_supported();
       static bool init = false;
 
       if (!init) {
--- a/include/targets/hip/tunable_kernel.h
+++ b/include/targets/hip/tunable_kernel.h
@@ -33,11 +33,28 @@ namespace quda
   protected:
     QudaFieldLocation location;
 
+    /**
+       @brief Set the maximum number of blocks that can reside on an
+       SM.  This is called when we are autotuning to allow us to work
+       out how many different shared memory over allocations we should
+       use to minimally cover all occupancy variations.
+     */
+    void setMaxActiveBlocks(const kernel_t &kernel, const TuneParam &tp) const
+    {
+      if (activeTuningWarmup() && tuneSharedBytes()) {
+        auto tp2 = tp;
+        setSharedBytes(tp2);
+        // only compute max number blocks when we have no shared memory over subscription
+        if (tp.shared_bytes == tp2.shared_bytes) max_active_blocks = qudaOccupancyMaxActiveBlocks(kernel, tp);
+      }
+    }
+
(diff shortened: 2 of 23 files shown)


include/dslash.h
```diff
@@ -86,7 +86,7 @@ namespace quda
 
     bool tuneSharedCarveOut() const override
     {
-      static bool tune_shared = true; // default is to do carve out tuning
+      static bool tune_shared = device::shared_carve_out_supported(); // default is to do carve out tuning
```
weinbe2: I'd update the comment to say "the default is to do carve out tuning if the architecture supports it"

include/targets/hip/tunable_kernel.h
```diff
@@ -12,26 +12,50 @@ namespace quda
...
+       SM.  This is called when we are autotuning to allow us to work
+       out how many different shared memory over allocations we should
+       use to minimally cover all occupancy variations.
+     */
+    void setMaxActiveBlocks(const kernel_t &kernel, const TuneParam &tp) const
+    {
+      if (activeTuningWarmup() && tuneSharedBytes()) {
+        auto tp2 = tp;
+        setSharedBytes(tp2);
+        // only compute max number blocks when we have no shared memory over subscription
```
weinbe2, line 47: This logic feels a bit fragile... I understand why this only gets called once (warm-up only happens once), but why does this equality guarantee there's no shared memory oversubscription? Is it because `tp` is at its default setting now, which is `shared_bytes == 0`?

weinbe2:
This PR looks good at a high level, with an outstanding question on some logic related to setting the max number of blocks. I'll give this the approve once that's cleared up and the CI passes, I know @bjoo is planning some hands-on testing.

include/targets/hip/tunable_kernel.h
maddyscientist: This works because `setSharedBytes` sets the shared memory to the default value with no oversubscription, e.g., the max of `sharedBytesPerThread()` and `sharedBytesPerBlock()`.  If we have this equality, then we know that we cannot be over allocating shared memory by definition, so that the max number of blocks per SM that the occupancy calculator returns is the true value, and not masked by oversubscription.

I think it's fairly robust.

include/dslash.h
maddyscientist: bc91f944d27cc9a73da407dbee0fe4ffc6980dff

bjoo:
Hmm. Commit [fdfafe2](https://github.com/lattice/quda/pull/1600/commits/fdfafe22da1653aa7a483f046f2da754b2e5124a) broke the compilation (according to a Git Bisect) with CUDA-13.0. This just adds the guard around the include of `<cub/block/block_reduce.cuh>`. Is it possible that the preproc symbol is not there in the compilation

```
#ifdef QUDA_MMA_AVAILABLE
#include <cub/block/block_reduce.cuh>
#endif
```
?

bjoo:
Looks like QUDA_MMA_AVAILABLE was not available. I included #include <quda_arch.h> and that fixed the build issue I had.

bjoo (approved):
Looks good to me and it worked with Chroma

weinbe2 (approved):
I'm happy about all the changes, thanks @maddyscientist !

PR #1602, opened 2025-10-21
Fix bug with complex::operator+ and complex::operator-

maddyscientist:
thanks to @SaltyChiang for reporting this bug


weinbe2 (approved):
Nice catch @SaltyChiang . Approved, we just need to wait until CI (presumably successfully) completes before merging.

SaltyChiang:
@maddyscientist @weinbe2 Thank you very much for the quick fix!

PR #1603, opened 2025-11-18
Fixed a compile-time error when compiling the openqcd interface QUDA_INTERFACE_NVTX

weinbe2:
It appears that a change that was made in `milc_interface.cpp` at some point wasn't propagated into `openqcd_interface.cpp`, so I just copy+pasted over the code from https://github.com/lattice/quda/blob/develop/lib/milc_interface.cpp#L23-L45 and called it a day. @chaoos , can you confirm these fixes work for you?

Also---I saw there are some warnings in `openQCD_qudaSolverGetHandle`:

```
/scratch/local/quda/lib/openqcd_interface.cpp: In function ‘void* openQCD_qudaSolverGetHandle(int)’:
/scratch/local/quda/lib/openqcd_interface.cpp:1631:31: warning: array subscript [-1, 31] is outside array bounds of ‘void* [32]’ [-Warray-bounds=]
 1631 |   if (qudaState.inv_handles[id] == nullptr) {
      |       ~~~~~~~~~~~~~~~~~~~~~~~~^
/scratch/local/quda/lib/openqcd_interface.cpp:65:9: note: while referencing ‘openQCD_QudaState_t::inv_handles’
   65 |   void *inv_handles[OPENQCD_MAX_INVERTERS]; /** Array of void-pointers to QudaInvertParam structs for the solver(s) */
      |         ^~~~~~~~~~~
/scratch/local/quda/lib/openqcd_interface.cpp:1631:31: warning: array subscript [-1, 31] is outside array bounds of ‘void* [32]’ [-Warray-bounds=]
 1631 |   if (qudaState.inv_handles[id] == nullptr) {
      |       ~~~~~~~~~~~~~~~~~~~~~~~~^
/scratch/local/quda/lib/openqcd_interface.cpp:65:9: note: while referencing ‘openQCD_QudaState_t::inv_handles’
   65 |   void *inv_handles[OPENQCD_MAX_INVERTERS]; /** Array of void-pointers to QudaInvertParam structs for the solver(s) */
      |         ^~~~~~~~~~~
```

I'm not sure how I missed this warning before. The code that checks if a solver id is valid, `check_solver_id`, expressly permits `-1`, so a value of `-1` won't trigger a runtime error... but `qudaState.inv_handles[-1]` is unambiguously out of bounds (there's no array pointer offset calculations). @chaoos can you sanity check this? Feel free to push any changes directly to my branch.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -101,7 +101,7 @@ set (QUDA_OBJS
   copy_gauge_offset.cu copy_color_spinor_offset.cu copy_clover_offset.cu
   staggered_oprod.cu clover_trace_quda.cu
   hisq_paths_force_quda.cu
-  unitarize_force_quda.cu unitarize_links_quda.cu openqcd_interface.cpp milc_interface.cpp milc_interface_internal.cpp
+  unitarize_force_quda.cu unitarize_links_quda.cu milc_interface.cpp milc_interface_internal.cpp
   tune.cpp
   device_vector.cu
   inv_gmresdr_quda.cpp
@@ -114,6 +114,11 @@ set (QUDA_OBJS
   instantiate.cpp version.cpp
   block_transpose.cu )
 # cmake-format: on
+
+# only add the OpenQCD interface file if we're compiling with it enabled
+if(QUDA_INTERFACE_OPENQCD)
+  list(APPEND QUDA_OBJS openqcd_interface.cpp)
+endif()
 
 # current workaround for nvshmem
 set(QUDA_DSLASH_OBJS 
--- a/lib/openqcd_interface.cpp
+++ b/lib/openqcd_interface.cpp
@@ -1628,15 +1628,24 @@ static void *openQCD_qudaSolverReadIn(int id)
 void *openQCD_qudaSolverGetHandle(int id)
 {
   check_solver_id(id);
-  if (qudaState.inv_handles[id] == nullptr) {
+
+  void *ptr = id == -1 ? qudaState.dirac_handle : qudaState.inv_handles[id];
+
+  if (ptr == nullptr) {
     if (id != -1) {
       WITH_COMM(logQuda(QUDA_VERBOSE, "Read in solver parameters from file %s for solver (id=%d)\n", qudaState.infile, id));
     }
-    qudaState.inv_handles[id] = openQCD_qudaSolverReadIn(id);
+    ptr = openQCD_qudaSolverReadIn(id);
   }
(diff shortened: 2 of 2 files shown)


chaoos (approved):
@weinbe2 I confirm that the changes do not break the interface to openqcd, I ran all tests on my side. And I fixed the compiler warning.

weinbe2:
Awesome, thank you for verifying @chaoos and for pushing a fix to the compiler warning.

@maddyscientist , if you don't mind giving this a review and, assuming there are no concerns, approving + merging, that would be awesome. Since @chaoos and I have both contributed code to this PR, we shouldn't be the ones to merge it (of course).

weinbe2:
Oh, quick pause---I'm now (maybe always?) hitting linker errors when compiling without MPI, calls that are also in the openqcd interface file... let me investigate this.

For reference:
```
/usr/bin/ld: ../lib/libquda.so: undefined reference to `ompi_mpi_comm_null'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `ompi_mpi_double'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `MPI_Bcast'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `MPI_Comm_split'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `ompi_mpi_byte'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `MPI_Comm_free'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `MPI_Query_thread'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `ompi_mpi_int'
/usr/bin/ld: ../lib/libquda.so: undefined reference to `MPI_Comm_rank
```

weinbe2:
Ah, okay, the issue is I was compiling QUDA without the OPENQCD interface enabled on cmake, but the file was still being included in the build so this `cmake` conditional wasn't being triggered:

```
if(QUDA_INTERFACE_OPENQCD AND NOT (QUDA_MPI))
  message(SEND_ERROR "OpenQCD requires QUDA built with MPI comms.")
endif()
```

The solution is removing the file from the list of default files to compile, and only adding it if QUDA_INTERFACE_OPENQCD is enabled in `cmake`. I'll implement this in a moment.

weinbe2:
@chaoos @maddyscientist fixed the issue with `a5d9941`, I'm not exactly sure how this slipped through previous testing but I manually tested a few things now...

maddyscientist (approved):

Issue #1605, opened 2025-12-10
Problem when policy tuning


chaoos:
Dear all,

I found a bug in policy tuning of Dslash kernels. For a minimal working reproducible example, see https://github.com/etmc/tmLQCD/issues/642.

The problem is the following:

1. In https://github.com/lattice/quda/blob/develop/lib/dslash_policy.hpp#L1937-L1958, the calls to 

    `(*dslashImp)(dslash, in, halo, profile);`

are meant to trigger tuning of constituent kernels that have not been tuned until then. Afterwards tuning of the policy starts. However the line above writes to AND reads from the output spinor field `out`. We have two cases:

1.1. All kernels are already tuned: then the line applies Dslash to the input field and read/writes to the output field. This makes the actual call to the (kernel- and policy-tuned) operator later in https://github.com/lattice/quda/blob/develop/lib/dslash_policy.hpp#L1991-L1992 acting on the wrong output field (which it also reads from) and thus the resulting output field is wrong.
1.2. Some kernels need tuning: then these kernels are tuned, however the operator is applied fully and again `out` is read/written. Same problem as in 1.1.

2. Later in the determination of the best policy (https://github.com/lattice/quda/blob/develop/lib/dslash_policy.hpp#L1982), we also need to backup/restore the output field explicitly, because the policy kernels are exterior by nature.

--------

I solved the bug in this branch: [`fix_tuning`](https://github.com/lattice/quda/tree/fix_tuning). There are multiple ways to fix this, most of which are hacks. The solution in [`fix_tuning`](https://github.com/lattice/quda/tree/fix_tuning) is the least ugly of what I was able to come up with.

1) In order to backup/restore the output field properly, I pass `out` explicitely to the `DslashPolicyTune` constructor.
2) Before the policy tuning phase (when the remaining constituents are tuned), I sandwich the region with `backup` and `restore` of the output field.
3) In order to fix the policy tuning phase, I reimplemented `preTune` and `postTune` to backup/restore the output field irrespective of the `KernelType` (as opposed to `Dlash`).
4) For this to work, `backup` and `restore` methods of `ColorSpinorField` need to be idempotent.

Best,
Roman

maddyscientist:
Hi @chaoos, thanks for diagnosing this issue, and understanding the underlying issues here.

[...]

PR #1606, opened 2025-12-11
Hotfix/Fix problem when policy tuning

chaoos:
For a description, see https://github.com/lattice/quda/issues/1605.


maddyscientist (approved):
Thanks for this fix @chaoos

PR #1607, opened 2025-12-30
Function to write gauge fields to disk

leonhostetler:
Adds a function `writeGaugeQuda` to `lib/interface_quda.cpp`. This function is similar to `saveGaugeQuda` but writes the gauge field to disk instead of just copying it to host.

The motivation for this is a desire to be able to write a gradient flowed gauge field (i.e. `gaugeSmeared`) to disk after using `performWFlowQuda` or `performGFlowQuda`, without going through the rigamarole of allocating space from MILC, copying the flowed field back to MILC, and then writing it to disk. This will also allow us to write the fat and long links to disk, which is something we've been working on as well.

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -909,7 +909,7 @@ void writeGaugeQuda(const char *file, QudaGaugeParam *param)
   gauge_param.ghostExchange = QUDA_GHOST_EXCHANGE_NO;
   gauge_param.location = QUDA_CUDA_FIELD_LOCATION;
   gauge_param.create = QUDA_NULL_FIELD_CREATE;
-  gauge_param.setPrecision(param->cuda_prec, true);
+  gauge_param.setPrecision(param->cuda_prec);
   GaugeField *cudaGauge = nullptr;
   switch (param->type) {
   case QUDA_WILSON_LINKS:
@@ -937,6 +937,7 @@ void writeGaugeQuda(const char *file, QudaGaugeParam *param)
   cpu_param.location      = QUDA_CPU_FIELD_LOCATION;
   cpu_param.order         = QUDA_QDP_GAUGE_ORDER;
   cpu_param.create        = QUDA_NULL_FIELD_CREATE;
+  cpu_param.setPrecision(param->cpu_prec);
   quda::GaugeField cpuGauge(cpu_param);
   cpuGauge.copy(*cudaGauge);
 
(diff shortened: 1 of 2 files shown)


weinbe2:
Thanks for the PR, Leon, this looks good to me and is a good contribution. I had an idea to simplify this which should be a relatively quick copy+paste job.

You could create a null CPU field at the start of the function, and then within the `switch` statement you could call `cpuField.copy(...);` as appropriate for the gauge field/fat links/long links. The smeared links would of course need the intermediate copy to a non-extended field, but in that case all of the logic can stay local, and you can create a `GaugeField` as opposed to a pointer thereto.

I think this should all "just work", but let me know what you think.

leonhostetler:
@weinbe2 I have revised and retested. Let me know what you think. I think it is indeed better now with the logic for each case easier to follow and more self-contained/local.

weinbe2:
Thanks @leonhostetler , this is exactly what I was picturing. I left one comment on the code for a trivial change. Otherwise, I'll do a cursory due-diligence build (I expect no surprises) and give an approve once the change is in.

lib/interface_quda.cpp
```diff
@@ -897,6 +897,59 @@ void saveGaugeQuda(void *h_gauge, QudaGaugeParam *param)
...
+    GaugeFieldParam cuda_param(*param);
+    cuda_param.location = QUDA_CUDA_FIELD_LOCATION;
+    cuda_param.create = QUDA_NULL_FIELD_CREATE;
+    cuda_param.ghostExchange = QUDA_GHOST_EXCHANGE_NO;
+    cuda_param.pad = 0;
+    cuda_param.setPrecision(param->cuda_prec);
+    GaugeField cudaGauge(cuda_param);
+    copyExtendedGauge(cudaGauge, *gaugeSmeared, QUDA_CUDA_FIELD_LOCATION);
+    cpuGauge.copy(cudaGauge);
+    break;
```
weinbe2: I believe this `break;` needs to be below the closing `}` on the following line, if anything as a "best practice". The `case QUDA_SMEARED_LINKS:` block will _always_ hit the `break;` in this case so it should be legal, but it still feels a bit too weird to me.

leonhostetler: Yes, I see what you mean. I have made the change.

weinbe2 (approved):
This looks great. Thanks @leonhostetler

PR #1609, opened 2026-01-12
fix compilation errors with modern GCC due to stricter template resolution and variable scoping

Jaden-Anderson:
- Variables, even pointer members in a Class, shall be declared first in the same scope before usage
- References of type `vector<Complex>&` initialized from `Complex*` will be regarded as invalid
- Ambiguous type conversion will NOT be inferred, if there are more than one constructor candidates
- A singleton `vector<double>` will NOT be converted to `double` implicitly when used as a scalar

Committer: WANG Ting-Xiao <wangtx@ihep.ac.cn>

On branch quda-v1.1.0-gcc13x
Changes to be committed:
- modified:   <a href=include/invert_quda.h>include/invert_quda.h</a>
- modified:   <a href=lib/inv_cg_quda.cpp>lib/inv_cg_quda.cpp</a>
- modified:   <a href=lib/interface_quda.cpp>lib/interface_quda.cpp</a>


maddyscientist:
Thanks for the contribution, however, this pull request in its current form is clearly unfinished:
* it deletes and renames several files
* there's a stray build log that's been added

Once this PR has been cleaned up, we can review it.

PR #1615, opened 2026-01-22
Hotfix for bug in shiftQuda

### LINKED ISSUE #1614: Bug in shiftQuda
Covariant shifts in the forward direction (`sym=1`) differ when done on CPU using MILC code and when done on GPU using `shiftQuda`. Looking through the code, I see the following structure in `shiftQuda` in `lib/interface_quda.cpp` :
```
  out = in;
  tmp = in;

  // Some code omitted

  if (sym & 1) {
    myCovDev.MCD(out, in, dir); // apply the operator
  }
  if (sym & 2) {
    myCovDev.MCD(tmp, in, dir + 4); // apply the operator
  }

  quda::blas::xpy(tmp, out);

  if (sym == 3) quda::blas::ax(0.5, out);
```
 I thought the line `quda::blas::xpy(tmp, out);` was suspect in its current context, so I commented it out and found that CPU and GPU results agreed now for `sym=1`. The part that should do the symmetric shift (`sym=3`) also looks suspect to me.

I wanted to bring this up as an issue so that people more familiar with how this function should work can comment.

Perhaps it should go like this (thanks @detar):
```
switch(sym) 
{
  case 1: // Forward shift
    myCovDev.MCD(out, in, dir);
    break;
  case 2: // Backward shift
    myCovDev.MCD(out, in, dir + 4);
    break;
  case 3:  // Symmetric shift
    myCovDev.MCD(out, in, dir);
    myCovDev.MCD(tmp, in, dir + 4);
    quda::blas::xpy(tmp, out);
    quda::blas::ax(0.5, out);
    break;
  default:
    errorQuda("Invalid shift type = %d\n", sym);
}
```
If people more familiar with this can confirm that this last bit of code is the intended behavior of this function, I'd be happy to do a PR.

leonhostetler:
This fixes bug #1614 .

I have compared CPU (MILC) and QUDA shifts for all three cases and they agree to machine precision now.

### DIFF
--- a/lib/interface_quda.cpp
+++ b/lib/interface_quda.cpp
@@ -2005,16 +2005,23 @@ void shiftQuda(void *h_out, void *h_in, int dir, int sym, QudaInvertParam *param
 
   GaugeCovDev myCovDev(diracParam); // create the Dirac operator
 
-  if (sym & 1) {
-    myCovDev.MCD(out, in, dir); // apply the operator
+  switch(sym) 
+  {
+    case 1: // Forward shift
+      myCovDev.MCD(out, in, dir);
+      break;
+    case 2: // Backward shift
+      myCovDev.MCD(out, in, dir + 4);
+      break;
+    case 3:  // Symmetric shift
+      myCovDev.MCD(out, in, dir);
+      myCovDev.MCD(tmp, in, dir + 4);
+      quda::blas::xpy(tmp, out);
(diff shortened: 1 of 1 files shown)


weinbe2 (approved):
This looks good and addresses the issue as was reported; the cscs-ci error seems like a red herring and it did indeed pass everything when I looked at the logs. Approved. Thanks for this fix @leonhostetler !

PR #1604, opened 2025-12-03
Feature/prefetch2

maddyscientist:
This work is latest towards optimizing QUDA for Blackwell:
* Adds supports for "spatial prefetching", where we over fetch data to L2 when issuing a global load.  Exposed as an optional template parameter to `vector_load`.  At present, not deployed anywhere.
* Add support for prefetching instructions, in the form of both per-thread prefetching (which works on all CUDA architectures), and TMA-based prefetching, which is Hopper+ only.  Prefetching type is set using `QUDA_DSLASH_PREFETCH` CMake parameter, with 0=per-thread, 1=TMA bulk, and 2=TMA descriptor
* Add an experimental L1 prefetch (using LDGSTS).  Disabled, but left for future experiments.
* Add single-threaded execution region helper function `target::is_thread_zero()` which should be used for TMA issuance.
* Optionally store the backward shifted gauge field.  This simplifies all dslash indexing, as all spatial indices thus correspond to "this" site.  Enabled with `QUDA_DSLASH_DOUBLE_STORE=ON` which is required for TMA-based prefetching (for alignment reasons).* Prefetching is exposed for both ColorSpinorFields and GaugeFields, though only latter actually used at present.
* Added prefetching support to both Wilson and Staggered dslash kernels, parameterized using `QUDA_DSLASH_PREFETCH_DISTANCE_WILSON` and `QUDA_DSLASH_PREFETCH_DISTANCE_STAGGERED` CMake parameters.
* Optimization of the neighbor indexing for the dslash kernels.  This reduced integer instruction overheads.
* Reduction in pointer arithmetic overhead (use more 32-bit integer operations where possible).  Added three operand and four operand variants of `vector_load` and `vector_store` to this end (respectively).
* Optimize FFMA2 issuance to reduce total number of floating point instructions on Blackwell
* Optimization of short <-> float conversation to reduce instruction overheads
* Optimization of staggered packing kernels (replace division by `int` with division by `fast_intdiv`)
* Extends some OpenMP parallelization on the host that was missing.

The end result of this work is that both Staggered and Wilson dslash kernels can saturate over 90% memory bandwidth for most variants.  Outstanding are half precision variants using reconstruction, that are still lagging.  These will be the focus of a subsequent PR.

PR #1618, opened 2026-03-10
Codifying host verify cleanup/reducing compiler warnings and errors

weinbe2:
This PR accomplishes a few quick things:
 * Cherry-picks various narrow bugfixes and compiler warnings from #1604, getting them merged in sooner as opposed to later
 * Partially homogenizes host verification routines, including standardizing adding "wiggle room" (described below)
   * This includes adding some extra `nan` checks
 * Temporarily disables MdagMLocal preconditioning tests for Mobius, knowing that it's fixed in #1604 , just to unblock some automated testing of `develop`

I don't endorse that last point being something we do typically---we shouldn't just disable tests that don't pass obviously---but I'd like to unblock testing given we know solving things in `develop` is eminent.

As for adding "wiggle room": empirically, there are cases where pure fp32 solves/etc "fail" because, for ex, a solve with a requested tolerance of 1e-5 converges at 0.99999e-5 on the device, then for order-of-reduction reasons is reported as not converging because the host check spits out 1.00001e-5. We've historically whack-a-mole'd these, which isn't sustainable. I've replaced this with a more blanket approach to all dslash and invertest tests. This is codified in the new function `checkReasonableHostDeviation` (happy to change the name) which checks for this, prints a _warning_ if the requested tolerance check technically does fail (so there aren't any silent exceptions), and then passes a convergence requirement with some extra padding back to gtest.

### DIFF
--- a/tests/blas_interface_test_gtest.hpp
+++ b/tests/blas_interface_test_gtest.hpp
@@ -77,8 +77,9 @@ TEST_P(BLASTest, verify)
     case QUDA_BLAS_DATATYPE_C: tol_gemm = 10 * std::numeric_limits<float>::epsilon(); break;
     case QUDA_BLAS_DATATYPE_D:
     case QUDA_BLAS_DATATYPE_Z: tol_gemm = 10 * std::numeric_limits<double>::epsilon(); break;
-    default: errorQuda("Unexpected BLAS data type %d", data_type);
+    default: ASSERT_TRUE(false) << "Unexpected BLAS data type " << data_type;
     }
+    ASSERT_FALSE(std::isnan(deviation_gemm)) << "Nan has propagated into the result";
     EXPECT_LE(deviation_gemm, tol_gemm) << "CPU and CUDA GEMM implementations do not agree";
     break;
   }
@@ -93,12 +94,13 @@ TEST_P(BLASTest, verify)
     case QUDA_BLAS_DATATYPE_Z: tol_lu_inv = 5000 * std::numeric_limits<double>::epsilon(); break;
     case QUDA_BLAS_DATATYPE_S:
     case QUDA_BLAS_DATATYPE_D:
-    default: errorQuda("Unexpected BLAS data type %d", data_type);
+    default: ASSERT_TRUE(false) << "Unexpected BLAS data type " << data_type;
     }
+    ASSERT_FALSE(std::isnan(deviation_lu_inv)) << "Nan has propagated into the result";
     EXPECT_LE(deviation_lu_inv, tol_lu_inv) << "CPU and CUDA LU Inversion implementations do not agree";
     break;
   }
-  default: errorQuda("Unexpected BLAS test type %d", test_type);
+  default: ASSERT_TRUE(false) << "Unexpected BLAS test type " << test_type;
   }
 }
 
--- a/tests/utils/gauge_utils.cpp
+++ b/tests/utils/gauge_utils.cpp
@@ -433,6 +433,7 @@ template <typename real_t> struct ApplyRandomU1Phase {
     auto gauge = reinterpret_cast<real_t *const *>(gauge_);
 
     for (int dir = 0; dir < 4; dir++) {
+#pragma omp parallel for
       for (int i = 0; i < Vh; i++) {
         for (int parity = 0; parity < 2; parity++) {
           // create a random phase
@@ -493,6 +494,7 @@ template <typename real_t> struct ConstructRandomMatrixGaugeField {
     };
 
     for (int dir = 0; dir < 4; dir++) {
+#pragma omp parallel for
       for (int i = 0; i < Vh; i++) {
         for (int parity = 0; parity < 2; parity++) {
           real_t *link = gauge[dir] + (parity * Vh + i) * gauge_site_size;
--- a/tests/utils/host_utils.cpp
+++ b/tests/utils/host_utils.cpp
@@ -181,6 +181,7 @@ void constructHostCloverField(void *clover, void *, QudaInvertParam &inv_param)
 template <typename real_t> struct ConstructCloverField {
   void operator()(void *res, double norm, double diag)
   {
+#pragma omp parallel for
     for (auto i = 0lu; i < static_cast<size_t>(Vh); i++) {
       for (auto parity = 0lu; parity < 2lu; parity++) {
         auto clover_matrix = reinterpret_cast<real_t *>(res) + 72 * (parity * Vh + i);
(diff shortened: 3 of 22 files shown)


tests/utils/gauge_utils.cpp
```diff
@@ -433,6 +433,7 @@ template <typename real_t> struct ApplyRandomU1Phase {
     auto gauge = reinterpret_cast<real_t *const *>(gauge_);
 
     for (int dir = 0; dir < 4; dir++) {
+#pragma omp parallel for
```
bjoo, line 436: would this be worth adding a `collapse(2)` to?

tests/utils/gauge_utils.cpp
```diff
@@ -493,6 +494,7 @@ template <typename real_t> struct ConstructRandomMatrixGaugeField {
     };
 
     for (int dir = 0; dir < 4; dir++) {
+#pragma omp parallel for
```
bjoo, line 497: worth a `collapse(2)` - it may not matter...

tests/utils/host_utils.cpp
```diff
@@ -181,6 +181,7 @@ void constructHostCloverField(void *clover, void *, QudaInvertParam &inv_param)
 template <typename real_t> struct ConstructCloverField {
   void operator()(void *res, double norm, double diag)
   {
+#pragma omp parallel for
```
bjoo, line 184: collapse(2) ?

tests/utils/staggered_gauge_utils.cpp
```diff
@@ -93,6 +94,7 @@ void constructFatLongGaugeField(void *const *fatlink, void *const *longlink, Gau
 
       // incorporate non-trivial phase into long links
       for (int dir = 0; dir < 4; ++dir) {
+#pragma omp parallel for
```
bjoo, line 97: collapse(2) ?

tests/utils/staggered_host_utils.cpp
```diff
@@ -186,6 +186,7 @@ void computeTwoLinkCPU(void **twolink, su3_matrix **sitelinkEx)
   for (int dir = 0; dir < 4; ++dir) E[dir] = Z[dir] + 4;
   const int extended_volume = E[3] * E[2] * E[1] * E[0];
 
+#pragma omp parallel for
```
bjoo, line 189: collapse(3)?

tests/utils/staggered_host_utils.cpp
```diff
@@ -698,6 +699,7 @@ void constructStaggeredTestSpinorParam(quda::ColorSpinorParam *cs_param, const Q
 // data reordering routines
 template <typename Out, typename In> void reorderQDPtoMILC(Out *milc_out, In **qdp_in, int V, int siteSize)
 {
+#pragma omp parallel for
```
bjoo, line 702: collapse(3) ?

tests/blas_interface_test_gtest.hpp
```diff
@@ -77,8 +77,9 @@ TEST_P(BLASTest, verify)
     case QUDA_BLAS_DATATYPE_C: tol_gemm = 10 * std::numeric_limits<float>::epsilon(); break;
     case QUDA_BLAS_DATATYPE_D:
     case QUDA_BLAS_DATATYPE_Z: tol_gemm = 10 * std::numeric_limits<double>::epsilon(); break;
-    default: errorQuda("Unexpected BLAS data type %d", data_type);
+    default: ASSERT_TRUE(false) << "Unexpected BLAS data type " << data_type;
     }
```
bjoo, line 81: Why do it this way?

tests/blas_interface_test_gtest.hpp
```diff
@@ -93,12 +94,13 @@ TEST_P(BLASTest, verify)
     case QUDA_BLAS_DATATYPE_Z: tol_lu_inv = 5000 * std::numeric_limits<double>::epsilon(); break;
     case QUDA_BLAS_DATATYPE_S:
     case QUDA_BLAS_DATATYPE_D:
-    default: errorQuda("Unexpected BLAS data type %d", data_type);
+    default: ASSERT_TRUE(false) << "Unexpected BLAS data type " << data_type;
```
bjoo, line 97: also here?

tests/blas_interface_test_gtest.hpp
```diff
@@ -93,12 +94,13 @@ TEST_P(BLASTest, verify)
...
     }
+    ASSERT_FALSE(std::isnan(deviation_lu_inv)) << "Nan has propagated into the result";
     EXPECT_LE(deviation_lu_inv, tol_lu_inv) << "CPU and CUDA LU Inversion implementations do not agree";
     break;
   }
-  default: errorQuda("Unexpected BLAS test type %d", test_type);
+  default: ASSERT_TRUE(false) << "Unexpected BLAS test type " << test_type;
```
bjoo, line 103: Also here?

tests/blas_interface_test_gtest.hpp
weinbe2: I wanted to switch to letting `gtest` handle the tear-down

bjoo (approved):
On a very cursory review this looks fine to me. Potential additions of collapse clauses to OpenMP #pragma locations, and there were one or through places I didn't understand why we went from errorQuda to ASSERT_TRUE(false). In any case I don't see any major issues.

tests/utils/gauge_utils.cpp
weinbe2: Outside of the scope of this PR, but per our offline discussion I filed an issue here: https://github.com/lattice/quda/issues/1619

tests/utils/gauge_utils.cpp
weinbe2: Logged here: https://github.com/lattice/quda/issues/1619

tests/utils/host_utils.cpp
weinbe2: Logged here: https://github.com/lattice/quda/issues/1619

tests/utils/staggered_gauge_utils.cpp
weinbe2: Logged here: https://github.com/lattice/quda/issues/1619

tests/utils/staggered_host_utils.cpp
weinbe2: Logged here: https://github.com/lattice/quda/issues/1619

tests/utils/staggered_host_utils.cpp
weinbe2: Logged here: https://github.com/lattice/quda/issues/1619

weinbe2:
@bjoo I believe I've addressed your comments. I also pulled in a few more commits, if you don't mind giving it a fresh look. Assuming it passes CI it should be good, pending your last thoughts.

bjoo (approved):
Looks Good to me.

maddyscientist (approved):

PR #1624, opened 2026-03-31
Add twisted clover support with non-zero mu to openQxD interface

fmargari:
This PR adds support for the twisted clover Dirac operator with non-zero twisted mass mu in the openQxD interface. The mu parameter is now read directly from the openQxD dirac parameters instead of being passed explicitly.


maddyscientist (approved):
LGTM.  Only thing I see is that clang-format needs to be applied on the diff.

chaoos:
@fmargari Looks good to too. Thanks for the addition! Only thing is, can you put

```c++
logQuda(QUDA_DEBUG_VERBOSE, ...);
```

instead of the raw debug `printfQuda` statements?

fmargari:
Yes!! Thanks for the feedback

maddyscientist:
@fmargari checking if you will apply the review suggestions so we can get this merged?

maddyscientist:
@fmargari one other addition you can make when you submit an update to this PR: please feel free to add your name to the list of authors in the README 😄

fmargari:
Sorry, closed by mistake, reopening! And thanks for the comment about the author lists :)

PR #1626, opened 2026-04-24
Add cassert include for assert function

dmcdougall:
This fixes a build error for debug builds.


maddyscientist (approved):
Thanks @dmcdougall 😄

PR #1628, opened 2026-05-05
Feature/blas prefetch

maddyscientist:
This PR is a stepping stone to lifting all boats in QUDA, to increase bytes in flight to improve bandwidth saturation for kernels that are otherwise lacking
* Add support for site unrolling to 1d, 2d, 3d, reduction and multi-reduction kernels
   * Enabled through the addition of a constexpr static variable `work_item_unroll` and kernel specialization
   * The functor for which we desire to implement loop unrolling must overload the value of `work_item_unroll`
   * It must also implement a specialized `operator()` method that specifies the unroll value as a template parameter, and accept as an additional stride parameter that declares the stride between unrolled elements
   * When using grid-stride form, the working set size does not need to be divisible by the loop unroll factor, however for the non-grid-stride case, this restriction is present (and caught at run-time if violated)
* Implemented for both streaming blas (blas_core and multi_blas_core) and reduction blas (reduce_core and multi_reduce_core) kernels
   * Enabled with CMake parameters `QUDA_BLAS_UNROLL_STREAMING` and `QUDA_BLAS_UNROLL_REDUCE`, respectively (present default is to keep both at value of 1)
    * Conversion of blas and multi_blas kernels to be non-grid-stride as this leads to better performance on recent architecture
   * For performance exploration, unroll value is part of tuning string (as is whether grid-stride or not)
   * Limited benefit is observed in the reduction kernels, and further optimization of the reduction kernels is deferred to a subsequent PR.
* As part of the PR we also bump up QUDA to use C++20 as the use of concepts makes it much easier to specialize unrolling on a per-kernel basis
* Fixes an issue introduced in #1604 which means a lattice with a local dimension greater than 128 leads to erroneous behavior if `QUDA_DSLASH_DOUBLE_STORE=ON`.
   * Limit has been raised to 256 and any larger will be caught at runtime


include/blas_helper.cuh
```diff
@@ -1,7 +1,21 @@
...
+#include <cstdio>
+#include <cstring>
+
+/* Defaults when CMake does not define these (e.g. IDE / standalone parse). */
+#ifndef QUDA_BLAS_UNROLL_STREAMING
+#define QUDA_BLAS_UNROLL_STREAMING 1
+#endif
+#ifndef QUDA_BLAS_UNROLL_REDUCE
+#define QUDA_BLAS_UNROLL_REDUCE 1
+#endif
```
bjoo: Do we want the raw defines here? Would it be better to square them away somewhere? On the other hand, they are not system specific but germane to the BLAS so I am fine with them here.

lib/targets/cuda/target_cuda.cmake
```diff
@@ -34,7 +34,7 @@ mark_as_advanced(CMAKE_CUDA_ARCHITECTURES)
 set(CMAKE_CUDA_HOST_COMPILER
     "${CMAKE_CXX_COMPILER}"
     CACHE FILEPATH "Host compiler to be used by nvcc")
-set(CMAKE_CUDA_STANDARD ${QUDA_CXX_STANDARD})
+set(CMAKE_CUDA_STANDARD ${CMAKE_CXX_STANDARD})
```
bjoo, line 37: Don't we want to force a specific standard here -- e.g. 20 so we don't have to add it on the command line?

bjoo: Nevermind, I see the toplevel CMakeLists.txt defaults this to C++20.

bjoo (approved):
Looks fine to me. Apart from the prefetch and bugfix, there is a lot of nice casting hygiene here. I only wonder about the CXX standard defaulting. We could make QUDA_CXX_STANDARD a non user settable value and initialize CMAKE_CXX_STANDARD. A user may set this to something wrong. 
Perhaps we can add a check to make sure CMAKE_CXX_STANDARD is at least a minimum? Ie we don't mind if people override it to be a more recent standard but we don't want to go below a minimum (e.g. 20)

include/kernels/gauge_shift.cuh
```diff
@@ -45,7 +45,7 @@ namespace quda
 
     __device__ __host__ void operator()(int x_cb, int parity, int dir)
     {
-      byte_array<int8_t, 4> x = {};
+      byte_array<uint8_t, 4> x = {};
```
weinbe2: Discussed offline: bumping this up to `uint16_t` to be infinitely future-proof, since we're seeing some lattices with local dimensions >= 256 (on the other hand, I'm not expecting to see any lattices with local dimensions >= 65536 in the near term...)

maddyscientist: 4c9673b8f2abb07e0b6e28f659fc568bc9dd169f fixes this.  We now have a generic `packed_array` type which can have either 8 or 16 bit elements, with total size either 32 or 64 bit supported.  Gauge shift kernel switched to using `packed_array<uint16_t, 4>` which I've confirmed allows for the doubles-the path to have large local dimensions.

include/blas_helper.cuh
maddyscientist: 469a286ecc7a5aae39bbff0ea7a3ac35e2d116a1

PR #1629, opened 2026-05-07
Cleanup stale IPC handles

dmcdougall:
Disclaimer: I did use AI to help me diagnose this problem.  The analysis looks sound, and the patch is easy to verify and explain, and it fixes a whole slew of P2P test failures (in addition to the ongoing bugfixes that are happening in https://github.com/ROCm/rocm-systems/pull/5575).

This PR does three things.

First, it cleans up stale IPC handles that were not being cleaned up on the path that resizes ghost buffers in
`LatticeField::allocateGhostBuffer`.  This method was freeing device memory without calling `destroyIPCComms()` first.  Later, when the buffer is resized, we call `qudaMalloc` and can land on the same VMA. When this happens, a subsequent `hipIpcGetMemhandle` gets a stale view of inter-process memory and was returning `hipErrorInvalidValue`.  The fix simply adds a call to `destroyIPCComms()` in the that resizes ghost buffers before freeing memory.  For context,
`LatticeField::freeGhostBuffer` is correctly calling `destroyIPCComms` before freeing ghost buffers.

Second, `LatticeField::createIPCComms` is leaking events.  When `initIPCComms` is true and `ghost_field_reset` is true, we just overwrite `ipcRemoteCopyEvent[b]` and `ipcCopyEvent[b]` without freeing the old one first.  The fix just inserts branch when both of these flags are true and inserts a call to `destroyIPCComms`.

Third, `comm_destroy_neighbor_event` only destroys local events created with `hipEventCreateWithFlags` and it doesn't destroy IPC events imported with `hipIpcOpenEventHandle`.  This leaks events.  The fix is simply to call `hipEventDestroy` on the remote event.

This PR only makes these changes in the hip backend.  Would you like me to make the same changes in the cuda backend too?  The analysis is agnostic to green/red/blue teams, but I didn't want to step on any toes.  Let me know what you'd like me to do.

### DIFF
--- a/lib/lattice_field.cpp
+++ b/lib/lattice_field.cpp
@@ -261,6 +261,13 @@ namespace quda {
           // before we free any comms buffers
           qudaDeviceSynchronize();
           comm_barrier();
+          // Tear down any existing IPC comms before freeing the ghost
+          // buffers.  Otherwise a peer rank can still hold an
+          // hipIpcOpenMemHandle mapping of a buffer we are about to
+          // hipFree, and a subsequent hipMalloc that lands at the same
+          // virtual address can then fail hipIpcGetMemHandle with
+          // hipErrorInvalidValue.
+          destroyIPCComms();
           for (int b=0; b<2; b++) {
             device_comms_pinned_free(ghost_recv_buffer_d[b]);
             device_comms_pinned_free(ghost_send_buffer_d[b]);
@@ -448,6 +455,13 @@ namespace quda {
   void LatticeField::createIPCComms() const
   {
     if ( initIPCComms && !ghost_field_reset ) return;
+
+    // If we already have IPC state and the ghost field has been reset,
+    // tear it down before recreating; otherwise the IPC mem/event handle
+    // arrays would be silently overwritten and the previous events
+    // (created with hipEventCreateWithFlags) and imported handles would
+    // leak.
+    if (initIPCComms && ghost_field_reset) destroyIPCComms();
 
     if (!initComms) errorQuda("Can only be called after create comms");
     if ((!ghost_recv_buffer_d[0] || !ghost_recv_buffer_d[1]) && comm_size() > 1)
--- a/lib/targets/cuda/comm_target.cpp
+++ b/lib/targets/cuda/comm_target.cpp
@@ -169,13 +169,28 @@ void comm_create_neighbor_event(array_2d<qudaEvent_t, QUDA_MAX_DIM, 2> &remote,
   }
 }
 
-void comm_destroy_neighbor_event(array_2d<qudaEvent_t, QUDA_MAX_DIM, 2> &, array_2d<qudaEvent_t, QUDA_MAX_DIM, 2> &local)
+void comm_destroy_neighbor_event(array_2d<qudaEvent_t, QUDA_MAX_DIM, 2> &remote,
+                                 array_2d<qudaEvent_t, QUDA_MAX_DIM, 2> &local)
 {
   for (int dim = 0; dim < 4; ++dim) {
     if (comm_dim(dim) == 1) continue;
     for (int dir = 0; dir < 2; dir++) {
+      if (!comm_peer2peer_enabled(dir, dim)) continue;
+      // First close our imported view of the neighbour's event (the
+      // counterpart to cudaIpcOpenEventHandle in
+      // comm_create_neighbor_event).  Without this, every IPC reset
+      // leaks the imported handle.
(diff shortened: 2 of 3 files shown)


leonhostetler:
@dmcdougall I'm curious, does this PR (and the related rocm bugfix) fix the issues with running QUDA on Frontier with P2P enabled?

As you may be aware (since OLCF staff escalated it to COE/AMD), I opened a ticket with OLCF last week about issues running QUDA with rocm/6.x and rocm/7.x on Frontier. With rocm/6.x, QUDA gives silent numerical errors in the dslash halo exchanges that go over P2P. This is apparently due to broken IPC events in rocm/6, which you're well aware of after investigating it last year. The hope was that the issue would be fixed in rocm/7. However, with rocm/7 QUDA now crashes with a `qudaStreamWaitEvent_` error due apparently to a rocm bug involving IPC event tracking. I provided a small QUDA-less reproducer (generated with the help of Claude Code) to OLCF that reproduces both the 6.x and 7.x bugs. With this PR and the rocm bugfix, should I expect QUDA with P2P enabled to work with the next version (7.2.4?) of rocm?

dmcdougall:
> I'm curious, does this PR (and the related rocm bugfix) fix the issues with running QUDA on Frontier with P2P enabled?

I'm fairly confident, yes.  This PR atop https://github.com/ROCm/rocm-systems/pull/5575 gave me a full passing suite of QUDA tests with P2P enabled.  My tests were done on two MPI ranks on two MI300X GPUs with OpenMPI+UCX, so I don't have cray-mpich and MI250X coverage, but I don't think that difference matters given what I believe is the root cause of the problem.

> With rocm/6.x, QUDA gives silent numerical errors in the dslash halo exchanges that go over P2P. This is apparently due to broken IPC events in rocm/6, which you're well aware of after investigating it last year.

Yes, this was because `hipEventSynchronize` was broken on IPC HIP events.  This problem was fixed from ROCm 7.0 onwards.  And indeed my minimal reproducer is a test in the hip test suite and this test is still passing.

> The hope was that the issue would be fixed in rocm/7. However, with rocm/7 QUDA now crashes with a qudaStreamWaitEvent_ error due apparently to a rocm bug involving IPC event tracking.

Yes.  This is because there is a limitation in the hip runtime that you cannot `hipStreamWaitEvent` on an IPC HIP event more than 32 times.  See [here](https://github.com/ROCm/rocm-systems/blob/rocm-7.2.0/projects/clr/hipamd/src/hip_event.hpp#L76-L84) and [here](https://github.com/ROCm/rocm-systems/blob/rocm-7.2.0/projects/clr/hipamd/src/hip_event_ipc.cpp#L107).  This limitation is removed in https://github.com/ROCm/rocm-systems/pull/5575.

This is a different breakage than the `hipEventSynchronize` breakage that I talked about above, with a different root cause.

>  I provided a small QUDA-less reproducer

Me too :), that's what started the work in https://github.com/ROCm/rocm-systems/pull/5575.

I should note that with https://github.com/ROCm/rocm-systems/pull/5575 but _without this QUDA PR_, we run into `hipIpcGetMemhandle` errors and other `qudaStreamWait` errors.  This is due to the leakage of events that cause the IPC mechanism to either run out of resources, or re-use stale handles.  There was additional work that went in to https://github.com/ROCm/rocm-systems/pull/5575 after my initial testing to mitigate some of those, but in my opinion the application should still be fixed to avoid the situation from ever happening again.

> With this PR and the rocm bugfix, should I expect QUDA with P2P enabled to work with the next version (7.2.4?) of rocm?

I can't talk about software release timelines publicly.  And I can't talk about what fixes will or will not appear in a given ROCm release.

Working with Henry Camacho, I think the best thing for users to do in the meantime is simply to turn off P2P with `QUDA_ENABLE_P2P=0` until we have all the fixes in place for you.

I've tried to answer your questions to the best of my ability, please let me know if something is unclear.

leonhostetler:
Thank you, your response was very helpful!

maddyscientist (changes requested):
Thanks for the contribution @dmcdougall.

Noting that the memory handle leak that is observed on ROCm is not observed in CUDA, but a little digging suggests this is due to difference between ROCm and CUDA in this regard (ROCm needing the memory handle to be closed and not simply freed).

Could you port the event locking fix to the CUDA comm_target.cpp?  We can then get this merged.

dmcdougall:
Sure, I'm happy to do that.

maddyscientist (approved):
Thanks @dmcdougall!

PR #1630, opened 2026-05-08
Early thread exit handling for ndeg twisted dslashes

jcosborn:
This is the next step in merging the early thread exit handling code from the SYCL branch.  This applies it to the ndeg twisted dshashes.  After this, there is still the DW and coarse operators to merge, but since these changes are somewhat complex, I wanted to do only a few files at a time.


include/kernels/dslash_clover_helper.cuh
```diff
@@ -214,15 +215,21 @@ namespace quda {
...
-
+      fermion in;
       int chirality = flavor; // relabel flavor as chirality
+      Mat A;
+      if (!allthreads || alive) {
+        in = arg.in[src_idx](my_flavor_idx, spinor_parity);
+        in.toRel(); // change to chiral basis here
+        A = arg.clover(x_cb, clover_parity, chirality);
+      } else {
+        in = fermion {};
```
maddyscientist: perhaps just the simpler approach would be better here 
```
in = {};
A = {};
```

maddyscientist (approved):
Apart from my minor style comment about initialization, this looks good to go.  Nice easy PRs are always welcome :)

include/kernels/dslash_clover_helper.cuh
jcosborn: a5840cabef4ad4ebe7d003b9832943ebd769b9b0

PR #1631, opened 2026-05-12
Fix `is_native` in `PackGhostArg`

SaltyChiang:
In `PackGhostArg`, the old way is still used to decide if a field order is native:
```c
    static constexpr bool is_native = order <= 8;
```

We should use the new value to set the value:
```c
    static constexpr bool is_native = (order == QUDA_NATIVE_FIELD_ORDER);
```

Another minor change in this PR is disabling `__grid_constant__` attribute on pre-Volta devices (with compute capability less than 7.0).


SaltyChiang:
The `is_native` issue caused a failure saying shared memory size mismatch while performing
- an inversion with a clover-Wilson type fermion action, on
- multiple Volta or post-Volta GPUs, with
- MMA and multi-RHS enabled.

maddyscientist (approved):
Thank you @SaltyChiang

PR #1632, opened 2026-05-19
add early exit handling to DW kernels

jcosborn:
Adding early exit handling from SYCL branch to DW kernels.


maddyscientist (approved):
Another nice short commit 😉

PR #1633, opened 2026-05-26
MILC interface: Exact current

leonhostetler:
This is the second and final PR in the series to enable exact current calculations for HVP in the MILC/QUDA interface. The previous PR was #1590 .

Changes:

1. Adds a new MILC interface function, `qudaExactCurrent`, in `lib/milc_interface.cpp`. This computes the exact low-mode contribution to the current densities needed for HVP current calculations.

2. Implements eigenvalue shifting in the MILC interface for preserved deflation spaces. Previously, when using a preserved deflation space across multiple solves, the eigenvalues were recomputed each time the mass changed. With this update, `computeEvals` is called only once, during the first solve. The resulting zero-mass eigenvalues are cached and shifted as needed for subsequent masses.

   This avoids many unnecessary mat-vecs and greatly reduces output volume: eigenvalues are now printed only once instead of once per solve, which previously could produce O(10^6) lines of eigenvalue output in a typical production run.

3. Makes `lib/eig_block_trlm.cpp` and `lib/eig_trlm.cpp` slightly more verbose by printing progress notifications every 10th or 100th Lanczos step. On production-size lattices, eigensolves can otherwise run for more than an hour without any visible progress, which can make a normally progressing job appear to be stalled.

Testing:

- Verified exact-current results against the corresponding MILC-side implementation.
- Verified that preserved deflation spaces reuse cached zero-mass eigenvalues and shift them correctly for subsequent masses.


lib/eig_trlm.cpp
```diff
@@ -77,7 +77,10 @@ namespace quda
     // Loop over restart iterations.
     while (restart_iter < max_restarts && !converged) {
 
-      for (int step = num_keep; step < n_kr; step++) lanczosStep(kSpace, step);
+      for (int step = num_keep; step < n_kr; step++) {
+        if (step % 100 == 0) logQuda(QUDA_VERBOSE, " starting lanczosStep %d\n", step);
```
maddyscientist: Why `step % 0` with block Lanczos but 10 with regular?

lib/milc_interface.cpp
```diff
@@ -1378,20 +1402,34 @@ void qudaLoadDeflationSpace(int external_precision, int quda_precision, const vo
...
+        // Opt-in residual sanity check
+        if (getVerbosity() >= QUDA_DEBUG_VERBOSE) {
+          dEig->M(temp, space->evecs[i]);
+          auto res = blas::caxpbyNorm(space->evals[i], space->evecs[i], n_unit, temp);
+          logQuda(QUDA_DEBUG_VERBOSE, "Eval[%04d] = (%+.16e,%+.16e), Residual = %+.16e (shifted)\n", i,
+                  space->evals[i].real(), space->evals[i].imag(), sqrt(res[0]));
+        }
+      } else {
+        // Compute eigenvalues, lambda_i = v_i^dag A v_i / (v_i^dag * v_i)
+        dEig->M(temp, space->evecs[i]);
```
maddyscientist: We could make this MRHS?

maddyscientist (approved):
This looks good to me.  I'll leave @weinbe2 to comment on the MILC interface specifics.  I've left a couple of comments.

lib/eig_trlm.cpp
leonhostetler: I think I originally did that because the loop for regular Lanczos increments by `1` whereas for block Lanczos it increments by `block_size`. I have revised both so they should now always print 10 notifications per restart.

lib/milc_interface.cpp
leonhostetler: With [4b9a4b6](https://github.com/lattice/quda/pull/1633/commits/4b9a4b6a32f2061388e440423a0bf8612e2f1c44) I have made the eigenvector and eigenvalue computations MRHS throughout `qudaLoadDeflationSpace`.

lib/milc_interface.cpp
```diff
@@ -1469,6 +1532,153 @@ void qudaLoadDeflationSpace(int external_precision, int quda_precision, const vo
...
+      break;
+    default: errorQuda("Wrong number of masses %d!", nmasses);
+    }
+
+    for (int mu = 0; mu < 4; mu++) {
+
+      // Do gauge covariant shift and flip sign on ODD sites
+      myCovDev.MCD(gr_mu, gr0, mu);
+      blas::copy(gr_mu.Odd(), gr0.Odd());
+      blas::ax(-1.0, gr_mu.Odd());
```
maddyscientist: You can replace these two lines with
```
blas::axy(-1.0, gr0.Odd(), gr_mu.Odd());
```

lib/milc_interface.cpp
```diff
@@ -1786,6 +2019,23 @@ void qudaInvertMsrcDeflatable(int external_precision, int quda_precision, double
...
+    // The empty() guard ensures we never overwrite an earlier snapshot from a different (or fresher) source.
+    if (preserved_evals_zero_mass.empty()) {
+      deflation_space *space = reinterpret_cast<deflation_space *>(preserved_deflation_space[local_parity]);
+      preserved_evals_zero_mass.resize(space->evals.size());
+      double m2_shift = 4.0 * mass * mass;
+      for (size_t i = 0; i < space->evals.size(); i++)
+        preserved_evals_zero_mass[i] = space->evals[i] - Complex(m2_shift, 0.0);
+
+      if (getVerbosity() >= QUDA_DEBUG_VERBOSE) {
+        logQuda(QUDA_DEBUG_VERBOSE, "Stored zero-mass eigenvalues (back-shifted from parity %d, mass %e)\n",
```
maddyscientist: These can be just `printQuda` since `logQuda` is redundant here.

lib/milc_interface.cpp
maddyscientist: perfect, thanks

maddyscientist:
Thanks for the tweaks @leonhostetler.  I've left a couple of non-essential comments that might be nice to update if you make another push.

@weinbe2 any thoughts before we merge?

lib/milc_interface.cpp
leonhostetler: Good catch, I have updated it.

lib/milc_interface.cpp
leonhostetler: I have replaced the redundant `logQuda`s with `printfQuda`s.

lib/milc_interface.cpp
```diff
@@ -1754,8 +1963,30 @@ void qudaInvertMsrcDeflatable(int external_precision, int quda_precision, double
     // Check that preserved eigenvalues are for this mass
     double epsilon = device_precision == QUDA_DOUBLE_PRECISION ? __DBL_EPSILON__ : __FLT_EPSILON__;
     if (fabs(mass - preserved_evals_mass[local_parity]) > epsilon) {
-      logQuda(QUDA_VERBOSE, "Resetting eigenvalues to mass %e\n", invertParam.mass);
-      qep.preserve_evals = QUDA_BOOLEAN_FALSE;
+      // If not, shift to correct mass using cached zero-mass eigenvalues
+      // NAIK CAVEAT: shifted evals are only approximate when inv_args.naik_epsilon != 0
+      if (!preserved_evals_zero_mass.empty()) {
+        assert(qep.preserve_deflation_space != nullptr);
```
weinbe2: This (and other `assert` calls) should be replaced with conditional checks and `errorQuda`

weinbe2:
Thanks for the PR, @leonhostetler , this largely looks good. I have a few questions/thoughts:

 * Resetting the eigenvalue mass shift to the sentinel value of `-1` seems a bit fragile still---`qudaCleanupDeflationSpace()` isn't triggered with a call to `invalidateGaugeQuda()`, or a fresh gauge field is loaded, for example. Maybe this is semi-intentional for heavy quarks where you want to plug in the heavy-quark-modified gauge fields, even if the eigenvectors technically change a bit, but maybe in that case we'd want a different routine for uploading those that _doesn't_ invalidate the eigenvalue mass shift. I guess these considerations apply to the eigenvectors as well, so I should have considered this in an earlier PR.
 * Did you try this with a `PRECISION=1` MILC build? I see you put in the proper plumbing to set the host precision in `qudaExactCurrent`, this is just making sure.

No other feedback, if you've verified that this works with the current workflow in MILC I trust you. It could be useful if you documented this on the QUDA Wiki in case other people wanted to use this, but the PR doesn't need to be blocked on that.

lib/milc_interface.cpp
leonhostetler: [e8275c3](https://github.com/lattice/quda/pull/1633/commits/e8275c3bdb6baacb1c781892ddff7de571b4933a)

leonhostetler:
Thanks for reviewing this, @weinbe2 !

> Resetting the eigenvalue mass shift to the sentinel value of -1 seems a bit fragile still

I agree. On the one hand, in practice, I don't know that we have any workflows where we load a completely different gauge config during a run. In principle though, we should probably guard against that possibility. As for the heavy quarks, yes, I figured that since the eigenvectors from the non-Naik'ed operator were already being used as an approximation to the real eigenvectors, then why not the eigenvalues as well? For deflated CG, I figured/assumed this would be fine in that they should be a good approximation and the CG should correct any errors. For other uses of the eigenvalues (as in `qudaExactCurrent` where the approximation is not corrected by CG), now that I think about it a little more, perhaps this is not a safe assumption at all. How do you think this should be handled? Maybe play it safe and force the eigenvalues to reset upon any gauge field change unless some explicit opt-in is set?

> Did you try this with a `PRECISION=1` MILC build?

I had neglected to test this after recent changes because an earlier investigation of single precision for our exact current calculations showed that it was not feasible for us. It was simply not precise enough. Nevertheless, I should've made sure that it works in principle. After fixing a number of bugs ([b35fe11](https://github.com/lattice/quda/pull/1633/commits/b35fe119e0183599afb60043360c3f819655ac05), [a132898](https://github.com/lattice/quda/pull/1633/commits/a13289823f400d36a09c3847de8f316465c6224f), [b2b5463](https://github.com/lattice/quda/pull/1633/commits/b2b54632dfcad6ce97cc3fffd3c8baa8db49e2b5)) it works now.

I have also taken the opportunity to make a few more updates to `qudaExactCurrent`. The commit [3b85f5c](https://github.com/lattice/quda/pull/1633/commits/3b85f5c55bf36aef2e6919f5a23191f2ab582c39) changes the function signature by changing the `masses`, `jlow_mu`, and `jlow_mu2` from type `double*` to type `void*` since MILC uses type `Real` which is either float or double depending on MILC precision. It had worked, albeit with compiler warnings on the MILC side, but this makes it more correct, I think.

The commit [8357c77](https://github.com/lattice/quda/pull/1633/commits/8357c77a0807dbd9898405b3ee38a459cc74f01c) does a refactoring of `qudaExactCurrent` to significantly reduce the number of small D2H transfers by replacing the per-eigenvector host accumulation with GPU-side accumulation and a small number of D2H transfers after the loop over eigenvectors.

weinbe2 (approved):
This all looks good now, thank you for checking the precision = 1 builds---sounds like it indeed help make things more robust!

This is good to merge from my end.

PR #1634, opened 2026-06-01
Unify host pinned memory under host_pinned_malloc

maddyscientist:
## Summary
- Replace separate `pinned_malloc` / `mapped_malloc` APIs with a single **`host_pinned_malloc`** that always registers host memory as page-locked and GPU-mapped (`cudaHostRegisterMapped | cudaHostRegisterPortable`, HIP equivalent).
- Remove **`QUDA_MEMORY_MAPPED`**; use **`QUDA_MEMORY_HOST_PINNED`** everywhere (with device view via `get_mapped_device_pointer` in `quda_ptr`).
- Collapse allocation tracking into **`host_pinned_allocated()`** / **`host_pinned_allocated_peak()`**; rename pool helpers to **`pool_host_pinned_malloc`** / **`pool_host_pinned_free`** / **`flush_host_pinned()`**.
- Improve CUDA/HIP malloc error reporting: log API name, error code, name, and string before `errorQuda`.
- **`QUDA_MEMORY_DEVICE`** and **`QUDA_MEMORY_DEVICE_PINNED`** are unchanged (device-side P2P/comms allocations).
## Breaking changes
- Removed macros: `pinned_malloc`, `mapped_malloc`, `pool_pinned_malloc`, `pool_pinned_free`
- Removed enum value: `QUDA_MEMORY_MAPPED` (subsequent enum values renumbered)
- Removed stats: `pinned_allocated()`, `mapped_allocated()`, and peak variants


maddyscientist:
@dmcdougall any concerns from you with this PR?  This PR unifies `mapped_malloc` and `pinned_malloc` with a single `host_pinned_malloc` which is both pinned to the host and mapped to allow for GPU to read / write a given CPU allocation.  Basically just a clean up, as having both was redundant as at least on CUDA on Linux64 all host pinned allocations are also available for the GPU to read / write directly.  I assume ROCm is similar?

This is a vestige from when QUDA was first written and we had 32-bit support, and allocations needed to be pinned _and_ mapped to be visible to the GPU address space.

lib/targets/cuda/malloc.cpp
```diff
@@ -20,7 +20,23 @@
 namespace quda
 {
 
-  enum AllocType { DEVICE, DEVICE_PINNED, HOST, PINNED, MAPPED, MANAGED, SHMEM, N_ALLOC_TYPE };
+  static void report_cuda_error(const char *api, cudaError_t err)
+  {
+    printfQuda("%s failed. Error code %d, error name %s, error string %s\n", api, static_cast<int>(err),
+               cudaGetErrorName(err), cudaGetErrorString(err));
```
weinbe2, line 26: I see you did a check for `nullptr` in `report_cu_error`, should we do the same thing here? More specifically, capture `cudaGetErrorName` and `cudaGetErrorString` to variables instead of passing them directly to `printfQuda`, then perform a check against `nullptr`?

lib/targets/hip/malloc.cpp
```diff
@@ -15,7 +15,13 @@
 namespace quda
 {
 
-  enum AllocType { DEVICE, DEVICE_PINNED, HOST, PINNED, MAPPED, MANAGED, N_ALLOC_TYPE };
+  static void report_hip_error(const char *api, hipError_t err)
```
weinbe2, line 18: Apply the same comment above about capturing `*GetErrorName` and `*GetErrorString` to temporary variables and comparing it against `nullptr`

lib/targets/cuda/malloc.cpp
```diff
@@ -98,7 +110,7 @@ namespace quda
 
   static void print_alloc(AllocType type)
   {
-    const char *type_str[] = {"Device", "Device Pinned", "Host  ", "Pinned", "Mapped", "Managed", "Shmem "};
+    const char *type_str[] = {"Device", "Device Pinned", "Host  ", "Host Pinned", "Managed", "Shmem "};
```
weinbe2: Not strictly necessary, but should we get rid of the extra spaces in `"Host  "` and `"Shmem "` or add extra spaces and padding to the other fields (and to those) as appropriate?

dmcdougall:
> at least on CUDA on Linux64 all host pinned allocations are also available for the GPU to read / write directly. I assume ROCm is similar?

Yeah, a call to `hipHostMalloc` will update the GPU's page table so that the GPU can read/write to that host memory.  If that's what you mean by "pinned" then, yes, no concerns from me.

lib/targets/cuda/malloc.cpp
weinbe2: Discussed offline that `cudaGetErrorName` and `cudaGetErrorString` are guaranteed to return non-`nullptr` as documented by the CUDA documentation.

lib/targets/hip/malloc.cpp
weinbe2: @dmcdougall are `hipGetErrorName` and `hipGetErrorString` guaranteed to return non-nullptr?

weinbe2 (approved):
This looks good to me, safe to merge from my perspective. Are there any concerns in the cscs failures?

weinbe2:
Answered my own question, the cscs failures are completely unrelated, it seems. Can we get that kicked off again?

lib/targets/cuda/malloc.cpp
maddyscientist: [4a88355](https://github.com/lattice/quda/pull/1634/commits/4a8835545d83effc9bfe5660e5e8db7297dfaeb3)

lib/targets/hip/malloc.cpp
maddyscientist: Answering this: the explicit API reference is here:

[HIP Runtime API — Error handling](https://rocm.docs.amd.com/projects/HIP/en/docs-7.0.2/reference/hip_runtime_api/modules/error_handling.html)

For the two functions QUDA uses in report_hip_error, the docs state:

`hipGetErrorName`

Returns: const char pointer to the NULL-terminated error name

`hipGetErrorString`

Returns: const char pointer to the NULL-terminated error string

so we're good with the code as is.

PR #1638, opened 2026-06-24
Changes required for the Wilson clover N_f=1 (RHMC) force computation…

kottnad:
… to work with tmLQCD. Removed hackish usage of detratio arg in computeCloverForce() and computeTMCloverForceQuda() signatures.

### DIFF
--- a/include/quda.h
+++ b/include/quda.h
@@ -1560,9 +1560,11 @@ extern "C" {
    * @param gauge_param Gauge field meta data
    * @param inv_param Dirac and solver meta data
    * @param detratio if 0 compute the force of a determinant otherwise compute the force from a ratio of determinants
+   * @param trlog Whether to compute the trlog term
+   * @param rhmc_flavors Number of flavors used in RHMC, i.e. 0: not using RHMC, 1: single flavor, 2 for e.g. non-degenerate tm doublet
    */
   void computeTMCloverForceQuda(void *h_mom, void **h_x, void **h_x0, double *coeff, int nvector,
-                                QudaGaugeParam *gauge_param, QudaInvertParam *inv_param, int detratio, int trlog, int rhmc);
+                                QudaGaugeParam *gauge_param, QudaInvertParam *inv_param, int detratio, bool trlog, int rhmc_flavors);
 
   /**
    * Compute the naive staggered force.  All fields must be in the same precision.
--- a/include/quda.h
+++ b/include/quda.h
@@ -1562,7 +1562,7 @@ extern "C" {
    * @param detratio if 0 compute the force of a determinant otherwise compute the force from a ratio of determinants
    */
   void computeTMCloverForceQuda(void *h_mom, void **h_x, void **h_x0, double *coeff, int nvector,
-                                QudaGaugeParam *gauge_param, QudaInvertParam *inv_param, int detratio);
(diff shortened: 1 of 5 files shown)


include/quda.h
```diff
@@ -1562,7 +1562,7 @@ extern "C" {
    * @param detratio if 0 compute the force of a determinant otherwise compute the force from a ratio of determinants
    */
   void computeTMCloverForceQuda(void *h_mom, void **h_x, void **h_x0, double *coeff, int nvector,
-                                QudaGaugeParam *gauge_param, QudaInvertParam *inv_param, int detratio);
+                                QudaGaugeParam *gauge_param, QudaInvertParam *inv_param, int detratio, int trlog, int rhmc);
```
maddyscientist: These `int` variables can be changed to `bool` types.  Historically we would have had these as `int` but since we have defined the interface to be C99, we can support `bool`.

lib/clover_force.cpp
```diff
@@ -11,11 +11,11 @@ namespace quda
   void computeCloverForce(GaugeField &mom, const GaugeField &gaugeEx, const GaugeField &gauge,
                           const CloverField &clover, cvector_ref<ColorSpinorField> &x, cvector_ref<ColorSpinorField> &x0,
                           const std::vector<double> &coeff, const std::vector<array<double, 2>> &epsilon,
-                          double sigma_coeff, bool detratio, QudaInvertParam &inv_param)
+                          double sigma_coeff, bool detratio, bool trlog, bool rhmc, QudaInvertParam &inv_param)
   {
     if (inv_param.matpc_type != QUDA_MATPC_EVEN_EVEN_ASYMMETRIC && inv_param.matpc_type != QUDA_MATPC_ODD_ODD_ASYMMETRIC)
       errorQuda("MatPC type %d not supported", inv_param.matpc_type);
-
+// error fuer trlog + detratio und rhmc vs detratio
```
maddyscientist: English comments please 😉

maddyscientist (changes requested):
Thanks for the submission @kottnad.  This looks good to me, just a few minor comments, but once those are addressed we can merge this,

include/quda.h
kottnad: Thanks! I have changed to bool for trlog in the updated commit f55458 . However, I renamed rhmc -> rhmc_flavors and kept it as int, because we (externally) need to be able to call this with 0, 1 or 2 flavors (for the non-deg tm doublet), i.e. the actual number of flavors in the RHMC.

lib/clover_force.cpp
kottnad: This one was just an overlooked reminder to my self. I deleted it :)

maddyscientist:
Thanks for changes.  I have tested the code, but it looks like the changes made in this PR are causing the clover_force_test to fail now for non-generate twisted clover fermions, so a bug has crept in there with these changes.

If you run `ctest` with a twisted-clover enabled build should see the bug pretty quickly.  Alternately this should also do the trick

```
mpirun -np 1 tests/clover_force_test --dim 2 4 6 8 --compute-clover  true --solution-type mat-pc-dag-mat-pc --matpc odd-odd-asym --dagger --enable-testing true --niter 1
```

Looks like bug was introduced with the first commit in this PR.....

PR #1639, opened 2026-06-26
Use `host_pinned_malloc` for all host ghost buffer allocations

weinbe2:
FYI, Claude Opus 4.8 helped me debug this.

In some rare, reproducible but somewhat non-deterministic across machines, etc, circumstances that could only be reproduced in a container so far (but are technically always on the table...), we would see a "double registration" error when performing a pinned host allocation: `cudaErrorHostMemoryAlreadyRegistered` while running QUDA test executables. Opus traced this down to an interesting intersection between how the `cuda_copy` transport in UCX registers and unregisters host pointers (yes, you read this correctly, host pointers) and the conditions under which `glibc`'s `free()` also calls `munmap()`.

First: for sufficiently small allocations, `free()` doesn't call `munmap()`.
Second: when the `cuda_copy` transport is passed an un-registered host pointer, it will register it: https://github.com/openucx/ucx/blob/ab1f057f9921955d59c5aa86a1cd801cf6ab3d71/src/uct/cuda/cuda_copy/cuda_copy_md.c#L140-L178 . It figures out when it should _de-register_ it by intercepting system calls, however, it doesn't intercept `free` --- it only intercepts `munmap`.

This means you can end up in a circumstance like this:
1. A program creates a "small enough" host (heap) allocation using `malloc`
2. The program passes it to an MPI call; the `cuda_copy` transport registers it
3. The program frees the allocation using `free` --- but, because it's "small enough", there's no `munmap` call; it doesn't get unregistered.
4. Life goes on, and later the program creates another heap allocation using `malloc` that _happens to, potentially completely coincidentally_ live in the same, already registered heap allocation.
5. This time, the program tries to register it using `cudaHostRegister` ... and we get the `cudaErrorHostMemoryAlreadyRegistered`.

...and that's what was happening. The first "small enough" allocation was a ghost buffer allocation, which is a `safe_malloc` -> technically a `posix_memalign`. The second allocation was a pinned allocation for a reduction buffer, which is a `host_pinned_malloc`.

Yes, this is bizarre, but it was verified in two ways.
1. Disabling the host registration cache: `export UCX_RCACHE_ENABLE=n` --- note this is `*_RCACHE`, not `*_CACHE`. No cache, no registration of the first allocation, no issue on the second allocation.
2. Forcing the `munmap` by inserting `malloc_trip(0);` at the start `host_pinned_malloc_`---this is the "nuclear" option since it kills of **every** mapped but not `malloc`'d allocation, so this is just a proof point and not a solution.

The solution in QUDA is straightforward: make sure the host communications buffers are `host_pinned_malloc`'d instead of just `safe_malloc`'d. If the host buffer is already pinned, UCX doesn't pin it/cache it, no opportunity for harm.

### DIFF
--- a/lib/color_spinor_field.cpp
+++ b/lib/color_spinor_field.cpp
@@ -964,9 +964,8 @@ namespace quda
       if (!initGhostFaceBuffer || resize) {
         freeGhostBuffer();
         for (int i = 0; i < nDimComms; i++) {
-          // Host communication buffers use host-pinned (registered) memory so that a CUDA-aware
-          // MPI/UCX registration cache does not register these pages itself and leave a stale
-          // mapping behind after they are freed (see cudaErrorHostMemoryAlreadyRegistered with UCX).
+          // Use host-pinned memory for host communication buffers so the `cuda_copy` transport in
+          // UCX doesn't own the pinning; see https://github.com/lattice/quda/pull/1639 for more context
           fwdGhostFaceBuffer[i] = host_pinned_malloc(ghostFaceBytes[i]);
           backGhostFaceBuffer[i] = host_pinned_malloc(ghostFaceBytes[i]);
           fwdGhostFaceSendBuffer[i] = host_pinned_malloc(ghostFaceBytes[i]);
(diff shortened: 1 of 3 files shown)


maddyscientist (approved):
Thanks for this amazing debug.

PR #1640, opened 2026-07-02
Simplify return statement in get_tensor_descriptor

maddyscientist:
Works around a compiler issue with CUDA 13.3.


weinbe2 (approved):
Verified good for CUDA 13.3 + `sm_80`, as a representative architecture that doesn't have TMA. This is good to merge once the tests finish up. Thanks @maddyscientist !

PR #1641, opened 2026-07-08
Feature/mnnvl p2p

bjoo:
# Summary

This work adds several modifications, the chief of which is stream-gated transport for MNNVL.  

## Description
Previous intra node P2P mechanism used CUDA/HIP IPC mechanism. The sender copied data to a shared buffer, and once the copy was completed created a remote IPC event. An MPI doorbell message was sent to the receiver. The receiver waiter first on the MPI doorbell and then began waiting on the remote IPC event to complete to know that the transfer was done. 

Multi-node NVLink use CUDA Fabric handles for remote memory operations which must be allocated using CUDA Virtual Memory Management (VMM) allocators. Remote memory allocated using VMM does not support remote IPC events. The solution to this is to use an approach we will refer to as **stream-gating**.  In this case once the transfer of the shared memory is complete the sender will issue `cuStreamWrite64` operation to a piece of remote memory which serves as the doorbell. The receiver sits in a polling loop over the various directions and can check the remote doorbell location and finally unblock with a `cuStreamWait64` operation. Follow on kernels can be queued on the stream behind this operation so once the `cuStreamWait64` operation fires subsequent kernels can dispatch without further host involvement, giving this method a very low latency. 

## Implementation notes
We implented the following:
* MNNVL build selected with CMake option `-DQUDA_MNNVL=ON`.
* Added stream gating as a user choice transport option selectable with the environment variable `QUDA_P2P_TRANSPORT=stream_gated|event`. The choice is only meaningful in CUDA builds on the non MNNVL path. For MNNVL builds, `event` is not available. On Non-MNNVL builds `event` is default.
* MNNVL comm buffer allocations proceed via VMM and allocate a single handle for their memory which they return as Fabric handles, both for P2P and even for regular MPI. MPI and P2P no longer share buffers with NVSHMEM when NVSHMEM is enabled.
* In MNNVL stream gating, P2P Remote Write operations rely on being able to flush streams using `cuStreamWait64` using a specific FLUSH flag. Whether the system in question provides the FLUSH capability is tested for at startup. On systems where the FLUSH capability is not available, P2P remote write operations are explicitly disabled and use Copy Engines only.
* in order to prevent proliferation of #ifdef lines in the main include/lib dirs the signaling API was abstracted and implementation was pushed down into the appropriate `include/targets` and `lib/targets` directory.
* The implementation should (in our basic testing did) can work accross MNNVL-Clique boundaries, gracefully falling back to MPI communications in directions where MNNVL-P2P is not available. In these use cases it may be worth setting variable `UCX_CUDA_IPC_ENABLE_MNNVL=n` so that 'cross-clique' MPI traffic doesn't attempt to use UCX's own MNNVL
* In order for fabric handles to work, the nodes must have an IMEX daemon running
* Testing was carried out using Wilson and Staggered Dslash tests, some of the hisq force tests and using short MILC trajectories. Chroma testing (especially MG) is currently still TBD. 

The implementation was carried out in large part by Claude Code (Opus 4.7 and Opus 4.8).  The expecation is that 'stream-gate' works best in the case where we have low latency requirements (e.g. Multi-Grid or strong scaling). For larger message sizes stream gating and the event based approach should perform similarly with each other as in that case  higher latencies can  be amortized (we chacked this on non-MNNVL builds where both are available).


bjoo:
All but CSCS test runs finished. CSCS build passed but test job didn't execute - for what may be non QUDA reasons

lib/targets/cuda/comm_target.cpp
```diff
@@ -1,15 +1,61 @@
...
+      return probe;
+    }
+
+    bool try_import_fabric_handle(const void *peer_handle)
+    {
+      CUmemFabricHandle fh;
+      memcpy(&fh, peer_handle, sizeof(CUmemFabricHandle));
+      CUmemGenericAllocationHandle h;
+      CUresult err = cuMemImportFromShareableHandle(&h, &fh, CU_MEM_HANDLE_TYPE_FABRIC);
+      if (err != CUDA_SUCCESS) return false;
```
hummingtree: Should we clear the CU error after this?

hummingtree: And should we check to see what is the error type to make sure we are seeing the expected error type when the remote is not accessible?

lib/targets/cuda/comm_target.cpp
```diff
@@ -71,128 +189,553 @@ namespace quda
...
+
+    // Receiver-side ordering for REMOTE_WRITE halos: add CU_STREAM_WAIT_VALUE_FLUSH
+    // so the wait drains outstanding remote writes to us before it is satisfied.
+    // Only where the device supports it (stream_gated_remote_flush_supported); on
+    // devices that reject the flag (GB200, CUDA_ERROR_NOT_SUPPORTED) remote-write is
+    // not offered as a policy (comm_p2p_remote_write_supported), so plain GEQ is
+    // correct here.
+    unsigned int wait_flags = CU_STREAM_WAIT_VALUE_GEQ;
+    if (stream_gated_remote_flush_supported) wait_flags |= CU_STREAM_WAIT_VALUE_FLUSH;
+    CUresult err = cuStreamWaitValue64(target::cuda::get_stream(stream), local, expected, wait_flags);
```
hummingtree: Why do we use `GEQ` here instead of `EQ`? I know by design they are the same but just wondering which one is better.

lib/targets/cuda/malloc.cpp
```diff
@@ -379,21 +385,201 @@ namespace quda
...
+    err = cuMemAddressReserve(&ptr_d, padded, granularity, 0, 0);
+    if (err != CUDA_SUCCESS) errorQuda("cuMemAddressReserve failed (%s:%d in %s())", file, line, func);
+    err = cuMemMap(ptr_d, padded, 0, h, 0);
+    if (err != CUDA_SUCCESS) errorQuda("cuMemMap failed (%s:%d in %s())", file, line, func);
+
+    CUmemAccessDesc access = {};
+    access.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+    access.location.id = device_id;
+    access.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+    err = cuMemSetAccess(ptr_d, padded, &access, 1);
```
hummingtree, line 455: Maybe it's time to add a macro/etc for these driver calls so we don't need to repeat the `if (err != CUDA_SUCCESS) errorQuda(` every time.

lib/targets/cuda/malloc.cpp
```diff
@@ -110,7 +113,7 @@ namespace quda
 
   static void print_alloc(AllocType type)
   {
-    const char *type_str[] = {"Device", "Device Pinned", "Host", "Host Pinned", "Managed", "Shmem"};
+    const char *type_str[] = {"Device", "Device Pinned", "Host  ", "Host Pinned", "Managed", "Shmem ", "Comm  "};
```
hummingtree, line 116: Just curious: why do we add space after `Host`, `Shmem` and `Comm`?

lib/color_spinor_field.cpp
```diff
@@ -746,6 +746,25 @@ namespace quda
...
+    // recv buffer (peer P2P writes, GDR NIC writes, unpackGhost staging).
+    // Aliases the symmetric buffer in non-NVSHMEM builds.
+    if (Location() == QUDA_CPU_FIELD_LOCATION) {
+      return nullptr;
+    } else {
+      if (bufferIndex < 2) {
+        return ghost_recv_buffer_p2p_d[bufferIndex];
+      } else {
+        return ghost_pinned_recv_buffer_hd[bufferIndex % 2];
+      }
```
hummingtree, line 760: Why would we have cases where `bufferIndex >= 2`?

lib/color_spinor_field.cpp
```diff
@@ -1232,15 +1268,64 @@ namespace quda
...
+    // mirrors sendStart's `if (!remote_write)` guard.  The signal still goes on this
+    // per-direction stream: issueGather inserts a qudaStreamWaitEvent(.., packEnd) on
+    // it, so the cuStreamWriteValue64 is correctly ordered after the remote write.
+    if (!remote_write) {
+      // P2P copy-engine (stream-gated): contiguous P2P send -> imported peer P2P recv.
+      void *ghost_dst
+        = static_cast<char *>(ghost_remote_send_buffer_p2p_d[bufferIndex][dim][dir]) + ghost_offset[dim][(dir + 1) % 2];
+      qudaMemcpyP2PAsync(ghost_dst, my_face_dim_dir_p2p_d[bufferIndex][dim][dir], ghost_face_bytes[dim], stream);
+    }
+    comm_p2p_signal_send_done(FieldKind::COLOR_SPINOR, bufferIndex, dim, dir, stream, QudaP2PSignal::STREAM_GATED);
```
hummingtree, line 1300: I am not sure but maybe we should check to see if `STREAM_GATED` is enabled/supported when we use the enum `QudaP2PSignal::STREAM_GATED` directly? (to be able to detect more errors from the decision trees up-stream)

hummingtree: Or maybe just rename this function to include `GatedStream` to make it clear.

lib/dslash_policy.hpp
```diff
@@ -559,16 +653,42 @@ namespace quda
...
-                          2 * i + dir, device::get_stream(dslashParam.remote_write ? packIndex : 2 * i + dir), false,
-                          dslashParam.remote_write),
-                        profile, QUDA_PROFILE_COMMS_START);
+                if ((dslashParam.p2p_signal == QudaP2PSignal::STREAM_GATED) && comm_peer2peer_enabled(dir, i)) {
+                  // Stream-mem-op signal send.  Queue qudaMemcpyP2PAsync + cuStreamWriteValue64
+                  // on the per-direction stream 2*i+dir.  issueGather already inserted a
+                  // qudaStreamWaitEvent(packEnd) on that stream, so the P2P copy serializes
+                  // with the pack output but runs in parallel with the other directions.
+                  PROFILE(if (dslash_comms)
+                            halo.sendStartStream(2 * i + dir, device::get_stream(2 * i + dir), dslashParam.remote_write),
```
hummingtree: Yeah I think renaming this to `sendStartStreamGated` (or something like that) would make it clear that this function is not general purpose.

lib/gauge_field.cpp
```diff
@@ -517,48 +590,85 @@ namespace quda {
...
+            sendStart(dim, dir, device::get_stream(2 * dim + dir));
+          }
+        }
+
+        // Stream-memory waits are not visible to the CUDA scheduler as a
+        // cross-device dependency.  Merge every local P2P copy+signal onto the
+        // default stream before allowing any per-dimension stream to enter a
+        // blocking remote wait.  Without this CUDA-visible edge, one dimension
+        // can occupy scheduler resources in WaitValue while another dimension's
+        // signal has not run yet, forming an intermittent cross-rank cycle.
```
hummingtree, line 634: Curious if you have seen such hang in practice?

hummingtree (approved):
Thanks @bjoo for the great work - it looks nice and neat to me. I have left a few comments out of my curiosity.

lib/targets/cuda/comm_target.cpp
bjoo: This is a reachability probe, so a failure is an expected outcome (peer not reachable over fabric → we symmetrically downgrade that link to MPI), not an error condition. On the driver API the status comes back in the return value rather than as a sticky per-thread error (unlike the runtime's cudaGetLastError), so there's nothing latched to clear here — but happy to add an explicit note/clear if you'd prefer defense-in-depth. We could refine the error checking though to see if there is something bad there.

lib/targets/cuda/comm_target.cpp
bjoo: This is the monotonic counter protocol:  the receiver waits for expected = ++recv_counter while the sender writes an ever-increasing ++send_counter into the slot. By the time a WaitValue64 is actually evaluated the peer may already have advanced the slot past expected (the next generation's write can land first). EQ would miss that exact value and block forever; GEQ is satisfied as soon as the slot reaches at-least the expected generation.

lib/targets/cuda/malloc.cpp
bjoo: Claude suggests that yes a macro is maybe worth it? It offered to generalize  target::cuda::set_driver_error(err, name, func, file, line)

lib/targets/cuda/malloc.cpp
bjoo: This was just for column alignment when displaying

lib/color_spinor_field.cpp
bjoo: So apparently this was there previously via Ghost2().

lib/gauge_field.cpp
bjoo: The invariant that comment guards — enqueue all sends before any blocking recv wait, and queue each WaitValue64 exactly once — is a design guardrail; in a correct build we haven't hit that pure GPU-stream self-deadlock. The hangs we did see during development were adjacent bugs that this discipline (and its fixes) resolved: (a) an asymmetric fabric-import "split-brain" where one rank ran P2P while its peer fell back to MPI, so a receiver waited on a signal that was never written — fixed by making the capability exchange symmetric (AND across the pair); and (b) a stale-halo/double-buffer hang tied to a bufferIndex reset. The HARD drain here exists because an async stream-merge alone lets a peer race ≥2 applies ahead and overwrite the ghost mid-read (monotonic counter + GEQ then passes on the overwritten halo), so it's as much about correctness under run-ahead as deadlock avoidance.

lib/targets/cuda/comm_target.cpp
hummingtree: Make sense. So when the remote is not accessible, just make sure that the error type is expected? From the API doc it looks like only [CUDA_ERROR_NOT_SUPPORTED](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggc6c391505e117393cc2558fff6bfc2e954756ae7ade0dfd09faeccb513dd831b) and [CUDA_ERROR_NOT_PERMITTED](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggc6c391505e117393cc2558fff6bfc2e9e23d0197c490ec332a43e55b167968a3) are expected?

lib/targets/cuda/comm_target.cpp
hummingtree: > By the time a WaitValue64 is actually evaluated the peer may already have advanced the slot past expected (the next generation's write can land first).

I think the double buffered `bufferIndex` is targeting to avoid this kind of over-by-one-step (I am inventing some words here). I am pretty sure it would work but maybe worth a documentation here.

lib/targets/cuda/comm_target.cpp
bjoo: Claude is on it :)

lib/targets/cuda/malloc.cpp
hummingtree: Up to you :)

bjoo:
Pushed some cleanup changes suggested by @hummingtree .

lib/comm_common.cpp
```diff
@@ -151,4 +153,43 @@ namespace quda
...
+        errorQuda("Unrecognised QUDA_P2P_TRANSPORT='%s' (expected 'stream_gated' or 'events')", env);
+      }
+      if (!p2p_signal_supported(requested))
+        errorQuda("QUDA_P2P_TRANSPORT='%s' is not supported by this backend/build "
+                  "(e.g. events are unavailable under QUDA_MNNVL, stream-gated under HIP)",
+                  env);
+      choice = requested;
+    }
+
+    if (getVerbosity() > QUDA_SILENT && comm_rank() == 0)
```
maddyscientist: Replace with this with `logQuda`: no need for the `if` then.

lib/targets/cuda/p2p_signal_defaults.cpp
```diff
@@ -0,0 +1,47 @@
+// lib/targets/cuda/p2p_signal_defaults.cpp
+//
+// Backend implementation of the QudaP2PSignal allow-list and default.
+//
+// CUDA backend rules:
+//   - MNNVL build (QUDA_MNNVL):  STREAM_GATED only.  cudaIPC event handles
+//     cannot traverse the MNNVL fabric; supported() returns false for
+//     REMOTE_IPC, and default() returns STREAM_GATED.
+//   - Non-MNNVL build:           both supported, and default() returns
+//     REMOTE_IPC (events) -- the established develop behaviour, least surprise.
```
maddyscientist: I think this is poorly worded.  What does "Remote IPC" mean?  IPC is the inter-process communication interface that is supported between multiple processes, the remote here seems superfluous.  All copying, reading or writing here is by definition remote that is done over IPC.  The established behaviors is synchronization via CUDA Events over IPC.

The language should be cleaned up.

include/comm_quda.h
```diff
@@ -264,26 +264,64 @@ namespace quda
...
+  void comm_create_stream_gated_comms();
+
+  /**
+     @brief Whether the REMOTE_WRITE P2P policy (the packing kernel stores halos
+     directly into the peer's buffer) is safe to offer to the dslash policy
+     tuner on this build/device.  On MNNVL/fabric builds remote-write halos land
+     in the peer's imported VMM buffer across multiple transactions and the
+     doorbell can be observed before the last data transaction commits; the only
+     safe guard is a receiver-side flush (CU_STREAM_WAIT_VALUE_FLUSH). Some hardware
+     can report CAN_FLUSH_REMOTE_WRITES=0. In those casesr emote-write is dropped from the
```
maddyscientist: Typo

include/comm_quda.h
```diff
@@ -423,6 +461,86 @@ namespace quda
...
+  */
+  enum class FieldKind { COLOR_SPINOR, GAUGE };
+  static constexpr int N_FIELD_KINDS = 2; // keep in sync with FieldKind
+
+  /**
+     P2P signalling mechanism (used by the comm_p2p_signal_* unified API
+     below to dispatch to the appropriate per-kind implementation inside the
+     backend).
+
+     - REMOTE_IPC  : cudaIPC event record/wait + MPI doorbell.  Single-node
```
maddyscientist: As noted elsewhere, the name `REMOTE_IPC` doesn't mean anything as all inter-process communication is done via IPC.  A more appropriate name would be `REMOTE_EVENT`.

include/communicator_quda.h
```diff
@@ -280,49 +293,172 @@ namespace quda
...
+          // Probe P2P reachability for our (up to) eight face-neighbours only:
+          // P2P is a nearest-neighbour halo exchange, so reachability to
+          // non-neighbour ranks is irrelevant (the old code imported all N peers
+          // but only ever consulted the 8 neighbours).  local_cap[dir][dim] is
+          // this rank's capability toward that neighbour -- true if we can import
+          // its fabric handle (P2P), false otherwise (MPI).  dir==0 is the -1
+          // (backward) neighbour, dir==1 the +1 (forward) neighbour (see
+          // comm_set_neighbor_ranks).
+          //
+          // Study hook: QUDA_P2P_FAKE_CLIQUE=<ranks-per-clique> soft-partitions the ranks
```
maddyscientist: What is a study hook?

include/gauge_field.h
```diff
@@ -343,6 +343,22 @@ namespace quda {
     */
     void commsComplete(int dim, int dir);
 
+    /**
+       @brief Queue a peer-to-peer ghost send via stream-mem-op signalling
+       (qudaMemcpyP2PAsync + cuStreamWriteValue64).  No MPI request posted;
+       the peer learns of arrival via the signal memory cell.  Caller must
+       ensure peer-to-peer is enabled for (dir, dim).
```
maddyscientist, line 350: Missing doxygen markup for parameters

include/gauge_field.h
```diff
@@ -343,6 +343,22 @@ namespace quda {
...
+    */
+    void sendStartStreamGated(int dim, int dir, const qudaStream_t &stream);
+
+    /**
+       @brief Queue a peer-to-peer recv wait via stream-mem-op signalling
+       (cuStreamWaitValue64).  No MPI wait; the signal is captured on the
+       GPU stream so downstream consumers gate on the stream / events.
+       Caller must ensure peer-to-peer is enabled for (1-dir, dim).
+    */
```
maddyscientist, line 365: doxygen markup

maddyscientist: Also noting that the comments here for gauge_field.h for these new functions are much more readable that the equivalent for color_spinor_field.h

include/targets/cuda/comm_target.h
```diff
@@ -0,0 +1,64 @@
...
+    size_t fabric_handle_size();
+
+    /**
+       @brief Allocate a small probe buffer suitable for fabric P2P and
+       write its local CUmemFabricHandle into `out_handle` (must point at
+       at least fabric_handle_size() bytes).  Returns an opaque handle the
+       caller passes to close_fabric_probe() when done.  Returns nullptr on
+       non-MNNVL builds.
+     */
+    void *open_fabric_probe(void *out_handle);
```
maddyscientist, line 32: Doxygen markup for function parameters

include/targets/cuda/comm_target.h
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+       @brief Attempt cuMemImportFromShareableHandle on the supplied peer
+       fabric handle (fabric_handle_size() bytes).  Returns true on success
+       (and releases the imported handle), false otherwise.  Safe to call
+       on a peer outside this rank's actual fabric reach -- the failed
+       import is the truth signal.  Returns false on non-MNNVL builds.
+     */
+    bool try_import_fabric_handle(const void *peer_handle);
```
maddyscientist, line 42: Doxygen

include/targets/cuda/comm_target.h
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+       @brief Release the probe buffer returned by open_fabric_probe().
+       No-op on non-MNNVL builds.
+     */
+    void close_fabric_probe(void *probe);
```
maddyscientist, line 49: Doxygen

include/lattice_field.h
```diff
@@ -748,30 +825,46 @@ namespace quda {
...
-       @brief Return base pointer to a remote device buffer for direct
-       sending in a given direction and dimension.  Since this is a
-       base pointer, one still needs to take care of offsetting to the
-       correct point for each direction/dimension.
+       @brief Return base pointer to the NVSHMEM remote pointer
+       (ghost_remote_send_buffer_d = nvshmem_ptr of the peer's symmetric recv
+       buffer), the Shmem direct-NVLink put dest.  Base pointer -- caller
+       offsets per dir/dim.  Non-null only when the peer is NVLink-reachable.
        @param[in] dir Direction we are requesting
        @param[in] dim Dimension we are requesting
```
maddyscientist, line 865: The @return markup has been deleted

include/lattice_field.h
```diff
@@ -748,30 +825,46 @@ namespace quda {
...
+       source).  Aliases the P2P send buffer in non-NVSHMEM builds.
+    */
+    void *myFaceShmem_d(int dir, int dim) const;
+
+    /**
+       @brief Return base pointer to the imported peer P2P recv buffer
+       (ghost_remote_send_buffer_p2p_d), the destination for P2P copy-engine /
+       remote-write.  Base pointer -- caller offsets per dir/dim.
        @param[in] dir Direction we are requesting
        @param[in] dim Dimension we are requesting
```
maddyscientist, line 854: @return deleted

include/lattice_field.h
```diff
@@ -748,30 +825,46 @@ namespace quda {
...
+    void *remoteFaceShmem_d(int dir, int dim) const;
 
     /**
-       @brief Return base pointer to the ghost recv buffer. Since this is a
-       base pointer, one still needs to take care of offsetting to the
-       correct point for each direction/dimension.
-       @return Pointer to remote memory buffer
+       @brief Return base pointer to the symmetric ghost recv buffer
+       (ghost_recv_buffer_d), used as the NVSHMEM put destination/reference.
+       Base pointer -- caller offsets per dir/dim.
```
maddyscientist, line 874: @return deleted

CMakeLists.txt
```diff
@@ -159,6 +159,7 @@ option(QUDA_QIO "build QIO code for binary I/O" OFF)
 # Multi-GPU options
 option(QUDA_QMP "build the QMP multi-GPU code" OFF)
 option(QUDA_MPI "build the MPI multi-GPU code" OFF)
+option(QUDA_MNNVL "enable MNNVL fabric P2P communication path" OFF)
```
maddyscientist: This option should be more clearly spelled out like we have with other CMake options, to make it more obvious what this does.  To those not in the know, `QUDA_MNNVL` is a bit of a random stream of characters.  How about `QUDA_MULTI_NODE_NVLINK`?  No problem keeping the compiler define variable inside QUDA as `QUDA_MNNVL` though.

tests/CMakeLists.txt
```diff
@@ -1124,6 +1124,54 @@ foreach(pol IN LISTS DSLASH_POLICIES)
...
+# Stream-gated P2P regression.  Guards the stream-gated flag-buffer lifecycle: on
+# an NVSHMEM build, QUDA_ENABLE_NVSHMEM=0 drives the non-NVSHMEM stream-gated P2P
+# path; if flag-buffer creation is ever suppressed, the lifecycle invariant in
+# comm_target.cpp aborts deterministically (no reliance on a numerical race).
+# --require-p2p makes it non-vacuous (fails if no P2P neighbour is enabled).
+#
+# Registered only when the configure-time grid has >1 rank (QUDA_TEST_NUM_PROCS,
+# derived from QUDA_TEST_GRID_SIZE which MUST be set BEFORE cmake configure). A
+# single-rank configure cannot exercise P2P, so we skip with a clear message.
+if(QUDA_NVSHMEM AND QUDA_DIRAC_STAGGERED AND (QUDA_MPI OR QUDA_QMP))
```
maddyscientist, line 1137: Just curious why doing this test with staggered fermions only?

lib/lattice_field.cpp
```diff
@@ -280,15 +339,47 @@ namespace quda {
...
           // silence any false cuda-memcheck initcheck errors
           qudaMemset(ghost_recv_buffer_d[b], 0, ghost_bytes);
 
           // gpu send buffer (use pinned allocator to avoid this being redirected, e.g., by QDPJIT)
-          ghost_send_buffer_d[b] = device_comms_pinned_malloc(ghost_bytes);
+          ghost_send_buffer_d[b] = ghost_comm_buffer_malloc(ghost_bytes);
           // silence any false cuda-memcheck initcheck errors
           qudaMemset(ghost_send_buffer_d[b], 0, ghost_bytes);
 
+          // Contiguous, RDMA-capable P2P buffers.  Under NVSHMEM ghost_*_buffer_d
```
maddyscientist, line 351: Does this mean that if we have an NVSHMEM build we end up burning more memory on comms buffers since we have separate RDMA and NVSHMEM comms buffers?

tests/utils/command_line_params.cpp
```diff
@@ -751,6 +754,26 @@ std::shared_ptr<QUDAApp> make_app(std::string app_description, std::string app_n
     },
     "Set the communication topology (X=1, Y=2, Z=4, T=8, and combinations of these)");
 
+  // Debug hook: per-dim P2P mask applied during comm_peer2peer_init.  CLI ->
```
maddyscientist, line 757: I seem to be missing the point of this.  Can you explain please?

lib/gauge_field.cpp
```diff
@@ -725,63 +857,104 @@ namespace quda {
...
-        // need to synchronize the copy streams before rejoining the compute stream - could replace with event post and wait
-        qudaDeviceSynchronize();
+        // Inject this dim's halo on the default stream (gated by the scatterDone
+        // events queued above) BEFORE the next dim is extracted, so its corners
+        // feed dim+1's extraction.
         extractExtendedGaugeGhost(*this, dim, R, recv_d, false);
       }
 
       bufferIndex = 1 - bufferIndex;
-      qudaDeviceSynchronize();
```
maddyscientist, line 784: Why was this sync deleted?

include/comm_quda.h
bjoo: Agreed. Adding to list.

include/communicator_quda.h
bjoo: It is a random Claude-ism. It is a testing knob really.  Will remove 'study hook'.

bjoo: Added to list.

tests/CMakeLists.txt
bjoo: I will test with Wilson too. Staggered came first just because of MILC pressures.

lib/lattice_field.cpp
bjoo: Yes. I think that is the price for having them separated. The issue being is that the symmetric heap may have more than one handle per allocation. This is to address @weinbe2 's request for single handle contiguous P2P buffers in the presence of NVSHMEM.

tests/utils/command_line_params.cpp
bjoo: This helped me to debug ordering issues which came when communicating in more than 1 dimension via P2P. We can get rid of this now if needed.

lib/gauge_field.cpp
bjoo: Claude's answer:

'develop' flagged each of those syncs in its own comment — "could replace with event post and wait". So that is what we do, swapping the two coarse intra-exchange device barriers for per-direction CUDA events:
- before the sends: gauge_extractEndEvent recorded on the default stream after the ghost extract (:603), each per-dir stream does qudaStreamWaitEvent on it (:612) — replaces develop's :527.
- before the inject: gauge_scatterDoneEvent[] recorded per-dir after the receive (:667), the default stream waits on each (:668) — replaces develop's :561.

Both events are recorded unconditionally (not under if (STREAM_GATED)), so the ordering holds for both the events (REMOTE_EVENT) and stream-gated transports. Two syncs were deliberately kept: the GDR one (:648) and the terminal per-exchange qudaDeviceSynchronize() (:686) that still brackets every exchange.

Why remove rather than keep under a guard: under stream-gated signalling a full-device host sync is potentially hazardous — it blocks the host on the per-dir streams that carry cuStreamWaitValue64 waits on a peer's not-yet-written signal memory, so the host can stall until the peer arrives (deadlock-prone under mutual run-ahead). The event edges keep the dependency on-device (stream-waits-stream) with no host barrier. The if (STREAM_GATED) sendDoneEvent/sendEndEvent block (:635-643) is an additional barrier for that transport (prevents a cross-stream WaitValue scheduler cycle) — it's additive, not a replacement.

bjoo:
Please block this merge until we merge hotfix/nvshmem-3.7.0. I will rebase after.

PR #1642, opened 2026-07-09
Cap `max_dynamic_shared_memory()` with `QUDA_MAX_SHARED_MEMORY` CMake override

maddyscientist:
## Problem

When `QUDA_MAX_SHARED_MEMORY` is set at cmake time, the compile-time `maximum_dynamic_shared_memory()` constexpr in `device.in.hpp` correctly respects it and bounds the tuner's shared memory search. However, the runtime `max_dynamic_shared_memory()` in `device.cpp` always returned the raw hardware-queried value, ignoring the override entirely.

This meant that `qudaLaunchKernel`'s `cuKernelSetAttribute(MAX_DYNAMIC_SHARED_SIZE_BYTES)` call was always set to the full hardware maximum, regardless of the user-specified cap. The cmake flag therefore had an incomplete effect: it constrained autotuning but not the actual attribute passed to the driver at launch time.

## Fix

`QUDA_MAX_SHARED_MEMORY` is now also passed to `device.cpp` as the preprocessor definition `QUDA_MAX_SHARED_MEMORY_OVERRIDE` via `target_compile_definitions` in cmake. (`device.cpp` compiles as C++, not CUDA, so it cannot include the generated `device.hpp` where the constexpr lives.) The runtime `max_dynamic_shared_memory()` then returns `min(hardware_max, override)` when the override is nonzero.

The existing code in `qudaLaunchKernel` (`quda_api.cpp:~153`) already correctly subtracts `sharedSizeBytes` (static shared memory) from the value returned by `max_dynamic_shared_memory()` before passing it to `cuKernelSetAttribute`. This means the fix is robust for kernels that mix static and dynamic shared memory — the total of the two will never exceed the user-specified limit.

## Testing

Compile with `-DQUDA_MAX_SHARED_MEMORY=102400` on a GPU with a higher hardware maximum (e.g. H100 at 228 KB) and verify that `cuKernelSetAttribute` is called with 100 KB rather than 228 KB minus static smem.

### DIFF
--- a/include/targets/cuda/device.in.hpp
+++ b/include/targets/cuda/device.in.hpp
@@ -9,7 +9,7 @@ namespace quda
     constexpr int maximum_dynamic_shared_memory_override = @QUDA_MAX_SHARED_MEMORY@;
 
     /**
-      @brief A constexpr function to returns the maximum dyanmic shared memory per block.
+      @brief A constexpr function to returns the maximum dynamic shared memory per block.
         See https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#feature-availability
      */
     constexpr int maximum_dynamic_shared_memory()
--- a/lib/targets/cuda/device.cpp
+++ b/lib/targets/cuda/device.cpp
@@ -4,6 +4,7 @@
 #include <quda_internal.h>
 #include <quda_cuda_api.h>
 #include <nvml.h>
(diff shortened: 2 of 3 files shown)


weinbe2 (approved):
Looks good to me, approved pending successful completion of the CI.

PR #1645, opened 2026-07-18
Hotfix/nvshmem 3.7.0 new

bjoo:
Minor changes to do with NVSHMEM-3.7 but which should work with 3.4.5 as well (3.6 had some header include issues). 
* CCCL Download option: Optionally download CCCL if not available in the toolkit. Currently QUDA downloads CCCL-3.1.4, but if available in toolkit, the toolkit version may be different. If NVSHMEM was built against the toolkit version of CCCL, then we may not want to download a different version
* Fix to use NVSHMEM's CMake support rather than using '-lnvshmem' flags
* Previously QUDA called nvshmem_initialize() in pool::init() and never explicitly called nvshmem_finalize(). With recent NVSHMEM versions it is a good idea to call nvshmem_finalize() otherwise finalization can occur in C++ __atexit() and this can result in segfaults in QUDA. I have added a pool::finalize() which explicitly calls nvshmem_finalize() -- guarded by #ifdef NVSHMEM_COMMS . I call pool::finalize() in endQuda(). This avoids the segfault.
* Rebased on QUDA develop.

### DIFF
--- a/include/malloc_quda.h
+++ b/include/malloc_quda.h
@@ -115,6 +115,12 @@ namespace quda {
     void init();
 
     /**
+       @brief Finalize the memory pool allocator.  Pairs with init(); when built
+       with NVSHMEM it also finalizes NVSHMEM (init() bootstraps NVSHMEM).
+    */
+    void finalize();
+
+    /**
        @brief Allocate device-memory.  If free pre-existing allocation exists
        reuse this.
        @param size Size of allocation
--- a/include/malloc_quda.h
+++ b/include/malloc_quda.h
@@ -115,10 +115,10 @@ namespace quda {
     void init();
 
     /**
-       @brief Finalize the memory pool allocator.  Pairs with init(); when built
+       @brief Destroy the memory pool subsystem.  Pairs with init(); when built
        with NVSHMEM it also finalizes NVSHMEM (init() bootstraps NVSHMEM).
     */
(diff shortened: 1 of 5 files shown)


bjoo:
It would be handy to merge this before PR #1641  and then I could rebase that.

maddyscientist (approved):
This looks good @bjoo.

One comment is that in other branches I'm working on, I have bumped the minimum CCCL version to get some newer features that 3.1.  If we need a higher version than comes with the CUDA toolkit, we'll always the version to build NVSHMEM to use the newer version as well, this use of the toolkit's version of CCCL, is just a convenience to avoid this step?

bjoo:
As per @mathiaswagner : renamed `pool::finalize()` to `pool::destroy()` and rebased on current `develop`. A quick test on Hecate looked good. Assuming all the builds work this should be ready for merging.

maddyscientist:
@bjoo can you merge in develop before we merge?  Also, why did you force push?

bjoo:
I already rebased onto 'develop' (since develpo has moved ahead I thought rebase may be better). The 'force push' was a result of the rebase otherwise it would've been rejected as not fast forward.

maddyscientist:
@mathiaswagner can you approve please?

mathiaswagner (commented):
looks good to me.

PR #1651, opened 2026-08-31
Hotfix/prefetch overflow

maddyscientist:
## Summary
This branch fixes inconsistencies within the gauge/spinor/clover `FloatNOrder` load/store and prefetch member functions which used different address arithmetic. 
 * `vector_load`/`vector_store` formed `ptr + (offset + N * idx)` with an unsigned 32-bit `offset`, so the add wrapped in 32 bits before the pointer add.
 * Prefetch used a similar `gauge + (full expression)`, however the expression was signed, thus products such as `dir * (M * N) * stride` could overflow while the true linear index still fit in 32-bit unsigned; prefetch then sign-extended into 64-bit and pointed at the wrong place.  This lead to pathological performance regressions at large volume with prefetching enabled.

This branch:
- Introduces a global `quda::index_t` (`uint32_t` by default, `uint64_t` with `QUDA_64BIT_INDEXING=ON`).
- Removes the per-accessor `huge_alloc` / `AllocType` template parameter so all native accessors use that one width.
- Makes `vector_load` / `vector_store` use a single 3-operand form that does `ptr + (offset + N * idx)` in `index_t`, matching prefetch.
- Uses unsigned 32-bit indexing by default so `dir * (M * N) * stride` cannot sign-extend into a wrong 64-bit prefetch address.

Default 32-bit codegen matches develop SASS on Wilson recon-18 (half/single/double, sm_70): identical instruction counts. 64-bit indexing adds ~5–7% instructions, all extra integer/address math (`IMAD.WIDE`, `LEA.HI.X`, `IADD3.X`); global load/store counts are unchanged and there are no extra spills.

Also on this branch:
- It was observed that the use of Mersenne Twister RNG in the test suite has pathological memory requirements (excess of 2 KiB per lattice site).  Thus the host test RNG now uses the 16-byte `xoroshiro128++` per site instead of `std::mt19937_64`, so large-volume tests do not OOM.
- `static const` → `static constexpr` in the color-spinor accessor (addresses linking errors with some versions of LLVM)
- Warning fixes when all Dirac operators are disabled.
- Vendor `json.hpp` update (cherry picked from the SYCL branch)

### DIFF
--- a/include/color_spinor_field_order.h
+++ b/include/color_spinor_field_order.h
@@ -11,7 +11,9 @@
  *  also.
  */
 
+#include <cstdint>
 #include <limits>
+#include <type_traits>
 #include <register_traits.h>
 #include <convert.h>
 #include <complex_quda.h>
@@ -1127,10 +1129,10 @@ namespace quda
       using GhostNOrder = GhostNOrder<Float, Ns, Nc, spin_project, huge_alloc, disable_ghost>;
       using real = typename mapper<Float>::type;
       using complex = complex<real>;
-      using AllocInt = typename AllocType<huge_alloc>::type;
+      using index_t = std::conditional_t<huge_alloc, uint64_t, uint32_t>;
       using norm_t = float;
       Float *field = nullptr;
-      AllocInt offset = 0; // offset can be 32-bit or 64-bit
+      index_t offset = 0; // offset can be 32-bit or 64-bit
       int volumeCB = 0;
 
       FloatNOrder() = default;
@@ -1358,10 +1360,10 @@ namespace quda
       using GhostNOrder = GhostNOrder<Float, Ns, Nc, spin_project, huge_alloc, disable_ghost>;
       using real = typename mapper<Float>::type;
       using complex = complex<real>;
-      using AllocInt = typename AllocType<huge_alloc>::type;
+      using index_t = std::conditional_t<huge_alloc, uint64_t, uint32_t>;
       using norm_type = float;
       Float *field = nullptr;
-      AllocInt offset = 0; // offset can be 32-bit or 64-bit
+      index_t offset = 0; // offset can be 32-bit or 64-bit
       int volumeCB = 0;
 
       FloatNOrder() = default;
--- a/include/color_spinor_field_order.h
+++ b/include/color_spinor_field_order.h
@@ -975,7 +975,7 @@ namespace quda
       }
     };
 
-    template <typename Float, int Ns, int Nc, bool spin_project = false, bool huge_alloc = false, bool disable_ghost = false>
+    template <typename Float, int Ns, int Nc, bool spin_project = false, bool disable_ghost = false>
     struct GhostNOrder {
       GhostNOrder() = default;
       GhostNOrder(const GhostNOrder &) = default;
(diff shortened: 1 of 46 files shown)


weinbe2:
I see that we removed the _template argument_ `spinor_direct_load` from a lot of the *.cuh files, which makes sense given the refactor. However, there are still lines that define it, such as:

```
static constexpr bool spinor_direct_load = false; // false means texture load
```

Can we get rid of all of these lines, or is there a reason I'm missing to keep them around?

maddyscientist:
> I see that we removed the _template argument_ `spinor_direct_load` from a lot of the *.cuh files, which makes sense given the refactor. However, there are still lines that define it, such as:
> 
> 
> 
> ```
> 
> static constexpr bool spinor_direct_load = false; // false means texture load
> 
> ```
> 
> 
> 
> Can we get rid of all of these lines, or is there a reason I'm missing to keep them around?

Good catch @weinbe2.  I'll fix that.

include/targets/generic/load_store.h
```diff
@@ -41,10 +43,11 @@ namespace quda
     return value_a;
   }
 
-  template <typename scalar_t, int N, size_t prefetch = 0>
-  __device__ __host__ inline array<scalar_t, N> vector_load(const scalar_t *ptr, unsigned int offset, int idx)
+  template <typename scalar_t, int N, size_t prefetch = 0, typename index_t>
```
weinbe2: I'm surprised this compiled because as far as I can tell, this isn't legal C++ because all default parameters have to come at the end of the instantiation list (even when there's mixed types/integral values). It should be:

```
  template <typename scalar_t, int N, typename index_t, size_t prefetch = 0>
```

I believe this never got triggered because `prefetch` is always explicitly specified?

include/color_spinor_field_order.h
```diff
@@ -386,25 +390,25 @@ namespace quda
...
-      const Float scale;             /**< Float to fixed-point scale factor */
-      const Float scale_inv;         /**< Fixed-point to float scale factor */
+      const Float scale;     /**< Float to fixed-point scale factor */
+      const Float scale_inv; /**< Fixed-point to float scale factor */
     public:
-      norm_t *norm;                  /**< Address of norm field (if it exists) */
-      const int norm_idx;            /**< Index into norm field */
-      const bool norm_write;         /**< Whether we need to write to the norm field */
+      norm_t *norm;                                                   /**< Address of norm field (if it exists) */
+      const int norm_idx;                                             /**< Index into norm field */
```
weinbe2: I think this should also be `index_t`.

maddyscientist: 67a1335e3d056ccc23f501bbb2189b86fc054ddf

include/targets/generic/load_store.h
maddyscientist: bc8bde3dac56c6a3f802e1f9ebc28f10a5f06790

maddyscientist:
@weinbe2 thanks for the review.  I've pushed fixes for all your suggestions (and gone the whole hog for using `index_t` in all colorspinor accessors).

Theres's more to be done in the future in this regard (optimizing 64-bit indexing when enabled, auto-detecting when one exceeds the indexing limit, etc.) but for another day.

maddyscientist:
cscs-ci run

weinbe2 (approved):
This looks good, approved, thanks @maddyscientist !

Issue #1636, opened 2026-06-18
HIP build adds non-existent include path for hipfft


kostrzewa:
In

https://github.com/lattice/quda/blame/914b462d802e515eb20b7313f2cd9d0b1264b8f4/lib/targets/hip/target_hip.cmake#L135

the path

```cmake
target_include_directories(quda PUBLIC ${ROCM_PATH}/hipfft/include)
```

is added to the include directories.

While QUDA builds fine, configuring another code linking against QUDA using CMake will fail like so:

```
CMake Error in src/lib/CMakeLists.txt:
  Imported target "QUDA::quda" includes non-existent path

    "/opt/rocm-6.3.4/hipfft/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

This may be a ROCm version issue (6.3.4 and 6.4.4 are the latest we can use on LUMI-G) but at least for those two versions, the correct path is `${ROCM_PATH}/include/hipfft`

kostrzewa:
@dmcdougall is the path to the include file the other way around with newer ROCm versions?

kostrzewa:
@kottnad I can thus confirm what you told me and the issue seems to be here.

dmcdougall:
I'm confused by this.  The header files were reorganised [quite a long time ago](https://rocm.docs.amd.com/en/docs-7.2.4/conceptual/file-reorg.html).

kostrzewa:
> I'm confused by this. The header files were reorganised [quite a long time ago](https://rocm.docs.amd.com/en/docs-7.2.4/conceptual/file-reorg.html).

Thanks for the quick reply!

This means that the "correct" `${ROCM_PATH}/include/hipfft` could simply be used in QUDA likely without ill effects, right?

dmcdougall:
> This means that the "correct" `${ROCM_PATH}/include/hipfft` could simply be used in QUDA likely without ill effects, right?

It depends what you mean by "ill effects".  I'm sure someone somewhere is using an ancient ROCm and expects it to still work.  But those configurations come with worse problems than a header file that now lives in a different location.

Yes, I highly encourage implementing your suggestion.  What confuses me is that I would have expected to hear about issues like this pretty soon after the change happened...

dmcdougall:
[...]

PR #1652, opened 2026-09-03
hip: drop the hardcoded hipfft include path

jakurzak:
`lib/targets/hip/target_hip.cmake` has:

```cmake
target_include_directories(quda PUBLIC ${ROCM_PATH}/hipfft/include)
```

`/opt/rocm/hipfft/include` has not existed since the ROCm 5 layout change; on ROCm 7.14 the headers are under `/opt/rocm/include/hipfft`.

Because the path is `PUBLIC` it ends up in the installed target's `INTERFACE_INCLUDE_DIRECTORIES`, and CMake hard-errors any project that consumes the QUDA package:

```
CMake Error in lib/CMakeLists.txt:
  Imported target "QUDA::quda" includes non-existent path

    "/opt/rocm/hipfft/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.
```

QUDA itself builds and installs without a warning, so this is invisible until something downstream calls `find_package(QUDA)` — here, Chroma, which died three seconds into its configure after a 17-minute QUDA build. An application that links QUDA directly rather than through the CMake package never sees it, which is why it has gone unnoticed.

The line is also unnecessary on any ROCm: `hip::hipfft` on the next line carries its own include directories, and `include/targets/hip/FFT_Plans.h` includes `<hipfft/hipfft.h>`, which resolves from the default HIP include path. QUDA compiles clean with the line removed — the `-I` was already pointing at nothing during that 17-minute build.

## Testing

ROCm 7.14, MI355X, `QUDA_TARGET_TYPE=HIP`, `QUDA_GPU_ARCH=gfx942;gfx950`, built with `QUDA_QDPJIT=ON -DQUDA_INTERFACE_QDPJIT=ON` against QDP-JIT.

- QUDA configures, builds and installs; `libquda.so` carries both architectures.
- Chroma's `find_package(QUDA)` now succeeds, and Chroma builds and links against the installed package. It did not before this change.

### DIFF
--- a/lib/targets/hip/target_hip.cmake
+++ b/lib/targets/hip/target_hip.cmake
@@ -135,7 +135,6 @@ set_source_files_properties( ${QUDA_CU_OBJS} PROPERTIES LANGUAGE HIP)
 target_link_libraries(quda PUBLIC hip::hiprand roc::rocrand hip::hipcub roc::rocprim_hip)
 target_link_libraries(quda PUBLIC roc::hipblas roc::rocblas)
 
-target_include_directories(quda PUBLIC ${ROCM_PATH}/hipfft/include)
 target_link_libraries(quda PUBLIC hip::hipfft)
 
 add_subdirectory(targets/hip)


kostrzewa:
should fix https://github.com/lattice/quda/issues/1636

maddyscientist (approved):

PR #1654, opened 2026-09-20
Feature/cccl fpmp

maddyscientist:
## Summary
Add CCCL experimental FPMP as first-class QUDA compute types:
* **fp64mp2** for double-double
* **fp32mp2 (float-float)** as a native double-width stand-in for IEEE `double` on low-FP64 GPUs (B300, Ada, consumer Ampere, etc.).
When `QUDA_FPMP_FLOATFLOAT=ON`, `QUDA_DOUBLE_PRECISION` register arithmetic and native double fields are `fp32mp2_*`. Reductions can use a different accuracy than local computation. IEEE `double` remains the host scalar (`real_t`) unless you pick another `QUDA_SCALAR_TYPE`.
### Accuracy (`low` / `mid` / `high`)
Each FPMP type is a `(hi, lo)` pair of IEEE limbs (`float`+`float` or `double`+`double`). CCCL exposes three arithmetic accuracies:
- **low** — unevaluated pair: cheapest, no renormalize after each op. Fine for local stencil work.
- **mid** — renormalized pair: extra robustness when many values are combined (reductions, heatbath / overrelax).
- **high** — same renormalized model, tighter error bound; not the default here.
The three accuracies are distinct C++ types, so bulk compute can be **low** while reductions are **mid**.
### CMake
- `QUDA_FPMP_DOUBLEDOUBLE` (default ON) / `_ACCURACY` (`low|mid|high`)
- `QUDA_FPMP_FLOATFLOAT` / `_ACCURACY`
- `QUDA_REDUCTION_TYPE`: `double` | `doubledouble` | `floatfloat` | `floatfloat_{low,mid,high}`
- **Low-FP64 arches** (`sm_72/75/86/87/89/103/110/120/121`) default to float-float **bulk low**, reductions **mid**
- **High-FP64** (`sm_70/80/90/100/107`) stay IEEE double
- `QUDA_DOWNLOAD_CCCL` still defaults ON (CCCL `main` until 3.6). Toolkit CCCL without `<cuda/fpmp>` only fails configure if an FPMP option actually needs it; the error tells you to update CCCL, download, or `-DQUDA_FPMP_FLOATFLOAT=OFF -DQUDA_FPMP_DOUBLEDOUBLE=OFF`
### Numerics / correctness
LOW float-float is stable for **local** work (dslash, most stencil arithmetic). A few places need the extra robustness of **MID**: global reductions, and heatbath / overrelax (those kernels promote LOW storage to MID in registers). Fields stored as LOW are **renormalized on write** so the `(hi, lo)` pair stays a valid pair across solver iterations, rather than drifting into an unnormalized state that MID/HIGH arithmetic would reject. Pair atomics use CCCL CAS for the same reason.
Use `compute_epsilon` (not `std::numeric_limits::epsilon`) so tolerances follow how QUDA was compiled — IEEE `double` vs fp32mp2, plus the usual half/quarter values. Tests use that same precision measure. Reproducible reductions (RFA) remain IEEE-double only.
- Explicit `real_t` / `double` conversions (no silent reinterpret)
### Performance
- CCCL double-double (**low**) is up to **2×** faster than the previous CUDA double-double for block reduction kernels
- The Wilson kernel is very large in float-float; the hop/dimension loop is **not** unrolled for fp32mp2 so the I-cache is not thrashed (staggered stays unrolled). Dslash kernels were also tightened so the coordinate/boundary logic does not spill to local memory.
- **B300 dslash**, same 8-byte fields, LOW float-float vs IEEE FP64:
  - Wilson: **2855** vs **842 GFLOPS** (3.4×)
  - Staggered: **2625** vs **1160 GFLOPS** (2.3×)
### Eigensolve
- `QudaEigParam::mat_norm` from a Krylov estimate of `||A||` (ARPACK `anorm`)
- A rewrite of the host eig tests is underway so they match how the solver computes its residual and so the host check takes the limit of precision into account (ARPACK-inspired). That work is not strictly coupled to this PR and may slip.


SaltyChiang:
Thank you for the excellent work!

I tested this PR with several `invertQuda` calls on my RTX 5080 Laptop (sm_120). Due to the instability of the laptop GPU (GPU clock speed, memory clock speed, power, etc.), I picked the maximum performance number from several runs. The pion mass for both actions is approximately 300 MeV.

|Performance (GFLOPS)|`develop`|`feature/cccl_fpmp`|Speedup|
|:-:|:-:|:-:|:-:|
|clover/strange/BiCGStab|1063.589|1478.391|1.39x|
|clover/light/GCR+MR|812.145|1264.239|1.56x|
|HISQ/light/CG|921.534|984.732|1.07x|

maddyscientist:
@SaltyChiang could you review the changes please and approve if you're happy?  Thanks

